PR #20, opened 2014-11-22
Add ability to use custom input options.

ebrentnelson:
When customOptions={true}, the user will be able to input any arbitrary value as a selection value. This works in single and multi-selection modes. Because this input represents the value AND the label (or rather, we can't distinguish between the two), the inputted value will be used for both label and value in the onChange handler.

Example:

```
var opts = [
    { value: 'hobo', label: 'Hobo' },
    { value: 'home alone', label: 'Home Alone' },
    { value: 'horchata', label: 'Horchata' },
    { value: 'hohoho', label: 'Ho! Ho! Ho!' }
];
<Select multi={true} customOptions={true} options={opts} />
```

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -302,8 +302,9 @@ var Select = React.createClass({
 	selectFocusedOption: function() {
 		if (this.props.customOptions && !this.state.focusedOption) {
 			return this.selectValue(this.state.inputValue);
+		} else if (this.state.focusedOption) {
+			return this.selectValue(this.state.focusedOption);
 		}
-		return this.selectValue(this.state.focusedOption);
 	},
 	
 	focusOption: function(op) {
--- a/src/Select.js
+++ b/src/Select.js
@@ -13,6 +13,7 @@ var Select = React.createClass({
 	propTypes: {
 		value: React.PropTypes.any,             // initial field value
 		multi: React.PropTypes.bool,            // multi-value input
+		customOptions: React.PropTypes.bool,    // custom value input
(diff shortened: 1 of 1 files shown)


magnusohlin:
Please add this feature!

TheLudd:
@ebrentnelson I have tried you patch and found some unexpected behavior. 
 1. If I enter something in the field and select it, the ```onChange``` callback is called as expected. But if I then re-edit the field and press enter the callback is **not** called again. Only if I drop focus of the input field and re-focus can I get the callback to fire again. 
 2. If enter custom option and then tab to the next element in the form, the callback is not called. This to me would be the way to "select" something that is not present in the pre existing options.

*Note:* The test are made against a rebase from the current master. You could update your fork if you are still interested in trying to have this merged or checkout my version here: https://github.com/mediatool/react-select

SrSandeepKumar:
Is their an option for `customOptions ` at present?
As is searched and could not find this functionality, i tried to `setState` of options array and then render back to `<select/>` but `onChange/onKeyDown` are not triggered.
What approach should i use to bring in the `customOptions` part in my application.

TheLudd:
@SrSandeepKumar We rolled our own selector to solve this.

SrSandeepKumar:
The whole `<select/>` ? I am new to react and from from <Select/> their are no events triggered other than `onChange` so can you give some reactjs-gyan!

agirton:
Closing as this has been resolved with the latest version.

PR #29, opened 2014-12-17
Add 3 config props

kompot:
Added 3 config properties:
— ability to disable "x" so that value can't be cleared
— configurable label for no results
— new label when input is empty (should probably by configured along with some new param like `minInputLengthToPerformRemoteReques`


julen:
Hm I realize there's some overlapping between this PR and #33, which was already merged few days ago. I believe having a separate PR for each concern would help to move this forwards.

JedWatson:
Agreed, @kompot if you could submit these additions as separate PRs that would be great.

kompot:
@JedWatson @julen Looks reasonable, will do separate PRs. Thanks.

PR #33, opened 2014-12-29
i18n: moved no results text to props

julen:
Allows customizing the no results text, required for scenarios where l10n is a must.


JedWatson:
Good idea, thanks.

PR #36, opened 2015-01-02
README: added missing bit to the example

julen:
The function is defined in the example but then it's unused.


JedWatson:
Thanks @julen

Issue #21, opened 2014-11-30
Focused option should scroll into view


JedWatson:
I've noticed that when you're navigating the list of options with the keyboard, it is possible to focus an option that's lower than the scrollHeight of the menu.

It should make sure that the currently focused option is always scrolled into view.

JedWatson:
Fixed by #37

moklick:
nice! thanks

lemurry:
Currently experiencing same issue with component with custom rendering items.

Arhane:
@lemurry Probably you didn't set `innerRef`

PR #37, opened 2015-01-02
scroll focused item into view on keyboard navigation

fmovlex:
fixes issue #21


fmovlex:
Just noticed I used 'selected' as a ref name where it should probably be 'focused', worth changing.

JedWatson:
Looks good @fmovlex, thanks!

PR #39, opened 2015-01-04
add support for touch event

fraserxu:
* For the option, I add an onClick event
* For the clear button, I add an onClick event
* For the wrapper, I add an onTouchEnd event
* For the Value.js, I add an onTouchEnd event

### DIFF
--- a/src/Value.js
+++ b/src/Value.js
@@ -17,7 +17,7 @@ var Option = React.createClass({
 	render: function() {
 		return (
 			<div className="Select-item">
-				<span className="Select-item-icon" onMouseDown={this.blockEvent} onClick={this.props.onRemove}>&times;</span>
+				<span className="Select-item-icon" onMouseDown={this.blockEvent} onClick={this.props.onRemove} onTouchEnd={this.props.onRemove}>&times;</span>
 				<span className="Select-item-label">{this.props.label}</span>
 			</div>
 		);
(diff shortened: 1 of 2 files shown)


src/Value.js
```diff
@@ -17,7 +17,7 @@ var Option = React.createClass({
 	render: function() {
 		return (
 			<div className="Select-item">
-				<span className="Select-item-icon" onMouseDown={this.blockEvent} onClick={this.props.onRemove}>&times;</span>
+				<span className="Select-item-icon" onMouseDown={this.blockEvent} onClick={this.props.onRemove} onTouchEnd={this.props.onRemove}>&times;</span>
```
JedWatson: Shouldn't the `onClick` handler still fire here on a touch device? or is it blocked by the `onMouseDown` handler?

JedWatson:
Thanks @fraserxu. I was initially concerned that binding both `mouseDown` and `onClick` to the same event handler might cause it to incorrectly fire twice for desktop users, but it may not be an issue... can you confirm this is correct?

src/Value.js
fraserxu: No, I tested, the `onClick` does not fire on a touch device.

Add the `onTouchEnd` will make it work.

I also tried to change the `onTouchEnd` handler to 
```
onTouchHanlder: function(e) {
  this.blockEvent(e)
  this.props.onRemove()
}
```
to prevent the input filed from focusing, but seems not work.

fraserxu:
Hi @JedWatson I checked the `mouseDown` and `onClick` on desktop, it will only fire once even after I added the `onTouchEnd` listener. 

Till now all seems work well for me.

JedWatson:
Cool. I'll merge it now so we can finally support mobile devices.

PR #42, opened 2015-01-06
Add `inputTooShortText` label.

kompot:


JedWatson:
Do you think `enterSearchText` or `noInputText` would be clearer for this prop name? `inputTooShort` implies that there's a minimum length setting to me.

kompot:
Well,
actually I think that minimum length setting should be a mandatory setting (defaulting to 1 probably). 

I can change it but creating a ticket for adding mandatory minimum input length seems also reasonable.

Up to you to decide.

JedWatson:
I've implemented this as `searchPromptText` because I think that's the clearest explanation for the text string itself. I also agree we could add a minimum length setting, but that requires some deeper thought because of the interaction with the autoloading of the default set of options.

The output is also dependent on `asyncOptions` being set, because otherwise in a situation where there are no options available (and no more may be loaded) it will say "Type to search..." which is not, I don't think, the correct behaviour.

Let me know if you think this should be tweaked.

JedWatson:
p.s. although I made the change directly rather than merging your PR, you got credit for this in the changelog :)

PR #43, opened 2015-01-06
Add `clearable` option to be able to restrict possibility of clearing

kompot:
input value.


JedWatson:
Great @kompot, thanks!

Issue #45, opened 2015-01-09
Allow any renderable node as an option label (int/string/element)


drewbanin:
It would be great if option labels weren't limited to only strings. I'd like to be able to pass in any React element and have it rendered as a regular option. The use case that I'm specifically thinking of is a bootstrap "badge" to show a count next to the select option 
http://getbootstrap.com/components/#badges

I made this change in a branch locally and would be happy to submit the PR!

julen:
Hmm there might be some overlap between this and #27.

drewbanin:
Yeah looks like it's a similar issue (sorry for not commenting there). The other PR suggest changing `propTypes` from

```javascript
propTypes: {
  label: React.PropTypes.string.isRequired
}
```
to
```javascript
propTypes: {
  label: React.PropTypes.oneOfType([
    React.PropTypes.string,
    React.PropTypes.number
  ]).isRequired
}
```

The docs (http://facebook.github.io/react/docs/reusable-components.html)  say that you can use `React.PropTypes.node` to permit "anything that can be rendered: numbers, strings, elements or an array containing these types."

I think allowing numbers + strings is great, but allowing numbers, strings, or React elements would be so versatile!

dmr:
We use Select with a custom component as a label and it sort of works meaning it renders.
Select reuses label also for search so search is broken now.
Is anybody experiencing similar things or has a solution?

wyattanderson:
> Select reuses label also for search so search is broken now. Is anybody experiencing similar things or has a solution?

Can confirm. Search is broken because it tries to match against the label directly, which doesn't work very well if label is a component. Would be nice to support this customization.

dmr:
@wyattanderson

I solved my own issue: You can implement your own filterOption function:

    var myFilterOption = function(op, filterValue) {
        var valueTest = String(op.value),
            labelTest = String(op.label ? op.label.props.object.name : '');  // HACK: use component data
        return !filterValue || this.props.matchPos === "start" ? this.props.matchProp !== "label" && valueTest.toLowerCase().substr(0, filterValue.length) === filterValue || this.props.matchProp !== "value" && labelTest.toLowerCase().substr(0, filterValue.length) === filterValue : this.props.matchProp !== "label" && valueTest.toLowerCase().indexOf(filterValue.toLowerCase()) >= 0 || this.props.matchProp !== "value" && labelTest.toLowerCase().indexOf(filterValue.toLowerCase()) >= 0;
    };

    return (
[...]

PR #50, opened 2015-01-21
Added possibility to add suggestions

vschoettke:
Added the possibility to add suggestions that are displayed instead of labels in the menu. This is useful if you want to display additional information in the menu that should not appear in the input box (e.g. zip code with city name for a zip code field)
You can either, use static suggestions one for each option by specifying the `suggestion` key in the options, or specify the property `suggestionGenerator` attribute. The suggestionGenerator is a function that is called with the input value and the current option object. Since react does not care what the child of an node is you can either return a string or a react element. This opens also the possibility to highlight the input in the suggestion, add images, etc.
This is probably related to #45 and allows implementation of #5.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -6,8 +6,12 @@ var _ = require('underscore'),
 
 var requestId = 0;
 
+function defaultSuggestionGenerator(inputValue, op) {
+	return op.suggestion ? op.suggestion : op.label;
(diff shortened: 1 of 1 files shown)


agirton:
Closing as we allow users to create a new option with the latest version.

PR #52, opened 2015-01-22
ensure the menu also uses the mousedown handler

### LINKED ISSUE #51: Manual drag of scrollbar causes blur (and closes dropdown)
Alright, this one is pretty freaking obscure, so pardon me as I try to explain this. I also will try to put together a simplified test case and hopefully even fix it, but I thought I would log it here :)

The issue does not present itself on the example page due to the relatively small height of the page; however, should you shrink your browser window so that the page is forced to scroll, scroll partially down the page, open the "States Select" and try to manually drag the scroll bar (via the handle) The select menu will close immediately due to handleInputBlur being called.

While I recognize the use case of clicking a scroll bar manually is probably few and far between, it is unexpected behavior.

If anyone has experienced this and knows a fix that does NOT involve patching react-select, I'd love to not need to go hacking around what I imagine is going to be a pretty nasty browser API issue.

nickhudkins:
Resolved #51


JedWatson:
:+1: good fix, thanks!

PR #53, opened 2015-01-23
Fixes an issue if options are updated

bruderstein:
If the props.value was set, but then options were set later (in
componentWillReceiveProps), the selected value (props.value) would not
be selected from the new options.


drewbanin:
:+1:, currently having this problem and this fix would be great!

dcousens:
Can you please rebase and only change the source files?  Cheers :)

bruderstein:
This is now fixed in a newer and better commit :)

PR #54, opened 2015-01-28
Do not ever addValue(undefined)

gja:


JedWatson:
Good catch, thanks @gja.

gja:
Sure, no problem :-). Would you mind doing a release? Need to update my bundle

JedWatson:
Yep, done.

PR #56, opened 2015-01-31
Added ability to disable the searching input

### LINKED ISSUE #55: Disabling input
I'm interested in using this plugin as a regular dropdown (like in the `react-dropdown` module, purely for styling purposes) and would love a way to disable the input field and autocomplete functionality. Is this something you would be interested in adding?

julen:
This PR adds a new prop, `searchable` (defaults to `true`), which will enable/disable the searching input functionality of a react-select element.

I haven't considered hiding the `input` element at all as suggested on #55, mostly because I quickly worked on this during xmas, therefore it was too early to read any suggestions made yesterday :sunglasses:
Instead, the `input` isn't rendered at all if `searchable` evaluates to false.

Fixes #55.

### DIFF
--- a/examples/src/app.js
+++ b/examples/src/app.js
@@ -18,11 +18,18 @@ var CountrySelect = React.createClass({
 });
  
 var StatesField = React.createClass({
+	getDefaultProps: function () {
+		return {
+			searchable: true,
+			label: 'States:',
+		};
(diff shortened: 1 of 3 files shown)


JedWatson:
Looks good! nice work @julen

PR #66, opened 2015-02-18
Improvement: Option label as a link

dmitry-smirnov:
Hi there. In my projects i use your component and i want to suggest little improvement.

**Description**: Add more possibilities for developers to customize react-select component. For example: when a user clicks on the label, I want description page to be opened. We also wrap the label into ```<a>``` tag so the developer can customize it.

Parameter name | Description | Default value
------------ | ------------- | -------------
onOptionLabelClick | function (Object option, Object event) | undefined

**Usage example**:
```jsx
var SelectedValuesField = React.createClass({
    onLabelClick: function (option, event) {
        console.log(option.value);
    },

    render: function() {
        var ops = [
            { label: 'A238 - Account1', value: 'a238' },
            { label: 'Vanilla', value: 'vanilla' },
            { label: 'Strawberry', value: 'strawberry' },
            { label: 'Caramel', value: 'caramel' }
        ];

        return (
            <div>
                <label>{this.props.label}</label>
                <ReactSelect
                    onOptionLabelClick={this.onLabelClick}
                    value="a238,vanilla,strawberry"
                    multi={true}
                    options={ops} />
            </div>
        );
    }
});
```

**Preview**:
screen-shot-2015-02-18-at-11 55 12-am


JedWatson:
Thanks @dmitry-smirnov! This looks really good, and it's something I've been meaning to add myself.

PR #71, opened 2015-02-23
Dependency changed to lodash, fixing requires

aaronpowell:
When the dep changed to lodash from underscore the `require` statements weren't updated and since this was a minor build anyone who us requiring `^0.3.2` will get the update


JedWatson:
sorry @aaronpowell, this shouldn't have slipped through on a minor. Thanks for the fix.

PR #72, opened 2015-02-24
Add handler for no found event

Lispython:
Added handler for no found event.

It's maybe need to collect not found input and user interests.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -27,6 +27,7 @@ var Select = React.createClass({
 		searchPromptText: React.PropTypes.string,  // label to prompt for search input
 		name: React.PropTypes.string,              // field name, for hidden <input /> tag
 		onChange: React.PropTypes.func,            // onChange handler: function(newValue) {}
+		onNoFound: React.PropTypes.func,           // onNoFound handler: function(value){}
 		className: React.PropTypes.string,         // className for the outer element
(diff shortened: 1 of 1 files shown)


dcousens:
@Lispython please rebase

dcousens:
@Lispython this approach sort-of conflicts with the `noResultsText` prop,  in that it seems odd that you would use both.

@JedWatson thoughts?

Lispython:
noResultsText for different purpose.
I use Select for cities list. Currently our service don't support all cities in out country. Then user try to find city it's view noResultsText message. We log all not found cities requests to Google Analytics via onNoFound callback. And when we collect many of requests we can add new city to out service.

dcousens:
@JedWatson feel free to re-open.  @Lispython this should be handle by your async request then.  There is no reason for the component to call a callback based on external state.

JedWatson:
I agree, best way to handle this is wrap the service call before providing it to `asyncOptions`

Lispython:
I don't use async options in this case. I get list of all cities and give its to options property. 

You say, that this is external state, but noResultsText and filterOptions incapsulated into Select Class.
If I try to handle this case in asyncOptions, I need to duplicate filterOptions and filterOption in asyncOptions function. That sad and not dry...

Issue #74, opened 2015-02-26
Setting value before options available doesn't update placeholder


farism:
If you set the value on the component before the options are available (e.g. from a parent component's state, which got them from an async store), when the options are finally retrieved and the component is rerendered, the select placeholder stays stuck on the option value rather than displaying the option label. This doesn't seem to be an issue when using the async prop.

3 sec delay example: 

http://jsfiddle.net/kb3gN/10048/

farism:
Submitted a PR to address this. I think the performance loss is fairly trivial.

https://github.com/JedWatson/react-select/pull/79

dcousens:
Ref #53

dcousens:
Resolved after https://github.com/JedWatson/react-select/pull/79 was submitted. 
ping @JedWatson to double check

PR #79, opened 2015-03-01
Support setting options after value thru props

farism:


JedWatson:
@farism can you explain this a bit please? looks like some important logic (e.g. `if (newProps.value !== this.state.value)`) has been removed.

farism:
From my understanding, that line of code is purely an optimization technique? The issue is that if options are set after value,  then the component will not be properly updated (and label placeholder does not change). See https://github.com/JedWatson/react-select/issues/74

It's hard to tell without tests, but I don't believe the code is harmful. I don't think there is an issue with calling `getStateFromValue` each time props change. I've been using this fork in a few development projects the past couple weeks, but I cannot make any guarantees.

farism:
Just realized that https://github.com/JedWatson/react-select/pull/53 pretty much addresses the same issue :)

dcousens:
Please rebase.  Happy to merge this or https://github.com/JedWatson/react-select/pull/53

Reggino:
I'm running into the same issue here. The fix looks good, thanks!

PR #80, opened 2015-03-03
Performance: Fixing repaint on scroll

jsmunich:
Before:
screenshot 2015-03-03 08 14 01

After: 
screenshot 2015-03-03 08 13 48

### DIFF
--- a/less/default.less
+++ b/less/default.less
@@ -23,6 +23,7 @@
 @select-arrow-width:               5px;
 
 @select-menu-zindex:               1000;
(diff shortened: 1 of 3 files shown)


JedWatson:
Thanks @jsmunich! that's a great improvement.

MarioJerkovicDickeys:
There is still an performance issue on select2 component. repaint on scroll. In my select2 comp I have loaded around 1200 items and its lagging really bad.

PR #81, opened 2015-03-04
add onBlur and onFocus event handlers

cloudkite:
Thanks for the great library!

Adding ability to hook into onBlur and onFocus events of the underlying input or div.
Needed this so that I could ensure form validation happens only once user has 'visited' a select box


mdressman:
+1

YannPl:
Need that too =)

JedWatson:
Thanks @cloudkite!

PR #82, opened 2015-03-04
Add custom props support for Input in select-control div

YannPl:
Allow dev to add custom props on the Input in the select-control div.
Useful to add data-foo props for example.


JedWatson:
Looks good @YannPl, thanks!

PR #85, opened 2015-03-08
Fix issue w/ remote callbacks overriding cached options

coreymcmahon:
Fixes an issue with the async callback overriding options populated from the cache.

Steps to reproduce:

1. User types `abcde`, callback issued for `?query=abcde`. Results displayed as expected.
2. User backspaces to `abc`
3. User *again* types `abcde`, slowly this time.
4. Options are correctly displayed from the cache (`cacheKey=abcde`), but an async request is raised for `abcd`
5. When the remote request is completed, the callback is initiated. Because `requestId` was not incremented for the last input change (as options were populated from the cache), the async results are pushed into the `options`.
6. User sees options for `abcd` even though they have typed `abcde`.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -348,6 +348,8 @@ var Select = React.createClass({
 
 	loadAsyncOptions: function(input, state) {
 
+		var thisRequestId = this._currentRequestId = requestId++;
+
 		for (var i = 0; i <= input.length; i++) {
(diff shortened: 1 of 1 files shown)


JedWatson:
Good catch, @coreymcmahon, thanks.

arnold-almeida:
:+1: :+1:

PR #93, opened 2015-03-13
Focus first element

esiqveland:
Focuses first element or the previous focused element on async data fetch.

Also focuses first element on menu_open if there isn't a focused one already.


JedWatson:
Thanks @esiqveland!

Issue #58, opened 2015-02-05
Return custom id instead of value


jvillasante:
It would be nice to have an option to pass custom data that would be passed on to handleSelect.
something like:

```javascript
options.push({
  label: 'label',
  value: 'value',
  id: 494949494
});
```

`id` being the value i want to get back later on selection/deletion.

Gregoor:
+1

cmnstmntmn:
:+1:

dcousens:
@jvillasante is there any reason you can't just have a look up?  Meaning we can leave the implementation as simple as possible?
I don't expect the lists to ever have duplicates?

dcousens:
Closing in favour of #194

PR #95, opened 2015-03-16
nameable value and label fields

lukemcgregor:
Change the names of the value and label fields on the data objects so you can use existing properties on richer collections of data.

NOTE: you may want to re-gen the bundle on your machine, I imagine the differences (like the source map) are due to different versions of things on our separate environments

Not too sure what SOP is for your auto-generated folders, I'm not sure they should even be in source.


dcousens:
Please rebase

dcousens:
Is this related to https://github.com/JedWatson/react-select/issues/58?

lukemcgregor:
Yeah it should resolve that also (although I hadn't seen that discussion) Ill rebase in a sec, is it easier to just exclude changes to autogen files and get you guys to regen?

dcousens:
Yeah.  Only change the source files,  otherwise review means double
checking.

dcousens:
@lukemcgregor I think #58 is a better solution to this.  This adds unnecessary complexity to the implementation IMHO,  you can always do a look up.

lukemcgregor:
@dcousens you are always going to have to trade between the complexity of the component and the complexity to use. I think this change reduces the amount of mapping consumers need which reduces usage complexity.

dcousens:
@lukemcgregor this is true,  but we should also strive to keep our components to abide by typical software design rules as much possible.
Limiting responsibilities of any software component is always ideal.

If you're having problems matching your data to what the component needs, perhaps you should be looking at ways to transform your data, rather then trying to bend the component to match your exact situation.

PR #100, opened 2015-03-22
Correction in README

Gregoor:
minor inconsistency


JedWatson:
Thanks for catching that @Gregoor

PR #101, opened 2015-03-22
Adding createable prop to allow creating of options which were not found

Gregoor:
I required this feature which is afaik also present in Selectize. Let me know if I missed anything or you want to merge it and also need documentatoin.

Also tried to stick to your coding style and lint went through fine.


JedWatson:
Thanks @Gregoor. I'd like to have a think on the implementation of this, as it would be good to also allow a callback for creating a new option (e.g. if a tag is required in the database to accept an ID - useful for SPAs). Will get back to you soon.

AlexMeah:
:+1: Need this for group selection/creation

guzart:
What are your thoughts so far @JedWatson ? 

Am thinking probably a `createOption` property that returns the option to be added. The function could have the following signature `function(value:string, [callback:func]):object`. 

By default the added option would be `{ value: _optionsFilterString, label: _optionsFilterString}`. If a `createOption` function is specified it could work synchronously by returning the option to be added, or asynchronously if a callback function is specified as the second parameter (arity of 2)

barsukov:
+1

dehbmarques:
+1

dcousens:
Closing, this was added in #151

PR #103, opened 2015-03-23
Ignore click events when menu container is closed

muromec:
For multiselect input it is possible to get into
race condition when user removes value while menu is visible.
Menu would be closed in Select.setValue() method, component
would be rerendered AND after this global click handler would fire.


JedWatson:
Good call @muromec, thanks.

PR #106, opened 2015-03-25
Fix an issue where options are not filtered at all

gsklee:
Prior to this fix, if both `props.value` and `props.options` are being altered at the same time, two consecutive `setState` calls will be fired with the latter one's `filteredOptions` not taking into account the newly set `props.value` at all. This PR fixes the issue by reversing the order of the `setState` calls so the correct `filteredOptions`, generated by `this.getStateFromValue`, can overwrite the wrong one.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -133,14 +133,14 @@ var Select = React.createClass({
 	},
 
 	componentWillReceiveProps: function(newProps) {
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @gsklee!

PR #107, opened 2015-03-25
Adding Jest unit test framework

mattdell:
Adding Jest unit test framework and one simple unit test to get started.

Run npm install and then npm test to run the test.

The unit test throws a ‘Maximum call stack size exceeded’ error, which seems to be caused by the Input component on line 631 of Select.js (change it to a div and you'll see). Input is a module of react-input-autosize, which is one of your other repositories so maybe you know more about why this happens?

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,9 @@
     "eslint": "^0.17.1",
     "eslint-plugin-react": "^1.5.0",
     "gulp": "^3.8.11",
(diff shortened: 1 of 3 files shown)


JedWatson:
Thanks @mattdell - that's really useful. I'll drop it into `react-input-autosize` too and see what I can work out, probably has something to do with the way that component updates its size by detecting DOM visibility - I suspect this causes an infinite loop in the test environment.

PR #116, opened 2015-04-03
Move react to peerDependencies and devDependencies

### LINKED ISSUE #88: Use peerDependencies for react reference
This means that it will use the version of react referenced by the application. Using different versions causes explode.

article about this here: http://stackoverflow.com/questions/27153166/typeerror-when-using-react-cannot-read-property-firstchild-of-undefined

ahutchings:
Fixes #88

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
   "dependencies": {
     "classnames": "^1.2.0",
     "lodash": "^3.5.0",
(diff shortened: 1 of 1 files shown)


JedWatson:
I had a lot of reservations about this initially but since it's how most other React libraries are handling it, I'm happy to change my position until a better workaround comes along.

PR #118, opened 2015-04-06
fix: set default options to null for correct async options state

akud82:
getStateFromValue returns wrong result with default options empty array in line 151-153 Select.js


JedWatson:
Thanks @akud82!

PR #124, opened 2015-04-09
Switch from lodash to ES5 native (+object-assign polyfill)

dennishall1:


src/Select.js
```diff
@@ -408,11 +412,11 @@ var Select = React.createClass({
 				var options = this._optionsCache[cacheKey].options;
 				var filteredOptions = this.filterOptions(options);
 
-				this.setState(_.extend({
+				this.setState({
 					options: options,
 					filteredOptions: filteredOptions,
-					focusedOption: _.contains(filteredOptions, this.state.focusedOption) ? this.state.focusedOption : filteredOptions[0]
-				}, state));
```
dcousens: Was this `extend` unnecessary?

src/Select.js
```diff
@@ -453,7 +457,7 @@ var Select = React.createClass({
 			return this.props.filterOptions.call(this, options, filterValue, exclude);
 		} else {
 			var filterOption = function(op) {
-				if (this.props.multi && _.contains(exclude, op.value)) return false;
```
dcousens: ACK

src/Select.js
```diff
@@ -463,8 +467,8 @@ var Select = React.createClass({
 					(this.props.matchProp !== 'label' && valueTest.toLowerCase().indexOf(filterValue.toLowerCase()) >= 0) ||
 					(this.props.matchProp !== 'value' && labelTest.toLowerCase().indexOf(filterValue.toLowerCase()) >= 0)
 				);
-			};
```
dcousens: Preferably just use a closure, but can fix after.

src/Select.js
```diff
@@ -463,8 +467,8 @@ var Select = React.createClass({
...
-			return _.filter(options, filterOption, this);
```
dcousens: ACK, thanks for removing the double bind.

dcousens:
Closing.  The use of lodash has been removed entirely.

PR #125, opened 2015-04-10
Prevent "document not defined" error in server side rendering.

wehriam:
See: https://github.com/facebook/react/issues/3620

### DIFF
--- a/lib/Select.js
+++ b/lib/Select.js
@@ -6,7 +6,8 @@ var _ = require('lodash'),
     React = require('react'),
     Input = require('react-input-autosize'),
     classes = require('classnames'),
-    Value = require('./Value');
+    Value = require('./Value'),
+    canUseDOM = require('react/lib/ExecutionEnvironment').canUseDOM;
 
 var requestId = 0;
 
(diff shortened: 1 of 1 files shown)


lib/Select.js
```diff
@@ -97,7 +98,10 @@ var Select = React.createClass({
 	componentWillMount: function componentWillMount() {
 		this._optionsCache = {};
 		this._optionsFilterString = '';
-		this.setState(this.getStateFromValue(this.props.value));
+		if(canUseDOM) {
+			this.setState(this.getStateFromValue(this.props.value));
```
sebmarkbage: Does this error actually fire here? If so, that is a bug in React 0.13.

The issue in facebook/react#3620 refers to callbacks to OTHER parent components from componentWillMount.

It should be fine to setState in componentWillMount, even on the server.

Perhaps one of the other callback fires or there's a bug in React 0.13 that can be fixed?

sebmarkbage:
This is not the right feature detection since I should be able to renderToString even in a DOM environment. E.g. to send to an iframe, across workers, in a chrome extension etc.

I should also be able to render to a string on the server even if I have JSDOM enabled for some legacy integration points.

xcatliu:
+1

xcatliu:
You should not update `lib` files directly, maybe you should read [CONTRIBUTING.md](https://github.com/JedWatson/react-select/blob/master/CONTRIBUTING.md) first.

agirton:

PR #131, opened 2015-04-19
Updating clear buttons to use the string code rather than raw html

bisrael8191:
When including this library in my project (react w/ webpack), the clear buttons displayed 'ã—' instead of 'x' (&times). I looked into it and found that the code was using raw HTML and the dangerouslySetInnerHTML option when it didn't need to. 

This pull changes the code to comply with the react site: https://facebook.github.io/react/docs/jsx-gotchas.html#html-entities 

Tested using the gulp scripts on Linux and Windows with Chrome and Firefox. Thanks for the great library!

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -609,7 +609,7 @@ var Select = React.createClass({
 		}
 
 		var loading = this.state.isLoading ? <span className="Select-loading" aria-hidden="true" /> : null;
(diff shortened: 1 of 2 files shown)


simpixelated:
I ran into this same problem but was able to fix it by adding
```html
<meta charset="UTF-8">
```
See discussion here: https://github.com/JedWatson/react-select/issues/314

agirton:

rsimonton:
@agirton This looks to me to still be an issue, i'm seeing it when using v1.0.0-rc.5:  https://github.com/JedWatson/react-select/blob/v1.0.0-rc.5/lib/Value.js#L83

PR #132, opened 2015-04-20
Fix autofocus on initial load & allow to disable cache with async

akud82:


cymen:
+1 -- I'd really like the focus not to happen after the async data callback is called (which this does of course)!

cymen:
The issue with focus is fixed on master by 50de293a.

dcousens:
Closing, please rebase and re-open without the focus fixes.

PR #137, opened 2015-04-21
Remove dependency on Lodash

### LINKED ISSUE #129: lodash dependency
This project looks pretty good, but is it possible to remove the lodash dependency? I don't really want to pull a large JavaScript library in just to use a typeahead component.

Daniel15:
Most of the stuff in here can just be done with vanilla JavaScript. After this change, the minified file only ends up ~450 bytes larger than the original version, and the total size is greatly reduced due to the removal of lodash.

If you really dislike for-in loops, they can be replaced with a [small Object.assign implementation](https://www.npmjs.com/package/object-assign)

Fixes #129

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
   },
   "dependencies": {
     "classnames": "^1.2.0",
-    "lodash": "^3.7.0",
     "react-input-autosize": "^0.4.3"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


JedWatson:
Looks great @Daniel15, I was meaning to do this so thanks for beating me to it :)

re: Object.assign, since we're using babel already [this plugin is probably a good option](https://www.npmjs.com/package/babel-plugin-object-assign) - even lighter than including the `object-assign` package. I'll add that in to the `react-component-gulp-tasks` package so we can use it in the future.

JedWatson:
I've added the object-assign plugin to `react-component-gulp-tasks` now, and published `react-select@0.4.7` without lodash.

Daniel15:
@JedWatson  Oh I didn't notice you're using Babel. If you enable [stage 1 (experimental)](https://babeljs.io/docs/usage/experimental/) stuff, you can use the spread operator:

```
var first = {
  hello: 'world',
};
var second = {
  ...first,
  something: 'else',
}
```

PR #141, opened 2015-04-22
Show "Searching..." instead of "no results found" while searching asynchronously

johnomalley:
"no results found" is displayed while an async search is running... this is misleading IMO.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -603,15 +603,15 @@ var Select = React.createClass({
 		if (ops.length) {
 			return ops;
 		} else {
-			var noResultsText = (function() {
-	            if (this.state.isLoading) {
-	                return this.props.searchingText;
-	            } else if (this.state.inputValue || !this.props.asyncOptions) {
-	                return this.props.noResultsText;
-	            } else {
-	                return this.props.searchPromptText;
-	            }
-	        }.bind(this))();
+			var noResultsText = (function () {
+				if (this.state.isLoading) {
+					return this.props.searchingText;
+				} else if (this.state.inputValue || !this.props.asyncOptions) {
+					return this.props.noResultsText;
+				} else {
+					return this.props.searchPromptText;
+				}
+			}.bind(this))();
 			return (
 				<div className="Select-noresults">
 					{noResultsText}
--- a/src/Select.js
+++ b/src/Select.js
@@ -23,6 +23,7 @@ var Select = React.createClass({
 		clearValueText: React.PropTypes.string,    // title for the "clear" control
 		clearAllText: React.PropTypes.string,      // title for the "clear" control when multi: true
 		searchable: React.PropTypes.bool,          // whether to enable searching feature or not
+		searchingText: React.PropTypes.string,     // message to display during an asynchronous search
 		searchPromptText: React.PropTypes.string,  // label to prompt for search input
 		name: React.PropTypes.string,              // field name, for hidden <input /> tag
 		onChange: React.PropTypes.func,            // onChange handler: function(newValue) {}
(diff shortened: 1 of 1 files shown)


src/Select.js
```diff
@@ -598,11 +600,24 @@ var Select = React.createClass({
...
+			var noResultsText = (function() {
+	            if (this.state.isLoading) {
+	                return this.props.searchingText;
+	            } else if (this.state.inputValue || !this.props.asyncOptions) {
+	                return this.props.noResultsText;
+	            } else {
+	                return this.props.searchPromptText;
+	            }
+	        }.bind(this))();
+			return (
```
Daniel15: Indentation is messed up, this line uses tabs but the line before it uses spaces.

johnomalley:
I don't use tab characters in my editor... there did seem to be some tab characters in that section so I did a reformat.  It's looking better but the indent is bigger than expected in the diff... i'm not sure why.  It shows up as four spaces in my editor.

dcousens:
@johnomalley can you please squash the commits?

src/Select.js
```diff
@@ -598,11 +600,24 @@ var Select = React.createClass({
...
+		} else {
+			var noResultsText = (function () {
+				if (this.state.isLoading) {
+					return this.props.searchingText;
+				} else if (this.state.inputValue || !this.props.asyncOptions) {
+					return this.props.noResultsText;
+				} else {
+					return this.props.searchPromptText;
+				}
+			}.bind(this))();
```
dcousens: nit: Can you use `var this = self` instead of `.bind(this)`
Happy to merge once you squash though.

src/Select.js
```diff
@@ -598,11 +600,24 @@ var Select = React.createClass({
...
 
-		return ops.length ? ops : (
-			<div className="Select-noresults">
-				{this.props.asyncOptions && !this.state.inputValue ? this.props.searchPromptText : this.props.noResultsText}
-			</div>
-		);
+		if (ops.length) {
+			return ops;
+		} else {
+			var noResultsText = (function () {
```
dcousens: Why make a closure for this at all?

src/Select.js
```diff
@@ -598,11 +600,24 @@ var Select = React.createClass({
 
 		}, this);
 
-		return ops.length ? ops : (
-			<div className="Select-noresults">
-				{this.props.asyncOptions && !this.state.inputValue ? this.props.searchPromptText : this.props.noResultsText}
-			</div>
-		);
+		if (ops.length) {
+			return ops;
```
dcousens:
```
if (ops.length) return ops

// continue as normal, no else
```

Don't bother with the `else` if you have an early exit `return`.
It is implied.

dcousens:
Happy to merge after squash and a fix up of the nits,  will otherwise merge in 24hrs and fix up myself.

johnomalley:
@dcousens why don't you just reject this request and implement it however you want.  I think your quibbles amount to a difference of style... i personally think code is more readable, for example, if you explicitly use if/else rather than relying on fall-throughs.  The self-executing anonymous block is a consequence of using languages that support richer expressions than JavaScript.  I thought the expression was getting too complex to embed in the JSX so I extracted it into a function.

But it doesn't matter... if you're that particular do it however you want.  You won't offend me.

dcousens:
@johnomalley fair enough.

@JedWatson feel free to rebase/merge this.
Otherwise close.

I think the feature is worth while though.

bruderstein:
I've rebased this under #379, but left the code pretty much as is, as I think @johnomalley is right, and it does make it more readable (just removed the IIFE, as I can't see that has any advantage, feel free to show me where I'm missing something).

Closing this so we can move to #379

PR #151, opened 2015-04-29
Adds tag creator

flovilmart:
Adds allowCreate=true prop so the field will allow option creation.
By default this option is disable to maintain backward compatibility

### DIFF
--- a/examples/src/app.js
+++ b/examples/src/app.js
@@ -169,6 +169,39 @@ var SelectedValuesField = React.createClass({
 	}
 });
 
+var SelectedValuesFieldCreate = React.createClass({
+
+	onLabelClick: function (data, event) {
+		console.log(data, event);
+	},
+
+	render: function() {
+		var ops = [
+			{ label: 'Chocolate', value: 'chocolate' },
+			{ label: 'Vanilla', value: 'vanilla' },
+			{ label: 'Strawberry', value: 'strawberry' },
+			{ label: 'Caramel', value: 'caramel' },
+			{ label: 'Cookies and Cream', value: 'cookiescream' },
+			{ label: 'Peppermint', value: 'peppermint' }
+		];
+		return (
+			<div>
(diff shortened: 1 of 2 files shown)


dcousens:
Hi @flovilmart, could you please update this PR without the dist files?
I think this is a cool addition in any case :)

flovilmart:
Ok. I'll do my best :)

flovilmart:
Rebased and removed the commits from the dist and lib

bsr203:
+1

beeant:
+1

flovilmart:
I've updated, and rebased

dcousens:
Thanks @flovilmart this is great.

dcousens:
@flovilmart should a callback be fired to notify of a new tag?

flovilmart:
Not that implémentation, just the on change is fired. That would be a nice feature (but I didn't need it at the time)

dcousens:
@flovilmart thoughts about making a PR for it? :)

flovilmart:
Haha :) you could do it too :p what callback signature would you like? 

Would you want if before or after the change?

Do you want it cancellable?

dcousens:
@flovilmart haha, before the state change is always better, that way the user can throw new props at it if necessary.

> Do you want it cancellable?
That is a good question,  if it was structured properly, we'd handle this like an `onChange` event for a typical field and manage the state in the parent,  but in this case,  the state is being managed internally...

flovilmart:
As for thé signature, I could go with the same as onChange.

Call it onAdd?

dcousens:
@flovilmart sounds good to me

flovilmart:
Just created another PR, maybe we should rename allowCreate to enableAdd, or canAdd or allowAdd, or addEnabled...

dcousens:
`canAdd` works for me.

nickbouton:
+1

PR #154, opened 2015-04-29
fix scope error

Yiner:
bind this when add and remove _closeMenuIfClickedOutside


dcousens:
Already fixed in https://github.com/JedWatson/react-select/commit/53beeaa9591e34951c71ed09514157e4332bafed

dcousens:
Thanks for the contribution though @Yiner

Issue #158, opened 2015-04-30
TypeError: this.getInputNode(...).focus is not a function


bannaN:
You get TypeError: this.getInputNode(...).focus is not a function when _focusAfterUpdate is true and the element is disabled. You cannot call focus on a disabled input element.

Inside componentDidUpdate

JedWatson:
Closed by your PR #159

PR #159, opened 2015-04-30
Doesnt try to set focus afterupdate if component is disabled

bannaN:
A attempt on a fix for the small issue i registered (#158).

If there are other issues with this fix please let me know and i will try to sort them out :)


JedWatson:
This is a good change, thanks @bannaN

PR #160, opened 2015-04-30
Lodash is no longer a dependency!

chrisvariety:
:star2:


JedWatson:
Thanks @chrisvariety :smile:

PR #162, opened 2015-05-03
Hitting Esc creates empty label if there are no previous options selected

### LINKED ISSUE #161: Hitting Escape when typing adds empty label
Hi,
This happens only when there are no previously added labels. SetValue inside resetValue function is called with empty string as this.state.value

rgrzelak:
Fixes #161


JedWatson:
:+1: Thanks @rgrzelak, good fix.

PR #163, opened 2015-05-04
Add Chinese pinyin support, exports as react-select/lib/SelectCN

xcatliu:
Hi, I added a support for Chinese pinyin search, exports as an other component `react-select/lib/SelectCN`, so it would not break your main component.

Please have a look, and see if it is able to be merged.

Thanks :)


xcatliu:
I have published this as another component, [react-select-pinyin](https://github.com/xcatliu/react-select-pinyin).
Because of I think this is a better way to use a select control with pinyin support.
Have fun with it :)

PR #177, opened 2015-05-08
Use react/lib/EventListener instead of document.addEventListener

xcatliu:
`document.addEventListener` doesn't work on IE8.

Use `react/lib/EventListener` instead of `document.addEventListener` will solve this problem.

### DIFF
--- a/lib/Select.js
+++ b/lib/Select.js
@@ -5,7 +5,8 @@ var _extends = Object.assign || function (target) { for (var i = 1; i < argument
 var React = require('react'),
     Input = require('react-input-autosize'),
     classes = require('classnames'),
-    Value = require('./Value');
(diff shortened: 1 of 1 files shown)


dcousens:
@xcatliu please see the CONTRIBUTOR file,  you should only edit files in `src`.  Not `lib`.

JedWatson:
The React team has made it clear that using React internals in other packages is not safe; as tempting as it is to reach in and access these utils, we should find a better way.

dcousens:
Closing,  will open a relevant issue though.  @xcatliu please re-open if you rebase and submit a new/different fix.

xcatliu:
@dcousens @JedWatson 
Ok, I'll find a better way. Thanks

Issue #175, opened 2015-05-07
Select uses addEventListener/removeEventListener, which don't exist in IE8


StefanBilliet:
Hi

In Select.js->componentWillMount(), you add/remove event listeners for when clicking outside the element. Apparently, IE8 doesn't support addEventListener/removeEventListener; it only supports attachEvent/detachEvent (the name of the click event is also different).
It's fairly simple to fix: https://github.com/StefanBilliet/react-select/commit/f0b004833b9f5a7545773425f36d28e4eea84ef5
My question is, are you open to these kinds of fixes?

With kind regards
Stefan Billiet

dcousens:
@StefanBilliet we are open to these kind of fixes.   Please make a PR and we'll put it through the review cycle :)
Before making the PR, please read the CONTRIBUTORS file.

julen:
Hey, this was fixed in #193, and it should be closed.

dcousens:
Thanks @julen

PR #183, opened 2015-05-12
Stop using addEventListener/removeEventListener for IE8 compatibility

StefanBilliet:
Hi

As discussed in [issue 175](https://github.com/JedWatson/react-select/issues/175), here is the pull request.
npm test, npm run lint and npm run style all pass, so I hope everything is in order.

With kind regards
Stefan Billiet


dcousens:
Please squash to 1 commit.

examples/src/index.html
```diff
@@ -37,7 +37,9 @@
 		<div class="footer">
 			Copyright &copy; Jed Watson 2014. MIT Licensed.
 		</div>
-	</div>
+	</div>	
+	<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/es5-shim/4.1.1/es5-shim.js"></script>
```
dcousens: Are these necessary?

StefanBilliet:
Without the shim, the example wouldn't run under IE8, so I couldn't verify wether the fix worked.

examples/src/index.html
StefanBilliet: Without the shim, the example wouldn't run under IE8, so I couldn't verify wether the fix worked.

dcousens: If we were going to use a shim,  why not just use what we had before?

dcousens:
@StefanBilliet whats the conclusion?

StefanBilliet:
That I was doing something very stupid with my branches; trying to rectify the situation -_-

examples/src/index.html
StefanBilliet: because the shim doesn't abstract away the difference in api of attaching an event handler between IE8 and all other browsers

StefanBilliet:
I created a new PR here: https://github.com/JedWatson/react-select/pull/193

PR #193, opened 2015-05-18
Fix for issue 175

StefanBilliet:
Hi

I mistakenly deleted the wrong branch in my repo, so it took some time to get everything back.
I created a new branch with the one squashed commit; sorry for my blundering.

With kind regards
Stefan Billiet

### DIFF
--- a/examples/src/index.html
+++ b/examples/src/index.html
@@ -37,7 +37,9 @@
 		<div class="footer">
 			Copyright &copy; Jed Watson 2014. MIT Licensed.
 		</div>
-	</div>
+	</div>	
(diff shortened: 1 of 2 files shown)


StefanBilliet:
The reason I added the shim to the example was because there was an error about "Array.isArray is undefined" under IE8. I don't think such things need to be "fixed" for IE8 support, since you can easily use a shim for that.
However, in IE8, binding a click eventhandler happens differently compared to other browsers. The function is different (attachEvent() <-> addEventListener()), but the eventnames differ also ("onclick" <-> "click).
I'm not aware of a shim that could fix that.

dcousens:
ACK.  @JedWatson please merge

JedWatson:
This is great, thanks @StefanBilliet

PR #198, opened 2015-05-19
enable consumers to require less files programmatically

alanclarke:
This pr unobtrusively enables consumers of this module to programmatically require this module's less files e.g. react-select/less/default.less using browserify

@JedWatson @giuse88


dcousens:
@alanclarke I'm not so sure this is unobtrusively,  this forces users to have `lessify` as a dependency.

alanclarke:
Well yes, but as a devDependency, which is needed to make the less file requireable with browserify. By unobtrusive I meant that makes the module more consumable without affecting the build output

Issue #108, opened 2015-03-25
Pushing enter without dropdown open adds a value


alex1712:
When I push enter being in the input field, but without writing any character, the first available option is added.

Thank you very much for your work :)

JedWatson:
@alex1712 this is the behaviour as designed, because by default the first option is focused when you open the control. You can use the keyboard to navigate up and down between different options when it is open.

Thanks :)

alex1712:
Hi again,

thank you very much for your quick answer.

I understand this behaviour when an option is selected but sometimes happens also when the options are not shown. 

Example from your demo page (http://jedwatson.github.io/react-select/) would be:

1- Click on the 'CLICKABLE LABELS (LABELS AS LINKS):' selector
2- Remove one of the labels with the keyboard
3- Then the options are hidden
4- Press enter and one option it is going to be added.

I am using the last version of chrome in MacOs

Thank you again :)

JedWatson:
I see - that's not as expected, thanks for the clarification!

I'll look into it.

PR #204, opened 2015-05-22
Prevent add  the last focused value when the drop down menu is closed.

giuse88:
Hi, 

Thank you very much for developing this nice module. I found a small bug when a user presses `ENTER` and the drop-down is closed. Basically, the module remembers the last selected value, and when a user presses `ENTER` this value is added. I think, this is incorrect behaviour as a user might accidentally add a value to the list of select values. 

This is the behaviour I am fixing ( I ma just pressing enter ) :
spxyk2ojvl

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -373,7 +373,9 @@ var Select = React.createClass({
 			break;
 
 			case 13: // enter
-				this.selectFocusedOption();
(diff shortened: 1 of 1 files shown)


dcousens:
Fixes https://github.com/JedWatson/react-select/issues/108

PR #197, opened 2015-05-19
Adds onAdd handler

flovilmart:
As per discussion on #151

PR #205, opened 2015-05-22
Enable add custom value when allowCreate is true

giuse88:
Hi, 

This PR enables the add custom field feature despite the number of the filter options. In addition, it moves the `Add new value?` as first option in case the list of possible options is big. It mimics the behaviour of selectize.js. 

This GIF shows the new behaviour:
znlsaffc6v

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -613,13 +613,13 @@ var Select = React.createClass({
 			focusedValue = focusedValue == null ? this.state.filteredOptions[0] : focusedValue;
 		}
 		// Add the current value to the filtered options in last resort
-		if (this.props.allowCreate && !this.state.filteredOptions.length) {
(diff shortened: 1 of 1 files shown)


dcousens:
Thanks @giuse88,  this is good :)

dcousens:
@giuse88 please be aware that `allowCreate` is going to change to `canAdd` (see https://github.com/JedWatson/react-select/pull/205)

giuse88:
@dcousens Thank you for merging it. Yeah, I will update my code to use `canAdd` when the version is out.

JedWatson:
I've renamed the prop to `canAddOptions`, which I think is clearest, while preparing to release `0.5.0`.

JedWatson:
Nevermind, just read the subsequent discussion in #197, reverted my change.

PR #206, opened 2015-05-22
Callback context is undefined. 

giuse88:
Hi, 

Diving into your code, I found another small bug. When options are loaded asynchronously, the context of the callback is not correctly set. Therefore, when `this._closeMenuIfClickedOutside` is called within a callback function ( for example, when a user types something) a javascript error is thrown.


dcousens:
Thanks @giuse88.

PR #209, opened 2015-05-26
#201

### LINKED ISSUE #201: Undefined values when using asyncOptions
First off, your component seems great. However, I did quickly run into the following issue.
The error only seems to occur when using the asyncOptions. I tried the regular [Options] and everything worked.

It looks like the values are null and this.state.values are never set, but checked in the filterOptions function. Even when passing in values thru the props, there may need to be a this.setState(... for values. The offending lines are on Select.js on starting on line 507.

``` javascript
	filterOptions: function(options, values) {
		if (!this.props.searchable) {
			return options;
		}

		var filterValue = this._optionsFilterString;
		var exclude = (values || this.state.values).map(function(i) {
			return i.value;
		});
           ...
```
The exclude function blows up when both values and this.state.values are null or undefined. It gets called on ComponentDidMount, and that is fine, but then again when loadAsync, which causes the failure.

I was able to work around the issue by extending Select component like so, but thought you may want to address the root issue.

``` javascript
class SelectEx extends Select {
  constructor(props) {
    super(props);
    this.state.values = [];
  }
}
```
Lastly, I also noticed  a couple of spots where you are pointing to how values should be populated.

``` javascript
getInitialState: function() {
		return {
			/*
			 * set by getStateFromValue on componentWillMount:
			 * - value
			 * - values
			 * - filteredOptions
			 * - inputValue
			 * - placeholder
			 * - focusedOption
			*/
			options: this.props.options,
			isFocused: false,
			isOpen: false,
			isLoading: false
		};
	},

getStateFromValue: function(value, options) {
		if (!options) {
			options = this.state.options;
		}

		// reset internal filter string
		this._optionsFilterString = '';

		var values = this.initValuesArray(value, options),
			filteredOptions = this.filterOptions(options, values);

		return {
			value: values.map(function(v) { return v.value; }).join(this.props.delimiter),
			values: values,
			inputValue: '',
			placeholder: !this.props.multi && values.length ? values[0].label : this.props.placeholder,
			focusedOption: !this.props.multi && values.length ? values[0] : filteredOptions[0]
		};
	},
```
Like I said, the code assumes values will be set downstream, but it is not.

Hope that helps, glad to provide further feedback if it helps.

bannaN:
Seems to fix #201. Might be better approaches though

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -130,15 +130,12 @@ var Select = React.createClass({
 			}
 		};
     
-		this.setState(
-        this.getStateFromValue(this.props.value),
-        function(){
-          //Executes after state change is done. Fixes issue #201
-		      if (this.props.asyncOptions && this.props.autoload) {
-      			this.autoloadAsyncOptions();
-      		}
-        }
-    );
+		this.setState(this.getStateFromValue(this.props.value),function(){
+		  //Executes after state change is done. Fixes issue #201
+		  if (this.props.asyncOptions && this.props.autoload) {
+				this.autoloadAsyncOptions();
+			}
+    });
 	},
 
 	componentWillUnmount: function() {
--- a/src/Select.js
+++ b/src/Select.js
@@ -93,11 +93,7 @@ var Select = React.createClass({
 	componentWillMount: function() {
 		this._optionsCache = {};
 		this._optionsFilterString = '';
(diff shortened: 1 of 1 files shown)


dcousens:
This should be fine.  Thanks @bannaN

PR #210, opened 2015-05-26
enable consumers to require less files programmatically

alanclarke:
moved from https://github.com/JedWatson/react-select/pull/198 to custom branch

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,11 +20,17 @@
     "eslint-plugin-react": "^2.2.0",
     "gulp": "^3.8.11",
     "jest-cli": "^0.4.2",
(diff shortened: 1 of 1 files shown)


dcousens:
Thanks @alanclarke

dcousens:
@alanclarke can you please add an example here as to how this could work for users?

dcousens:
@JedWatson please double verify,  can still revert if necessary.

alanclarke:
heya @JedWatson, thank you for merging! :+1: 

[Here](https://github.com/alanclarke/lessify-react-select-example) is a working example of how this can be used.

Issue #196, opened 2015-05-19
It seems like setting matchPos="start" suddenly makes the component only recognize lower-case inputs


mhintz:
Great component! Really liking it, above all the other options out there. I found one thing which seems like it's a problem, but I'm not sure. 

It looks like when I set the matchPos="start" attribute, then the component only matches lower-case inputs (even when the labels start with upper case letters). I'm pretty sure it has something to do with the way line 519 works, but that's such a monster boolean statement that I don't really feel qualified to tackle it. Hopefully someone with more knowledge of the component can take a look.

matchPos="start"

input:
"Somevalue"
expected:
<suggested> "Somevalue"
got:
<suggested> (none)

input:
"somevalue"
got:
<suggested> "Somevalue"

Thanks much!

dcousens:
Fixed in #215

PR #215, opened 2015-05-28
Support case-insensitive filtering when matchPos="start"

wesrage:
Fixes Issue #196
Consistent behavior with default (`matchPos="any"`)


dcousens:
Perhaps this should be an option

wesrage:
I think that makes a lot of sense, with `true` as default.

dcousens:
@wesrage care to add it?

wesrage:
@dcousens Sure.

dcousens:
Thanks @wesrage

PR #218, opened 2015-05-29
add autoreload config, stopPropagationKeyDown config and fixed issue #201

snadn:
add reload async options after select.
stopPropagationKeyDown config.
in componentWillMount. the callback of setstate will not be called. so the old fix for #201 is Invalid.
let Select-noresults hide if the content is empty.


dcousens:
@snadn please separate these features into different commits, and, ideally if possible, into different pull requests.

snadn:
i know

dcousens:
Thanks @snadn :)

PR #219, opened 2015-05-29
Fixes issue #201

snadn:
in componentWillMount. the callback of setstate will not be called. so the old fix for #201 is Invalid.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -130,12 +130,14 @@ var Select = React.createClass({
 			}
 		};
     
-		this.setState(this.getStateFromValue(this.props.value),function(){
-		  //Executes after state change is done. Fixes issue #201
-		  if (this.props.asyncOptions && this.props.autoload) {
-				this.autoloadAsyncOptions();
-			}
-    });
+		this.setState(this.getStateFromValue(this.props.value));
+	},
+
+	componentDidUpdate: function() {
+		//Executes after state change is done. Fixes issue #201
+		if (this.props.asyncOptions && this.props.autoload) {
+			this.autoloadAsyncOptions();
+		}
 	},
 
 	componentWillUnmount: function() {


dcousens:
@bannaN can you please confirm this?

bannaN:
Sorry for the late response. Well.. I am not 100 percent sure... What i can say is that.

- If you dont have the autoloadAsyncOptions call in the componentWillMount method it will not be called for the initial render since componentDidUpdate is not called for the initial render

- Seems to me that this means that the autoloading of the options for the default value ''(emptry string) will not actually be kicked off when the component is mounted, but when it does its first update. Probably this happends relativly quickly, depending on the code. I havent gone through everything. 

- Where is it documented that the setState callback is not called? Seemed to work for me when i tried it out. React can squash together state updates for performance gains, but the setState callback should be called?

If you want i can try to look more into the issue tomorrow (CET)

snadn:
First of all, I'm sorry, because my English is poor, may not well understand what you mean.

in my project.the result of test is this:
qq 20150603091419

in document of react I find this:
image

so, i guess the callback in setState before first render will not be called.

bannaN:
Hi,
From the test it seems like you are correct... I always thought that it would be executed after the render, but i was obviously mistaken. It should be pointed out in the react docs

dcousens:
Interesting find!

I wasn't aware of this either,  I wonder if it is a bug?
@JedWatson thoughts?

bannaN:
Hm, regarding the patch.. Instead of moving the logic to componentDidUpdate
		if (this.props.asyncOptions && this.props.autoload) {
			this.autoloadAsyncOptions();
		}
wouldnt it be better to move it to a componentDidMount method? Or will it cause other issues?

agirton:

PR #220, opened 2015-05-29
add autoreload props to load async options after select.

snadn:


dcousens:
@bruderstein / @JedWatson can you confirm if this PR is still relevant?

bruderstein:
I *think* you can do this with `cacheAsyncResults=false` now, but I'm not sure what the original intention of this was - maybe @snadn can comment?

dcousens:
Going to close,  if @snadn can clarify the original intention and if necessary,  we can easily re-open.

snadn:
Now, this feature is not necessary for me already. 

At first I get asynchronous data when each input.
enable multi select. if list is ['ab', 'ac', 'cd']. input 'a'. list is ['ab', 'ac']. select 'ab' item. focus input, the list is ['ab', 'ac'] still and I can only see 'ac' item, the 'cd' is not in the list now.

Now, I get list once before react-select mount, and let react-select do the filter. But some still only suitable for sophisticated suggestion made in the service side.

with google translate.

bruderstein:
I think that's a bug, that should always refilter. I'll write a test or two
to check, but I don't think we need the prop, it should just always do it.

dcousens:
@bruderstein shall I re-open the PR or did you want to just create a new issue?

bruderstein:
I'll open an issue when I've tested that it's actually a bug.

PR #221, opened 2015-05-29
do not return Select-noresults if noresult content is empty.

snadn:


dcousens:
Thanks @snadn,

dcousens:
@bruderstein is this still worth merging/rebasing?

bruderstein:
I was going to say I'm not sure what the use case is, as you could always add a `.Select-noresults { display: none; }`, to your CSS, but we actually use the class for wrapping the `Searching...` method in #141 / #379, so that wouldn't work with them both. I'm inclined to fix #379, such that this is taken into consideration, but rather fix it by not abusing the `.Select-noresults` class, and use something else for that.

If there's a genuine reason to not render the div if it's empty (that can't be solved if necessary from user-css), then why not.  

However, I would expect the `noResultsText` being empty, and not being displayed is probably a rare use case?

bruderstein:
I'll leave this open until the fate of #379 is decided.  If #379 gets merged, I'd be inclined to close this, as I think you can achieve everything with that.

snadn:
when `noResultsText` is set ,i get 
image

but if i not need `noResultsText` and set it to '', i get 
image

bruderstein:
Ah, thank you, now I see the use case. This should probably go in, but after #379 has been accepted/rejected, as this will conflict with that.

agirton:

Issue #224, opened 2015-06-01
Add happiness


dcousens:
We need to add https://github.com/JedWatson/happiness to this repository.

Pull requests welcome.

dcousens:
Closing in favour of #373

PR #225, opened 2015-06-01
Update eslint and cleanup warnings.

AlexKVal:
Update eslint dev dependency.
Clean up eslint warnings.
And impose new "object-curly-spacing" rule.


dcousens:
THanks @AlexKVal,  note we will be switching to adding [happiness](https://github.com/JedWatson/happiness) soon.  

See https://github.com/JedWatson/react-select/issues/224 if you want to help out.

AlexKVal:
Good idea. 
It is best to maintain one set of rules in one place than maintain rule sets for every project :+1:

PR #226, opened 2015-06-01
allow key to be set for options

johan-olsson:
The idea is to allow options to have the same value without weird workarounds.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -651,9 +651,9 @@ var Select = React.createClass({
 			var mouseDown = this.selectValue.bind(this, op);
 
 			if (op.disabled) {
-				return <div ref={ref} key={'option-' + op.value} className={optionClass}>{op.label}</div>;
+				return <div ref={ref} key={'option-' + op.key || op.value} className={optionClass}>{op.label}</div>;
 			} else {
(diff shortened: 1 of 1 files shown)


dcousens:
@johan-olsson can you please rebase on master?  This is already out of date it would seem.

johan-olsson:
@dcousens rebased it now.

JedWatson:
I'm wondering if we should be introducing a way to customise the key used for `value` completely, rather than defaulting it to either `key` or `value`? could also be `id`, etc...

johan-olsson:
These changes should make it default to `value` if no `key` is set. The problem I had in the project I'm using react-select was multiple `options` with the same `value` but different `labels`, as aliases if you will. I can't think of any case where one would like `options` with duplicate `values` to just get removed by react.

Another solution would be to always generate a unique `key` for each `option` but that feels a bit heavy handed with the slim amount of cases it would be useful in.

AlexKVal:
why not ?
```js
key={op.key || ('option-' + op.value)}
```
`op.key` is the key by itself, it doesn't need that `option-` part.

johan-olsson:
Good point @AlexKVal. That would probably be true when setting the key from the value as well in that case.

dcousens:
Rebase?

PR #223, opened 2015-05-30
adding support for customizable background color

JohnMorales:
Hey Jed, 

Great work on this, I was able to fully skin the control to match the look and feel of my design. I'm using a gulp task that overwrites your variables in the select.less and creates a custom stylesheet for my project. 

The only variable that was missing was the background-color. Now it looks good even on dark backgrounds..
screen shot 2015-05-30 at 6 00 06 pm

Figured it might be useful for the next guy.

PR #230, opened 2015-06-02
Fixed syntax error in select.less

brianreavis:
Introduced in #223


AlexKVal:
Nice catch :+1:

dcousens:
Thanks @brianreavis

PR #231, opened 2015-06-02
Added ability to set props on hidden input

brianreavis:
This allows for properties to be set on the hidden `<input>` that holds the value of the control. One use is [Stripe](https://stripe.com/) integration ([guide](https://stripe.com/docs/tutorials/forms)), which requires a `data-stripe` property to be set. 

**Note:** this is a breaking change. I renamed the previous `inputProps` prop to `inputTextProps`. I think this is a little clearer (I originally thought `inputProps` was what I was looking for).


JedWatson:
For clarity, planning to merge this as soon as we're ready to bump a breaking change release.

agirton:
Closing as this is no longer applicable. We allow `inputProps`

Issue #211, opened 2015-05-27
Close dropdown when clicking on select


MaaikeB:
right now the dropdown only closes when clicking somewhere outside of the select-control.
so: the user opens the dropdown by clicking on the select element, but has to move the mouse outside the element in order to close it (if he doesn't want to select any of the options)

it would be nice to add functionality so the dropdown opens when clicking on the select-control, and closes when clicking on it while the dropdown is open

nkbt:
Clicking on the select itself should not close menu (only focus cursor inside the input), here is a solution - click on arrow should close menu.

JedWatson:
I agree with @nkbt, that seems like the correct behaviour. Otherwise placing the cursor at a different point in the text (with the mouse) would also close the menu, which would be unexpected.

If there's strong disagreement let me know and we can reopen the issue for discussion.

raulmatei:
Hi guys, I feel I should add my input on this too. My opinion is that for non-searchable version it should have the HTML `<select/>` behavior, which is closing the dropdown when user clicks on the select-control and probably for the searchable version, it should close when clicking on the arrow.

nkbt:
@raulmatei that totally makes sense. I am going to update component to support this behavior.

brianreavis:
> for the searchable version, it should close when clicking on the arrow.

On this I disagree, for the the same reason that clicking a textbox shouldn't remove its focus.

MaaikeB:
created PR #243  with suggested fix for this issue
for searchable select elements the feature will not apply.

raulmatei:
@brianreavis - I understand your point, but isn't the arrow indicating that you can hide/close the dropdown by clicking on it (without losing the focus on the select-control input element)?

nkbt:
That seems like a usual behavior of `<select>`. if you click on the input itself (not only the arrow) - it will toggle popout. So this additional PR looks fair.

JedWatson:
Sorry I left this for so long - fell completely off my radar. I agree this behaviour makes sense when the select is not searchable, so I've merged #243 and will release a patch soon.

PR #241, opened 2015-06-10
 Close dropdown when clicking on select #211

nkbt:
Clicking on the select itself should not close menu (only focus cursor inside the input), here is a solution - click on arrow should close menu.

The only issue is - arrow is too small, so invisible `.Select-arrow-zone` element was added under it. You don't have to be a sniper to click on the arrow anymore.

Clicking on `.Select-clear` did not prevent default action and because it goes slightly over arrow, menu was also closed, so I added `stopPropagation()` and `preventDefault` for it.

### DIFF
--- a/less/control.less
+++ b/less/control.less
@@ -149,6 +149,17 @@
 
 // arrow indicator
 
+.Select-arrow-zone {
+	content: " ";
+	display: block;
+	position: absolute;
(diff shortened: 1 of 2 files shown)


JedWatson:
Looks good @nkbt, thanks!

PR #243, opened 2015-06-10
proposed fix #211 - close dropdown when clicking on select

### LINKED ISSUE #211: Close dropdown when clicking on select

MaaikeB:


nickbouton:
+1, would love to see this merged in.

anton-kabysh:
+1, expect the same functionality.

borisd:
+1, please merge, this is how select should work

JedWatson:
I'm on board. Merging.

JedWatson:
This has broken several tests, sorry @bruderstein!

```
Select with multi=true and searchable=false selects multiple options
Select with multi=true and searchable=false calls onChange when each option is selected
Select with multi=true and searchable=false removes the selected options from the menu
```

Not sure at a quick glance between the PR and the tests why these now fail, I'll have to look into it more closely. Will come back to it soon.

bruderstein:
I'll take a look. If it's an easy/obvious fix, am I OK to simply commit it?

JedWatson:
Thanks - direct commit is fine, do tag this issue so we can see what you changed though, I'd love to know what the fix is :)

bruderstein:
Guessing because the tests assume the menu is still open after the first
selection. Should be trivial to fix (just need to open the menu after each
selection)

bruderstein:
....aaaaaand we're green :smile:

Issue #244, opened 2015-06-10
Scroll on touch screen on top of a select will open it.


YoYuUm:
If you scroll on touch screen on top of a select, this will open itself when the scroll ends. 

I would say the expected behaviour is that this happens only on click.

jljorgenson18:
+1 , this behavior is fairly annoying on mobile

azaharakis:
I've had a play with this here:
https://github.com/azaharakis/react-select/commit/b4a04c750b53cc8edacd22cc5e35cdd4fc63712a
The commit message states where i'm at

PR #245, opened 2015-06-10
Use onClick event instead of onTouchEnd (refs #244)

YoYuUm:
refs #244

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -771,7 +771,7 @@ var Select = React.createClass({
 		return (
 			<div ref="wrapper" className={selectClass}>
 				<input type="hidden" ref="value" name={this.props.name} value={this.state.value} disabled={this.props.disabled} />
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @YoYuUm, this will lead to always double-calling `onMouseDown` on 
desktops though (once on down, once on click...)

It will also introduce a delay on mobile which is not desirable.

I think a better solution might be to use either [react-tappable](https://github.com/JedWatson/react-tappable) or some of the logic from it to fix this.

azaharakis:
Hi Jed, I'd really like to fix this problem and happy to attempt a PR, with some guidance on how you would like it solved. Have you given any thought since on this issue? 
I think you are right in that the delay is not desirable, however could it not be solved with libraries that are designed to address this specific issue i.e fastclick (https://github.com/ftlabs/fastclick). 

It seems more appropriate to just rely on onClick to apply the menu functionality, than attempting to handle both use cases with the buggy behaviour that occurs on mobile devices? Or at least until the issue can be resolved within this code properly.

agirton:

PR #251, opened 2015-06-16
Get tests passing on node again

craigdallimore:
Hello!

This change will lower the JSDOM version so tests will run on node again.

It turns out [JSDOM from 4.x and above will not work on node.](https://github.com/tmpvar/jsdom/blob/master/Changelog.md#400)

If _node_ is not intended to be supported we should probably put a note in the contributions documentation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: node_js
+node_js:
+    - "0.10"
+    - "0.11"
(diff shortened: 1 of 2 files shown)


JedWatson:
I would say that node is *definitely* intended to be supported. Thanks for getting to the bottom of this @craigdallimore.

PR #253, opened 2015-06-17
Resolve #252 by adding travis configuration

### LINKED ISSUE #252: Add continuous integration
Hey @JedWatson and team

Would you be amenable if we add in some config for a CI service? I'm presently using [Travis-CI](https://travis-ci.org/craigdallimore/react-select/builds/67097990) on a fork to run the tests on a few versions of node - and iojs - once I've pushed. It will send me an email if the tests fail, so I'm less likely to miss broken tests.

If you follow that link you might see a branch failing and a branch passing - the difference being the presence of a `.travis.yml` [file](https://github.com/craigdallimore/react-select/blob/master/.travis.yml) that configures the service.

There are a couple of services out there - I only have experience with Travis, but I see [Codeship](https://codeship.com/pricing) has a free tier as well. No doubt there are others.

Your thoughts?

craigdallimore:
Relates to [Issue 252](https://github.com/JedWatson/react-select/issues/252)

Here is a travis-ci config for _react-select_. It will take an administrator to turn on the hook in [travis](https://travis-ci.org).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: node_js
+node_js:
+    - "0.10"
+    - "0.11"
+    - "0.12"
+    - "iojs"
+before_script:
+    - npm install


dcousens:
Thanks @craigdallimore

dcousens:
Ping @JedWatson

Issue #111, opened 2015-03-28
Disabled state


ondrejhlavacek:
When passing the disabled attribute with `true` value, the component with a value looks weird. 

![screen shot 2015-03-27 at 17 05 33](https://cloud.githubusercontent.com/assets/497675/6878893/9f361fc0-d4a3-11e4-86a0-71267f31294d.png)

ribeiroct:
I don't think react-select disabled state is meant for use as a readonly component.

I ran into the same trouble, which i solved through the following css hack on my stylesheet (may be useful for other users until it's fixed on the code).

        .Select.is-disabled .Select-item-icon,
        .Select.is-disabled .Select-placeholder,
        .Select.is-disabled .Select-input,
        .Select.is-disabled .Select-arrow{
            display: none;
        }

JedWatson:
@jossmac can you take a look at this? looks like something's gone out of sync since we added the disabled support.

dcousens:
@ondrejhlavacek can you confirm this still exists?

ondrejhlavacek:
@dcousens yes, still happening
![development_-_transformation_t2](https://cloud.githubusercontent.com/assets/497675/7874848/4650804c-056a-11e5-9cc1-87a10e82074b.png)

dcousens:
ping @jossmac

JedWatson:
Should be closed by #258

PR #258, opened 2015-06-20
Fix disabled state for is-multi [correct]

ondrejhlavacek:
As reported in #111.

I modified the example so the first multiselect contains a linked disabled checkbox. The disabled multiselect still keeps all values, but grays out the color and disables the remove action.

Had to commit a new build in the examples/dist as well. Please advise if was done wrong.


JedWatson:
Nice work @ondrejhlavacek, looks good to me.

Usually you only need to include the `src` and `less` changes in your PR, and I periodically add new builds, but it's fine how you've done it this time.

Thanks!

Issue #250, opened 2015-06-16
Allow clicking mid-input to modify input contents (have a patch ready)


bx98pjni9g88jnz:
Currently when typing in the input you can use the arrow keys to navigate around the string and make adjustments to the input.

It is currently not possible to click in the middle of some text in the input to make the same modifications (you have to use the arrow keys to get to where you want to make the modification).

This can be fixed in the `handleMouseDown` function with a `setTimeout 0`, e.g. (rough es5 / non es6 playing around with it modification)

``` javascript
// ... begin of fn
setTimeout(function() {
    if (this.state.isFocused) {
        this.setState({
            isOpen: true
        }, this._bindCloseMenuIfClickedOutside);
    } else {
        this._openAfterFocus = true;
        this.getInputNode().focus();
    }
}.bind(this), 0);
// ... end of fn
```

If this is acceptable I'll make a pull request with the proper changes following the guidelines.

JedWatson:
Thanks @trevorsenior, would definitely appreciate a PR for that!

karavaev-evgeniy:
+1

Firfi:
+1 
Please make PR appreciable again

PR #265, opened 2015-06-22
Allow clicking mid-input to modify input contents.

trevorsenior:
Related issue: #250 

---

Commit text:

```
Removed `event.stopPropagation()` & `event.preventDefault()` to allow
the default behavior of clicking in the middle of the input element
to edit text where the cursor clicked.

This then required a `setTimeout` to push the logic of opening the
select to the next tick.
```

---
- While running `gulp dev`, type some things in the multi select input, then click mid-way through the text to see that this change works & that the other inputs work as well.
- Linting passes
- Testing `Select.js` works as well. 

It should be noted that the `Value` test fails, but it also fails without any changes: 

```
> react-select@0.5.1 test /Users/trev/tmp/select/react-select
> mocha --compilers js:babel/register



  Select test
    ✓ should assign the given name

  Value component
Warning: Failed propType: Required prop `option` was not specified in `Value`.
    1) "before each" hook


  1 passing (640ms)
  1 failing

  1) Value component "before each" hook:
     TypeError: Cannot read property 'label' of undefined
      at [object Object].render (src/Value.js:27:32)
      at Context.<anonymous> (test/Value-test.js:29:21)
```

---

@JedWatson - Let me know if this needs any modifications before it can be merged in. Thanks!

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -311,16 +311,16 @@ var Select = React.createClass({
 			return;
 		}
 
-		event.stopPropagation();
-		event.preventDefault();
-		if (this.state.isFocused) {
-			this.setState({
-				isOpen: true
-			}, this._bindCloseMenuIfClickedOutside);
-		} else {
-			this._openAfterFocus = true;
-			this.getInputNode().focus();
-		}
+		setTimeout(function() {
+			if (this.state.isFocused) {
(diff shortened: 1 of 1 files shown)


agirton:

PR #269, opened 2015-06-23
Fix bug where Select shows the value instead of the label

sdemjanenko:
This happens if there is no matching option on initial render.  Future
updates to the options do not fix this condition.

### DIFF
--- a/lib/Select.js
+++ b/lib/Select.js
@@ -152,12 +152,13 @@ var Select = React.createClass({
 
 	componentWillReceiveProps: function componentWillReceiveProps(newProps) {
 		if (JSON.stringify(newProps.options) !== JSON.stringify(this.props.options)) {
+			var optionsChanged = true;
 			this.setState({
 				options: newProps.options,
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @sdemjanenko!

jketcham:
Hi, really appreciated this fix as I was also experiencing this bug, but it doesn't look like it's in master anymore? Did some other change overwrite this or something? Not in the current build. @JedWatson this commit 7c6223e530d44da2267dc2434ea7f56c5d236ada might've not included the most recent changes when it pushed directly to master.

stratease:
I'm using multi with async and I'm seeing this issue too. It eventually loads with the appropriate label, but it seems undesirable to display the value while it fetches the label data.

JedWatson:
I just realised what's going in here, I didn't pay close enough attention to the PR :disappointed: 

It updated `lib/Select.js` which is the _built_ version, not `src/Select.js`. I've just reapplied the fix and will push another update shortly.

PR #275, opened 2015-06-26
adds a callback option after creating a new value

walidvb:
Have read the contributing guidelines only after making the (tiny) change, apols for this.

Adds a callback option after creating a new value. 
My usecase is the following. Students belong to a Studio, and you can assign the studio in a table. If the studio doesn't exist, adding it in the field should make a POST request to create one on the server.

`npm run test` failed on the previous commit already, I haven't tried to fix it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,34 +140,35 @@ For multi-select inputs, when providing a custom `filterOptions` method, remembe
 ### Further options
 
 
-	Property			|	Type		|	Description
+	Property         | Type   | Description
 :-----------------------|:--------------|:--------------------------------
-	value 				|	any			|	 initial field value
-	multi 				|	bool		|	 multi-value input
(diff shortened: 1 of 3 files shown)


walidvb:
This actually needs to handle better the priority of `onChange` and `onCreateValue`. In my case, i need to make sure that the Studio is created before assigning the Studio. Maybe i should leave to the user to trigger `onChange`?

josefanostylus:
Hi, I have a similar case. Did you figure out how to do it?

walidvb:
No, i haven't looked at this in more details. Does my commit not help?

josefano:
Sorry , I didnt notice the commit. I figured a solution by just using "onNewOptionClick" event handler and passing a callback to my ajax request inside that  handler. Thanks though :)

PR #276, opened 2015-06-26
Add 'is-selected' class to the selected option

miraks:
Sometimes you may want to add additional styles to the selected option.


JedWatson:
Makes sense @miraks, thanks!

PR #277, opened 2015-06-27
Will it be better if the default async options set is auto-loaded with the default value?

Neson:
Just like where this PR has changed. I think it's useful to autoload the correct options set base on the initial value (if it exists).


JedWatson:
:+1: good call @Neson, thanks.

PR #281, opened 2015-06-29
Various tweaks and features

cgestes:


agirton:
Hi @cgestes since the project took a different route, these changes don't look like they are applicable any longer. Closing out.

PR #289, opened 2015-07-01
Allow entering of commas when allowCreate is on but multi is off.

adinardi:
The keydown handler for commas was introduced in #151, but it is preventing created values from having commas in them.

This allows commas to be included in created values as long as multi isn't enabled.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -418,7 +418,7 @@ var Select = React.createClass({
 			break;
 
 			case 188: // ,
-				if (this.props.allowCreate) {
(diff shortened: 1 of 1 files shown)


adinardi:
Is there anything I can do to help get this reviewed?

JedWatson:
Sorry for the delay @adinardi! Somehow missed the notification on this PR when it came in. Might actually be a better long-term solution to make the comma keydown handler option via a prop, but for now if this fix works for you I'll get it released.

Issue #291, opened 2015-07-02
Options having same key when trying to add options with the same values.


mannebusk:
When you are running create mode and create two options with the same value. React throws a warning "Encountered two children with the same key", and only renders the first one. The state still holds all typed options.

the4dpatrick:
+1 

Just ran into the same problem today

brianreavis:
What's the use case for allowing multiple copies of the same option?

pindia:
I get this even before creating the option, as soon as I mouse over the "Add {label}?" item. This is for an input with `allowCreate` true, but `multi` false.

mannebusk:
I don't think you should be able to add the same option twice. And if you wanted to add the same option multiple times, they would have to have an id to differentiate them even if they have the same label.

mannebusk:
@pindia I updated the pull request with a fix for that. The "Add {label}?" option had the same key as the "normal" options when you type in a option vlaue that already exists.

I also changed so that an option can't be added multiple times.

somlor:
Also having this issue. :+1:

randavidovitz:
I think #421, fixes it ?

jossmac:
Version 1 of react-select is no longer supported.

In the best interest of the community we've decided to spend the time we have available on the latest version.

We apologise for any inconvenience.

Please see:

1. [v1 to v2 upgrade guide](https://react-select.com/upgrade-guide)
1. [v1 documentation and examples](v1.react-select.com)

PR #292, opened 2015-07-02
Fixed key error on options

mannebusk:
When you are running create mode and create two options with the same value. React throws a warning "Encountered two children with the same key", and only renders the first one. The state still holds all typed options. This is a fix that gives each option a unique key, issue #291.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -565,7 +565,7 @@ var Select = React.createClass({
 	},
 
 	selectFocusedOption: function() {
-		if (this.props.allowCreate && !this.state.focusedOption) {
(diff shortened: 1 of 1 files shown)


mannebusk:
I have changed it so that you can't add the same value multiple times. I also fixed a bug where the "Add xx?" option had the same key as the normal options some times.

JedWatson:
Thanks @mannebusk, sorry this took so long to get back to. I've added a fix for duplicate option values to the new version. Adding duplicate values would now be handled at a higher level in your app, as Select is now a controlled component.

Issue #293, opened 2015-07-02
Abillity to style options and to have different colours on each option.


mannebusk:
It would be good to have a way to specify style for each option. Specially if you are using react-select for managing tags or labels.

gwyneplaine:
resolved in v2

PR #294, opened 2015-07-02
Adds abillity to set color on each option and general option style

mannebusk:
Issue here #293.


brianreavis:
This feels a little overkill...why not write CSS?

agirton:
Closing out as this can be done in user land.

PR #295, opened 2015-07-02
Add a property called backspaceRemoves

lehikol2:
The standard behavior has been that pressing backspace removes the last item if
there is no input value present. This property allows the user to disable this
behavior by setting backspaceRemoves to false. Default behavior remains the
same, i.e. backspace removes an item.


JedWatson:
:+1: looks good to me, thanks!

pragya04:
Is there a method where I can capture the backspaceRemoves event?

PR #302, opened 2015-07-05
Reset height and box-shadow

Fauntleroy:
While using react-select in a website that uses Foundation, I ran into a problem where the hidden input would be absurdly tall:

screen shot 2015-07-05 at 12 05 51 pm

This is because Foundation modifies the height of all `input` elements. I simply ensured that react-select had its own height set, and the situation is now resolved:

screen shot 2015-07-05 at 12 06 08 pm

**Addendum**: I also removed and extra padding declaration and added a reset for box-shadow.

### DIFF
--- a/less/control.less
+++ b/less/control.less
@@ -86,6 +86,7 @@
 	> input {
 		cursor: default;
 		background: none transparent;
+		height: auto;
 		border: 0 none;
 		padding: 0;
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @Fauntleroy, looks good.

Issue #303, opened 2015-07-05
Undefined values, as no longer set up on componentWillMount


alistairjcbrown:
Issue introduced in the latest version `v0.5.3` (not present in `v0.5.2`)

```
Uncaught TypeError: Cannot read property 'forEach' of undefined
```
Which references this line: https://github.com/JedWatson/react-select/blob/23f0436094432986392aadbb1482cd12ef36a627/lib/Select.js#L720

Looks like this is related to the change in PR #280
https://github.com/JedWatson/react-select/pull/280/files#diff-04191649796841200b33375bdff7c78fL133

This PR removes the state setup code which sets up `values` - there is a comment in the removed code which references `Fixes issue #201` - #201 also details that `this.state.values` is not set up, causing an error.

Manually adding the removed code back in solves the issue on my build, with the latest version `v0.5.3`.

alistairjcbrown:
@JedWatson I think too much was code removed in that last version which included changes to `autoloadAsyncOptions` - I've added a PR #304 which should fix this issue.

JedWatson:
Fixed by #304

PR #304, opened 2015-07-06
Add setState back for componentWillMount

alistairjcbrown:
Adds `this.setState` call back into `componentWillMount` to set the initial values. This prevents attempting to loop over `this.state.values` when it is `undefined`, causing an exception.

Fixes issue #303


brianreavis:
Good catch! Instead of this approach, how about we just call `getStateFromValue` in `getInitialState`? That way we wouldn't be regressing back to the old – broken in its own way – pattern. Would you mind putting together a pull request for that?

JedWatson:
I've merged this as-is because the latest version is broken and I'm about to jump on a plane from London to Sydney and don't want to leave it in this state. Would appreciate a PR as @brianreavis suggests though, we should clean this up for another release asap.

alistairjcbrown:
@brianreavis @JedWatson Thanks for the feedback - I just threw this together late last night to prevent the error coming out in my local build.
I've created PR #308 which moves the creation of state into `getInitialState`.

PR #306, opened 2015-07-06
allow the translation of the 'Add label ?' line on multi-value input

Fenntasy:
This 'Add label ?' line was the only one that I found that was hardcoded and was then preventing a translation on react-select. 

I choose this simple system of replacing {label} to accomodate different languages that might need to replace the entire sentence and not just the "Add" part (the verb could be at the end, punctuation could differ...). 

I only touched to the src folder. Please let me know if I need to change other places to make it work :)


JedWatson:
Looks good @Fenntasy, thanks!

PR #308, opened 2015-07-06
Generate state object in getInitialState

alistairjcbrown:
Followup to PR https://github.com/JedWatson/react-select/pull/304

Instead of setting state in `componentWillMount`, this generates the state object as part of `getInitialState`.

/cc: @brianreavis @JedWatson


alistairjcbrown:
I'm not sure what's changed in the 6+ months since this was put up, so I'll close it off.

PR #313, opened 2015-07-13
Add tests

bruderstein:
I'm starting to add some tests to react-select, as we're beginning to rely on it as an excellent "native" implementation of select2. I'm submitting this PR to see how the land lies - we've had previous success using [unexpected](https://github.com/unexpectedjs/unexpected) and [unexpected-dom](https://github.com/Munter/unexpected-dom) testing react components.

I'm aiming towards writing tests for where I've seen a couple of issues, but are difficult to prove/check without tests. However, in order to get there, I thought I'd start with a few general tests to cover the basics.

Coverage has improved from about 24% to about 62% :)

I know changing assertion libraries is not necessarily seen as a positive thing, but as the previous tests weren't really being extended, I thought I'd give it a go.

Comments welcome, obviously!

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,10 @@
     "react": ">=0.13.3",
     "react-component-gulp-tasks": "^0.7.0",
     "sinon": "^1.15.4",
-    "sinon-chai": "^2.8.0"
+    "sinon-chai": "^2.8.0",
+    "unexpected": "^9.2.1",
+    "unexpected-dom": "^1.1.3",
+    "unexpected-sinon": "^6.4.1"
   },
   "peerDependencies": {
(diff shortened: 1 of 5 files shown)


JedWatson:
These look really good @bruderstein. Given that you're stepping up to add tests I'm happy for you to make calls like the assertion library. I'll possibly have more comments / questions in the next few days but wanted to merge this & give you the green light.

Thanks!

bruderstein:
Great, thanks for the green light!  I'll see what I can come up with in the next few days. I've got a couple of weeks holiday coming up, so if it goes quiet, it's not because this won't be continued! :)

PR #319, opened 2015-07-16
Filter options regardless of searchable setting

### LINKED ISSUE #318: onChange firing twice with multi={true} & searchable={false}
I'm running the following setup:

- `react-select-0.5.5`
- `react-input-autosize-0.5.0`
- `classnames-2.1.3`

When using a `multi={true}` select with `searchable={false}`, the `onChange` event appears to fire twice on every selection. It also doesn't seem to be properly removing selected options from the dropdown menu as expected based on [what's in the related readme section](https://github.com/JedWatson/react-select#multiselect-options).

**Note:** The `onChange` event doesn't seem to double-fire when `searchable={true}`, FYI, only when `false`.

I pared down my code to an absolutely basic example (basically what's in the readme + the two extra flags mentioned) and it's still happening. Not sure if this is a recently introduced issue or if it's been happening for a while, I only switched my multi-select to non-searchable today and ran into the issue. Tried rolling back to `react-select-0.5.2` to double-check and had the same issue.

Here's the code I tried (mine was inside a component so slightly altered for readability):

```
var Select = require('react-select');

var options = [
    { value: 'one', label: 'One' },
    { value: 'two', label: 'Two' }
];

function logChange(val, opts) {
    console.log("Selected: %o, %o", val, opts);
}

<Select
    name="form-field-name"
    value="one"
    options={options}
    multi={true}
    searchable={false}
    onChange={logChange}
/>
```

... and here's the console output in Chrome from clicking both list items once.

image

NSinopoli:
Fixes #318.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -532,10 +532,6 @@ var Select = React.createClass({
 	},
 
 	filterOptions: function(options, values) {
-		if (!this.props.searchable) {
-			return options;
-		}
-
 		var filterValue = this._optionsFilterString;
 		var exclude = (values || this.state.values).map(function(i) {
 			return i.value;


JedWatson:
@NSinopoli would you mind helping out by explaining how this fixes the problem?

bruderstein:
I can confirm that 
a) The problem exists, and is easy to replicate
b) This fixes it.

However, I can't yet replicate the double onChange issue in a test. I can replicate that existing selections aren't removed when searchable=false, and this fixes that too (and that the test then passes)

I'll keep on it - when I can replicate it in a test, I should be able to see what causes it, and therefore why this fixes it.

bruderstein:
I'm a little bit further - leaving this here in case anyone has any ideas:

The mouseDown event on the option is fired twice, but on two different "instances", ie. two different renders.

mouseDown fires, adds the value, calls setState, rerenders, render calls buildMenu(), 
mouseDown fires on the re-rendered instance of the option.

It "feels" like a timing issue, as it's not always, but is mostly reproducible for me on the selection of the second item, but occasionally it works, and it works if I step it through via the debugger.

bruderstein:
I'm basically giving up :(

This fixes the issue because the item isn't there after the first mouseDown for the second mouseDown event to fire. Why it fires twice is still a mystery to me. I've tried rebuilding a little example with the same idea - regenerating options with mouseDown, mouseEnter and mouseLeave events that call setState, even tried making render() slow, and everything works perfectly.

I think what this really needs, is a bit of a refactor so that the menu is a separate component. This could potentially simplify the main component.

TL;DR: I'm :+1: for this to be merged :)

JedWatson:
Thanks for the review @bruderstein - happy to go with your :+1:

Issue #262, opened 2015-06-20
Disabling while empty in multiselect mode causes box to collapse


ianobermiller:
Probably missing a min-height when collapsed. Reproduced in the examples.

![react-select-disable](https://cloud.githubusercontent.com/assets/897931/8267323/97757556-1710-11e5-8948-681166b56c26.gif)

JedWatson:
@ianobermiller I've fixed this and #261, and pushed a new version `0.5.1`.

Thanks a lot for the clear issues and gif examples, they made it really easy to understand and track down what was wrong!

ianobermiller:
Absolutely! I know how frustrating to have to ask for lots more info just
to repro!

PR #320, opened 2015-07-16
Specify a min-height for .Select-control

NSinopoli:
The select box collapses (similar to what is seen [here](https://github.com/JedWatson/react-select/issues/262#issue-89765704)) when `searchable` is false and no values are selected.

Specifying a `min-height` fixes this.


JedWatson:
@NSinopoli we've tried really hard to make the styling of `react-select` controllable with variables, and this will cause issues with the control at different font sizes / padding values / etc. We should either use `em` units, or (preferably) find a better fix for this (e.g rendering a `&nbsp;` value in that case, as per the solution to #262)

agirton:
This is no longer an issue and has been fixed.

PR #321, opened 2015-07-16
More tests

bruderstein:
Some more tests, this time based around options and value.

I've split this into two commits - the first adds some pretty general test cases, the second adds a single failing test, that shows the bug if the value is set, then the matching option arrives.

I've checked, and this issue is resolved by a rebased PR #79

(Realistically, this is all I'm going to manage for the next couple of weeks - but, I'll be back - there's more to come :)

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -407,7 +407,7 @@ var Select = React.createClass({
 			case 27: // escape
 				if (this.state.isOpen) {
 					this.resetValue();
-				} else {
+				} else if (this.props.clearable) {
 					this.clearValue(event);
 				}
 			break;
--- a/src/Select.js
+++ b/src/Select.js
@@ -381,7 +381,7 @@ var Select = React.createClass({
 	},
 
 	handleKeyDown: function(event) {
-		if (this.state.disabled) return;
+		if (this.props.disabled) return;
 
 		switch (event.keyCode) {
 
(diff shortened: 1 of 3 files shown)


bruderstein:
@JedWatson - not sure if you've seen this, or are leaving it as the build is failing.  How would you like to deal with these?  I've got a few more failing tests, which I've fixed in the same commit - I'll push those when I've got a decent connection again. I've not yet found anything else that needs a lot of fixing (just been one liners), so I'm hoping there's no more where it doesn't make sense to include the fix with the test.

bruderstein:
Whilst there are still some props not tested (notably valueRenderer), this is now a more complete test suite. There are certainly cases and combinations that aren't covered, but coverage is up in the 90s, and a good number of the common cases are now covered.  I'm not sure how the late options issue has been fixed - probably via the placeholder PR, but it now works too.

I'll send a separate PR for the currently untested props when I get chance.

I've tried to keep everything black-box, so the only thing it relies on, is that options have .Select-option, and the search input has .Select-input etc. This should make it easy to refactor stuff :)

Comments obviously welcome.

JedWatson:
> Comments obviously welcome.

Epic. Thank you. This is awesome.

JedWatson:
(also sorry for the delay - I was really sick for a week or so after getting home from React Europe and had a huge number of notifications, this did get lost amongst them and I'm working through getting back on top of everything now!)

bruderstein:
No worries - hope you're feeling better!

PR #325, opened 2015-07-20
Bugfixed update input-value when options are renewed

Reggino:
When using `react-select` for a single-select, with static options, the `input`-field isn't updated properly when new options are set via `props`.

This change fixes that issue.


agirton:
Closing as this is using the old version of the package and no longer applicable.

PR #326, opened 2015-07-21
Update Select.js

goodzsq:
fix a bug when allowCreate=true, console Warning: flattenChildren(...): Encountered two children with the same key, `.$option-a`. Child keys must be unique; when two children share a key, only the first child will be used.


seanmheff:
Can we get this merged?

seanmcl:
+1

JedWatson:
Done, thanks @goodzsq.

PR #331, opened 2015-07-22
Stopped eslint from linting bower_components

0xCMP:
Hey, I know there isn't an issue for this but `npm run lint` reports a ton of errors from bower_components. 

It still reports problems but it's only 3 from [wallaby.js](https://github.com/JedWatson/react-select/blob/master/wallaby.js)


JedWatson:
Thanks @0xCMP, I hadn't noticed this one before.

PR #332, opened 2015-07-22
valueRenderer now works when multi={false} #296

### LINKED ISSUE #296: valueRenderer works only if multi={true}
I tried using ```valueRenderer``` without ```multi``` set to true but it didn't work.
I gave a look at the code and found out that ```valueRenderer``` is only used inside [this if block](https://github.com/JedWatson/react-select/blob/bbefe623c26d476eae75aef7b0263f98f89d620d/src/Select.js#L712-L723)

0xCMP:
* Updated the README to mention the custom renders
* Updated examples with custom renderer that has multi disabled
* Added style for custom value rendering `Select-value` (basically `Select-placeholder`, but different so it can be selected)
* Using valueRenderer to render `label` to display as value which fixes #296
* Works with custom inputs as well.

Are there any tests to create or anything else I need to update?

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,9 +143,11 @@ For multi-select inputs, when providing a custom `filterOptions` method, remembe
 	Property			|	Type		|	Description
 :-----------------------|:--------------|:--------------------------------
 	value 				|	any			|	 initial field value
+	valueRenderer		|	func		|	 function which returns a custom way to render the value selected
 	multi 				|	bool		|	 multi-value input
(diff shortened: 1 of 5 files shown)


0xCMP:
Made a tiny mistake and referenced the wrong issue. Nothing a force push didn't fix, but somethings weren't properly notified.

JedWatson:
This looks good, @0xCMP, thanks.

PR #334, opened 2015-07-23
Updating Select's props.placeholder doesn't get reflected by the component - fix for this bug

alesn:
When Select receives updated 'placeholder' in its props, the component would not reflect it. The initial placeholder would remain displayed. This code change fixes this bug


JedWatson:
Looks good @alesn, thanks!

PR #336, opened 2015-07-24
Fix the call to the "clearValue" method.

mkotelnikov:
Press Esc button rises an exception. The reason is that the "clearValue()" method is called without parameters.


JedWatson:
Thanks @mkotelnikov!

Issue #274, opened 2015-06-26
disabled value is selectable by keyboard enter


chandu0101:
http://jedwatson.github.io/react-select/

country = US 

i am able to select ``Alabama`` with keyboard enter , but not through mouse click though.

bruderstein:
This has been fixed in >=0.6.0.

PR #346, opened 2015-07-30
Allow disabled option

mondaychen:
It will be useful to allow `{value: "some value", label: "some label", disabled: true}`.

I noticed that this is already partially implemented. A disabled option gets a classname `is-disabled` (with proper css style). And you can no longer select it by mouse click. However you can still select it using keyboard. This PR will fix it. (as described in #274)


mondaychen:
Should I commit the the changes of the dist files? I got more diff than expected. Seems that I'm using some newer version npm packages.

JedWatson:
Great - thanks @mondaychen. No need to commit the dist files, they're generated when I publish releases.

PR #348, opened 2015-07-31
Placeholder tests

bruderstein:
Adds tests for the placeholder functionality, including the fix in the recent PR (#334)

Also corrects the comparison - this was causing extra unnecessary work by comparing
to state.placeholder instead of props.placeholder

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -161,7 +161,7 @@ var Select = React.createClass({
 				filteredOptions: this.filterOptions(newProps.options)
 			});
 		}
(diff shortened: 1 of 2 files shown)


bruderstein:
I've got the rest of the props tested too (see [bruderstein/react-select/tree/even-more-tests](https://github.com/bruderstein/react-select/tree/even-more-tests)), but they sadly rely on the small test refactoring that is in this PR. When/if this is merged, I'll rebase and send the rest.

JedWatson:
Brilliant as usual, @bruderstein :grinning: thanks!

PR #352, opened 2015-08-02
Extract components

jniechcial:
New PR due to rebasing and fixing conflicts, but the discussion [available in this PR](https://github.com/JedWatson/react-select/pull/328).

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -49,7 +49,8 @@ var Select = React.createClass({
 		valueRenderer: React.PropTypes.func,       // valueRenderer: function(option) {}
 		optionComponent: React.PropTypes.func,     // option component to render in dropdown
 		valueComponent: React.PropTypes.func,      // value component to render in multiple mode
-		singleValueComponent: React.PropTypes.func // single value component when multiple is set to false
+		singleValueComponent: React.PropTypes.func,// single value component when multiple is set to false
+		newOptionCreator: React.PropTypes.func     // factory to create new options when allowCreate set
 	},
 
 	getDefaultProps: function() {
@@ -80,7 +81,8 @@ var Select = React.createClass({
 			value: undefined,
 			optionComponent: Option,
 			valueComponent: Value,
-			singleValueComponent: SingleValue
+			singleValueComponent: SingleValue,
+			newOptionCreator: undefined
 		};
 	},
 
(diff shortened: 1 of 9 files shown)


bruderstein:
Cool, I've just rebased this on my even-more-tests branch (this will be sent as a PR if/when PR #348 is merged, as it relies on it), and one test fails. (The tests for valueRenderer are sadly only in this branch)

It's a pretty easy fix though, you just need to move the 
```
var val = this.state.values[0] || null;
```
from line 767 to outside the deepest if, otherwise in the case of a single value with a valueRenderer, you pass option={val}, but val isn't defined.  After doing that, the test passes, and everything is green.

Tests are nice :)

src/Select.js
```diff
@@ -720,29 +739,39 @@ var Select = React.createClass({
...
 			if(this.props.valueRenderer && !!this.state.values.length) {
-				var val = this.state.values[0] || null;
 				value.push(<Value
 						key={0}
 						option={val}
 						renderer={this.props.valueRenderer}
 						disabled={this.props.disabled} />);
 			} else {
-				value.push(<div className="Select-placeholder" key="placeholder">{this.state.placeholder}</div>);
+				var val = this.state.values[0] || null;
```
bruderstein: Need to move this line up

src/Select.js
```diff
@@ -720,29 +739,39 @@ var Select = React.createClass({
...
+					onRemove: onRemove,
+					disabled: this.props.disabled
+				});
+
+				value.push(valueComponent);
 			}, this);
 		}
 
 		if(!this.state.inputValue && (!this.props.multi || !value.length)) {
 			if(this.props.valueRenderer && !!this.state.values.length) {
```
bruderstein: line 767 should be here

src/Select.js
```diff
@@ -720,29 +739,39 @@ var Select = React.createClass({
...
-				var val = this.state.values[0] || null;
 				value.push(<Value
 						key={0}
 						option={val}
```
bruderstein: val used here, without being defined

jniechcial:
@bruderstein fixed, good to have wide test coverage :)

JedWatson:
#348 is merged :) this looks great @jniechcial, thank you. and thanks to @bruderstein for the review!

PR #355, opened 2015-08-03
Added allowCache option

### LINKED ISSUE #354: Cache option
Currently there is no way to set caching to be off. My use case changes the asyncOptions results based on previous values in the select. Therefore, the input string when there is 1 value vs 2 values might be different. There is no elegant way to allow for such logic.

khankuan:
Closes #354

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ The function takes two arguments `String input, Function callback`and will be ca
 
 When your async process finishes getting the options, pass them to `callback(err, data)` in a Object `{ options: [] }`.
 
-The select control will intelligently cache options for input strings that have already been fetched. Async options will still be filtered like the normal options array, so if your async process would only return a smaller set of results for a more specific query, also pass `complete: true` in the callback object.
+The select control will intelligently cache options for input strings that have already been fetched. Cache can be disabled by setting `allowCache` to `false`. Async options will still be filtered like the normal options array, so if your async process would only return a smaller set of results for a more specific query, also pass `complete: true` in the callback object.
 
 Unless you specify the property `autoload={false}` the control will automatically load the default set of options (i.e. for `input: ''`) when it is mounted.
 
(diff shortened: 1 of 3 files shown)


bruderstein:
Hi, it'd be awesome if you could add a test or two for when this prop is false :smile:

It'd be worth noting in the README too, that setting completed=true has no effect when allowCache=false

khankuan:
Done :) Is the test sufficient?

bruderstein:
Perfect, thanks!

If you rebase on master, you'll see the `createControl()` helper can be used now to create the control with the given props, which shortens the boilerplate code in the `beforeEach()`

This means your beforeEach() becomes simply
```js
instance = createControl({
    value: '',
    asyncOptions: asyncOptions,
    allowCache: false

```js
instance = createControl({
    value: '',
    asyncOptions: asyncOptions,
    allowCache: false
});
```

(`onChange` is automatically set by `createControl`)

khankuan:
Alrights updated ;)

JedWatson:
Great - thanks for the PR @khankuan and the review @bruderstein!

JedWatson:
As I was writing up the release notes for 0.6.1, I realised that I think it's clearer if we flip this around so there's a `disableCache` option, rather than having `enableCache` default to true.

Having an option that needs to be set to turn something off (caching) seems more intuitive if it's written in the negative, rather than having to specify a false value for a positive option.

If the prop should be in the positive by default, we should name the feature (similar to `autoload`); possibly `cacheAsyncResults` would be clearer.

I want to release a new version with the PRs that have just been merged, but will leave some time for feedback on this so we don't have to release a breaking change if there's a better option.

I've committed the change to make it `disableCache`, let me know what you think.

khankuan:
Ah good point ;) cacheAsyncResults sounds great. Now that you've mentioned, disableCache seems to be ambiguous as well since cache only applies to async results in this case.

bruderstein:
I'm generally against toggles named negatively, they remind me of the opt-out boxes on marketing forms ("tick here if you don't want to not receive email from us" :smile:), but I see the clarity problem here. 

I think I'm with @khankuan on this, and cacheAsyncResults with default as true would be my vote.

JedWatson:
I thought of `cacheAsyncResults` about a minute after I committed the change to `disableCache`... :smile: 

@bruderstein that marketing form connection is pretty funny, I may never use a negative toggle again...

Just committed the change, so we now have `cacheAsyncResults`.

PR #356, opened 2015-08-03
Final batch of tests

bruderstein:
This is the last batch of tests for the existing functionality.  I think that pretty much everything is covered now, although there are obviously many combinations of options that are not covered - we'll add these as bugs are reported, would be my suggestion.

Currently there are 8 failing tests (one for searchable=false and multi=true, and 7 for disabled options).

PR #319 corrects the searchable=false/multi=true test, and PR #346 corrects 4 of the 7 disabled option tests.

I'm sending another PR which fixes the bug causing the last three to fail (but that PR doesn't make sense without #346, hence the separate PR).

I've also got a change for #346, to show the disabled options in a search - I'll send that separately as that's a matter of taste as to whether disabled options should appear after a search. It just "felt" wrong when I tried it in the examples.

I've rebased these PRs and tried out the examples, and all LGTM 

Happy to rebase this PR when the fixes are in, if you'd prefer to "stay green" :smile:

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,14 +23,12 @@
     "istanbul": "^0.3.17",
     "jsdom": "^3.1.2",
     "mocha": "^2.2.5",
-    "mocha-jsdom": "^1.0.0",
     "react": ">=0.13.3",
     "react-gravatar": "2.0.0",
     "react-component-gulp-tasks": "^0.7.0",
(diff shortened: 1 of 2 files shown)


bruderstein:
Rebased.

PR #357, opened 2015-08-03
Fix for disabled option being the first option

bruderstein:
This is a fix for disabled options being the first (or only) option.

This PR only makes sense after PR #346 has been merged, and the tests are in PR #356 (3 of them fail without this commit)


JedWatson:
Working through this chain of PRs :smile:

Thanks @bruderstein!

Issue #227, opened 2015-06-01
Potential Refactor


brianreavis:
First, react-select is slick! It's been awesome to work with. I've been wanting to make some bugfixes, but have ended up with my hands tied because of a few things that are hard to reason about. This list also contains some other notes:

- [ ] Drop the "dist" files altogether. React's already basically impossible to use without a bundler, so I'm not sure if it's doing much good? "Please don't update the dist files" gets really old to say to the million people who don't read the contributing notes.
- [ ] Build "lib" when publishing, but ignore in repo (same point as above)
- [ ] Reduce/reconsider usage of `componentWillMount`? It's triggered on the server, so calling `autoloadAsyncOptions` and such here seems strange. Also, I'm not sure `this._bindCloseMenuIfClickedOutside = ...` etc are assigned in this way here?
- [ ] Enforce `"options"` prop to be immutable. Whenever a prop changes, there's two `JSON.stringify` calls in `componentWillReceiveProps` that are expensive for large datasets. It'd be more efficient if there could just be a strict equality check here.
- [ ] Move functionality not directly related to the control out to separate files/modules. The includes:
   - [ ] DOM event binding
   - [ ] Option searching (for [selectize](https://github.com/brianreavis/selectize.js) I split this out to [sifter.js](https://github.com/brianreavis/sifter.js) because it ballooned and needlessly complicated the library)
- [ ] Consolidate menu opening code to happen in one spot. As it is right now, the menu's events are bound multiple places throughout the library. I *think* the intent is to consolidate `setState` calls? React will batch these so I think's a bit moot and leading to more problems than it's solving? It's tough to track down where events are bound/unbound.
- [ ] Don't modify state directly (esp in render methods). e.g. [L630-L634](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L630-L634)
- [ ] Use `...spread` in cases like [these](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L737-L741).
- [ ] Why does the `this` context matter on callbacks? E.g. `callback.call(this, {});` All React element methods are auto-bound already.
- [ ] Consider https://github.com/JedWatson/react-select/issues/25. Given that React's point is abstracting away DOM craziness, I think that the "values must be strings" is quite limiting... it leads to lots of application-level casting in `onChange` events. It feels very weird to cast an ID to a string for `<Select>` then cast it back to a number.

[...]

PR #359, opened 2015-08-05
Search ignoring character variations

pedroseac:
This change is [Reviewable](https://reviewable.io/reviews/jedwatson/react-select/359)

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -560,6 +560,19 @@ var Select = React.createClass({
 
 		});
 	},
+	
+	removeAccent: function (string) {
+		string = typeof string != "undefined" ? string : this;
+		string = string.replace(/[á|à|â|ã]/gi, "a");
+		string = string.replace(/[é|è|ê]/gi, "e");
+		string = string.replace(/[í|ì|î]/gi, "i");
+		string = string.replace(/[ó|ò|ô|õ]/gi, "o");
+		string = string.replace(/[ú|ù|û]/gi, "u");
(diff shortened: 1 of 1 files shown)


bruderstein:
I'm half torn between giving this the :+1: , and saying that if you want to do this, you should be using your own custom `filterOptions` function.  I can see this would be something that would grow significantly (e.g. would be good to have umlauts in there, speaking from Germany :smile:).  I'd also make this optional, so add a prop to turn this on or off (e.g. ignoreAccentsInFilter={true}).  

Note you don't need the `|` in a group, and you can chain the replacements, so this works fine
```js
string = string.replace(/[áàâã]/gi, "a")
    .replace(/[éèê]/gi, "e")
    .replace(/[íìî]/gi, "i")
// etc
```

There is a suggestion from @brianreavis in #227 to use [sifter.js](https://www.npmjs.com/package/sifter) for the filtering, which would solve this sort of thing nicely.

pedroseac:
sorry for the insanely delayed respose!

thanks for the comment and corrections @bruderstein! I gonna have a better look at it. 

My thinking at the time was: "Google does like this, why not you?". Huahaha!

pedroseac:
https://github.com/JedWatson/react-select/blob/master/lib/utils/stripDiacritics.js =)

Lenne231:
Is there any progress on this? I would really like to use umlauts as well...

agirton:
This has been resolved by using `ignoreAccents` prop and setting it to false.

Issue #200, opened 2015-05-20
Uncached values, not labels, prepopulating the field


elevee:
I'm not sure if this is a bug or just my misunderstanding of what should be done, but consider a multiselect async-fetched field that fetches options fine, and displays the already-fetched labels as user feedback:

![screen shot 2015-05-20 at 11 11 05 am](https://cloud.githubusercontent.com/assets/4307773/7733020/2665c4c0-fee1-11e4-9b0c-bd46d8f2d732.png)

![screen shot 2015-05-20 at 11 11 25 am](https://cloud.githubusercontent.com/assets/4307773/7733041/4be9ad9c-fee1-11e4-9928-5e646d1b791a.png)

However, upon a refined query: 
![screen shot 2015-05-20 at 11 11 44 am](https://cloud.githubusercontent.com/assets/4307773/7733066/7dcf6ee6-fee1-11e4-933f-0cc0d9fb8b89.png)

it seems to set the previous selections (grabbing from selectedOptions argument on onChange) to label: "16985", value: "16985", and so on. It seems to only preserve the label data if it's part of that specific response. How can one cache it all? 

![screen shot 2015-05-20 at 11 12 00 am](https://cloud.githubusercontent.com/assets/4307773/7733089/ab3301e0-fee1-11e4-955e-e569c6bad3b5.png)

dcousens:
Thanks for the verbose report @elevee

elevee:
@dcousens Haha hope that wasn't sarcasm. Happy to help. 

So the library doesn't support this behavior yet? I just figured I was missing some sort of setting. Do you know how this could be implemented in the meantime?

dcousens:
> @dcousens Haha hope that wasn't sarcasm. Happy to help.

Haha, it wasn't.  This is a great bug report.

> So the library doesn't support this behavior yet? 

It is meant to, but clearly this is a bug.  It should be fixed before the next release though.

JedWatson:
Fixed by #360 - sorry we didn't get to it sooner @elevee!

elevee:
Thx for addressing @JedWatson!

PR #360, opened 2015-08-05
Issue #200 - calling loadAsyncOptions when receiving new props

drfeelgoud:
This resolves issue #200

When passing new props to component, asyncOptions are fetched, because the passed values are not guaranteed to be contained in the current options cache.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ coverage
 
 # Dependency directory
 node_modules
+bower_components
 
 # Publish directory
 .publish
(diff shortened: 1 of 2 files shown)


JedWatson:
Thanks for the fix @drfeelgoud! much appreciated.

bruderstein:
Oh, can you hold off on 0.6.1 (PR #355) for a little longer - I've spotted a bug with the fix for requesting asyncOptions again on value change - PR on the way :smile:

drfeelgoud:
Happy to help, thanks for merging

PR #363, opened 2015-08-06
Disabled options appear in the search results

bruderstein:
Previously (following PR #346) a disabled option would show up in the list, until searching, then it would disappear.
This commit changes that behaviour, so that the option is still shown (when it matches the search), but still cannot be selected.

Test added.

This is a very subjective PR, so it'd be good to get a few opinions on this - I find it feels weird when you click the arrow to show the states, type 'a', and 'Alabama' disappears.  If the caller didn't want to display the option, they'd have removed it from the options, not set it to disabled, but that's just my opinion.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -572,7 +572,6 @@ var Select = React.createClass({
 			var filterOption = function(op) {
 				if (this.props.multi && exclude.indexOf(op.value) > -1) return false;
 				if (this.props.filterOption) return this.props.filterOption.call(this, op, filterValue);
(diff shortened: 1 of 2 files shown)


JedWatson:
I'm inclined to agree with you, @bruderstein. I think that if the option shouldn't be visible, don't include it. If it's included but disabled there must be some information that is intended to convey, so it should remain during search. To do otherwise is confusing.

I'm happy to merge this. If there's a compelling reason to support removing disabled options during searches we could always introduce a prop to control this behaviour (not necessary though, IMO).

PR #365, opened 2015-08-08
Use correct parameters for loadAsyncOptions

bruderstein:
Fix for PR #360 / #355 (the two combined cause a potential issue)

If the cache is disabled, and the value is updated, the new options received
from the callback should be used for the display value, not the options from
state.options, as the state after the callback is stale (this.setState() has
been called, but won't have taken effect due to update batching)

Tests added.


JedWatson:
Sweet. This is what we needed to release `0.6.1` right? (ref #360)

bruderstein:
Yep :-)

PR #368, opened 2015-08-09
Add coveralls badge

bruderstein:
This adds a coverage badge to the readme, integrating [coveralls.io](http://coveralls.io).  

The badge links and API key need changing if this gets the thumbs up, which I think I can do. You can see the result at [bruderstein/react-select](https://github.com/bruderstein/react-select)

I've been testing it out with my repo and keys, hence the need to update the keys & links, but I didn't want to sign up to coveralls without the :+1: 

The downside is it adds a `devDependency` ([coveralls](https://www.npmjs.com/package/coveralls)) that is only really needed for travis.


JedWatson:
We've already got some heavy dev dependencies for the build process, this doesn't look like it would add much more so I'm happy to put it in there. Worth wearing that badge with pride :)

:+1: let me know when it's ready to merge.

bruderstein:
Good to go :+1:

JedWatson:
Thanks!

PR #369, opened 2015-08-09
Fix trim_trailing_whitespace in editor_config

bruderstein:
Any reason why trim_trailing_whitespace was set to false?

After seeing the whitespace fixes in 835caa3fb, I thought there must be something wrong with my editor, but it all looked like it was configured correctly, then I spotted this :smile:

I've left it false for *.md where it might be meaningful.


JedWatson:
No reason at all, it really should have been true!

Issue #270, opened 2015-06-24
dist/react-select doesn't work in an AMD environment


pixeldrew:
Due to browserify's (or browserify-shim) way of referencing global objects, in an AMD environment React isn't available as window.React. Which will cause the module to fail to load.

webpack has the ability to transform to UMD which would work in an AMD and commonJS scenario.

paulkoerbitz:
I am having the same issue. Using [webpack](http://webpack.github.io/docs/) in the react-select base directory as follows

    webpack --output-library "react-select" --output-library-target umd . react-select.js

solved the problem for me. I am now having a different error (prevComponentInstance is null), but I think this is unrelated.

bannaN:
I solved this problem with webpack aswell.
A short webpack config like this should do the trick:

module.exports = {
  entry : {
    "react-select" : "react-select/lib/Select.js"
  },
  output : {
    path: "web/static/dist/external/",
    filename : "[name].bundle.js",
    chunkFilename : "[id].bundle.js",
    libraryTarget : "amd"
  },
  externals : [
    "react"
  ]
};

mondaychen:
@paulkoerbitz Thanks for the workaround

jvmlet:
+1 for this issue, meanwhile here is my workaround :
````
 requirejs.config({
shim: {
"react-select":{
	deps:['reactSelectDep']
}
}
 paths: {
'react-select':'webjars/react-select/0.6.7/dist/react-select.min',
'react-input-autosize':'webjars/react-input-autosize/0.5.0/dist/react-input-autosize.min',
'classnames':'webjars/classnames/2.1.3/index'
}
}

define('reactSelectDep',['react-input-autosize','classnames'],function(autoSize,classNames){
			window['classNames'] = classNames;
			window['AutosizeInput'] = autoSize;
			return {};
		})
````

Please support react-select in AMD env.
Thanks

jochenberger:
Unfortunately the workaround will not work with current versions as react-input-autosize also requires React to be globally available.

jvmlet:
I've posted the idea, you can extend it to define React globaly for react-input-autosize

pixeldrew:
I've created a webpack fork that includes react-input-autosize. Since it's using babel you'll need to have babelHelpers included in the page. See https://github.com/Razorfish-East/react-select

jochenberger:
Still an issue with 1.0.0-beta12

jochenberger:
[...]

PR #374, opened 2015-08-13
Remove browserify-shim configuration

sdeframond:
This PR removes all `browserify-shim` configuration so that we don't have to add react-select's dependencies to the global namespace. I believe it is still possible to shim those libraries from the requiring `package.json` file if necessary.

It is related to https://github.com/JedWatson/react-select/issues/270

edit: typos

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,11 +39,6 @@
     "react": "^0.14.0",
     "react-dom": "^0.14.0"
   },
-  "browserify-shim": {
-    "classnames": "global:classNames",
-    "react": "global:React",
-    "react-input-autosize": "global:AutosizeInput"
-  },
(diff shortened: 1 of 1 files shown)


agirton:
Hi @sdeframond it's been nearly two years. Is this still applicable?

sdeframond:
Hi @agirton , I changed project and I haven't been working with this lib for a while. I don't know if it is still applicable or not.

agirton:
Gotcha. Ok going to close out and will revisit if someone else is faced with the same issue. Thanks!

jochenberger:
It's still a valid fix for #270, but not necessarily an optimal one.

PR #379, opened 2015-08-13
Searching text prop (Rebased #141)

bruderstein:
This is a rebased version of #141 from @johnomalley, with some tests added.

This makes sense to me, if you change `autoload` to `false` in the remote options example, and type `a`, you get `No results` and a spinner for half a second, then the results come in. With the patch, it shows `Searching...`, customisable by the `searchingText` prop.

I've left the code pretty much as is (I'm with @johnomalley on it, the `if` makes it more readable). I've removed the IIFE though, as there doesn't seem any advantageous, and creates another function scope and another bind - feel free to show where I'm missing something.

Tests for the new functionality added in a separate commit.

### DIFF
--- a/test/Select-test.js
+++ b/test/Select-test.js
@@ -2414,7 +2414,7 @@ describe('Select', function() {
 				typeSearchText('abc');
 				expect(asyncOptions, 'was called');
 				expect(React.findDOMNode(instance), 'to contain elements matching', '.Select-loading');
-				expect(React.findDOMNode(instance), 'queried for first', '.Select-noresults',
+				expect(React.findDOMNode(instance), 'queried for first', '.Select-searching',
 					'to have text', 'Testing async loading...');
 			});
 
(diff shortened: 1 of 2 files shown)


dcousens:
@bruderstein looks good!   I personally find that the less indentation and syntax there is,  the less I have to grok in my head,  but I guess that is just a style thing :+1: ,  it was a nit,  my main gripe was the unnecessary IIFE.

Thanks for rebasing!
Over to you @JedWatson

test/Select-test.js
```diff
@@ -2389,6 +2389,69 @@ describe('Select', function() {
 			});
 		});
 
+		describe('searchingText', function () {
+
```
dcousens: Why the function block padding by the way?  Note this will end up failing happiness once the [bug](https://github.com/feross/standard/issues/170) is fixed in eslint.

dcousens: Note this function padding will end up failing happiness once the [bug](https://github.com/feross/standard/issues/170) is fixed in eslint.

bruderstein: because of the blank line after the `function () {`?

Eeee. Funnily enough, that's the [hapi](http://hapijs.org) style, which I like, and find it makes it much easier to read, I also saw that there is a blank line after many functions.

We'll need to do a clean up of all the tests if that gets accepted.  I really don't like that it seems to fail on node 0.10 though.

JedWatson:
This looks great, thanks for cleaning it up @bruderstein!

Issue #290, opened 2015-07-02
Add className support in options objects


emattias:
I need to add an image to every option in the select (see my attached image). My idea is that if you add a className property to every option object the class will be added to the .Select-option div.

![screen shot 2015-07-02 at 09 57 20](https://cloud.githubusercontent.com/assets/351537/8472341/c2990ace-20a0-11e5-9889-d1b37e199cdf.png)

brianreavis:
I'd check out the relatively new "optionRenderer" property for cases like this. It allows for even more granular control – each option can have its own markup that you define.

bruderstein:
Yes, this is exactly what optionRenderer is for - see the example with gravatars. 

I was going to close this, but actually I'm thinking supporting this would only be a matter of adding to the default `<Option>` component, so we're not really adding any unnecessary complexity to `<Select>`.  I know @JedWatson has some plans for being more flexible with options, but not sure if that fits in or not.

Issue #273, opened 2015-06-25
Add optional title attribute to options?


HankMcCoy:
For my use case it would be nice to add an optional title attribute to select options. Theoretical example use:

```js
<Select
  value="one"
  options={[
    {
      value: 'one',
      label: 'One',
      // Adds a title attribute to the option DOM element.
      title: 'The first element in this example list.'
    }
  ]}
/>
```

Is this something you folk would be open to? If so I'd be happy to work on a PR.

shanecalhoun:
This is great. Just what I needed. Now it would be nice to be able to one line ellipse the text instead of wrap it but we will find a css solution for that.

PR #380, opened 2015-08-13
Styled options

bruderstein:
This is to address #290, adding `className` and `style` support to the options objects.

Obviously, you can now do this with the renderers or custom components, however, I wanted to have the option that the default components support it.  I'm entirely agnostic about this - on the one hand it would be better to keep stuff like this up to custom renderers, on the other hand, it'd be nice that the defaults do some basics like this.

I've added `title` support too, to address #273 

Just to reiterate, I really am agnostic about this - this was a quick fix that if thrown away, really doesn't matter. I'm just trying to clear some older issues.  Either, this is merged and #290 and #273 can be closed, or this is not merged, and #290 and #273 can be closed because it can be done via the custom component props.  - Either way, 2 issues down :smile:

### DIFF
--- a/src/Option.js
+++ b/src/Option.js
@@ -20,11 +20,13 @@ var Option = React.createClass({
 		return obj.disabled ? (
 			<div className={optionClasses}>{renderedLabel}</div>
 		) : (
-			<div className={optionClasses} style={obj.style}
-				onMouseEnter={this.props.mouseEnter}
-				onMouseLeave={this.props.mouseLeave}
-				onMouseDown={this.props.mouseDown}
-				onClick={this.props.mouseDown}>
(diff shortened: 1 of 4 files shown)


JedWatson:
Two more issues, and one PR, closed :grinning: 

I think this is a good change - there are two ways to achieve it, but this is much lighter weight if it's all you want to customise. Let's go with it.

amit1911:
it would be good to add this to the documentation.

maxtotheguenther:
@bruderstein i ran into the same problem. I need an optional field for each option field. Beside value and label i need for example an optionalData Object. Is there an approach or rather a best practice for my problem? #2807

Issue #381, opened 2015-08-14
Clicking a disabled item closes the menu


bhstahl:
Not sure if this is your intended action, but it feels like click events on disabled items should be completely ignored.

** Observed**:
- Clicking a disabled menu item closed the menu
![disabled_broke](https://cloud.githubusercontent.com/assets/1383979/9280796/cacb9242-428e-11e5-852d-e1599837ef7f.gif)

** Expected**:
- Clicking a disabled menu completely ignores the click but keeps the window open

![disabled_fix](https://cloud.githubusercontent.com/assets/1383979/9280809/ed83f6a8-428e-11e5-8fa8-79ff298ec80c.gif)

PR #382, opened 2015-08-14
Completely ignore clicks on disabled items

bhstahl:
Add functionality to completely ignore clicks on disabled items, unless the target of the click is a link (super edge case, I know).


bhstahl:
Intends to close #381

bruderstein:
Great catch!  Thanks for adding the extra tests too. The test that disabled option links are clickable needs to actually check something though - I'm not sure how to do that, but there should be an `expect` after the `TestUtils.Simulate.click(link)`.  Maybe you could check window.location.href afterwards?  Or spy on `window.open` with a link with target?

Probably be worth adding a test for the links with `target` too, which is where the coverage drop comes from I guess. 

There might be issues with the slightly dumb DOM emulation here - there's a [better one](https://github.com/chantastic/react-media-object/blob/mocha-tests-with-jsdom/test/utils/document.js) that's been suggested by @chantastic for [generator-react-component](http://github.com/JedWatson/generator-react-component), I'd like to integrate that, then maybe testing this sort of thing will be easier.

bhstahl:
Hey @bruderstein, thanks! Ah, sorry I meant to leave a note in there that it's 80% of the way through but somehow forgot before having to dash on Friday.

Looks like `window` isn't currently available in the tests, so I looked into something similar to `unexpected-dom` for `document` but didn't really get anywhere. I also tried mocking the `window` object to no avail. Will look into `react-media-object` though!

bhstahl:
@bruderstein Totally missed that `window` is on the `global` namespace so I was able to test clicking the link does navigate the location :). The only thing missing is testing that links with `target="_black"` actually open a new window (couldn't find a something like phantomjs's [onPageCreated](http://phantomjs.org/api/webpage/handler/on-page-created.html) in jsdom) but I am testing that click those links doesn't navigated the current window. Let me know if you want me to add anything else!

JedWatson:
@bhstahl that looks great, it looks like there are conflicts with recent changes though - would you mind checking that and rebasing so I can merge this? Thanks!

bhstahl:
@JedWatson good to go! Thanks

JedWatson:
Thanks @bhstahl! Would you mind removing the build / dist files? We just need the `src` changes in PRs.

Nearly there :)

bhstahl:
Whoops, sorry about that as well @JedWatson . Fixed and also squashed some of the redundant test file commits.

JedWatson:
Much better @bhstahl - thanks :D

PR #390, opened 2015-08-19
Fix multi select bug in IE - fixes #92

### LINKED ISSUE #92: Selecting an option sometimes selects mutiple options in IE
When using react-select on IE, I've noticed an issue where sometimes clicking the first option in the dropdown will select both that and the next option simultaneously. As far as I can tell this only happens in IE (version 11 is what I've tested it on). The first example on the demo page seems to have this issue.

jwarning:
This seems to resolve the weird behaviour in IE when clicking options in the drop down list too quickly can cause a single click to add two options. There may be a better solution, but this seems to be the simplest that fixes it without a lot of extra complication.

I'm not sure if the onClick part of the option is necessary if the mouseDown event fires, however there may be a reason it is there?


JedWatson:
@jwarning I think that may be there for mobile device support, we need to look into how touch handles this change before it's safe to merge. I think you've identified the cause of the issue in IE, but I don't want to fix it at the cost of mobile support. Let's make sure the change works for both.

jwarning:
@JedWatson Yeah that would have been my guess, so that sounds like a good idea to do first.

jwarning:
I have checked this change on a Samsung Galaxy tablet through Chrome and it seems fine. That being said I did notice that the clear buttons on all the inputs (the cross) either doesn't work or is really hard to tap properly.

jwarning:
@JedWatson After a bit of further investigation, including on iPads, this seems to work fine as only the mouseDown event seems to matter. This PR: https://github.com/JedWatson/react-select/pull/39 seems to have introduced the issue and even though the mousedown and onclick weren't causing it to fire more than once on most browsers, IE (11) seems to suffer from this issue.

agirton:
Haven't seen this issue yet for the latest release. Closing as this PR is for an old version.

Issue #378, opened 2015-08-13
Disabled options are keyboard selectable


bhstahl:
**Observed**:

Setting an option as disabled blocks mouse focus/selection, but when tabbing through the list with a keyboard, the option enters a focus state and can still be selected.

**Expected**:

Setting an option as disabled blocks mouse and keyboard focus/selection.

bruderstein:
Are you on the latest version (0.6.2) this was fixed around 0.6.0, and there's an explicit test for this behaviour, so it shouldn't have come back!

bhstahl:
Ah sorry about that! Somehow missed #274 and was < 6.0. Works great, thanks!

cosminnicula:
@bruderstein go to the example page http://jedwatson.github.io/react-select. For the first select called "STATES" choose "US". Type inside the select "Alabama"  and hit enter. The disabled option is selected.

Expected behavior: the disabled option should not be selectable via keyboard

bruderstein:
@cosminnicula this is the case for me at least, we fixed this in the last release (hence the issue is closed)

Is "Alabama" still selectable via keyboard for you?  If so, can you double check you've refreshed, and then let us know the browser & version?

cosminnicula:
I've tested this directly on the examples website http://jedwatson.github.io/react-select

bruderstein:
Ah, I see now the issue, when we type and it's the only option available. Apologies, I hadn't properly followed your instructions. I've reopened the issue, and I'll aim to get a PR in tonight.

Thank you for reporting.

PR #395, opened 2015-08-20
Fix disabled after search

bruderstein:
This is a fix for the (reopened) #378, where a disabled option can be selected with enter when the search results in only (or a single) disabled option.  Tests added to cover this case.

I've refactored a little in order to extract a `getFirstFocusableOption` function, which simply returns the first none-disabled option from the given options. This is now called in 2 places.  This really needs simplifying, such that we're only setting the focused option in one place, but that's a bigger issue and this fixes the immediate issue.

This will probably have an effect on #382, but should definitely go in together.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -237,12 +237,7 @@ var Select = React.createClass({
 			focusedOption = values[0];
 			valueForState = values[0].value;
 		} else {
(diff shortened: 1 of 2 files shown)


JedWatson:
Nice work @bruderstein - I can't merge #382 yet because of conflicts, but this looks like it's safe to go in first.

PR #406, opened 2015-08-25
Include built CSS file in bower.json's main block.

seansfkelley:
The `main` field is used by build tools to bring in the right resources; as it stands the default CSS is not included and leads to all kinds of mangled UI if your bower + build system setup depends on files being listed in `main`.


JedWatson:
Thanks @seansfkelley - good call!

PR #412, opened 2015-08-26
Enable lessify transform (compatability)

MattMcFarland:
[Per the Browserify Handbook](https://github.com/substack/browserify-handbook#browserifytransform-field) + [node_modules](https://github.com/substack/browserify-handbook#node_modules)  transforms don't apply across module boundaries. 

Since you are using less, it makes good sense to add the browserify transform field to your package.json so other modules that use react-select as a dependency may transform react-select's .less files simply using browserify's native transform configuration.  I personally find it more useful to use transforms in the package file instead of adding tasks to my task runner that use external plugins that may go out of date and lose compatibility with browserify.  

This pull request adds the transform field to improve compatibility.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,11 @@
   "peerDependencies": {
     "react": ">=0.13.3 || ^0.14.0-beta3"
   },
+  "browserify": {
+    "transform": [
+      "lessify"
+    ]
+  },
   "browserify-shim": {
     "classnames": "global:classNames",
     "react": "global:React",


MattMcFarland:
Please read here too: https://github.com/substack/browserify-handbook#node_modules
>  If your application had transforms configured in package.json, you'll need to create a separate package.json with its own transform field in your node_modules/foo or node_modules/app/foo component directory because transforms don't apply across module boundaries. This will make your modules more robust against configuration changes in your application and it will be easier to independently reuse the packages outside of your application.

Again, I really thought react-select was broken because the stylesheets would not load when I used require('react-select') - I thought it was just another dud react component that doesnt work but after further investigation I found that your package.json is missing the transform field for browserify.  Please add this, if not at least explain why it would be unhelpful to have this

MattMcFarland:
@JedWatson have you looked at this PR yet?

JedWatson:
Thanks for the PR @MattMcFarland 

This shouldn't be necessary anymore with our new build system. Please let me know if it's still relevant.

PR #416, opened 2015-08-27
Add or clear current input value on blur and allow auto select on tab (for new values)

akum32:
When new values are allowed:
1. On blur, add the current input value, so when the user clicks away, it will be automatically added.
2. On tab key press, add the current input value.

When new values are NOT allowed:
1. On blur, clear the current input value.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -312,6 +312,19 @@ var Select = React.createClass({
 		this.setValue(this.state.values.concat(value));
 	},
 
+	addOrClearInputValue: function() {
+		var inputValue = this.state.inputValue;
+		if((inputValue || '').length === 0) {
+			return;
+		}
+		if (this.props.allowCreate) {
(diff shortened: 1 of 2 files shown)


bruderstein:
I'm really not sure if I agree with this change in behaviour. 

I'd be more inclined to offer some hooks so that the caller can do things like this on blur. We'd actually only need an onBlur hook (might be one already, but I'm on a phone, so can't easily check), and a `getCurrentSearchText()` public method.

erikras:
It's very unclear to the user whether or not a value is selected, or what the selected value is when the search input is not cleared on blur.

react-select-weird

JedWatson:
Thanks for the PR @akum32 

We've made a lot of changes since it was opened, so I'm going to close it. I do think our new props give you a similar level of control, but please let me know if this is still relevant.

Issue #399, opened 2015-08-22
Expose input onChange event to ReactSelect


mattbreeden:
Would love to be able to easily access the onChange event for the input component (typing in the bar, not selecting a value).

anton-kabysh:
+1. Eah, highly expected feature. For example, selector used inside popover which is collapsed by any click outside it. Now I am should intercept mouse events on the top of selector to avoid popover collapse. This is really hated boilerplate code.

lomteslie:
I'd love to have this feature as well. Just submitted a pull request that would add this functionality.

PR #425, opened 2015-08-28
Adding onInputChange handler that returns the current input value.

lomteslie:
I needed this feature to proceed and I noticed others were requesting it as well – #399. Thought it might be useful.


JedWatson:
This looks good @lomteslie, thanks - and sorry the PR took so long to review! going to try and pick up the pace around here :)

PR #428, opened 2015-09-03
Bugfix: prevents value splitting on non-multi-value select

### LINKED ISSUE #429: Unexpected value splitting on ',' even when props.multi is not set to true

alanrsoares:
Scenario:
By default, the delimiter is set to ','.
When a value contains ','. Even if props.multi is set to false/undefined, it was splitting on ','.

Solution:
Just check in a guard condition on props.multi before splitting.

cc @JedWatson 

Fixes #429


JedWatson:
This looks good @alanrsoares, thank you :)

PR #438, opened 2015-09-09
Add maxOptions option.

mull:


agirton:
Closing as this is using an old version of the package and not the direction we would like to take.

Issue #441, opened 2015-09-11
Support React v0.14-rc1


RafalFilipek:
http://facebook.github.io/react/blog/2015/09/10/react-v0.14-rc1.html

Thank you

JedWatson:
Just realised this should have been closed when #444 was merged...

PR #444, opened 2015-09-14
Support react 0.14.0-rc1 instead of 0.14.0-beta3

Calvein:
#441


JedWatson:
we should be able to do this with a broader range of `^0.14.0-beta`, I'll push a new patch version out now. Thanks @Calvein!

PR #447, opened 2015-09-16
prevent default behavior of key down

davidpene:
Details:

- Preventing backspace navigating to previous page

cc: @JedWatson


JedWatson:
This looks safe to me, thanks @davidpene

PR #450, opened 2015-09-16
simplifying handleKey function

davidpene:
Details:

- removing top call to preventDefault
- adding break when backspace, as preventDefault is called at the bottom of function

cc: @JedWatson


JedWatson:
Good call, cheers @davidpene

PR #457, opened 2015-09-18
only prevent default if there is no input

davidpene:
details:

- If there's an input value, break will continue the flow and eventually call preventDefault, which will prevent the removal of the input characters
- Therefore, only prevent default if there is no input value, and so it will not return to previous screen

cc: @JedWatson 

NB: This should be the final fix. Tested and run.


JedWatson:
@davidpene :+1:

PR #465, opened 2015-09-21
Fixed closing of dropdown for multi select

urvashi01:
For multi-select, the menu should remain open for the user to select more options. Closing it after every selection counteracts the intuition to select multiple options.


gwyneplaine:
thanks @urvashi01  this functionality has since been implemented in the following commit c7fe6f4
you can opt out of the default behaviour by setting the with a closeOnSelect prop to false.

PR #467, opened 2015-09-21
use wrapper instead of control

bezreyhan:
Our select components are designed so that the menu covers the control when the menu is open. This entails changing the css of `.Select-menu-outer` to have a top of `0px`.

However, when we make that change the menu doesn't open. It flickers but remains closed. I noticed that the menu will close if the user clicks outside of the `.Select-control`. I'm not sure why, but if the menu covers the control, when the user clicks on the control to open the menu the `event.target` === `Select.wrapper` This means that the menu will never open because the event will always occur outside of the control since it is occurring in the wrapper.


examples/dist/standalone.js
```diff
@@ -187,10 +187,10 @@ var Select = React.createClass({
 				return;
 			}
 			var menuElem = React.findDOMNode(_this.refs.selectMenuContainer);
-			var controlElem = React.findDOMNode(_this.refs.control);
+			var wrapperElem = React.findDOMNode(_this.refs.control);
```
RafLeszczynski: I have the same issue and your solution work for me. Just a small typo issue. I think you wanted to get reference to the `_this.refs.wrapper` instead of `_this.refs.control`

src/Select.js
```diff
@@ -118,10 +118,10 @@ var Select = React.createClass({
 				return;
 			}
 			var menuElem = React.findDOMNode(this.refs.selectMenuContainer);
-			var controlElem = React.findDOMNode(this.refs.control);
+			var wrapperElem = React.findDOMNode(this.refs.control);
```
RafLeszczynski: the same case as in `standalone.js`:
`this.refs.wrapper` instead of `this.refs.control`

src/Select.js
```diff
@@ -118,10 +118,10 @@ var Select = React.createClass({
...
 
 			var eventOccuredOutsideMenu = this.clickedOutsideElement(menuElem, event);
-			var eventOccuredOutsideControl = this.clickedOutsideElement(controlElem, event);
+			var eventOccuredOutsideControl = this.clickedOutsideElement(wrapperElem, event);
 
 			// Hide dropdown menu if click occurred outside of menu
 			if (eventOccuredOutsideMenu && eventOccuredOutsideControl) {
```
RafLeszczynski: In this case you probably don't need to check both  for `eventOccuredOutsideMenu && eventOccuredOutsideControl` cause both `selectMenuContainer ` and `control` are children of `wrapper`. This should be enough:

`if (eventOccuredOutsideControl) {`

src/Select.js
```diff
@@ -118,10 +118,10 @@ var Select = React.createClass({
 				return;
 			}
 			var menuElem = React.findDOMNode(this.refs.selectMenuContainer);
-			var controlElem = React.findDOMNode(this.refs.control);
+			var wrapperElem = React.findDOMNode(this.refs.control);
 
 			var eventOccuredOutsideMenu = this.clickedOutsideElement(menuElem, event);
-			var eventOccuredOutsideControl = this.clickedOutsideElement(controlElem, event);
+			var eventOccuredOutsideControl = this.clickedOutsideElement(wrapperElem, event);
```
RafLeszczynski: please consider variable name to `eventOccuredOutsideWrapper`

RafLeszczynski:
BTW @bezreyhan do you know what is the timeline for merging this pull request?

JedWatson:
Sorry this never got attended to. With the release of 1.0.0 I'm closing a lot of old PRs and Issues that either may be fixed / or won't be fixed given the major changes since.

Please re-open if you still want to address this in the latest version.

Thanks!

Issue #459, opened 2015-09-18
Option to wait before sending queries for asyncOptions


oliversong:
It would be nice if we could specify a time delay to wait after the last keystroke before calling asyncOptions, so it doesn't make unnecessary requests.

PSkierniewski:
Bump, as far as I can see asyncOptions are fired on change, is it possible to fire it when user stops typing?

oliversong:
It's implemented in my branch https://github.com/JedWatson/react-select/pull/473 @PSkierniewski

PSkierniewski:
@oliversong Thx I used it and it works.

tommyalvarez:
@oliversong i need this too. Could you submit a PR to this library so the owner can merge it?

phamthaibaoduy:
For those who might need, for now you can have debouncing by wrapping your loadOptions callback by lodash/debounce 

```
const wait = 500; // milliseconds
const loadOptions = input => dataStore.getData(input)
const debouncedLoadOptions = _.debounce(loadOptions, wait)

return <Async loadOptions={debouncedLoadOptions} ... />
```

PR #473, opened 2015-09-23
Implement rolling delay before loadAsyncOptions feature

oliversong:
Addresses https://github.com/JedWatson/react-select/issues/459

Adds the ability to pass in a timeout prop. react-select will now wait `delayAsyncMs` milliseconds before calling loadAsyncOptions, and keep resetting the delay on every new input change until `delayAsyncMs` milliseconds passes without an input change.

Needs tests. Sorry, didn't have time to figure that out yet, but hopefully this will be useful for somebody!

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -526,13 +526,13 @@ var Select = React.createClass({
 
 			this.setState({
 				isLoading: true,
-				inputValue: event.target.value
+				inputValue: this._optionsFilterString
 			});
 		} else {
 			var filteredOptions = this.filterOptions(this.state.options);
 			this.setState({
 				isOpen: true,
-				inputValue: event.target.value,
+				inputValue: this._optionsFilterString,
 				filteredOptions: filteredOptions,
 				focusedOption: this._getNewFocusedOption(filteredOptions)
 			}, this._bindCloseMenuIfClickedOutside);
--- a/src/Select.js
+++ b/src/Select.js
@@ -26,6 +26,7 @@ var Select = React.createClass({
 		clearAllText: React.PropTypes.string,      // title for the "clear" control when multi: true
 		clearValueText: React.PropTypes.string,    // title for the "clear" control
 		clearable: React.PropTypes.bool,           // should it be possible to reset value
+		delayAsyncMs: React.PropTypes.number,      // time delay before querying after a keyup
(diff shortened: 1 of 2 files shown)


src/Select.js
```diff
@@ -494,23 +497,42 @@ var Select = React.createClass({
 		this._optionsFilterString = event.target.value;
 
 		if (this.props.onInputChange) {
-			this.props.onInputChange(event.target.value);
+			this.props.onInputChange(this._optionsFilterString);
 		}
 
 		if (this.props.asyncOptions) {
+			var callAsyncLoad = function() {
+				this.loadAsyncOptions(this._optionsFilterString, {
```
oliversong: Quick note that `this._optionsFilterString` is preferable to `event.target.value`, since in the case of this function it doesn't require the argument to be redefined in its closure.

mmahalwy:
any uck with this?

tommyalvarez:
@oliversong can you fix the coverage downgrade?

JedWatson:
Thanks for the PR @oliversong but I'd prefer this was left for users to implement in the function that's passed to `loadOptions`

You can do this really simply by wrapping the method in lodash's `debounce` method (or similar) without adding any more logic to react-select

craigmichaelmartin:
For those doing so, be aware that Lodash's debounce function isn't suitable for this. A promise-returning debounce method where subsequent calls return promises which will resolve to the result of the next func invocation in needed. See #3075 (comment)

ebonow:
> For those doing so, be aware that Lodash's debounce function isn't suitable for this. A promise-returning debounce method where subsequent calls return promises which will resolve to the result of the next func invocation in needed. See #3075 (comment)

For those looking for a solution, please feel free to try the solution suggested here: https://github.com/JedWatson/react-select/issues/614#issuecomment-679056254
``` js
const loadOptions = React.useCallback(
  debounce((inputValue, callback) => {
    getOptions(inputValue).then(options => callback(options))
  }, 500),
  []
);

return <AsyncSelect loadOptions={loadOptions} {...otherProps} />
```

Demo: [codesandbox](https://codesandbox.io/s/react-select-async-debounce-usecallback-oeixm)

AndrejGajdos:
@ebonow take into account `useCallback` throws `React Hook useCallback received a function whose dependencies are unknown. Pass an inline function instead.` warning

Issue #392, opened 2015-08-20
Is there a way to set the inputValue via props when using asyncOptions


sambomartin:
My scenario is that I fetch options from web service, label and value.
I then select the options, handle the change and keep value in store.
The value of the select is being set via the value property, which when it changes to the value (id) to replaces the value of the input, then triggers another async call to find values using the value in the input (which is now the id).

Is there a way of setting the value and inputValue so the inputValue doesn't get changed to the ID and trigger call to ayncOptions?  I don't want to do a fetch against the remote item, when I already have the value and label in the store.

Thanks

TigerC10:
When the react-select component is fed a value prop on initial render (consider a parent component's getInitialState feeding in an empty string), and then a call to some backend service is launched and changes the value in the parent state to be an option object - the asyncOptions function is fired off despite the fact that options have not been requested. Traced the issue to componentWillReceiveProps(). When the props change on react-select, it should probably only fire asyncOptions if the autoload prop is true.

Opened a PR to address https://github.com/JedWatson/react-select/pull/477

oluckyman:
Have the similar problem. I have a grid with lots of react-select components. And they all simultaneously call `asyncOptions` on grid rerender despite of `autoload: false`

JedWatson:
Reviewing #477 now

jossmac:
Version 1 of react-select is no longer supported.

In the best interest of the community we've decided to spend the time we have available on the latest version.

We apologise for any inconvenience.

Please see:

1. [v1 to v2 upgrade guide](https://react-select.com/upgrade-guide)
1. [v1 documentation and examples](v1.react-select.com)

PR #477, opened 2015-09-23
#392 - Prevent autoload of asyncOptions when value changes

TigerC10:
When the react-select component is fed a value prop on initial render (consider a parent component's getInitialState feeding in an empty string), and then a call to some backend service is launched and changes the value in the parent state to be an option object - the asyncOptions function is fired off despite the fact that options have not been requested.  Traced the issue to componentWillReceiveProps().  When the props change on react-select, it should probably only fire asyncOptions if the autoload prop is true.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -177,7 +177,7 @@ var Select = React.createClass({
 					newProps.placeholder)
 				);
 			};
-			if (this.props.asyncOptions) {
+			if (this.props.asyncOptions && this.props.autoload) {
 				this.loadAsyncOptions(newProps.value, {}, setState);
 			} else {
 				setState();


src/Select.js
```diff
@@ -177,7 +177,7 @@ var Select = React.createClass({
 					newProps.placeholder)
 				);
 			};
-			if (this.props.asyncOptions) {
+			if (this.props.asyncOptions && this.props.autoload) {
```
JedWatson: should this be ` && (this.props.autoload || newProps.value)`?

If a value is passed, we do want the async options to load, otherwise the label for the value won't be expanded correctly

TigerC10: I don't think so, the use case is that the component renders with a null value and then a RESTful promise returns to populate the value of the react-select as an object with a label and a value.  With the object, react-select has enough information to display the correct value without polling the `this.props.asyncOptions` function.

In the proposed case of checking for `newProps.value`, I think it is only beneficial to call `this.props.asyncOptions` if it is a string since it won't have the label to display.

So if `this.props.asyncOptions` is specified and `newProps.value` is a string then it should always call `this.props.asyncOptions` to get the full label/value object.  If `newProps.value` is an object then it should only call `this.props.asyncOptions` if `this.props.autoload` is true.

oluckyman: > So if `this.props.asyncOptions` is specified and `newProps.value` is a string then it should always call `this.props.asyncOptions` to get the full label/value object

But what about string values that appears as labels? Such a values doesn’t need options to show label because the value is the label. So that logic will trigger unnecessary asyncOptions call.

I think we should respect flag `autoload` and never load options automatically before user’s actions.
Consider the grid with lots of react-select components. They all will simultaneously request options on initial value update.

TigerC10: In that case, I think this PR would work as is to accomplish that goal.

AugustinLF: I agree with @TigerC10. It's something I specifically need, another use case to consider is in poor network situations. What happens right now is that the react-select will requests options instantly, but no value will be displayed, as long as the call is not done. Which is obviously not satisfactory.

JedWatson:
Sorry to leave this PR hanging - this issue should be comprehensively resolved by the rewrite I did for 1.0, please let me know if you have any issues with it.

bvaughn:
Closing. This issue should no longer be valid.

PR #478, opened 2015-09-24
Separate mouseDown handlers for menu and main control. Fixes #471

### LINKED ISSUE #471: Menu disappears when scrollbar is touched (when not searchable)
The fix for #397 doesn't seem to work for the case where the selector is not searchable (searchable=false). The menu vanishes when the scrollbar is touched.

chooser-bug

burtyish:
Here's my stab at a fix for JedWatson/react-select#471. 

There's some code duplication here. I'd love for any suggestions for improvement.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -389,9 +389,6 @@ var Select = React.createClass({
 		}
 		event.stopPropagation();
 		event.preventDefault();
-
-		this._openAfterFocus = true;
-		this.getInputNode().focus();
 	},
 
 	handleMouseDownOnArrow: function(event) {
--- a/src/Select.js
+++ b/src/Select.js
@@ -269,7 +269,7 @@ var Select = React.createClass({
 			if (typeof values === 'string') {
 				values = values === ''
 					? []
-					: this.props.multi 
+					: this.props.multi
 						? values.split(this.props.delimiter)
 						: [ values ];
 			} else {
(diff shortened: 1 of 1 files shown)


hugotox:
please merge this. I doesn't looks like coverage is adding value when preventing bugfixes to be merged

hugotox:
actually this PR fixes the bug in all browsers except IE9

burtyish:
> actually this PR fixes the bug in all browsers except IE9

@hugotox thanks for the feedback. I'll take a look at IE9 when I have a chance, but it'll take a while since my free time is limited. Also I don't have access to an actual IE9 browser (will probably try browserstack). Any contributions are, of course, welcome.

I've been trying to add some tests as well, but running into some technical difficulties.

JedWatson:
I'm happy with this change - thanks @burtyish, and @hugotox for the review.

@bruderstein if you want to jump in with advice on how to test this, please do, I'm having trouble running tests on node.js 4.x (and about to open another issue for that)

slestang:
Why does only one commit of this PR has been merged?
[Here](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L645) we can see that handleMouseDownOnMenu is called on mouse down on the menu but this function is actually never define :-(
So bug #471 is still here on 1.0.0-beta6. I slightly modify 30cbe9451003b4876b20a11da526a0168d68767e to make it work with current master:
```javascript
	handleMouseDownOnMenu(event) {
		// if the event was triggered by a mousedown and not the primary
		// button, or if the component is disabled, ignore it.
		if (this.props.disabled || (event.type === 'mousedown' && event.button !== 0)) {
			return;
		}
		event.stopPropagation();
		event.preventDefault();

		this._openAfterFocus = true;
		this.focus();
	},
```
I test it with success. Tell me if a new PR is required.

Issue #434, opened 2015-09-09
TypeError: undefined is not an object (evaluating '_this3.getInputNode().focus')


liamcardenas:
<img width="507" alt="screen shot 2015-09-08 at 6 00 22 pm" src="https://cloud.githubusercontent.com/assets/5027235/9750952/9208a426-5653-11e5-9bf0-779921965681.png">

I get this when a user selects a delete option in the dropdown, which then delete the entire row of the table containing the select box. Do you think this is the cause of the error? Everything is working properly, but I would like to not get this error if possible.

Thanks

jorrit:
Same as #208 ?

jossmac:
Version 1 of react-select is no longer supported.

In the best interest of the community we've decided to spend the time we have available on the latest version.

We apologise for any inconvenience.

Please see:

1. [v1 to v2 upgrade guide](https://react-select.com/upgrade-guide)
1. [v1 documentation and examples](v1.react-select.com)

Issue #208, opened 2015-05-25
When clicked fast enough, component throws error like "cannot call function focus() of undefined"


voroshkov:
I found this issue during writing tests with React.TestUtils.
The issue is annoying, but the solution is simple:
in ComponentDidUpdate you must add clearTimeout(this._focusTimeout) before overwriting this._focusTimeout variable with new timeout id.

dcousens:
@voroshkov can you please add a PR?

jochenberger:
I think that this can be closed since #481 was merged.

PR #481, opened 2015-09-25
Fix for #208 and #434. Reset timeout before setting new one.

PetrGlad:
As suggested by @voroshkov, this patch resets existing timeout before setting new one.
In fixed the problem for me. 
However I see similar pattern in handleInputBlur. It might be also worth looking at.

I ran lint on it, there are pre-existing errors in examples code and tests, changed file is OK.
Test is not included, issue is related to timing. If you know how to implement it I'd happy to add test.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -186,6 +186,7 @@ var Select = React.createClass({
 	componentDidUpdate: function() {
 		if (!this.props.disabled && this._focusAfterUpdate) {
 			clearTimeout(this._blurTimeout);
+			clearTimeout(this._focusTimeout);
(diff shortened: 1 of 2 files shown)


JedWatson:
:+1: thanks @PetrGlad!

JedWatson:
@PetrGlad could you expand on your concerns about `handleInputBlur`? I'm not seeing an issue with it, but it's late so I might just be missing it.

PetrGlad:
handleInputBlur - also sets timers on event, just similar pattern. 
It may well be that there are no problems.

PR #482, opened 2015-09-26
Add `isLoading` prop, resolves #476

### LINKED ISSUE #476: Add `isLoading` prop
I'm working on a project that async loads options for react-select. The async behavior is complex and needs to start **before** react-select is rendered, so using the built-in `asyncOptions` functionality doesn't really work for me.

I'd love to be able use the `searchingText` and `isLoading` indicator, and the best way I can think of is to pass down an `isLoading` prop. Since this could override the built-in `asyncOptions` behavior, maybe `forceIsLoading` would be a better prop name? Not sure.

I'm happy to submit a PR, but since this is a new feature I thought a discussion was in order. So... Let the discussion begin!

jgautsch:
Just like @rileyjshaw (#476) we async load options for react-select. This PR allows for complex async behavior that starts before before react-select is rendered and loads the options to be used in the `Select`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,22 @@ var getOptions = function(input, callback) {
 />
 ```
 
+### Async options loaded externally
+
+If you want to load options asynchronously externally from the `Select` component, you can have the `Select` component show a loading spinner by passing in the `isLoading` prop set to `true`.
+
+```
+var Select = require('react-select');
+
+var isLoadingExternally = true;
+
+<Select
+  name="form-field-name"
+	isLoading={isLoadingExternally}
+	...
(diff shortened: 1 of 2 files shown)


rileyjshaw:
Cool! This looks good to me - you might want to squash it into a single commit, but other than that this looks good to merge. @JedWatson, have you seen this?

JedWatson:
Nice - thanks @jgautsch, and @rileyjshaw for the review. Looks great, will release now.

rileyjshaw:
Yeehaw! Nice work @jgautsch

PR #483, opened 2015-09-27
Get ready for React 0.14

daviferreira:
Had to do this for a personal project, so might as well open a PR. Fantastic work like always Jed :)

On this PR:

* update dependencies
* use `react-dom` package / `ReactDOM.findDOMNode`
* user `react-addons-test-utils` package

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,9 @@
     "istanbul": "^0.3.20",
     "jsdom": "^3.1.2",
     "mocha": "^2.3.2",
-    "react": ">=0.13.3 || ^0.14.0-beta",
+    "react": "^0.14.0-rc1",
+    "react-addons-test-utils": "^0.14.0-rc1",
+    "react-dom": "^0.14.0-rc1",
     "react-gravatar": "^2.0.1",
     "react-component-gulp-tasks": "^0.7.1",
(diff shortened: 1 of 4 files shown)


JedWatson:
Thanks @daviferreira!

I'd like to keep the main package compatible with 0.13 for a little while, particularly while 0.14 is still in `rc`... any ideas how we could do this neatly? Obviously this update would go out with a big version bump, and we could keep a `react-0.13` branch around to make patch releases from with other updates, but I've found that PRs are most often made against master and hard to merge back into old branches / releases.

Any feedback on how we could manage this well over the broader community transition from 0.13 to 0.14 would be very welcome.

daviferreira:
Yeah, I don't know, we could go the other way around now and and have a branch for 0.14, or just keep this PR waiting :)

sirreal:
I would _love_ to have this available on npm under a dist-tag: `react-select@next`.

I don't have a great solution for getting PRs in the right place. The best thing I can think of is to tell contributers that all PRs must be against X branch, and changes will have to be merged. For now that would likely continue to be the 0.13 branch until 0.14 is released.

chicoxyzzy:
:shipit:

PR #484, opened 2015-09-27
update README with isLoading prop

jgautsch:
follow on from #482


jgautsch:
Actually it seems there are several mismatches between `Select` props and those in the "Further Options" table in README. Going to close for a PR with more corrections.

PR #500, opened 2015-10-07
ES6 syntax for tests

dmatteo:
While I was working on the tests, I thought it could be helpful to move to a more ES6 syntax, even if not complete.

I basically moved to arrow functions and member functions but I haven't touched all the `var`.
I've also added a couple of missing traling semicolons


dmatteo:
what the hell. I should have branched from master, I guess

PR #501, opened 2015-10-07
ES6 syntax for tests + housekeeping

dmatteo:
While I was working on the tests, I thought it could be helpful to move to a more ES6 syntax, even if not complete.

I basically moved to arrow functions and member functions but I haven't changed `var` to `let/const`.
I've also added a couple of missing traling semicolons

### DIFF
--- a/test/Select-test.js
+++ b/test/Select-test.js
@@ -45,59 +45,59 @@ class PropsWrapper extends React.Component {
 	}
 }
 
-describe('Select', function() {
+describe('Select', () => {
 	var options, onChange, onInputChange;
 	var instance, wrapper;
 	var searchInputNode;
 
-	function getSelectControl(instance) {
+	var getSelectControl = (instance) => {
 		return React.findDOMNode(instance).querySelector('.Select-control');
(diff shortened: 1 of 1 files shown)


JedWatson:
@dmatteo can you rebase this for us? and @bruderstein what are your thoughts on this?

Seems safe to adopt ES6 for tests now that we're requiring node 4.

bruderstein:
I'm all for it!

The tests are compiled with babel anyway, because of jsx, so there really are no limits :-)

We could const all the things too 😃

JedWatson:
_facepalm_

haha _of course_ we're using Babel.

Awesome, let's do this :grinning:

dmatteo:
Yeah, I saw Babel, that's why I safely assumed it was ok :)

I'll rebase now

dmatteo:
I'm not sure how could the coverage fell when converting to ES6...

JedWatson:
Great, thanks @dmatteo!

dmatteo:
Cheers!

PR #503, opened 2015-10-07
Close select on touchend if touched outside menu/control.

neilzo:
Add touchend listener so the select can close when touched outside the control. This helps touch devices where scrolling triggers the select open, and the user does not want to clear/reselect an option to close the menu.


neilzo:
Closing due to 1.0.0-beta rewrite.

PR #506, opened 2015-10-07
Update package.json

davidmarkclements:
peer deps suck!

```
npm ERR! peerinvalid Peer react-addons-css-transition-group@0.14.0 wants react@^0.14.0
npm ERR! peerinvalid Peer react-select@0.6.12 wants react@>=0.13.3 || ^0.14.0-rc1
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,9 +33,6 @@
     "unexpected-dom": "^1.3.0",
     "unexpected-sinon": "^7.1.1"
   },
-  "peerDependencies": {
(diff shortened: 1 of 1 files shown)


JedWatson:
Unfortunately I don't think this is a reasonable change at this stage, react-select actually _requires_ React 0.14 as of `0.7.x`, and dropping this would incorrectly allow use with prior incompatible versions.

PR #522, opened 2015-10-16
Clear search input on blur

### LINKED ISSUE #490: Searchable Inputs should Reset When _closeMenuIfClickedOutside is Called
I'm not sure if this is a standard or simply a desire I have, but I find it odd that there is inconsistent behavior between hitting escape and clicking outside when the results are searchable.

If you start searching for something and the menu comes up with a few options, you haven't selected anything and the input looks like a value that may not be valid.  If you click outside of the Select control, the menu disappears.  This is standard html select behavior.  What isn't standard is that the input still shows the search term, which isn't a valid select value.  

Conversely, if you do the same steps above and instead of clicking, hit escape, the input resets to the default value that was selected before you began searching, which is what i would expect from an input with a limited number of options.

ooflorent:
Fix #490

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -440,6 +440,7 @@ var Select = React.createClass({
 		this._blurTimeout = setTimeout(() => {
 			if (this._focusAfterUpdate || !this.isMounted()) return;
 			this.setState({
+				inputValue: '',
 				isFocused: false,
 				isOpen: false
 			});


bcardiff:
+1

JedWatson:
Thanks @ooflorent!

Issue #460, opened 2015-09-18
Focus and expand the select list


approots:
![react-select-expand](https://cloud.githubusercontent.com/assets/235342/9973176/6b6afca6-5e23-11e5-8a5f-33f1cd11e12b.png)

When I focus the component instance (`this.refs.mySelectList.focus()`), the list will focus but not expand. Is there a way to also expand the list programatically?

jurgob:
+1

thebillg:
+1

olwen1987:
I figured out the solution for this. We need to set the this._openAfterFocus = true in the focus() method of select.js

gcedo:
+1

bjoernm:
+1, would be great to merge the solution of @olwen1987

bjoernm:
Is there a reason the solution of @olwen1987 is not being merged into master?

panetta-net-au:
The focus function is a little different now. I'm looking at version 1.0.0-beta12 and it references the property "openAfterFocus" to determine if focus() should open the drop down when focus() is called.

Try setting the openAfterFocus (not openOnFocus) prop to "true" and see if that has an effect. It still doesn't work for me but that could be due to Redux Form (I'm still investigating but might put it on the backburner for a bit).

panetta-net-au:
After more testing, I've decided that openAfterFocus works fine on 1.0.0-beta12 if you click on the Select Control. However, if the Select Control receives focus via clicking a label (with a ref) or by pressing tab, then it does not open the menu. Depending on what you're going for UX-wise, this might actually be the desired functionality.

To make the menu respect the openAfterFocus for any type of focusing, the Select control needs to be modified or monkey-patched. 

If you really want to monkey patch it, you can put this code in your Component that contains the Select Component. 
Note that `this._reactSelectRef` refers to the Select Component
All I've done is include `|| this.props.openAfterFocus` for the isOpen check.

  
      //HACK: Awful awful monkeypatching so that the handleInputFocus method respects the openAfterFocus property
      componentDidMount() {          
          var handleInputFocusMonkey = function(event) {
              var isOpen = this.state.isOpen || this._openAfterFocus || this.props.openAfterFocus;
              if (this.props.onFocus) {
                  this.props.onFocus(event);
              }
              this.setState({
                  isFocused: true,
                  isOpen: isOpen
              });
              this._openAfterFocus = false;
          };
          
[...]

PR #524, opened 2015-10-16
Focus and expand the select list #460

olwen1987:
**Issue:Focus and expand the select list #460**

This is the pull request to provide the feature of expanding the list when focus is invoked explicitly.
The solution is just setting ._openAfterFocus to true in the focus() method. This is the same action that takes place in handleMouseDown method . 

Please let me know if this can be part of the main code or i need to do more research to handle this issue.

Thanks,
Aaron

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -203,7 +203,7 @@ var Select = React.createClass({
 	},
 
 	focus: function() {
-		this._openAfterFocus="true";
+		this._openAfterFocus=true;
 		this.getInputNode().focus();
 	},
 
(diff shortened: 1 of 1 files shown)


JKillian:
Hi @olwen1987, I believe https://github.com/JedWatson/react-select/pull/891 implemented this feature, so you most likely can close this

JedWatson:
We have new props that enable this functionality. Thanks for the original PR @olwen1987

olwen1987:
Thanks @JedWatson

PR #527, opened 2015-10-19
Simplify control ui

JedWatson:
PR on behalf of @jossmac, who has kindly simplified and improved the DOM structure and CSS of the whole control so it should be more robust in all the various browsers and platforms.

In particular:

* no more `position: absolute` for inner controls
* `display: table` is used for layout, which works in IE8 and above, and [all other modern browsers](http://caniuse.com/#feat=css-table)
* less "magic numbers" used for layout, should fix various browser-specific alignment issues
* clear "x" control now animates in
* clearer `.Select--multi` className replaces `.Select.is-multi`
* new height & theme variables
* "dropdown" indicator chevron is no longer displayed for multi-select controls

Examples have also been retouched.

### DIFF
--- a/examples/src/app.js
+++ b/examples/src/app.js
@@ -38,7 +38,7 @@ ReactDOM.render(
 		<SelectedValuesField label="Clickable labels (labels as links)" options={FLAVOURS} hint="Open the console to see click behaviour (data/event)" />
 		<SelectedValuesField label="Disabled option" options={FLAVOURS_WITH_DISABLED_OPTION} hint="You savage! Caramel is the best..." />
 		<DisabledUpsellOptions label="Disabled option with a link"/>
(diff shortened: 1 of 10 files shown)


JedWatson:
For discussion, I've restored the dropdown arrow in `multi` mode, @jossmac I know you prefer it without because it gives more room for options but it's breaking the tests. I'm interested in whether anyone else thinks this should be a permanent change we make before we go refactoring them.

JedWatson:
All tests passing, I'm going to publish this now with a minor version bump (no breaking JavaScript / React API changes but if we consider the DOM structure itself as a public API, especially for custom stylesheets, this may have impacts)

As per SemVer `0.x` range standards, `minor` = "(potentially) breaking API change or significant update", `patch` = "new fix or feature"

PR #528, opened 2015-10-19
Set default value for input in loadAsyncOptions (fixes #439)

### LINKED ISSUE #439: loadAsyncOptions raises TypeError in setup (example included)
I have a setup in which the internal `loadAsyncOptions` of react-select raises a `TypeError`, which I think is a bug in react-select.

#### Setup description

Please have a look at the following setup: https://gist.github.com/dotcs/99041355fbae0d5c757c

I have defined two components. `RecipientSelect` is a wrapper around `Select` from react-select and makes use of its async call mechanism. `AsyncDemoComponent` uses `ReciepientSelect` together with one input field and a textarea to create some basic setup to write messages.

If I search for `Test` in the `RecipientSelect` component, I get the three options `Test 1`, `Test 2` and `Test 3` from which I then choose one. The click then triggers a `setState` call in `onSelectChangeHandler`.

#### What I expect

I expect is that the `name` of the selected entry is written in `this.state.recipients` of the `AsyncDemoComponent`.

#### What I get

The `Select`-Component raises a TypeError (although `this.state.recipients` is updated correctly):

> Uncaught TypeError: Cannot read property 'length' of undefined

```javascript
loadAsyncOptions: function loadAsyncOptions(input, state, callback) {
	var _this6 = this;
	//if (!input) { return; } // this would fix the problem

	var thisRequestId = this._currentRequestId = requestId++;
	if (this.props.cacheAsyncResults) {
		for (var i = 0; i <= input.length; i++) { 
        // ...
```

The exact line of the error is [this one](https://github.com/JedWatson/react-select/blob/master/lib/Select.js#L536).

#### First analysis

I have noted a possible fix in the example above, which I have used as a quick fix for the problem. But I think the actual problem is buried somewhere deeper in the code. I think the call to `loadAsyncOptions` isn't necessary at all at this point, which is why `input` is undefined which raises the error.

Interestingly in the setup (without my fix) every attempt to fill one of the other fields ends in the same error which seems to me that somehow this might be related to a problem with how `this` is treated in the component. But as far as I can tell I have linked all contexts correctly.

pancham348:


JedWatson:
:+1: thanks @pancham348

PR #529, opened 2015-10-20
Add possibility to render menu in another container

neoziro:
- To be able to use it in modal
- Or in an overflow: hidden container


neoziro:
@JedWatson I need this feature because I want to use it in a modal and I can't use only CSS to be out of the flow. I tried to respect your coding standard, I added an example and updated readme. If you are not OK with that feature tell me but I will have to fork it and create another select module... not very good.

I see coverage is no good, maybe some tests are missing, I can add it.

jsdf:
As an alternative to passing in a DOM node `menuContainer` to render the menu into, you could pass in a function which is called as a side effect of `render()`, taking the menu ReactElement tree as an arg. This would allow arbitrary rendering logic to be performed to render the menu, which seems like a more versatile option.

neoziro:
I will use another plugin, this branch is stale. I close it.

talbenbasat:
Hi @jsdf ,

I'm facing the same problem.
It's very hard to debug it since every time I try to see the style of the menu the menu dissapreas.

I believe it should be very simple to solve this using css.

Can you please help.
replacing the menu seems too complex solution for this simple problem

Thanks
Tal

talbenbasat:
Hi @neoziro ,

I managed to debug using F8 for pausing the chrome scripts.
this might do the trick https://css-tricks.com/popping-hidden-overflow/
another thing is using float

Tal

praveenaj:
For anyone still having this issue: 
https://github.com/JedWatson/react-select/issues/810#issuecomment-250274937

PR #539, opened 2015-10-26
IE-fix: don't blur if document.activeElement is menuDOM

### LINKED ISSUE #495: IE10/11: Menu closes when trying click+drag scroll indicator
In IE10 (Win8) and IE11 (Win8.1), when you try to click and drag the scroll indicator in the dropdown menu, the menu closes instead of scrolling. This behavior can be observed on the live demo page (http://jedwatson.github.io/react-select/) with the very first example, the "States" select.

petetnt:
Fixes #495.

Still missing a test though: took multiple different stabs to simulate the situation #495 creates when the scrollbar is clicked (`menuDOM` is the `document.activeElement`) in a testable fashion, but couldn't find anything (at least yet) that would actually test that behavior.

Any ideas for the test? The other tests pass just fine.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -437,6 +437,11 @@ var Select = React.createClass({
 	},
 
 	handleInputBlur (event) {
+		var menuDOM = ReactDOM.findDOMNode(this.refs.menu);
+		if (document.activeElement.isEqualNode(menuDOM)) {
+			return;
+		}
+
 		this._blurTimeout = setTimeout(() => {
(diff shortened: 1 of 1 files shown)


JedWatson:
I'm happy to merge this and get the fix out quickly, then circle back for the test. @bruderstein might have an idea on how to approach it.

JedWatson:
Thanks for this fix, by the way!

petetnt:
Cool :+1:

PR #542, opened 2015-10-27
Fix math operations to use less --strict-math=on

vincentfretin:


JedWatson:
Thanks @vincentfretin

PR #544, opened 2015-10-28
Convert LESS to SCSS and re-organise

subshad:
Convert LESS to SCSS using the less2sass gem. Fix references to some
mixins in the individual scss files, and allow the importing of the
variables separately from the components so an end user can override the
variables before use.

### DIFF
--- a/scss/components.scss
+++ b/scss/components.scss
@@ -1,0 +1,5 @@
+@import "control";
+@import "menu";
+@import "mixins";
+@import "multi";
+@import "spinner";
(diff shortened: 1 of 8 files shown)


subshad:
Hi :)

I'm not sure how/if you'd like to handle compiling the scss -> css since it seems like the LESS files are still your active stylesheet. This is probably really specific to my use case, but I needed to use this in an existing SASS ecosystem.

Use case is that I can now import the variables and over-ride them in my own sass file before importing the component styles.

I'm currently testing this locally, but it seems to be a 1-1 conversion. Is this an addition you're happy with?

JedWatson:
Thanks - I'm not sure how we'll approach maintaining both versions, but at the moment I'm glad to have this more friendly for SASS users!

PR #552, opened 2015-10-30
added a flag to preserve input value when options change

alexeib:
I was trying to use react-select as basically an autocomplete component that goes to the server and fetches search results as the user types. Unfortunately whenever results arrive, react-select removes the text that the user has typed in. This change preserves input that the user typed to filter/search until the selected value changes, or they decide to remove the text themselves


JedWatson:
I like this, do you think it should be on by default? Seems like clearing the input when new options arrive is more a bug than a feature.

alexeib:
I agree. I already added 2 additional check ins that basically remove the flag and preserve the user input until selection changes. I am travelling right now but once i am back i can flesh this out and add some tests if you'd like.

JedWatson:
Thanks for the PR @alexeib. I've taken this into account in the new version and from my testing it works properly in the beta, let me know if you have the same experience.

alexeib:
it works great now. thanks a lot

Issue #566, opened 2015-11-02
ClearValue Function is Bound to Multiple Click Events


Fireworks:
Currently the `clearValue` handler is bound to both `onClick` and `onMouseDown` events.  

https://github.com/JedWatson/react-select/blob/master/src/Select.js#L846

Binding it to `onClick` causes `mouseUp` events to also trigger a clear.  I'm having an issue with this because I am attempting to clear a value which causes reordering of my options in such a way that a new ReactSelect component (with a set value) appears in place of the cleared element.  Thus, the `mouseDown` will clear the first value, then a new component is rendered underneath the mouse pointer which causes the `mouseUp` to clear the second value as well.

This could be solved by simply removing the binding of `clearValue` to `onClick` on the line above.  I would PR this change directly, but I wanted to check to see if there was a specific reason the clear is also being bound to `onClick`.

JedWatson:
Resolved in the rewrite for 1.0, let me know if you find any issues in the new version related to this.

PR #569, opened 2015-11-03
Remove onClick clearValue

Fireworks:
I made an issue about this (https://github.com/JedWatson/react-select/issues/566) but went ahead and PR'd the change because I couldn't see a drawback to changing this and I need this fix ASAP.

To summarize: having the `clearValue` bound to onClick triggers clearing on a `mouseUp` event, causing issues in my case.

I know there is a rewrite in progress for 1.0, but I'm trying to avoid having to maintain a forked version for this minor change.


JedWatson:
@Fireworks Thanks for the PR! I've taken this into account in the rebuild, let me know if you have any issues with the new beta.

PR #577, opened 2015-11-06
Added !default to scss variables to more easily customize

naddeoa:
This will allow people to override the variables in their own scss files
without having to touch the source.


JedWatson:
Good call! Thanks @naddeoa.

PR #599, opened 2015-11-16
Fix keyboard nav for non-searchable multi controls

MrLeebo:
This is a simple fix based on @azaharakis's comments on #595.


JedWatson:
Thanks @MrLeebo

PR #609, opened 2015-11-20
Fix failing Value.js tests

MrLeebo:
* Fixes label onTouchEnd mistake (`this.props.handleMouseDown` should be `this.handleMouseDown`)
* Drops event propagation test (method no longer exists)
* Fixes other tests


JedWatson:
Thanks!

Issue #571, opened 2015-11-04
Get the tests working with 1.0.0


JedWatson:
There are a lot of failing tests at the moment. @bruderstein, if you've got time at the moment I could really use your help, it feels like a lot of them should be passing but there's something I'm missing (updates on render don't seem to complete before the DOM is queried, for one)

I've already fixed quite a few tests, updated others to use new class names, etc. and added some TODO notes where major API changes mean tests need to be rewritten.

bruderstein:
I'll see what I can do - stretched for time at the moment myself, between real-work and a new testing lib that should make these sort of tests easier. I'll take a look though, maybe there's some quick wins.

bruderstein:
OK, had a play around, and I think I understand the main issue :-)

The tests are still assuming the selected value is uncontrolled - ie. When we click on an option, that option is then displayed. Of course, that isn't the case any more. I'm inclined to remove a lot of those tests, as we need to check that `onChange` is correctly called, and when `value` is set, that the right option is displayed, but we shouldn't be checking that the selected value changes after the `onChange` is called, as that is no longer our responsibility (yay, by the way!)

I'll go through the tests and work out what doesn't make sense any more, and where we're maybe missing stuff. Prob be the weekend before I get chance though.

bruderstein:
Just a quick update - I've started to get somewhere with this. I'll send a PR as soon as I've got the bulk done.  One regression I've found so far is when `value` is set to a value that is not in the options, it doesn't select anything, previously it would use the value as the label. Not sure how you want to handle that now, but I've left the test in so you can decide.

nkbt:
I've started to debug tests too today =).

First one issue found:
```
UnexpectedError: 
expected
<div class="Select is-focused is-pseudo-focused is-searchable has-value" data-reactid=".l">
  <input data-reactid=".l.0" name="form-field-name" type="hidden" value="one">
  <div class="Select-control" data-reactid=".l.1">
    <div class="Select-value" data-reactid=".l.1.0">...</div>
    <div class="Select-input" data-reactid=".l.1.1" style="display: inline-block">...</div>
    <span aria-label="Clear value" class="Select-clear-zone" data-reactid=".l.1.3" title="Clear value">...</span>
    <span class="Select-arrow-zone" data-reactid=".l.1.4">...</span>
  </div>
</div>
queried for .Select-placeholder to have text 'Select...'
```

Legitimate issue since select has no value and does not render placeholder

```
[...]

PR #610, opened 2015-11-21
Fix tests 1.0

bruderstein:
This is where I am up to. Please don't merge this yet, this is just to keep public where I'm up to. Should hopefully finish this weekend.

### DIFF
--- a/examples/src/components/States.js
+++ b/examples/src/components/States.js
@@ -20,7 +20,8 @@ var StatesField = React.createClass({
 			country: 'AU',
 			disabled: false,
 			searchable: this.props.searchable,
-			selectValue: 'new-south-wales'
+			selectValue: 'new-south-wales',
+			clearable: true,
 		};
 	},
 	switchCountry (e) {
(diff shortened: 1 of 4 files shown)


bruderstein:
This will be the fix for #571

bruderstein:
@JedWatson This is ready to go now - together with #609 from @MrLeebo, we're back to green \o/  - Thank you @MrLeebo for the PR!

I think there's only two bug fixes - 9767a8d and 4d5eaf5

The rest, as usual, I've tried to keep each change or fix in a separate commit, so it's possible to see why each change was made. Hope that's ok.

I'll open a new issue for the async tests, as I'd like to do those separately. At least we're back to green, and can work from there. There's probably one or two other bits that need tests now (`simpleValue={true}`, for example), but I'll add those as I see them.

JedWatson:
Brilliant - thank you so much @bruderstein and @MrLeebo!

I promise not to do any more major test-breaking refactoring on react-select this year :wink:

MrLeebo:
@JedWatson @bruderstein Glad to see this done. When I made my PR I flirted with fixing the Select tests, but I quit when I realized how much work that was going to be.

Would you say this merits an NPM release?

bruderstein:
@MrLeebo I don't think there's any pressing reason to. There are only two tiny fixes, that fix edge cases (neither of which are serious), so can easily just be lumped in with the next release.  Not sure when I'll get the async tests done, but when they're in, they may turn up some stuff, which would may make a release worth it.

PR #611, opened 2015-11-21
Clicking on scrollbar will hide the options (a bug reappear?).

albertoclarit:
this.handleMouseDownOnMenu might be an oversite. Changing to handleMouseDown will solve the issue


gwyneplaine:
Thanks for this @albertoclarit, the root behaviour addressed by this PR has been fixed in the following PR https://github.com/JedWatson/react-select/pull/408

PR #620, opened 2015-11-23
Add clearableValue option for Multi-Select

### LINKED ISSUE #458: Question: Disabling removal of items in multiselect
Is there a way to disable the removal of an individual value in multiselect mode?  I have a use case where a user can select multiple things, but some items are locked in, meaning they cannot be removed from the list.  I'd really need two things here:

1. Disable or hide the "X" when the item is in the list of current values.
2. Apply a different class to a disabled value so that I can style it differently.

slybridges:
Fixes #458

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,14 @@ You can enable multi-value selection by setting `multi={true}`. In this mode:
 * The `onChange` event provides an array of the selected options as the second argument
 * The first argument to `onChange` is always a string, regardless of whether the values of the selected options are numbers or strings
 * By default, only options in the `options` array can be selected. Setting `allowCreate` to true allows new options to be created if they do not already exist.
+* By default, selected options can be cleared. To disable the possibility of clearing a particular option, add `clearableValue: false` to that option:
+```javascript
+var options = [
+	{ value: 'one', label: 'One' },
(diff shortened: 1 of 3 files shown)


slybridges:
On this PR I've added a note in the `README` that the `clearableValue` should only be used along with the `clearable` prop set to `false`.

This is because if we want not to clear the options that have `clearableValue=false` when pressing the `Clear All`button, we have to iterate over each `valueArray` all the time which I felt could have some perf issues.

slybridges:
@jedwatson - anything you'd need from me to get this merged ?

JedWatson:
Sorry for the delay @slybridges - this looks great, I'm merging it now.

slybridges:
:+1: @JedWatson.
No worries for the delay, I understand open source can be tough on you sometimes ;)

shivamjjha:
Seems this is removed in latest version of react-select ?
https://codesandbox.io/s/codesandboxer-example-forked-pn5c9c?file=/example.tsx

vivekrekhadia:
> Seems this is removed in latest version of react-select ? https://codesandbox.io/s/codesandboxer-example-forked-pn5c9c?file=/example.tsx

How you solved it then, I am having the same issue.

PR #621, opened 2015-11-23
Pressing Enter on open Select should stop propagation

MrLeebo:
I don't expect pressing Enter on an open menu to bubble back to the Select's container.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -247,6 +247,7 @@ const Select = React.createClass({
 			break;
 			case 13: // enter
 				if (!this.state.isOpen) return;
+				event.stopPropagation();
 				this.selectFocusedOption();
 			break;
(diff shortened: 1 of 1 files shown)


MrLeebo:
@JedWatson Is this OK?

MrLeebo:
@JedWatson This trivial one-liner prevents pressing Enter on an open menu from also submitting the form that the Select is a part of. May I ask what I can do to expedite it being merged? I am forced to maintain a fork of react-select until it does.

JedWatson:
Sorry for the delay @MrLeebo, and thanks for the fix! I'll release an update with this included tonight.

MrLeebo:
@JedWatson Awesome, thank you!

PR #630, opened 2015-11-26
Fixing access to asynchronously loaded data

mkotelnikov:
This construction "options: data && data.options || []," does not work (it provides null).
It was changed to "options: !!data ? data.options || [] : [],"


gwyneplaine:
Thanks @mkotelnikov, as far as I can see the existing code is now working as expected, and does not produce a null case. 

PR Closed.

PR #631, opened 2015-11-26
Change input back to div

azaharakis:
This essentially reverts the change made here
https://github.com/JedWatson/react-select/pull/595 By applying a tabIndex on a
div element allows it to be focused. This was the reason the div was changed to
an input, by putting it back to a div mobile devices display consistent behaviour
when the select box recieves focus. IOS would display a chrome at the bottom
of the page where as chrome wouldn't.


JedWatson:
Thanks @azaharakis, that's much cleaner.

PR #632, opened 2015-11-26
Feature to scroll the viewport to display the menu

azaharakis:
Our project required the menu of an enabled dropdown to always be in view when the menu is activated. This PR introduces this feature. The viewport will only be shifted if the menu isn't in focus. 

We discussed different implementations such as the dropdown menu displaying above the select box. This implementation required the least amount of change to the code.

The implementation also uses a prop to calculate how much space to place below the menu and the viewport base. We discussed whether or not this should be calculated dynamically from css rather then a prop (incorporating any margin-bottom thats applied to .Select-menu) as this is arguably a styling concern but this semantically might not make sense as margins don't apply to elements taken out of document flow. 

If you are interested in this feature but would like some changes please let us know!

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,8 @@ For multi-select inputs, when providing a custom `filterOptions` method, remembe
 	labelKey	|	string	|	'label'		|	the option property to use for the label
 	matchPos 	|	string	|	'any'		|	(any, start) match the start or entire string when filtering
 	matchProp 	|	string	|	'any'		|	(any, label, value) which option property to filter on
+	scrollMenuIntoView |	bool	|	true		|	whether the viewport will shift to display the entire menu when engaged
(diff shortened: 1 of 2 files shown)


JedWatson:
Looks good @azaharakis but I think it would be best if we make this optional, and enable the behaviour with a prop, maybe `scrollMenuIntoView`?

JedWatson:
Thanks @azaharakis!

fritz-c:
This is a really handy feature, however it doesn't seem to work when the select element is inside a scrolling container within the page. Something like:
scroll
Perhaps offer an option to specify the parent to scroll?

xabikos:
The auto scroll could be a solution but as I mentioned in the issue I created using a library like [tether](https://github.com/HubSpot/tether) is probably better choice as it makes sure the drop down menu is always shown in the correct place

Issue #635, opened 2015-11-30
bower.json incorrectly pointing to "dist/default.css", file doesn't exist


omgaz:
`react-select/bower.json` references `dist/default.css` in `main`, when the actual path is `dist/react-select.min.css`.

JedWatson:
Fixed by #636

PR #636, opened 2015-11-30
Fix bower.json css dist path

omgaz:
- From `dist/default.css` to `dist/react-select.min.css`

Closes issue: https://github.com/JedWatson/react-select/issues/635


JedWatson:
Thanks @omgaz!

PR #652, opened 2015-12-07
Fix a problem with border inside table

### LINKED ISSUE #619: Border is omitted when rendering in table-element
When wrapping the element in a table cell:
```
<table>
  <tbody>
    <tr>
      <td>
        <Select
          options={options}
          onChange={value => this.handleChange(value)} />
      </td>
    </tr>
  </tbody>
</table>
```
The border of the `div.Select-control` is not drawn. This can be fixed via removing `display: table` from the mentioned class which obviously breaks the control in other places. Is this an known issue with a fix or a more fundamental problem?

Best,
  Bastian

### LINKED ISSUE #646: react-select breaks in tables that have their own border-spacing
Because of the new `display: table;` handling react-select breaks in some situations where react-select itself is put inside a table.

Namely if a surrounding table has its own border-spacing, react-select will inherit this and end up like this:

screen shot 2015-12-03 at 1 00 24 am

.Select-control should explicitly declare the `border-spacing: 0; border-collapse: separate;` it requires for its styles.

rwrz:
Fixing a problem where the border isn't showing when the element is inside a table.

fixes #646, fixes #619

### DIFF
--- a/scss/control.scss
+++ b/scss/control.scss
@@ -39,6 +39,8 @@
 	color: $select-text-color;
 	cursor: default;
 	display: table;
+	border-spacing: 0;
+	border-collapse: separate;
 	height: $select-input-height;
 	outline: none;
 	overflow: hidden;


JedWatson:
Thanks @rwrz!

PR #656, opened 2015-12-09
Check if menu ref is not null

gaastonsr:
This fix should go to the commit with tag v0.9.1.

This fix is related to #539. I couldn't narrow the exact error but my function handleInputBlur is called twice, one with this.refs.menu set to a div element and the other to null.

Tested on IE11.

### DIFF
--- a/lib/Select.js
+++ b/lib/Select.js
@@ -448,7 +448,7 @@ var Select = React.createClass({
 		var _this7 = this;
 
 		var menuDOM = ReactDOM.findDOMNode(this.refs.menu);
-		if (document.activeElement.isEqualNode(menuDOM)) {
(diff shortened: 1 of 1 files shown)


JedWatson:
@gaastonsr you mean this is a fix for the 0.9.1 version? or the new 1.0 version in Master?

Unfortunately we've got conflicts, If you can update your branch I'll merge it in.

gaastonsr:
This is a fix for 0.9.1 version. Because as far as I can see, master doesn't have the changes introduced by #539.

If you merge it with the [v0.9.1 commit](https://github.com/JedWatson/react-select/commit/d8fe544f32dc28e1f72592671f7c70bddae3258a) you shouldn't have problems.

gaastonsr:
Closing this because I updated the wrong file...

PR #657, opened 2015-12-10
Missing onChange prop prevents setting the value.

kromit:
`onChange` prop should not be mandatory. Not everybody want to track the change event.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -329,11 +329,12 @@ const Select = React.createClass({
 	},
 
 	setValue (value) {
-		if (!this.props.onChange) return;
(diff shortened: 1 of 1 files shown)


JedWatson:
I'm not sure I understand this change @kromit -

1. The component is now [controlled](https://facebook.github.io/react/docs/forms.html#controlled-components), so if you don't provide an `onChange` the select won't work
2. It already isn't mandatory - this just moves the early return to an if statement; which means a local variable is always calculated but then nothing is done with it.

kromit:
yeah my bad. forget it.

PR #659, opened 2015-12-10
fixed importing of non existing style in package.json

vzaidman:


gwyneplaine:
Thanks for this @vzaidman this has since been fixed by https://github.com/JedWatson/react-select/commit/84a74505d7c74e3afa41c7e80783e6638e6a287c

Issue #649, opened 2015-12-04
Microsoft Edge / IE11 Focused Fields


hnordt:
Hello.

I'm using React Select `v1.0.0-beta6`.

In IE11 and:

![captura_de_tela_120415_024412_pm](https://cloud.githubusercontent.com/assets/1625399/11595220/7c95e104-9a95-11e5-8d35-f72be5031e0b.jpg)

When I selected a value, looks like field always stay focused:

![captura_de_tela_120415_024505_pm](https://cloud.githubusercontent.com/assets/1625399/11595247/9c8085e6-9a95-11e5-95ce-d489db48bc44.jpg)

The problem is that in Microsoft Edge / IE11 the onBlur event doesnt dispatches.

cloudkite:
@hnordt most likely related to https://github.com/JedWatson/react-select/pull/661

JedWatson:
I really need to get a Windows machine :grimacing:

Can anyone confirm what's going on between this issue and #661?

gaastonsr:
I just verified and #661 fixes this.

The problem is when the blur handler is called on IE it fails because isEqualNode expects a DOM node and it receives an undefined/null object. Which in modern browsers instead of fail it returns false.

JedWatson:
Fixed by #661

PR #661, opened 2015-12-10
Fix blur handler on IE where menu ref is undefined

gaastonsr:
On IE when the input is blurred it fails because on isEqualNode does not accept undefined or null values.

This re-introduces the fix on #539 and checks menuDOM is not undefined/null.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -208,9 +208,12 @@ const Select = React.createClass({
 	},
 
 	handleInputBlur (event) {
-		if (document.activeElement.isEqualNode(this.refs.menu)) {
-			return;
-		}
+ 		let menuDOM = ReactDOM.findDOMNode(this.refs.menu);
+
+ 		if (menuDOM && document.activeElement.isEqualNode(menuDOM)) {
+ 			return;
+ 		}
+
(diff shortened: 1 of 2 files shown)


src/Select.js
```diff
@@ -208,9 +208,12 @@ const Select = React.createClass({
 	},
 
 	handleInputBlur (event) {
-		if (document.activeElement.isEqualNode(this.refs.menu)) {
-			return;
-		}
+ 		let menuDOM = ReactDOM.findDOMNode(this.refs.menu);
```
cloudkite: @gaastonsr findDOMNode is unnecessary in react 0.14 as this.refs.menu is a DOM Node. see https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html

gaastonsr: Thanks, I'll update this and check if I can fix #649.

gaastonsr:
I updated the branch with the fixes pointed out by @cloudkite and verified it fixes #649.

JedWatson:
Thanks @gaastonsr! really appreciate this :)

gaastonsr:
Thank you for merging. :)

PR #665, opened 2015-12-12
Check menu node is thruthy when calling isEqualNode

cloudkite:
calling isEqualNode with undefined in IE9 causes browser to throw `no such interface supported`


cloudkite:
Sorry just noticed this is a duplicate of https://github.com/JedWatson/react-select/pull/661

PR #668, opened 2015-12-14
Allow customization of menu option background color

Naveg:
Introduces a new variable, $select-option-bg that defaults to the value
of $select-input-bg to maintain existing behavior.

### DIFF
--- a/less/menu.less
+++ b/less/menu.less
@@ -38,6 +38,7 @@
 
 .Select-option {
 	box-sizing: border-box;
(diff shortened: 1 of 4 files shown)


JedWatson:
Thanks @Naveg - would you mind adding the corresponding variable to the LESS stylesheet as well? We need to keep them both in sync.

Naveg:
fixed

JedWatson:
Great! thanks :+1:

PR #670, opened 2015-12-14
added inputValue property

vzaidman:
fixed #669

### DIFF
--- a/lib/Select.js
+++ b/lib/Select.js
@@ -199,6 +199,9 @@ var Select = React.createClass({
 				setState();
 			}
 		}
+		if(newProps.inputValue){
+			this.setState({inputValue: newProps.inputValue});
+		}
 	},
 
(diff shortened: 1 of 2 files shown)


vzaidman:
so now you can pass 
<state 
  inputValue={this.state.inputValue}
  onInputChange={ this,setState({ inputValue, isLoading:true }) }
/>

wereHamster:
This seems like a really useful and straightforward change. What's the holdup?

mmmveggies:
Do you think you could rebase this PR?
If not, I will open a new one - this would make using `<Select.Async />` better when having large result sets because `loadOptions` gets called with `''` when the component re-mounts, so if your selection is from a query that starts with `xx` the `value` does not match up with the new `options` and the placeholder is displayed, which is erroneous.

vzaidman:
I didn't use this library for a long time and don't remember what's going here.

PR #675, opened 2015-12-16
v0.6.x - Keep multi-select open after selection

Jhays10:
This pull request is to keep the drop-down open for multi-selects after a selection. If you could make a release for the v0.6.x branch, that would be great. I fixed the tests to replicate the dropdown staying open. If there is anything else you need, let me know.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -308,7 +308,9 @@ var Select = React.createClass({
 			this._focusAfterUpdate = true;
 		}
 		var newState = this.getStateFromValue(value);
-		newState.isOpen = false;
+		if (!this.props.multi) {
+			newState.isOpen = false;
+		}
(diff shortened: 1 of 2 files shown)


Teresast:
We have faced same issues, hope this PR will be merge soon

Jhays10:
@Teresast We just ended up using my fork version: 
https://github.com/Jhays10/react-select/tree/fix/keep-multi-open

It's the most current until v0.6 gets updated.

gwyneplaine:
This functionality has since been implemented in the following commit https://github.com/JedWatson/react-select/commit/c7fe6f44763d680313ef7620e3c5373c87b8e319

you can opt out of the default behaviour by setting the with a closeOnSelect prop  to false.

PR #677, opened 2015-12-17
Expose focus() in the Async component. Fix #676

### LINKED ISSUE #676: <instance>.focus() is not available from the Async component
Right now, the focus method is only exposed in the <Select> component, but not in the Async one. I'll submit a PR to fix that.

AugustinLF:
I didn't add any test, didn't really know how to do so, since the `<Select />.focus()` method is not tested either. If someone (@bruderstein?) can advise me on that, I'll gladly do so.

I've seen that you use string refs in this project, however, the React team considers this approach as [mostly legacy](https://facebook.github.io/react/docs/more-about-refs.html) now. So I used one of their [example](https://facebook.github.io/react/docs/more-about-refs.html#the-ref-callback-attribute) as a reference.

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -84,6 +84,9 @@ const Async = React.createClass({
 			});
 		}
 	},
+	focus () {
+		this._select.focus();
+	},
 	resetState () {
 		this._currentRequestId = -1;
 		this.setState({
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks for catching this @AugustinLF!

Interesting - I hadn't noticed that comment in the docs about string references being "mostly legacy". It seems like a more performant approach when you don't need to use the function refs, as you're not creating additional functions in the render cycle... I might update this one for consistency (we could easily transition them all at once) but I'm going to learn more about the background here.

PR #684, opened 2015-12-20
Fixes input focus bug when toggling disabled prop

### LINKED ISSUE #683: Toggling Disabled Prop
Seems as though there is a bug if you have the component go from disabled={false} to disabled={true} and then back to disabled={false}.  It's related to the fact that this.state.isFocused doesn't get switched to false when the input becomes disabled. Therefore when you undisable it, you cannot click the input because it thinks it is still focused. I'll try to make a pull request later tonight.

dcurletti:
fixes #683


JedWatson:
Thanks @dcurletti!

Issue #671, opened 2015-12-14
Clicking clear on adjacent input interrupts onBlur


etburke:
I have a form: 
![screen shot 2015-12-14 at 1 53 00 pm](https://cloud.githubusercontent.com/assets/436070/11795250/93e42d52-a26a-11e5-9c02-0d83e8f49a00.png)
Where the flight number field is disabled unless an airline value is selected. When I select the airline field, open the flight number field, and then click the airline clear X the flight number field is disabled but left open.
![screen shot 2015-12-14 at 1 53 20 pm](https://cloud.githubusercontent.com/assets/436070/11795304/d01203f8-a26a-11e5-8001-7dd68d052c75.png)
![screen shot 2015-12-14 at 1 53 31 pm](https://cloud.githubusercontent.com/assets/436070/11795314/d51031d6-a26a-11e5-963c-7a8491a55007.png)
The flight number field is left unresponsive, even if the disabled prop is toggled back to false.

JedWatson:
Fixed by #686

PR #686, opened 2015-12-21
Removed event.stopPropagation() from Select's clearValue

etburke:
See Issue #671. 

It wasn't clear to me why both stopPropagation and preventDefault were being called on this event. Removing stopPropagation allowed the event to make it to the adjacent field's blur providing the expected behavior. The examples continue to work as expected and the tests all pass. If stopPropagation is necessary for a case that I'm unfamiliar with I'd be happy to try a different approach. Thanks for the awesome components!

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -390,7 +390,6 @@ const Select = React.createClass({
 		if (event && event.type === 'mousedown' && event.button !== 0) {
 			return;
 		}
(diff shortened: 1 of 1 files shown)


JedWatson:
I've looked into this and I agree @etburke - this line doesn't seem to be necessary and can have unintended side-effects. Merging, thanks for the PR!

PR #700, opened 2016-01-08
Update README onOptionLabelClick name change

panw:
Update README onOptionLabelClick with function name change


JedWatson:
Thanks for catching that @panw

panw:
@JedWatson happy to help. thanks for the great component.

PR #708, opened 2016-01-14
Enter selects and retains focus, Tab selects and shifts focus

### LINKED ISSUE #643: Question: Why press tab select the current choice?
I am using 1.0.0-beta6 with a multi value select. The select is part of a form displayed in a modal dialog.
I want to be able to navigate and fill my form only with the keyboard.

When tabulation is press, currently the current choice is added to the list of values (see https://github.com/JedWatson/react-select/blob/master/src/Select.js#L242). It's seems a weird behavior to me as we usually expect tab to move the focus to the next input/selectable element.
Do we have a good reason for this current behavior or a PR is welcome to change it?

I my specific case I cannot use escape to close the choices list as it aslo close the modal dialog and I want to keep this behavior for consistency with the other modal.

rdsubhas:
Hi, the Tab behavior is really confusing users, since in all other Form Controls, "Tab" will switch focus, but for this one single control alone a user has to press Tab twice. Almost every user so far has asked "Why?" when I've asked them to press tab twice here.

I propose a simple enhancement (or rather simplification):

* Enter selects and keeps focus
* Tab selects and shifts focus

Simple and clear. That's how all forms work. Even goodness Excel and LibreOffice!

Fixes #643

### DIFF
--- a/lib/Select.js
+++ b/lib/Select.js
@@ -291,7 +291,7 @@ var Select = _react2['default'].createClass({
 					return;
 				}
 				this.selectFocusedOption();
-				break;
+				return;
 			case 13:
 				// enter
 				if (!this.state.isOpen) return;


JedWatson:
@rdsubhas I'm happy with that logic. Good improvement. Thanks!

rdsubhas:
Thanks @JedWatson for merging, and for the great component!

fhelwanger:
Seeing this PR, the change was made on lib/Select.js. Shouldn't it be src/Select.js?

I'm afraid it'll be overwritten on next build.

rdsubhas:
Thanks @fhelwanger! I've raised a new PR

fhelwanger:
Great! Thanks!

vsverlov:
Hello,
I see that 'Tab selects and shifts focus' functionality disappeared in react-select v. 2.4.1.
Is it a bug? or there is a reason why functionality was changed?

gilcardenas:
Also seeing this issue on  v. 2.4.3

sgoyal100:
do we have solution for this yet ?
to make one Tab - select and move focus to next control ?

kartik2014:
Hi @JedWatson : could you please let us know if "Enter selects and retains focus, Tab selects and shifts focus" functionality was turned off in recent versions of this library because I can see it is not working .
Is there any workaround I can do to make it work with the latest version?

kartik2014:
> do we have solution for this yet ?
> to make one Tab - select and move focus to next control ?
> using 3.0.3 version
Did you find any solution to this yet?

shlgug:
According to the original commit it seems this line should be changed from "break" to "return".
https://github.com/JedWatson/react-select/blob/9606a92fc236d00d79fa6e7b4b021e74118f8ac2/packages/react-select/src/Select.js#L1207

kartik2014:
@JedWatson @fhelwanger

srevenant:
Can we re-open this?  There is now a regressed feature that people want to see fixed.  What is a workaround in the meantime?

kartik2014:
I can see this pull request is merged. But a new version of this library hasn't been released. Do we know when it will be released?

kartik2014:
Please help me guys. Until the new version is released , what is the workaround?

shlgug:
Use version 2.4.4 - it works for me.

kartik2014:
Hi @shlgug  : went back to version 2.4.4 still have to tab twice to select a value and move to next field

shlgug:
Actually I think going back to 2.4.4 solved a different issue for me - namely selecting current value not working correctly.  Here is the code for the workaround I used (using React Hooks):

`
import React from 'react'

let tabPressedTarget = undefined;

const handleKeyDown = event => {
    if (event.keyCode === 9) {
        tabPressedTarget = event.target;
        setTimeout(() => {
            tabPressedTarget = undefined;
        }, 100)
    } else {
        tabPressedTarget = undefined;
    }
};

const focusNextElement = currentElement => {
    //Put this in a timeout to prevent an issue where the focus was sometimes going to the previous element
    setTimeout(() => {
        const allElements = document.querySelectorAll('input, button, a, area, object, select, textarea, [contenteditable]');
        const currentIndex = [...allElements].findIndex(el => currentElement.isEqualNode(el));
        allElements[currentIndex + 1].focus();
    }, 100);
};

const MyComponent = props => {
    const handleSelectChanged = data => {
        if (tabPressedTarget) {
            //A single press of the tab key should cause moving to the next element besides the default functionality of selecting the currently highlighted item
            focusNextElement(tabPressedTarget);
        }
        valueChanged(data);
    };

    return (
        <Select
            onKeyDown={handleKeyDown}
            onChange={handleSelectChanged}
        />
    )
}
`

leikoilja:
Inspired by @shlgug solution, here is my component wrapper:
```jsx
// AsyncCreatableSelectPatched.tsx
import AsyncCreatableSelect from "react-select/async-creatable";

let tabPressedTarget: HTMLElement | undefined = undefined;

const handleKeyDown = (event: any) => {
  if (event.keyCode === 9) {
    tabPressedTarget = event.target;
    setTimeout(() => {
      tabPressedTarget = undefined;
    }, 100);
  } else {
    tabPressedTarget = undefined;
  }
};

const focusNextElement = (currentElement: any) => {
  // Put this in a timeout to prevent an issue where the focus was
  // sometimes going to the previous element
  setTimeout(() => {
    const allElements = document.querySelectorAll(
      "input, button, a, area, object, select, textarea, [contenteditable]"
    );

    // @ts-ignore
    const currentIndex = [...allElements].findIndex((el) =>
      currentElement.isEqualNode(el)
    );
    // @ts-ignore
    allElements[currentIndex + 1].focus();
  }, 100);
};

export default function AsyncCreatableSelectPatched(props: any) {
  const handleSelectChanged = (data: any, originalOnChangeHandler: any) => {
    if (tabPressedTarget) {
      //A single press of the tab key should cause moving to the next element
      // besides the default functionality of selecting the currently highlighted item
      focusNextElement(tabPressedTarget);
    }
    originalOnChangeHandler(data);
  };
  return (
    <AsyncCreatableSelect
      {...props}
      onKeyDown={handleKeyDown}
      onChange={(data) => handleSelectChanged(data, props.onChange)}
    />
  );
}
```
You can replace `AsyncCreatableSelect` with `AsyncSelect` or `Select` as you please with corresponding props
And the usage:
```
      <AsyncCreatableSelectPatched
        cacheOptions
        defaultOptions
        defaultValue={{
          label: "Search or create",
          value: initValue,
        }}
        className="w-full"
        onChange={handleValueChange}
        loadOptions={...}
      />
```

Issue #716, opened 2016-01-18
Suggestion: Call `onMenuScrollToBottom` when result set length reduced by filter.


rhys-vdw:
Version: v1.0.0-beta-8.

I'm using the undocumented prop `onMenuScrollToBottom` in combination with `isLoading`. Which seems the natural place to hook into my Redux store.

The idea here being that whenever the end of the result set is reached, I dispatch a 'fetch more items' action with the correct search term. This temporarily switches the `isLoading` flag and ultimately updates the `options` prop and resets `isLoading` to `false`.

The query to grab more options has a search parameter. Ideally I'd just fire off this query whenever the end of the results is visible, which works in most situations. However if there are only a few options and the result set is not scrollable, changing the filter to match a subset of these does not trigger the event (since the check is done on [`handleMenuScroll`](https://github.com/JedWatson/react-select/blob/v1.0.0-beta8/src/Select.js#L313-L315)).

I suggest having a test to see if the bottom of the menu is visible after any change to the number of results (in `componentDidUpdate`) and firing off the same event.

Happy to contribute a PR.

bladey:
Hello @rhys-vdw,

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest versions.

If you feel this issue is still relevant and you'd like us to review it, please leave a comment below.

bladey:
Hello @rhys-vdw -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you feel this issue / pull request is still relevant and you'd like us to review it, please leave a comment and we'll do our best to get back to you.

PR #727, opened 2016-01-21
General optimizations

rhys-vdw:
Closes #722 and #719. Includes #716.

Sorry, this is not a nice atomic PR. Also I have not updated tests as I had an older version of node installed and could not justify the time overhead at work. I wouldn't necessarily expect this to be merged as is, but I'd highly recommend taking on some or all of the changes here.

I would consider this library to be essentially unusable before these changes. Previously this library would re-render every element in the menu every time the mouse position changed while hovering. To make matters worse, it would re-run the filter every render!!! I'm not sure if this is an issue outside of the beta releases, but I'm shocked nobody had picked up on this.

Now only the minimum number of elements are re-rendered when and only when they need to be re-rendered, and only the ones that have changed are re-rendered. The filter is only run when its dependencies are modified, and will not be run every render.

I have only done manual testing with the following props, so it's possible that I've broken other features.

``` js
      <ReactSelect
        filterOptions={this.filterJobs}
        isLoading={isFetchingJobs}
        onChange={this.handleChange}
        onInputChange={this.handleSearchInputChange}
        onMenuScrollToBottom={this.handleMenuScrollToBottom}
        optionRenderer={this.renderOption}
        options={jobs}
        placeholder={placeholder}
        value={selectedJobId}
        valueKey='id'
        valueRenderer={this.renderValue}
      />
```


JedWatson:
@rhys-vdw thanks for this - all very valid points, and things that were missed in the rewrite late last year. Really appreciate the thorough review. I'm going to work through the various fixes you've included and test them with the other cases and get a new release out asap.

For simplicity I might cherry-pick them rather than merging the PR as is, but will see how it goes!

rhys-vdw:
@JedWatson Hopefully it can be of some help.

One of theother optimizations is not using `props.children`, as this necessarily requires rendering before the `children` value is passed to the component and its `shouldComponentUpdate` is checked. Instead I pass the `renderOption` callback as a prop, so that the invocation is deferred until after `shouldComponentUpdate` returns `true`.

Two big remaining problems are inability to click on the scrollbar in the dropdown (it dismisses the menu instead), and inability to click and drag to select text. I'm putting those on my TODO list for my fork, but will try to push them upstream to this repo first.

Also I suggest removing the compiled output from version control and adding an `npm run build` to the `prepublish` script.

wmertens:
One more optimization, for large option lists, would be using something like https://github.com/seatgeek/react-infinite in the option menu

rhys-vdw:
@wmertens Yes, that looks great.

JedWatson:
Thanks again for this @rhys-vdw. With the release of 1.0.0 I'm closing a lot of old PRs and Issues that either may be fixed / or won't be merged given the major changes since. If there's anything left here that you're interested in bring forward, please feel free to open new (atomic 😉) Pr's

Issue #717, opened 2016-01-18
Suggestion: add showNoResultsText


fhelwanger:
We'd like to be able to **not** show "No results found" text.

I think this could be done implementing a new bool prop: `showNoResultsText`.

What do you think? If this is OK, I can make a PR to add it.

fhelwanger:
I was thinking... maybe it'll be simpler to just use the existing `noResultsText` prop.

If it's a falsy value, don't show the text.

Thoughts?

PR #730, opened 2016-01-21
Hide noResultsText when value is falsy

fhelwanger:
When the `noResultsText` prop is a falsy value, don't show it when typing.

References #717


JedWatson:
Nice, @fhelwanger, thanks!

PR #731, opened 2016-01-22
feature(.props.required): add HTML5 required attribute capabilities

dmatteo:
This PR adds the ability to specify a prop named `required` on the `Select` component that make use of the HTML5 required attribute in order to ensure that at least one input has been submitted.

The logic is pretty straightforward: I basically add and remove the `required` attribute from the `Input` component of `react-input-autosize`, which in turn applies it and removes it from the real `<input />`.

It obviously defaults to `false`, and I've added quite a few tests in order to make sure I didn't break any existing behaviour and that the thing I wrote actually works.

You can see a live example using the new feature here:
https://nextpodio.dk/webforms/14728164/987369

**Note:**
In the specs that are testing mutations (adding and removing values) I couldn't test on the real HTML attribute because React doesn't update it, when using `TestUtils` to change the node.
I had to resort on testing `.state.required` which is suboptimal, but better than nothing

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -75,6 +75,7 @@ const Select = React.createClass({
 			React.PropTypes.string,
 			React.PropTypes.node
 		]),                                         // field placeholder, displayed when there's no value
+		required: React.PropTypes.bool,             // applies HTML5 required attribute when needed
 		searchable: React.PropTypes.bool,           // whether to enable searching feature or not
 		simpleValue: React.PropTypes.bool,          // pass the value to onChange as a simple value (legacy pre 1.0 mode), defaults to false
 		style: React.PropTypes.object,              // optional style to apply to the control
@@ -112,6 +113,7 @@ const Select = React.createClass({
 			onBlurResetsInput: true,
 			optionComponent: Option,
 			placeholder: 'Select...',
+			required: false,
 			searchable: true,
 			simpleValue: false,
 			valueComponent: Value,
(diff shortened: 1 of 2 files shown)


JedWatson:
Looks good to me @dmatteo, thanks!

dmatteo:
Well that was fast :)

Thanks!

colepacak:
Just what I needed. Thanks, @dmatteo and @JedWatson.

dmatteo:
Hey @JedWatson, how about a `beta-9` to get this out of the door?

snackycracky:
I just wanted to use this. Please release :+1:

src/Select.js
```diff
@@ -320,6 +323,11 @@ const Select = React.createClass({
 		}
 	},
 
+	handleRequired (value, multi) {
+		if (!value) return true;
+		return (multi ? value.length === 0 : Object.keys(value).length === 0);
```
dayletter: @dmatteo @JedWatson Why check if a single value has any keys here? It breaks when the value is a number.

dmatteo: multi value is an array, while single value is an Object like `{ label: 'Whatevz', value: 'Bananas' }`

I guess there is one case I may have missed, and that is when `simpleValue === true`. If I remember correctly, in that situation value can be a number or string.

If you create a bug please ping me and I'll be happy to look into that.

PR #735, opened 2016-01-25
Add touch & drag  behaviour

### LINKED ISSUE #244: Scroll on touch screen on top of a select will open it.

Paveltarno:
This fixes #244.
Iv'e noticed PR #245 but i'm not a big fan of introducing dependencies in core infrastructure components.
This is a rather clean an minimal fix that  addresses the same issue in 3 situations:
- Dragging over the input won't toggle the options menu
- Dragging over the 'clear' icon won't clear the selection
- Dragging over the remove icon of a Value won't remove the value

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -161,6 +161,34 @@ const Select = React.createClass({
 		this.refs.input.focus();
 	},
 
+	handleTouchMove (event) {
+		// Set a flag that the view is being dragged
+		this.dragging = true;
+	},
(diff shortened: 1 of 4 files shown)


JedWatson:
Looks good @Paveltarno, thanks!

Issue #709, opened 2016-01-14
Option to preserve inputValue after select item in multiselect


nosovsh:
Use case: user types search text in text field to filter items in multiselect. Then he wants to select few of filtered items(more than one). In current implementation dropdown will be closed after selecting one of the items so user has to type search text again.

What about adding an option "preserveInputValue"? if it is true, search text will not be cleared after item selection.

I'm currently working on PR because I need it anyway in my project. Please tell me if this feature fits in your roadmap or you have any advices/suggestions.

wmertens:
Would `onBlurResetsInput	bool	true	whether to clear input on blur` do what you want? And then another option would be needed on multi-select that keeps the option list open on select?

trashgenerator:
no, onBlurResetsInput is responsible for clearing or not clearing input when widget looses focus. I want the input value to be not cleared when some option is selected. 
I create PR #750. Check it please
You can pass onSelectResetsInput=false to achieve this

mthomas46:
I would also like to see this, even thought the list is preserved, when working with large volumes of data, the user can easily forget what they typed to get that list

PR #750, opened 2016-01-31
#709 do not clear input value in multiselect if prop `onSelectResetsInput={false}` is passed

nosovsh:
I can add tests and update readme and example if you are going to accept this

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -393,9 +393,11 @@ const Select = React.createClass({
 	selectValue (value) {
 		if (this.props.multi) {
 			this.addValue(value);
-			this.setState({
-				inputValue: this.props.onSelectResetsInput ? '' : this.state.inputValue,
-			});
+			if(this.props.onSelectResetsInput) {
+				this.setState({
+					inputValue: '',
+				});
+			}
 		} else {
 			this.setValue(value);
 			this.setState({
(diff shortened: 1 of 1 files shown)


src/Select.js
```diff
@@ -392,7 +394,7 @@ const Select = React.createClass({
 		if (this.props.multi) {
 			this.addValue(value);
 			this.setState({
-				inputValue: '',
+				inputValue: this.props.onSelectResetsInput ? '' : this.state.inputValue,
```
wmertens: When you're leaving the value as-is, there is no reason for the setState, so better put it in an `if` around it

src/Select.js
```diff
@@ -571,8 +573,8 @@ const Select = React.createClass({
 	renderClear () {
 		if (!this.props.clearable || !this.props.value || (this.props.multi && !this.props.value.length) || this.props.disabled || this.props.isLoading) return;
 		return (
-			<span className="Select-clear-zone" title={this.props.multi ? this.props.clearAllText : this.props.clearValueText} 
-						aria-label={this.props.multi ? this.props.clearAllText : this.props.clearValueText} 
+			<span className="Select-clear-zone" title={this.props.multi ? this.props.clearAllText : this.props.clearValueText}
+						aria-label={this.props.multi ? this.props.clearAllText : this.props.clearValueText}
```
wmertens: I don't know @JedWatson's preferences, but is seems to me that lint-type changes should go in a separate commit…

src/Select.js
trashgenerator: fair enough. Changed in the next commit

src/Select.js
trashgenerator: hmm. seems it was added by IDE. Is it possible to remove it from the commit after push?

JedWatson:
Implemented by #1952

PR #754, opened 2016-02-03
Add prop to open Select on focus

willdavidow:
Adds boolean prop called openAfterFocus which, if true, will open the select component's dropdown container when focused.


willdavidow:
@JedWatson Anything I need to change/can do to get this PR accepted? If you've got a few, let me know - thanks!

Issue #738, opened 2016-01-26
Lazy loading of options


UrsBisang:
I searched in the docs, but could not find anything about the following:

Is it possible to load the options lazily, i.e. to defer the ajax call, until the dropdown is actually openened or text is typed? Ideally it should be possible to display a default value when the select component is created, so that the select box is not empty at the beginning.

Is something like this possible?

We have a form with several select boxes on it and the call for retrieving the options is expensive and we try to avoid the overhead when nothing is changed.

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you feel this issue / pull request is still relevant and you'd like us to review it, please leave a comment and we'll do our best to get back to you.

feyst:
I've created a topic in the Ideas section under Discussions please upvote: 
[https://github.com/JedWatson/react-select/discussions/5928](https://github.com/JedWatson/react-select/discussions/5928)

Issue #697, opened 2016-01-06
Render to "other" container & use Tether to position overlays


Andreyco:
_I am not using screenshot of react-select in action, but they do apply to react-select as well_

In some scenarios you have to use `overflow: hidden`. This causes undesired side effects such as cropping list of options. See this picture...
![](http://i.stack.imgur.com/siaKj.png)

This can be solved by attaching select's overlay to outer element, `document.body` preferably.
**Tether** does this job very well. Bootstrap 4 is moving to Tether as well.
https://github.com/HubSpot/tether

Would you consider use of Tether to display overlays?

**Edit**
It's was not clearly stated. Tether does not render elements to "other" tree. It just positions those elements to their targets. Rendering should happen via "portals".

thejameskyle:
I doubt that tether will work with this, but just to note if you want to make this type of rendering work universally (isomorphically) you can use https://github.com/cloudflare/react-gateway

Andreyco:
Yes, you need to render element via "portals" to target container. Then, with Tether positions those elements next to their targets on scrolling, window resize, ...

jsdf:
A more extensible option here would be to add a prop like `renderMenu` which allows developers using react-select to implement the actual rendering of the menu, eg. via a portal. They could use tether if they wish, or implement their own rendering logic. This would also allow stacking context problems to be solved by rendering the dropdown at the end of `<body>`.

Andreyco:
Oh yes, this is an excellent option too!

jsdf:
https://github.com/JedWatson/react-select/pull/529 would allow this, though I am not sure it is the best approach, as I've mentioned on the PR.

jossmac:
Version 1 of react-select is no longer supported.

In the best interest of the community we've decided to spend the time we have available on the latest version.

We apologise for any inconvenience.

Please see:

1. [v1 to v2 upgrade guide](https://react-select.com/upgrade-guide)
1. [v1 documentation and examples](v1.react-select.com)

Issue #637, opened 2015-11-30
Menu list is not shown in a scrollable div


xabikos:
When we render the select component inside a div that has overflow auto and there is not enough space for the menu to appear then we need to scroll to be able to see all the options. You can see an example here 
![select](https://cloud.githubusercontent.com/assets/487416/11473084/c57812fc-9770-11e5-9f31-46a8c6aee34b.png)

Instead a behavior like on that follows would be better

![date](https://cloud.githubusercontent.com/assets/487416/11473104/e674d3aa-9770-11e5-8c15-d7c093cafe0a.png)

Both of the pop ups are absolute positioned but I couldn't find any other differences

xabikos:
After investigating a bit more and dig into the code of react-datepicker to see how they do it I found that this control uses [tether library](https://github.com/HubSpot/tether) to have the pop up absolute positioning and always to the right spot. Is there any interest to submit a pull request and follow the same approach to solve the issue? @JedWatson

christophediprima:
I am very interested on this feature as I am having the same issue in my list !

![capture d ecran 2015-12-22 a 10 30 23](https://cloud.githubusercontent.com/assets/3352687/11952252/2259dd66-a897-11e5-906f-53f5971eeb9c.png)

winterbe:
Is this fixed in latest 1.0-beta? I would love to give react-select a try but this issue would be a job stopper for us.

guilleCM:
Any fix this issue?

matthewcorreia:
@guilleCM if you use react-select v2, there is an option to use portals to solve this issue:
https://react-select.com/advanced#portaling

Issue #554, opened 2015-10-31
Lazy-render large option lists (and reuse option panels)


mathew-kurian:
Something similar to this http://devblog.orgsync.com/react-list/

mwkurian:
I have added it my repo: https://github.com/mwkurian/react-select
Example of its benefit: http://4playdb.me/platform/94

asides:
Great work! Waiting in the official branch...

bvaughn:
Check out PR #859 (and the demo linked) which adds react-virtualized support.

bvaughn:
**Update**: If you're able to use react-select 1.0.0 beta, then you might be interested in trying a new HOC, [react-virtualized-select](https://bvaughn.github.io/react-virtualized-select/). Does windowing/lazy rendering so it should help with the issue you're reporting.

jossmac:
Version 1 of react-select is no longer supported.

In the best interest of the community we've decided to spend the time we have available on the latest version.

We apologise for any inconvenience.

Please see:

1. [v1 to v2 upgrade guide](https://react-select.com/upgrade-guide)
1. [v1 documentation and examples](v1.react-select.com)

PR #763, opened 2016-02-08
[Enhancement] Add `menuComponent` option for custom Menu rendering.

Craga89:
Currently there's no easy way to override the `Menu` component used internally, in a similar way to the `Option` and `Value` components. 

This is handy especially in cases where we want to implement, say, a virtualized scrolling approach when rendering a large amount of items (#738, #554) using `react-virtualized` or `react-list`. It's may also be a possible opportunity to implement `react-tether` to position the menu dynamically based on viewport constraints (#637, #697).

I've added new example to the `examples` files that implements the aforementioned `react-virtualized` approach to render a large list of 300 items (also included).

### DIFF
--- a/examples/src/app.js
+++ b/examples/src/app.js
@@ -6,6 +6,7 @@ import Select from 'react-select';
 
 import Contributors from './components/Contributors';
 import CustomComponents from './components/CustomComponents';
+import CustomMenuComponent from './components/CustomMenuComponent';
 import CustomRender from './components/CustomRender';
 import Multiselect from './components/Multiselect';
 import NumericSelect from './components/NumericSelect';
@@ -19,6 +20,7 @@ ReactDOM.render(
 		<NumericSelect label="Numeric Values" />
 		<CustomRender label="Custom Render Methods"/>
 		<CustomComponents label="Custom Placeholder, Option and Value Components" />
+		<CustomMenuComponent label="Custom Menu Component" />
 		{/*
 		<SelectedValuesField label="Option Creation (tags mode)" options={FLAVOURS} allowCreate hint="Enter a value that's NOT in the list, then hit return" />
 		*/}
--- a/src/Select.js
+++ b/src/Select.js
@@ -51,6 +51,7 @@ const Select = React.createClass({
 		matchPos: React.PropTypes.string,           // (any|start) match the start or entire string when filtering
 		matchProp: React.PropTypes.string,          // (any|label|value) which option property to filter on
 		menuBuffer: React.PropTypes.number,         // optional buffer (in px) between the bottom of the viewport and the bottom of the menu
+		menuComponent: React.PropTypes.func,        // menu component to render instead of the default
 		menuContainerStyle: React.PropTypes.object, // optional style to apply to the menu container
 		menuStyle: React.PropTypes.object,          // optional style to apply to the menu
 		multi: React.PropTypes.bool,                // multi-value input
@@ -310,9 +311,12 @@ const Select = React.createClass({
 	},
 
 	handleInputBlur (event) {
- 		if (this.refs.menu && document.activeElement.isEqualNode(this.refs.menu)) {
- 			return;
- 		}
+		if (this.refs.menu) {
+			let menuDOM = ReactDOM.findDOMNode(this.refs.menu);
+			if (document.activeElement.isEqualNode(menuDOM)) {
+				return;
+			}
+		}
 
 		if (this.props.onBlur) {
 			this.props.onBlur(event);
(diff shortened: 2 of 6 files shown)


Craga89:
Doh, accidentally included the `dist` files in this PR. Would you like me to re-PR once I've fixed?

JedWatson:
Hey @Craga89, thanks for this!

Sounds great, would you mind updating the PR just to include the changes to the `src` files? hard to review the diff with everything included. I'll update the build & release a new version once it's merged.

Craga89:
Rebased with just the `src` and `example` file changes, sorry for the delay :)

Craga89:
Rebased into single commit with only the `src` and `example` files changed and pulled from master, too :+1:

src/Select.js
```diff
@@ -707,14 +712,25 @@ const Select = React.createClass({
...
 		}
+
+		let Menu = this.props.menuComponent || 'div';
+		return (
+			<Menu
+				ref="menu"
+				className="Select-menu"
+				style={this.props.menuStyle}
+				onScroll={this.handleMenuScroll}
+				onMouseDown={this.handleMouseDownOnMenu}>
```
bvaughn: Looks like you have an extra `>` ?

examples/src/app.js
```diff
@@ -6,6 +6,7 @@ import Select from 'react-select';
 
 import Contributors from './components/Contributors';
 import CustomComponents from './components/CustomComponents';
+import CustomMenuComponent from './components/CustomMenuComponent';
```
bvaughn: Where's this file? :)

cloudkite:
@Craga89 are you still able to work on this? If you don't have the capacity, I'm happy to contribute a pull request instead as it's becoming an urgent requirement at work. I think @bvaughn was interested in spending some effort on it as well. Let me know thanks :)

Craga89:
Added the missing files, removed the syntax error and rebased onto current master :)

cloudkite:
@Craga89 nice work :+1: really hoping this gets merged @bvaughn you mentioned in https://github.com/JedWatson/react-select/pull/808#issuecomment-193618111 that you feel that creation of children should be given over to react-virtualized using a  **factory function or something similar**. My understanding is that ReactElements are fairly cheap to create (the expensive part is attaching and rendering them to the DOM) so this PR should get us most of the way. Maybe functionality to provide a factory function for creating options could be a seperate PR as a further optimisation for scenarios when you have very very long list of data and don't want the overhead of creating a whole bunch of ReactElements upfront.

bvaughn:
@Craga89: React elements do seem cheap to create but it still wastes cycles to create them unnecessarily. The main reason I mentioned this is that `react-virtualized` integration was mentioned on this PR (and a related issue) and that requires the ability to manage/defer the creation of children. :)

Craga89:
Agreed, it's somewhat of a "bodge" in that sense but really this PR is just to give a general way to replace the `menuComponent` wholesale without worrying about specifics, inline with similar `props` the plugin provides.

We're using the approach mentioned here using `react-virtualized` in production at th moment to render 600+ items and it's usable, but could definitely benefit from the deferred approach. I'll take another pass at it soon if I get time at work.

cloudkite:
@Craga89, @bvaughn I would advocate trying to get this merged as it is now if @JedWatson is happy with the changes so far. Since its a workable solution for relatively large lists and solves other issues mentioned such as dynamic menu positioning. Then in future another seperate PR could be raised to implement deferred creation of menu options to allow further optimisation, thoughts?

bvaughn:
Hi again. :) I created a fork of react-select this evening and tried integrating react-virtualized just to see how difficult it would be. In case you're curious...

* **Demo**: https://rawgit.com/bvaughn/react-select/master/examples/dist/index.html
* **Source**: https://github.com/bvaughn/react-select

My fork is obviously a direct integration but it was just intended as a proof of concept. I feel pretty confident about the approach this PR takes, now that I've kind of played with it a it more.

I would be happy to release and maintain an adapter library that sits between react-select and react-virtualized for users who want to enable virtualization by going this route. (I've already reserved an NPM package for this purpose in fact.)

bvaughn:
FWIW I went ahead and tried what I suggested in my own branch.
* New `Select` property: https://github.com/bvaughn/react-select/blob/master/src/Select.js#L706
* Example usage: https://github.com/bvaughn/react-select/blob/master/examples/src/components/Cities.js#L53
* Live demo: https://rawgit.com/bvaughn/react-select/master/examples/dist/index.html

I think this is the direction to go. I'd be happy to submit a PR (or a patch to this PR) if there's interest.

bvaughn:
FYI I went ahead and submitted a PR with the approach I outlined for easier discussion purposes ~ #859

Craga89:
Great job @bvaughn, this looks much more flexible than this approach. Happy to close this if everyone agrees? Though, perhaps having a `menuComponent` prop services a simpler purpose that could live side-by-side?

bvaughn:
Thank you @Craga89! Fwiw I also released a HOC to make the integration between the 2 libraries easier: https://github.com/bvaughn/react-virtualized-select/

Maybe that helps address your concerns about "simpler use case"?

Craga89:
Perfect! I'll be using that come Monday!

PR #764, opened 2016-02-09
Adds missing handleMouseDownOnMenu function and fixes #648

### LINKED ISSUE #648: Select Menu Scrollbars don't play nice with React-Bootstrap Modal
This happens in v1 but not in v0.9.1. 

Expectation: When a Select control is inside a React-Bootstrap Modal, clicking on the Select Menu scrollbars will scroll the menu.

Actual: The menu closes.

Repro:
```
import React from 'react';
import { Modal } from 'react-bootstrap';

const options = ['one', 'two', 'three', 'four', 'five', 'six', 'seven', 'eight'].map(text => ({ label: text, value: text }));

export default React.createClass({
  getInitialState() {
    return { selected: null }
  },

  onChange(option) {
    this.setState({ selected: option && option.value });
  },

  render() {
    return (
      <Modal show>
        <Select value={this.state.selected} options={options} onChange={this.onChange}/>
      </Modal>
    );
  }
});
```

The culprit appears to be related to this line:
https://github.com/JedWatson/react-select/blob/master/src/Select.js#L238
For some reason, the `document.activeElement` value is the modal/backdrop container.

MrLeebo:
The function `handleMouseDownOnMenu` was referenced in the definition of the `menu` node, but the source code was missing an implementation. This is @slestang's implementation from https://github.com/JedWatson/react-select/pull/478. This PR fixes the issues reported in https://github.com/JedWatson/react-select/issues/648.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -247,6 +247,19 @@ const Select = React.createClass({
 		this.closeMenu();
 	},
 
+	handleMouseDownOnMenu (event) {
+		// if the event was triggered by a mousedown and not the primary
+		// button, or if the component is disabled, ignore it.
+ 	  if (this.props.disabled || (event.type === 'mousedown' && event.button !== 0)) {
+		  return;
+		}
+		event.stopPropagation();
+		event.preventDefault();
+
+		this._openAfterFocus = true;
+		this.focus();
+	},
+
 	closeMenu () {
 		this.setState({
 			isOpen: false,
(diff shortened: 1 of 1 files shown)


jorgechen:
@MrLeebo I tried this PR on Chrome and Micrsoft Edge on Windows 10 and it doesn't seem to work. Will double check later, maybe I set it up wrong.

MrLeebo:
@jorgechen The PR is just the change to the src folder, so if all you did was change the reference in your package.json, you won't see a fix. Either build it (if you've cloned locally) or copy the `handleMouseDownOnMenu` function into `node_modules/react-select/lib/Select.js` if you want to test within the context of your application.

jorgechen:
@MrLeebo Thanks! 
I built this locally and the scroll bar works now on Chrome (OSX / Win10) and IE11

JedWatson:
Thanks for picking this up and getting it working @MrLeebo!

PR #769, opened 2016-02-13
Use camel case for autofocus prop

X455u:
Changed the autofocus prop to camelCase to match with React's `autoFocus` attribute/prop.

[React HTML attributes](https://facebook.github.io/react/docs/tags-and-attributes.html#html-attributes)


gwyneplaine:
thanks for this @X455u, great catch! I've applied what you've done here to a separate PR with additional deprecation logic here https://github.com/JedWatson/react-select/pull/2002.

PR #770, opened 2016-02-15
Fix typo in Contributing.md

schnerd:
Pretty sure this is supposed to be "chance" instead of "change"


JedWatson:
Thanks @schnerd!

PR #779, opened 2016-02-18
Load options should receive original input [SUGGESTION]

Bertg:
If the input does not receive the original input there is no way to do a proper lookup based on it.
The lookup filtering takes care of matching based on case and accent ignoring.


gwyneplaine:
@Bertg thanks for this, an alternative to this solution that resolves the root issue has been made here https://github.com/JedWatson/react-select/pull/914, and has since been merged into master.

Issue #783, opened 2016-02-20
Provide the optionRenderer access to the inputValue


ijohnson888:
In one of my cases, I found that I needed access to the filter's inputValue in the optionRenderer function. As far as I can tell, in the current code there isn't another way to provide it to this function outside of tracking it separately via onChange in the parent component, which seems excessive.

Orangeyness:
This would be useful for me too. I would like to have my custom optionRenderer highlight the segment of the label which matches the filter.

JedWatson:
I've implemented this now, will be published in 1.1.0

PR #784, opened 2016-02-20
optionRenderer inputValue

ijohnson888:
The optionRenderer will now have access to the inputValue from the
state. #783


jnleonard3:
:+1:

thomasbottonari:
:+1:

mrqcmc:
👍

Tommy10802:
:+1:

JedWatson:
Thanks for the PR @ijohnson888 

This is a good idea but the rendering code has been completely rearranged so unfortunately I'm going to have to close this. Will be looking to simplify how custom rendering can be implemented with access to more state very soon though, so this feature should be supported when that lands.

PR #785, opened 2016-02-22
Fixed onMenuScrollToBottom not working in all cases

sstrong-pica9:
When my browser was zoomed to anything other than 100%, the scroll to bottom wasn't working. I've never used that calculation being exactly equal so I think this is what we need.


flexdinesh:
Hey @sstrong-pica9 [v1 of react-select is no longer maintained](https://github.com/JedWatson/react-select/pull/3786). 

PR #790, opened 2016-02-24
Update hidden input for default html form behavior

james4388:
When submit form with multiple select, It will be submit as an array of field, not a single value with delimiter
```
<select id="inscompSelected" name="myInput" multiple="multiple">
               <option value="1">1</option>
               <option value="2">2</option>
               <option value="3">3</option>
               <option value="4">4</option>
</select>
```
will be sumitted as 
myInput=2
myInput=3

This change make it compatible with javascipt FormData and html form submit.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -719,8 +719,14 @@ const Select = React.createClass({
 
 	renderHiddenField (valueArray) {
 		if (!this.props.name) return;
-		let value = valueArray.map(i => stringifyValue(i[this.props.valueKey])).join(this.props.delimiter);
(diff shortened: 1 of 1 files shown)


JedWatson:
This is a breaking change, which should be fine for the update to 1.0, might be worth implementing an option to enable the old behaviour though.

Other than that, I think it's a good call - I do want the API to become basically a "drop in" upgrade to the standard HTML `<select>` element. Thanks!

jochenberger:
I'm not sure that this is a good solution. It changes the semantics of the `name` attribute.
I think there should be a single hidden `input` containing either a single value or a serialized array depending on the `multi` prop.

PR #792, opened 2016-02-25
Add `joinValues` prop to bring old join values behavior back

james4388:
I just add `joinValues` prop that will alow enable old joining values behavior. Please review.


gwyneplaine:
@james4388 thanks for this, joinValue has already been added back in as a legacy prop in a separate body of work.

Issue #791, opened 2016-02-25
Test needed: new `joinValues` prop


JedWatson:
Form values in `multi` mode are now submitted in multiple fields, see #790

The new `joinValues` prop (boolean) re-implements the old behaviour of joining the values with the `delimiter` prop value.

Both these cases need to be covered by our tests

JedWatson:
Added by #793

PR #793, opened 2016-02-25
Test hidden values for multi option.

ruiaraujo:
This is a try at solving #791.
I might be adding more tests later.

joinValues in #792 hasn't been merged but when it is I wouldn't mind doing tests like this.


JedWatson:
Thanks a lot for this @ruiaraujo! Took a while, but I cleaned up the conflicts and merged it 😄

PR #794, opened 2016-02-25
allow tabbing out if onBlurResetsInput property is false

vakopian:
currently, when onBlurResetsInput is set to false, typing a non-existing value in the select box then pressing tab does not let the focus to move to the next input. (Incidentally, pressing shift+tab does move the focus to the previous input). This small patch fixes this behavior.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -334,7 +334,7 @@ const Select = React.createClass({
 				}
 			return;
 			case 9: // tab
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks for the PR @vakopian 

Our tab behaviour has been changed quite a bit since this was opened, and is now more inline with native select controls. Please let me know if you still think it's something we should add.

PR #724, opened 2016-01-20
Make sure the selected option is immediately visible when the menu is open

jooj123:
Makes sure that when you select an option, next time you open the menu that selected option is immediately visible. 
I made sure it doesn't unnecessarily re-render

I noticed a few people asking for this - hopefully its as per guidelines :+1:  

Also included fixing existing linting errors

PR #797, opened 2016-02-26
Fixed bug with getFocusableOption that prevents selected option to be immediately visible

jooj123:
This fixes a bug in the getFocusableOption func where on some circumstances it would not return the appropriate focusable option.
The current indexOf check wont work for arrays of objects - it will only work if the object references are the same, example here shows it wont work correctly: http://jsfiddle.net/7B7dQ/1/).
 
This has caused this feature PR to not work: https://github.com/JedWatson/react-select/pull/724

This could be reproduced in the example page by replacing:

```
var options = STATES[this.state.country];
```

in States.js `render()` with:

```
var options = [
	{ value: 'australian-capital-territory', label: 'Australian Capital Territory', className: 'State-ACT' },
	{ value: 'new-south-wales', label: 'New South Wales', className: 'State-NSW' },
	{ value: 'victoria', label: 'Victoria', className: 'State-Vic' },
	{ value: 'queensland', label: 'Queensland', className: 'State-Qld' },
	{ value: 'western-australia', label: 'Western Australia', className: 'State-WA' },
	{ value: 'south-australia', label: 'South Australia', className: 'State-SA' },
	{ value: 'tasmania', label: 'Tasmania', className: 'State-Tas' },
	{ value: 'northern-territory', label: 'Northern Territory', className: 'State-NT' },
];
```

If you select (click) option 'Northern Territory' for example and then re-open the selector - it always focuses on the first option.
This fails because its a new reference on each render.


martin-jujou-sirca:
@JedWatson please let me know if there is anything wrong with this PR which may be stopping it from being merged.
Im not familiar with coveralls - from what i can see from the Select.js coverage - all the lines I added have test coverage (ie. there is no red lines)

jooj123:
@JedWatson please let me know if there is anything wrong with this PR which may be stopping it from being merged.
Im not familiar with coveralls - from what i can see from the Select.js coverage - all the lines I added have test coverage (ie. there is no red lines)

xeniatay:
+1

PR #809, opened 2016-03-02
Update react-select.css

Yoonseonwoong:
Hello, JedWatson

I think about the value of the cursor.
I think that value is a no-drop right.
What do you think?

### DIFF
--- a/dist/react-select.css
+++ b/dist/react-select.css
@@ -23,7 +23,7 @@
   box-shadow: none;
 }
 .Select.is-disabled .Select-arrow-zone {
-  cursor: default;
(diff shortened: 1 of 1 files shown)


JedWatson:
This is a personal preference, I like disabled inputs not to change the cursor. This is standard browser behaviour. Bootstrap (maybe other frameworks earlier too) introduced the no-drop cursor for disabled fields, but for me that is opinionated and distracting when the cursor moves across the page. Disabled fields _shouldn't_ grab your attention, imo.

Thanks for the PR, feel free to override this in your own stylesheet if you prefer :)

Issue #813, opened 2016-03-03
For Async, call 'props.onInputChange' before loadOptions


adorsk:
I think we should call 'props.onInputChange' before loadOptions.

This will allow for capturing raw input, as well as doing other things related to input changes.

This is also an alternative to https://github.com/JedWatson/react-select/pull/779 . Instead of passing the rawInput to loadOptions, users can capture it themselves via onInputChange.

I will submit a PR with a possible implementation.

jasonh42:
This is a pretty big issue for me as well - I need to know when the input changes for the Async component.

JedWatson:
Added by #914

PR #814, opened 2016-03-03
wrap onInputChange

adorsk:
RE: https://github.com/JedWatson/react-select/issues/813


mgroeneman:
👍  I could use this feature as well.

JedWatson:
Added by #914

PR #817, opened 2016-03-04
add isOpen prop that allows owners to override the isOpen state

### LINKED ISSUE #189: Ability to close dropdown list
Hi,

Thanks for all the hard work putting this together. One issue that I have is that I would like to be able to close the dropdown list when it is open, either at will or through a parameter that would close it when a selection is made.

Currently, if @props.mutli is true, the list stays open, requiring the user to click elsewhere in the UI in order to move on. While I understand the logic behind this, this behavior is not always desirable and it would be nice to have some control over it.

bobbyrenwick:
This closes #189.

You may want to change the name of the prop to something like `forceOpen` in order to disambiguate it from the state, but that's up to you!

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -45,6 +45,7 @@ const Select = React.createClass({
 		ignoreCase: React.PropTypes.bool,           // whether to perform case-insensitive filtering
 		inputProps: React.PropTypes.object,         // custom attributes for the Input
 		isLoading: React.PropTypes.bool,            // whether the Select is loading externally or not (such as options being loaded)
+		isOpen: React.PropTypes.bool,               // Pass a boolean to start controlling the open state externally, null to let react-select manage it
 		labelKey: React.PropTypes.string,           // path of the label value in option objects
 		matchPos: React.PropTypes.string,           // (any|start) match the start or entire string when filtering
 		matchProp: React.PropTypes.string,          // (any|label|value) which option property to filter on
@@ -99,6 +100,7 @@ const Select = React.createClass({
 			ignoreCase: true,
 			inputProps: {},
 			isLoading: false,
+			isOpen: null,
 			labelKey: 'label',
 			matchPos: 'any',
 			matchProp: 'any',
@@ -122,7 +124,7 @@ const Select = React.createClass({
 			inputValue: '',
 			isFocused: false,
 			isLoading: false,
-			isOpen: false,
+			isOpen: this.props.isOpen !== null ? this.props.isOpen : false,
 			isPseudoFocused: false,
 			required: this.props.required && this.handleRequired(this.props.value, this.props.multi)
 		};
@@ -131,6 +133,14 @@ const Select = React.createClass({
 	componentDidMount () {
 		if (this.props.autofocus) {
 			this.focus();
+		}
+	},
+
+	componentWillReceiveProps(nextProps) {
+		if (nextProps.isOpen !== null) {
+			this.setState({
+				isOpen: nextProps.isOpen,
+			});
 		}
 	},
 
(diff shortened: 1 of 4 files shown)


Paveltarno:
+1

src/Select.js
```diff
@@ -122,7 +124,7 @@ const Select = React.createClass({
 			inputValue: '',
 			isFocused: false,
 			isLoading: false,
-			isOpen: false,
+			isOpen: this.props.isOpen !== null ? this.props.isOpen : false,
```
JKillian: will this work properly if `isOpen` is `undefined`?

JKillian:
+1 an `isOpen` prop would be great and would solve my issue in #962

src/Select.js
bobbyrenwick: Why would it be `undefined`, it defaults to `null`.

JKillian: I guess you're right, I was thinking in the case of `<Select isOpen={undefined}/>`, but I suppose React uses the default value in this case?

bobbyrenwick: I didn't know, so I made a [fiddle](https://jsfiddle.net/bobbyrenwick/a70Loco4/) that demonstrates what you suggested is the true behaviour!

charlesbodman:
+1 need this badly. It's making it very difficult to style .

ericbiewener:
@charlesbodman To force it open temporarily for styling, just set `isOpen` to `true` in the source code [here](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L908).

adidahiya:
bump @JedWatson, any word on this?

dkh215:
Is this on ice?

tonyjmnz:
This feature would be really useful for an use case I'm working on. I also think it would make the component much more flexible when paired with a flux architecture.

vaclav-zeman:
Is this going to be implemented anytime soon?

kajmagnus:
I'd just want to tell you about my use case: When React-Select opens, it hides the Save button in my dialog. So the user needs to understand that s/he has to either select an option, _or_ click somewhere outside React-Select to close it, for the Save button to become visible again. However, if s/he clicks outside the _dialog_, in order to close React-Select then the whole dialog closes and disappears.

This would likely confuse people (even I feel a bit confused). I think this can be solved, by leaving React-Select open always, and giving it enough space, so it won't hide the Save button.

(I don't want it to be "only" open — I want it to push the stuff below downwards too so it won't overlap the Save button. But this last thing seems to be a simple CSS style: `.Select-menu-outer { position: relative; }` (instead of `absolute`.) works fine for me.)

ConneXNL:
Waiting for this one...

urisavka:
Hey guys. Any news on this one? 
Or maybe is there any other way to force open dropdown from the start?

tnrich:
@JedWatson  Please add this feature!

This PR is doing the same thing as https://github.com/JedWatson/react-select/pull/1015

aleccool213:
bump

thiagocoelho:
There is any fix/way to force the open state? I need to style the options but I can't inspect the elements to discovery what I need to change.

MindRave:
Any news on this one? Would be a very useful prop indeed.

haio:
+1 really need this feature!

EnriqueVidal:
Is there still interest in this? I just pushed #1950, it works a little different though, rather than using `isOpen` it implements `alwaysOpen`, the test suite is green and I seem to be passing the coverage test, however I cannot run the example components via the `start` script, not sure if I'm missing anything in there.

Any advise/suggestion is more than welcomed.

JedWatson:
Thanks for the PR @bobbyrenwick, I'm going to close it until we have more clarity on #1989

leojacob7:
Why did you close this despite so many requests @JedWatson

Issue #753, opened 2016-02-03
Disable autosize


ghost:
This lib is amazing but i would like to know if we can disable autosize fixture
Thanks in advance

wmadden:
This would be really useful :+1:

JedWatson:
Done! Thanks @wmadden

PR #830, opened 2016-03-14
Use a normal input if `autosize` prop is false

wmadden:
Addresses #753


JedWatson:
Awesome, thanks @wmadden - I've been meaning to add this :)

wmadden:
No problem @JedWatson, glad to help :)

wmadden:
One quick question: do we need `react-input-autosize` at all? Is it useful to have an autosizing dropdown, or would that be something better added by the consumer?

PR #834, opened 2016-03-15
Update required when Select gets new props

andyhite:
**Bug**: updating props on the Select does not update `required` in the component's state. This means that if a Select is marked as required, initially mounted with an empty value, and then a subsequent render updates the value, the field is improperly marked as being invalid upon submit.

**Solution**: when the component receives new props, re-evaluate `handleRequired` and update 'required' in the component's state.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -138,6 +138,14 @@ const Select = React.createClass({
 		}
 	},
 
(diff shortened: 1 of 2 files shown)


JedWatson:
Awesome, thanks @andyhite!

PR #839, opened 2016-03-17
fix(docs): update the propType for placeholder

kumarharsh:
Updated the propType of placeholder to show that it does support both a `string` or a `node` value.


JedWatson:
Thanks @kumarharsh!

PR #840, opened 2016-03-17
Fix scrollMenuIntoView option

### LINKED ISSUE #799: scrollMenuIntoView option broken on IE
in IE, when you open the select, instead of scrolling to the menu it scrolls to the top of the page.  This is because the window.scrollY property doesn't exist in ie.

This is fixed by changing 
```javascript
if (window.innerHeight < menuContainerRect.bottom + this.props.menuBuffer) {
    window.scrollTo(0, window.scrollY + menuContainerRect.bottom + this.props.menuBuffer - window.innerHeight);
}
```
to 
```javascript
if (window.innerHeight < menuContainerRect.bottom + this.props.menuBuffer) {
    var scrollY = window.pageYOffset || document.documentElement.scrollTop
    window.scrollTo(0, scrollY + menuContainerRect.bottom + this.props.menuBuffer - window.innerHeight);
}
```
as in https://github.com/reactjs/react-router/issues/605

aij:
Switching to relative scrolling avoids two problems:
 1. IE does not support window.scrollY
 2. window.scrollTo(0, ...) scrolls to the left

This fixes #799.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -180,7 +180,7 @@ const Select = React.createClass({
 		if (this.props.scrollMenuIntoView && this.refs.menuContainer) {
 			var menuContainerRect = this.refs.menuContainer.getBoundingClientRect();
 			if (window.innerHeight < menuContainerRect.bottom + this.props.menuBuffer) {
-				window.scrollTo(0, window.scrollY + menuContainerRect.bottom + this.props.menuBuffer - window.innerHeight);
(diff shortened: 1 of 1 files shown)


aij:
Please let me know if you'd prefer some alternate solution. We are currently using react-select and need to support IE10.

JedWatson:
Looks great @aij, thanks for the fix and sorry for the delayed merge!

PR #843, opened 2016-03-18
Update README.md

jaimeagudo:
Rename `cacheAsyncResults` to `cache`


JedWatson:
Thanks for picking that up @jaimeagudo!

jaimeagudo:
Thanks for sharing the whole thing @JedWatson!

PR #847, opened 2016-03-23
Fix issue with value object not being passed to handleRequired

andyhite:
I discovered that my previous patch for `required` was failing in some cases (for instance, in Internet Explorer) but passing in others for some reason, so this PR addresses that issue. The problem was that the value being passed to `handleRequired` was the value directly from `this.props.value` instead of the expanded value object, which `handleRequired` expects.

Don't ask me how this ever worked at all, haha.

This PR also fixes the same issue with the original `handleRequired` implementation on `getDefaultState`, and removes the [props-in-getDefaultState anti-pattern](https://facebook.github.io/react/tips/props-in-getInitialState-as-anti-pattern.html).

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -128,8 +128,18 @@ const Select = React.createClass({
 			isLoading: false,
 			isOpen: false,
 			isPseudoFocused: false,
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks for that @andyhite!

Issue #848, opened 2016-03-23
In IE11 the focus of the select field is lost when scrolling through the list of options


elijahmanor:
A strange IE11 specific bug was found in the react-select component when trying to scroll through the filtered options. You don't technically have to scroll the list to see the behavior, you can just click the scrollbar as shown in the following animated gif. You'll notice the field is blurred, which causes the cursor to be lost and if you click the scrollbar again focus will be regained and you can see the cursor blinking again.

![ngbfguo4op](https://cloud.githubusercontent.com/assets/86454/13999626/31656854-f10b-11e5-8c7d-ba1330718d25.gif)

I have a code fix and tests to address the IE bug and plan to submit a Pull Request soon. I wanted to give a heads up about the issue before I submitted the PR. Thank you for your component and the time you have invested in this project.

JedWatson:
Fixed by #849

PR #849, opened 2016-03-23
Fix IE11 issue where focus is lost when scrolling through options

elijahmanor:
This PR fixes issue #848 and the following animated gif shows the fix in action...

bruc7pz2bb-after

You can read more about the details of the bug in the related #848 issue and watch the broken behavior there (animated gif)

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -311,6 +311,7 @@ const Select = React.createClass({
 
 	handleInputBlur (event) {
  		if (this.refs.menu && document.activeElement.isEqualNode(this.refs.menu)) {
+			this.focus();
  			return;
  		}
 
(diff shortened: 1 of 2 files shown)


herrstucki:
I ran into the same issue, and this worked.

Maybe you could use `this.focus()` instead of `this.refs.input.focus()`, since that's used throughout the rest of the component.

elijahmanor:
@herrstucki good catch! i'll update the code and make sure the behavior still works and the tests pass.

elijahmanor:
@herrstucki I made the update to use `this.focus` and all tests still pass and the behavior is still as intended. Thanks for the suggestion

HenriJ:
We still encounter this bug with IE (11). This PR seems to fix this issue. Please merge it when possible :)

JedWatson:
Thanks @elijahmanor! Merged it, will publish a new release shortly.

PR #853, opened 2016-03-30
feat(Option): support addition className

iam4x:
Hello @JedWatson 

I'm using https://github.com/ftlabs/fastclick to remove the 300ms click delay on iOS devices, but I'm getting incompatibilities with `react-select` which is listening on `touchEnd` event.

Fastclick can ignore certain elements but they need a `needsclick` class (https://github.com/ftlabs/fastclick#ignore-certain-elements-with-needsclick)

This PR adds the support to pass custom classnames to the Option element, I could have used the `optionComponent` prop but I find it overkill for this use case :+1:


JedWatson:
Hi @iam4x, thanks for the PR!

I've just manually merged it due to conflicts, sorry about not merging directly - it looks like I missed a step in the process, but the changes are now in `master`.

iam4x:
Thank's @JedWatson :+1:

PR #855, opened 2016-03-30
handleMouseLeave fix

seergiioo6:


JedWatson:
@seergiioo6 would you mind giving some more context here for the fix you're suggesting? I don't think `onFocus` should be firing when the `mouseLeave` event does..?

JedWatson:
@seergiioo6 bump :)

Issue #807, opened 2016-03-01
Input filtering is laggy for large lists of options on v1.0.0-beta10


janraasch:
I have an options list with about 2000 items, e.g. 

```js
const options = Array.apply(0, Array(2000)).map(function (_i, i) { return i; }).map((i) => ({ 
  value: ('my_value_numero_' + i), label: ('MyLabelStart - (Numero' + i + ')')
}))
```

Then, simply render the `Select` component via

```js
function onChange(values) {
  console.log('Not even doing anything')
}

ReactDOM.render(<Select
  multi=true
  value={[]}
  options={options}
  onChange={onChange} 
/>, domNode)
```

Typing something into the input feels very laggy on `v1.0.0-beta10`. On `v0.9.1` it is fine.

It is really quiet significant of a difference. `v0.9.1` is really smooth, while `v1.0.0-beta10` is laggy.

(If you need an actual production example I can sent you my data via e-mail, but I tried to construct data good enough to reproduce the problem above.)

bvaughn:
Check out PR #859 (and the demo linked) which adds react-virtualized support.

bvaughn:
**Update**: If you're able to use react-select 1.0.0 beta, then you might be interested in trying a new HOC, [react-virtualized-select](https://bvaughn.github.io/react-virtualized-select/). Does windowing/lazy rendering so it should help with the issue you're reporting.

bvaughn:
Chiming in after some time.

I _think_ this problem (slow filtering) comes down to how filtering is done within `react-select` (aka [synchronously, in the UI thread](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L826-L864). For large lists, this approach is going to hurt FPS.

I've created some JS search libs (eg [`js-worker-search`](https://github.com/bvaughn/js-worker-search), [`redux-search`](https://github.com/treasure-data/redux-search)) that manage filtering in web-workers, using data structures optimized for fast look-up, and the response time for filtering drop-down lists using them is _very fast_.

Perhaps that would be interest in trying to integrate one of these libs with `react-select`? Perhaps even as an external HOC (similar to `react-select-virtualized`) that just exported a `filterOptions` function that could be passed as a prop to override the built-in `react-select` filter.

If there's interest, I would be happy to take a stab at it. Let me know!

bvaughn:
To add a little more context...

[...]

Issue #771, opened 2016-02-15
Virtual scrolling to improve performance of Options


gabrielenosso:
There are some issues opened about react-select performance, when it is used with a lot of options.
Main problem is simply that it renders a <div> for each option, even if you have more than 500 options (In my case I have more than 2000 options).

The best solution would be to use the "virtual scrolling" (sometimes called "buffered scrolling"): it's the technique used for large Data Tables.
It practically renders only a part of the options (50 of them for example), while it simulates all the others with one large and empty <div> (so that the vertical scrollbar is always consistent).

You can find a complete explanation of the concept here:
http://datatables.net/blog/2011-06-11

It would be awesome to have it in react-select!

jk171505:
+1.
I have 4k elements on the list and Select works 2-3 times slower than chosen...

ahoiin:
+1 I have 9K elements and it need 2seconds to open up. 
For lists I'm using [https://github.com/bvaughn/react-virtualized](https://github.com/bvaughn/react-virtualized)

Dtraft:
+1 I have 2K elements and while it does open, there is noticeable delay.  Great component over all though!

bvaughn:
+1

I would also love to see react-virtualized integrated. Happy to help if help is needed. :)

bvaughn:
FYI folks, I've added this support in PR #859. Let's hope it gets merged. :)

bvaughn:
**Update**: If you're able to use react-select 1.0.0 beta, then you might be interested in trying a new HOC, [react-virtualized-select](https://bvaughn.github.io/react-virtualized-select/). Does windowing/lazy rendering so it should help with the issue you're reporting.

jakeboone02:
I was going to create a new ticket for this but I'll just mention it here.  The README file in this project references the new `menuRenderer` option, as well as @bvaughn's `react-virtualized` library, but does not mention `react-virtualized-select`.  I think it would be better to mention the latter, otherwise people will just end up recreating what @bvaughn has already done.

By the way, `react-virtualized` is awesome!

bvaughn:
Thanks @jakeboone02. Good suggestion on the README. I had considered updating that but then I must have forgotten. I'll post a PR. :)

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

[...]

Issue #674, opened 2015-12-15
doesn't work with 1000's of entries


natehardy:
Don't know if it was intended for this purpose but I just tried to load a few thousand entries into the select and the fist scroll through is very jumpy. After that scrolling is fine.
However, selecting an item doesn't work.

CosticaPuntaru:
i think there should be a hook that allows you to render options,
in this way you can use something like [virtual-list](https://www.npmjs.com/package/react-virtual-list) to handle large number of options

bvaughn:
Check out PR #859 (and the demo linked) which adds react-virtualized support.

bvaughn:
**Update**: If you're able to use react-select 1.0.0 beta, then you might be interested in trying a new HOC, [react-virtualized-select](https://bvaughn.github.io/react-virtualized-select/). Does windowing/lazy rendering so it should help with the issue you're reporting.

jossmac:
Version 1 of react-select is no longer supported.

In the best interest of the community we've decided to spend the time we have available on the latest version.

We apologise for any inconvenience.

Please see:

1. [v1 to v2 upgrade guide](https://react-select.com/upgrade-guide)
1. [v1 documentation and examples](v1.react-select.com)

PR #859, opened 2016-03-31
[Enhancement] Add `menuRenderer` prop to support windowing for large lists

bvaughn:
I believe this PR resolves the following open issues: #554, #674, #739, #771, #807.

This PR adds support for windowing libraries like [react-virtualized](https://github.com/bvaughn/react-virtualized) to be used to render large lists of options. It is similar to the approach mentioned in PR #763 but it avoids creating all of the menu options when rendering and so it performs better.

The size of the PR is mostly due to the fact that it contains a JSON data file of the largest 1,000 cities in Europe (used to demonstrate custom menu windowing). The actual changes to the `Select` component were very minimal (~40 new lines).

#### This change-set includes:
* New `renderMenu` prop added to `Select` component.
* New tests added to `test/Select-test.js` for new property. (Previous tests all pass.)
* README [updated to mention](https://github.com/bvaughn/react-select/#effeciently-rendering-large-lists-with-windowing) newly added property.
* Example page updated to demonstrate new property as well. (Live demo of this functionality can be seen [here, under "World's Largest Cities" header](https://rawgit.com/bvaughn/react-select/demo/examples/dist/index.html).

Note that [react-virtualized](https://github.com/bvaughn/react-virtualized) was added to the `devDependencies` section (not `dependencies`) so it could be used in the example to demonstrate a more efficient way of rendering select menus with large amounts of data.

If this PR gets merged, or one like it, I plan to release a small adapter library with a single HOC that provides the react-virtualized drop-down menu automatically and passes through other props to `Select`. (In other words, I'm happy to take over the ownership of the feature.)


README.md
```diff
@@ -250,6 +296,7 @@ For multi-select inputs, when providing a custom `filterOptions` method, remembe
 	optionRenderer	|	func	|	undefined	|	function which returns a custom way to render the options in the menu
 	options 	|	array	|	undefined	|	array of options
 	placeholder 	|	string\|node	|	'Select ...'	|	field placeholder, displayed when there's no value
+	renderMenu | func | undefined | Renders a custom menu with options; accepts the following named parameters: `renderMenu({ focusedOption, focusOption, options, selectValue, valueArray })`
```
cloudkite: maybe `menuRenderer` would be better choice since it follows existing naming convention of `optionRenderer `

cloudkite:
Nice work! :+1: 

I think it would be good idea not to include `examples/dist` and `lib` folders in the PR, makes it alot more readable

bvaughn:
Truth. I wanted a live demo to point out so I had to push them to my fork. But maybe I'll clean up this PR and use another branch for the demo. Thanks for the input :)

README.md
bvaughn: Renamed. Thanks for the suggestion.

bvaughn:
Okay. Dist files have all been removed. Only large files still around are the example cities list.

JedWatson:
This looks fantastic @bvaughn! Thanks for the PR, and for owning the feature :)

Naveg:
This is great. It seems like it should be possible to use `react-tether` in the `menuRenderer` function, but I'm having trouble getting that to work. Problem is that I think the tether target should be the `Select-control` component, which isn't available here. Only thing I can think of it to add a dependency on `react-tether` and wrap the `Select-control` and menu in a `TetherComponent`, which we definitely don't want. Any ideas?

Craga89:
:+1: to this. This is way more flexible than the `menuComponent` PR I made, great job @bvaughn! I'd love to see this merged so we can translate our current implementation over!

PR #869, opened 2016-04-06
Change noResultsText prop type to stringOrNode for Async component

mgroeneman:
Matches type of same prop on standard Select component. Thanks!


JedWatson:
Good catch @mgroeneman, cheers!

mgroeneman:
@JedWatson thanks for a really helpful piece of software!

PR #878, opened 2016-04-09
Added ability to override resetValue when clearing the control

LubergAlexander:
Currently when you clear the control it sets the value to null, which is a little bit opinionated. I think it would be useful to be able to override the value with props.

I had a use case when I needed to set the value to 'unset'. I also assume that it may be required to reset to 0 or ''.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,7 @@ Check out the demo site for a more complete example of this.
 	clearable 	|	bool	|	true		|	should it be possible to reset value
 	clearAllText 	|	string	|	'Clear all'	|	title for the "clear" control when `multi` is true
 	clearValueText 	|	string	|	'Clear value'	|	title for the "clear" control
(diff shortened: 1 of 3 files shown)


LubergAlexander:
I am not sure though what to do with the failed travis. Installation fails on NPM2, since react 15.0.1 is out, and there are conflicts with transitive dependencies. Works on NPM3, all tests pass & examples work.

LubergAlexander:
@JedWatson do you know, by any chance, if I should fix package versioning in the scope of this pull request or this is going to be fixed in master(since it is irrelevant to this changeset) ?

LubergAlexander:
@JedWatson rebased, thanks for updating dependencies. Now travis tests passed.

JedWatson:
Great, thanks @LubergAlexander! Ready to merge + release now :)

PR #883, opened 2016-04-10
Styles for .is-selected and fixed indentations

dherran:
Just added a few styles there to the SASS and LESS files for the .is-selected class. Gray for now.

There were some spaces on the indentations for Select.js. Fixed those too.


JedWatson:
Thanks @dherran! much appreciated :)

Issue #841, opened 2016-03-18
Breaking error in IE 8


Dharmoslap:
![virtualbox_ie8 - win7_18_03_2016_10_50_19](https://cloud.githubusercontent.com/assets/3543614/13875677/8e46df68-ecff-11e5-92f8-2d1b0e492a0a.png)

How to replicate this issue

1. run Windows 7 with IE 8 on VirtualBox [https://dev.windows.com/en-us/microsoft-edge/tools/vms/](https://dev.windows.com/en-us/microsoft-edge/tools/vms/)
2. clone this repo https://github.com/Dharmoslap/react-select
3. run "npm start" and open http://10.0.2.2 in IE 8
4. try to select different options and the error will appear

However it's not fancy at all, I still need to support IE 8 for a few more months.

Thanks

Dharmoslap:
It seems to be React related issue, so I created the same one in React repository as well.
https://github.com/facebook/react/issues/6291
Do you know how to run unconcatenatded library inside your stack, to provide more information about the error?

gaearon:
Please see https://github.com/facebook/react/issues/6291#issuecomment-199023995. It is not a React-related issue. IE8 doesn’t support `isEqualNode()` which is used in https://github.com/JedWatson/react-select/blob/18758f78655b4004a3a22fc049f5706cd25bffcd/src/Select.js#L313.

PR #889, opened 2016-04-12
Fix for #841 : replace isEqualNode by strict equality

abalhier:
Fix for #841 IsEqualNode is not available in IE8 and cannot be shimmed.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -323,7 +323,7 @@ const Select = React.createClass({
 	},
 
 	handleInputBlur (event) {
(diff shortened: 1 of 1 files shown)


kurttheviking:
fwiw: Our team forked react-select internally then applied this patch; I can confirm everything is working as expected

abalhier:
Same for us. It works from IE8 to IE11. No regression observed on multiple versions of Chrome and FF (tested on our app via Browserstack).

JedWatson:
Thanks @abalhier and @kurttheviking for verifying!

PR #891, opened 2016-04-13
always open on focus option

HuysentruytRuben:
`alwaysOpenOnFocus` prop that opens the options if the input gets focus in any way (including tabbing and clicking the matching label) when set to true

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,7 @@ Check out the demo site for a more complete example of this.
 	autosize  | bool | true  | If enabled, the input will expand as the length of its value increases
 	autoBlur	|	bool | false | Blurs the input element after a selection has been made. Handy for lowering the keyboard on mobile devices
 	allowCreate	|	bool	|	false		|	allow new options to be created in multi mode (displays an "Add \<option> ?" item when a value not already in the `options` array is entered)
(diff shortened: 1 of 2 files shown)


srph:
The tests seem to be broken. You might need to add an `defaultProps.alwaysOpenOnFocus`.

JedWatson:
Thanks @HuysentruytRuben!

I think this would be better as `openOnFocus` but will merge, clean up + make that change so I can include it in the release I'm about to push out.

ozoesono:
This seems to have broken the component, as I did a fresh npm install on this and now the dropdown is always open when clicked and does not select other options. I downgraded back to 0.9.1 and it works.

HuysentruytRuben:
@ozoesono do you still have this problem, and is it maybe possible to make a jsfiddle? what other options do you use

caitwood:
I'm having the same issue as @ozoesono. If I set openOnFocus to false, it still shows the menu on click. I am using version 1.0.0-beta13. Here is my JSX snippet...

`<Select
                    autoBlur={true}
                    disabled={!this.state.members.length}
                    isLoading={this.state.isLoading}
                    name='member-search'
                    onChange={this.props.onMemberChange}
                    openOnFocus={false}
                    options={this.state.members}
/>`

PR #901, opened 2016-04-18
Add option for custom input renderer function

leftmostcat:
This partially mimics the menuRenderer, valueRenderer and optionRenderer props and allows the user to use a custom input in place of the existing `renderInput()` code. The existing behavior is retained verbatim if the inputRenderer prop is not set, and delegates to the provided function otherwise. Changes in behavior should be minimal and isolated.

I haven't provided updates to the documentation for this change as yet. If the maintainers want to pull this change into master, I'll gladly make the necessary changes myself.


JedWatson:
Thanks @leftmostcat - I'm happy to add this!

hhaidar:
Sorry to resurrect this, but how would one pass the required props for onChange/etc?

Doesn't seem like there's a mechanism for it right now.

tecnolocoEdan:
Any news for the documentation of inputRenderer?

PR #914, opened 2016-04-26
Drd/wrap on input change for async

drd:
Based on #814, but adds test for `onInputChange` and fixes a couple of style/stylistic issues.


JedWatson:
Thanks @drd, looks great.

PR #922, opened 2016-04-29
Enhance IE 8 compatibility

rockallite:
This PR tries to enhance IE 8 compatibility.
* It includes some CSS fixes for positioning and coloring on IE 8.
* It adds `.Select--single` CSS class to single select components. Meanwhile, it replaces `:not(.Select--multi)` CSS selectors with `.Select--single` classes in CSS files. The former does not work on IE 8, and causes sibling CSS rules (separated by commas) fail.
* It maintains usability in Chrome and Safari without side effects.

There are other `:not` CSS selectors which are not replaced. Although they do not work on IE 8, I consider them appearance enhancements which are too trivial to fix.


JedWatson:
I think IE8 support is limited, given that React v15+ have dropped it (ref https://facebook.github.io/react/blog/2016/01/12/discontinuing-ie8-support.html) but this is great, @rockallite, so I'm happy to support it for a bit longer thanks to your help :)

PR #924, opened 2016-04-29
Clarify behavior in NumericSelect example

jakeboone02:
The actual behavior of the example is:

```
[x] Match value only
[ ] Match label only
```
^ matches value only

```
[ ] Match value only
[x] Match label only
```
^ matches label only

```
[ ] Match value only
[ ] Match label only
```
^ matches both

```
[x] Match value only
[x] Match label only
```
^ matches both

So, the checkboxes each indicate that their respective attribute is matched or not matched, independent of the other checkbox (except when both options are unchecked).  I think removing the word "only" from the labels makes them more accurate.

Or you could make it three radio buttons:
```
(*) Match value only
( ) Match label only
( ) Match both
```


JedWatson:
Thanks @jakeboone02, that _is_ clearer

PR #925, opened 2016-04-29
Update README to mention react-virtualized-select

bvaughn:
@jakeboone02 made a [suggestion](https://github.com/JedWatson/react-select/issues/771#issuecomment-215851474) on issue #771 to update the README to point out `react-virtualized-select` as the easiest way of integrating react-virtualized with react-select. Seemed like a reasonable suggestion since it prevents users from having to reinvent the wheel, so I took a stab at updating the README. :)


JedWatson:
Cheers @bvaughn!

PR #926, opened 2016-05-01
Remove unnecessary onUnfocus

gojohnnygo:
This PR fixes the `Uncaught TypeError: this.props.onUnfocus is not a function` error on the "Custom Placeholder, Option and Value Components" example. 

`handleMouseLeave()` and `onUnfocus` is removed.


JedWatson:
Thanks @gojohnnygo

PR #930, opened 2016-05-02
Support for react components in searchPromptText

hellaeon:
We needed to add this change to support styling of the searchPromptText.
Since **searchPromptText** and **noResultsText** now both support text and rendered components, I would like to suggest these are renamed to **searchPromptLabel** and **noResultsLabel**


JedWatson:
Really good point, @hellaeon. Thanks for the PR, and I agree - `{x}Label` would be clearer!

I think for `1.0.0` it would be fine to support the `{x}Text` prop as a fallback with a deprecation warning to the console (can be removed in 2.x) - would you mind submitting that change in another PR?

PR #931, opened 2016-05-02
Fix MIT License

ilkome:


JedWatson:
:+1: thanks @ilkome, good eye for detail!

PR #932, opened 2016-05-03
Change text nodes to labels

hellaeon:
In order to render html as a part of the Select, we need to be able to provide styled HTML elements for the props searchPromptText and noResultsText. We also should then rename the props to indicate their new type.

This PR:
- adds a new prop searchPromptLabel
- adds a new prop noResultsLabel
- adds a deprecation warning for prop searchPromptText
- adds a deprecation warning for prop noResultsText

As per comment on #930, most of this code can be removed after version 2 of react select is released.
See also #933


hellaeon:
This PR is in relation to https://github.com/JedWatson/react-select/pull/930

JedWatson:
Hey @hellaeon, thanks for the PR, and sorry it didn't get attended to. I'm going to clean up all these extensibility props now that 1.0.0 is out, but closing this because it's way out of date.

PR #933, opened 2016-05-03
Add support for searchingLabel props.

hellaeon:
In order to render html as a part of the Select, we need to be able to provide styled HTML elements for the props searchingText. We also should then rename the props to indicate their new type.

This PR:
- adds a new prop searchingLabel
- adds a deprecation warning for prop searchingText

As per comment on #930, most of this code can be removed after version 2 of react select is released.
See also #932


hellaeon:
This PR is in relation to https://github.com/JedWatson/react-select/pull/930

JedWatson:
As per my comment on #932, I'm going to close this as out of date, but want to do a sweeping fix for all our text/component extension props in the near future. Thanks for the PR.

PR #938, opened 2016-05-04
Adding the fix for missing border inside table to LESS

jackhsu978:
The problem where the border isn't showing when the element is inside a
table was only added to the SCSS version in [this commit](https://github.com/JedWatson/react-select/commit/4b8d12832b3f5c6c9d3ca8
bd003a8c0cb8fd8aea). Adding it to the LESS version.

### DIFF
--- a/less/control.less
+++ b/less/control.less
@@ -51,6 +51,8 @@
 	color: @select-text-color;
 	cursor: default;
 	display: table;
+	border-spacing: 0;
+	border-collapse: separate;
 	height: @select-input-height;
 	outline: none;
 	overflow: hidden;


jackhsu978:
@JedWatson Is there any reason to have the `scss` directory at all if the build process is going to ignore it? Having both `scss` and `less` seems to be redundant and can be confusing to people.

JedWatson:
Thanks for this @jackhsu978!

The reason for having both SCSS and LESS is so that users of either preprocessor can include react-select and use variables to control the appearance.

Having only LESS wouldn't be very useful for people with SCSS stylesheets, for example.

It's not ideal from a maintenance point of view but my impression is that having both is valuable so for now, I'm persisting with it.

Thanks for helping keep them in sync!

jackhsu978:
I see, so it's the "LESSer" evil. Thank you for your detailed explanation @JedWatson!

gojohnnygo:
ps. This PR gives you the option to build the SCSS. 

https://github.com/JedWatson/react-component-gulp-tasks/pulls

Issue #940, opened 2016-05-05
[1.0.0-beta13] Async options with Promises displays [object Promise] as input value


matejlauko:
When loading options async with Promises, like in the example `[object Promise]` string is set as input value of Select
```js
const getOptions = (input) => {
  return fetch(`/users/${input}.json`)
    .then((response) => {
      return response.json();
    }).then((json) => {
      return { options: json };
    });
}

<Select.Async
    loadOptions={getOptions}
/>
```

It is because of #907.

`loadOptions` from Async component attaches to `onInputChange` on Select component.
so this
```js
if (this.state.inputValue !== event.target.value && this.props.onInputChange) {
    let nextState = this.props.onInputChange(newInputValue);
    if (nextState != null) {
        newInputValue = '' + nextState;
    }
 }
```
sets newInputValue to return value of loadOptions, which is `[object Promise]`

VladimirPal:
I've got the same error.
https://github.com/VladimirPal/react-select/tree/async_promise_bug - here i've added example with fetch.js

matejlauko:
@VladimirPal thx for the fix

JedWatson:
Fixed by #941, thanks @VladimirPal!

And thanks @matejlauko for the comprehensive report too :)

dvreed77:
I'm running 1.0.0-beta13 and am still experiencing this bug.

VladimirPal:
@dvreed77 Currently this fix only in master branch

dvreed77:
maybe dumb question, but can I force use this branch from npm?

VladimirPal:
Try this:
`"react-select": "git+ssh://git@github.com:JedWatson/react-select.git"`

VladimirPal:
@dvreed77 But it will not help you:( You have to build lib folder manually.
`cd node_modules/react-select`
`npm install && npm build`

dvreed77:
I am not sure what is happening, but I can't npm to install.  NPM just hangs.  Could be an issue with one of the packages?  Here is a snippet

```
npm http request GET https://registry.npmjs.org/xml-name-validator
npm http 200 https://registry.npmjs.org/array-equal
npm http 304 https://registry.npmjs.org/abab
npm http 304 https://registry.npmjs.org/parse5
npm http 304 https://registry.npmjs.org/nwmatcher
npm http 304 https://registry.npmjs.org/cssom
npm http 304 https://registry.npmjs.org/whatwg-url
[...]

PR #941, opened 2016-05-06
Async options with Promises - Promise object bug

VladimirPal:
https://github.com/JedWatson/react-select/issues/940

### DIFF
--- a/examples/src/app.js
+++ b/examples/src/app.js
@@ -5,6 +5,7 @@ import ReactDOM from 'react-dom';
 import Select from 'react-select';
 
 import Contributors from './components/Contributors';
+import GithubUsers from './components/GithubUsers';
 import CustomComponents from './components/CustomComponents';
 import CustomRender from './components/CustomRender';
 import Multiselect from './components/Multiselect';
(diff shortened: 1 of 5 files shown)


VladimirPal:
I've broken tests, trying to figure out...

JedWatson:
@VladimirPal wow, thanks for not just fixing this but adding a compete example! great work 👍

f2net:
In the /lib folder I still see the old Select.js without this fix. Why?

Thank you,
Matteo Frana

VladimirPal:
@f2net It's because lib folder builds automatically after releasing new version, looks like you have to build it manually:
`cd node_modules/react-select`
`npm install && npm build`

VladimirPal:
@JedWatson Maybe you can build current state of project or release new version? I think this bug is pretty important

f2net:
@JedWatson @VladimirPal I agree, I think the current state should be built, otherwise who installs the package via npm without building gets the [Promise object] bug. 
Thank you for this great component.

eliseumds:
May we have a release?

f2net:
Could you update the /lib version too, since the file in the /lib folder is still bugged?
Every time I distribute an app with react-select async I have to manually change line 390 of Select.js in /lib to:
```javascript
if (nextState != null && typeof nextState !== 'object') {
```
Thank you,
Matteo

jeffreybrowning:
I would love a release.

khankuan:
@JedWatson any chance of a release soon? :)

TrevorBurnham:
FYI I've published a release of the [react-select-plus](https://github.com/HubSpot/react-select-plus) with this patch, `1.0.0-beta13.patch1`.

PR #942, opened 2016-05-06
Fix searchingText bug

tonydecat:
When using the Async component with a large number of options, it's possible that refining the search input doesn't match any previously loaded results, but may match new results returned by the server. There was a bug in the text that's displayed

For example, let's say a search of "a" asynchronously returned options

- "aardvark"
- ...
- "aaz"

The user then refines their search to "ab"
This shows a loading spinner and no options, while it makes an asynchronous request with the new search query.
There was a bug where "No results found" was displayed under the search input while waiting for a response, instead of "Searching...". This PR fixes that bug.

I'd be happy to clarify what the bug is if it's still not clear to anyone.

(https://github.com/JedWatson/react-select/pull/941 or similar should get merged first since that fixes a more general bug with async)

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -141,9 +141,10 @@ const Async = React.createClass({
 		let { isLoading, options } = this.state;
 		if (this.props.isLoading) isLoading = true;
 		let placeholder = isLoading ? this.props.loadingPlaceholder : this.props.placeholder;
-		if (!options.length) {
-			if (this._lastInput.length < this.props.minimumInput) noResultsText = this.props.searchPromptText;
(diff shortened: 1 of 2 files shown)


JedWatson:
That's clear, thanks @tonydecat!

JuanCrg90:
hi. Is the package.json updated to this commit?

tonydecat:
@JuanCrg90 If you're using npm I believe you'll have to wait for the next release (probably v1.0.0-beta14) to get this fix. Perhaps @JedWatson can confirm

PR #943, opened 2016-05-06
Fix a focus bug on iOS

tonydecat:
The bug meant that calling `focus()` programmatically (or using `autofocus={true}`) could render the select component useless on iOS, since it refused to focus on a tap event if it thought it was already focused.

There are some other ways to fix this, like using `document.activeElement` to detect if the input is focused rather than maintaining `isFocused` in the component state, but I wasn't sure if there could be downsides to that approach or if there was a reason that was not done in the first place. This seemed like the simplest change.

I'm also not sure if there's a great way to unit test this, since it's a fairly iOS specific issue. I'm open to suggestions and discussion.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -271,6 +271,11 @@ const Select = React.createClass({
 		}
 
 		if (this.state.isFocused) {
+			// On iOS, we can get into a state where we think the input is focused but it isn't really,
+			// since iOS ignores programmatic calls to input.focus() that weren't triggered by a click event.
+			// Call focus() again here to be safe.
+			this.focus();
+
 			// if the input is focused, ensure the menu is open
 			this.setState({
 				isOpen: true,


JedWatson:
Thanks @tonydecat - seems like a pretty elegant fix.

@bruderstein might have an idea on how to test this, I'm happy to merge though so we can get the fix out!

bruderstein:
@tonydecat I'm just trying to understand the issue (I've not got an iOS device, so it's tricky to try things out to see what's going on..)

I'll just summarise my understanding, then if you could confirm that's correct, or correct it where it's inaccurate, I can try to come up with a test for it :)

So if I've understood correctly, element.focus() is ignored by iOS unless it is contained within a click handler (or some more advanced thing that means "originated from a click handler").  This means that we can get into a state where Select.state has `isFocused === true`, but actually we're not focused.

That kinda makes sense. So  `isFocused` is only set to true in `handleInputFocus`, which is called for `onFocus` for the input or input placeholder div. So if the user calls `document.getElementById('my-react-select-div').focus()`, the `onFocus` get's called, but we're not actually focused?  In order to do this, iOS must be triggering the `onFocus` handler, even though it has no intention of focussing the element? Or am I just confused?

I suspect we can mimic the behaviour by sending the relevant events, but we need to know what events are actually sent by iOS in order to get into this state.

Thank you!

tonydecat:
@bruderstein Yes, your summary of this issue matches my understanding. And yes, from my testing iOS does fire the `onFocus` event even when it doesn't actually give focus (which I found a bit strange, and is what causes this bug).

(I believe it would be possible to get around this behaviour by creating an artificial click event and using that to focus, although I'm not sure if that's a great idea. From what I understand Apple's handling of `focus()` calls in this manner is quite intentional, and designed to prevent the keyboard from opening without the user taking some action, as `focus()`ing an element puts the cursor there and the iOS keyboard slides up from the bottom.)

I don't know if there's any official documentation from Apple about this, but [here's a stack overflow link discussing this issue](http://stackoverflow.com/questions/12204571/mobile-safari-javascript-focus-method-on-inputfield-only-works-with-click).

As far as testing goes, one possible approach might be to manually trigger react-select's `onFocus` handler, and then call focus() and make sure that the react-select component is open and editable. Curious to hear if you have more thoughts

PR #946, opened 2016-05-08
Make input act mmore natively

gojohnnygo:
This PR allows the `react-select` input to behave more similarly to native HTML input.

1. Users can now move the cursor by clicking the text in the input.
2. Clicking empty space to the right of text moves the cursor to end of the text.


JedWatson:
These look great @gojohnnygo, thanks!

PR #950, opened 2016-05-10
Always pass valueArray to renderOuter

aselbie:
I am currently extending react-select to allow the options to stay visible in the dropdown menu with an indication as to whether they've been selected or not (see below). I'm able to do most of this with `filterOptions: false` and `optionsComponent: MyCustomOptionComponent`.

It appears, though, that if the `multi` option is set to `true` that the `valueArray` will not be passed down through `renderOuter` to `renderMenu`, which means that the `isSelected` value that gets passed to the `Option` components will **always** equal `null`

image

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -891,7 +891,7 @@ const Select = React.createClass({
 					{this.renderClear()}
 					{this.renderArrow()}
 				</div>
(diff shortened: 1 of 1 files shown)


nraynaud:
Hi I just stumbled on the same bug, did the maintainer have any reaction to this PR ?

PR #952, opened 2016-05-10
Add support for bold format in label matches

### LINKED ISSUE #5: String matches bold
Perhaps the results could bold the matching parts of the search string? eg.

Search: Sou
Results: New **Sou**th Wales

Xaxatix:
Possibly fixes #5. Currently not extensible, but support for custom classes can easily be added. Also is disabled if a `optionRenderer`  prop is passed.


Siyfion:
If this fixes #5, any idea what the hold-up is in getting it merged @JedWatson?

JedWatson:
Closing this as it can easily be implemented with a custom options renderer (we now pass `inputValue`)

PR #954, opened 2016-05-11
add delay options for Select.Async component, which will reduce network cost while loading remote options

alvinyao:
rt


Grsmto:
Hey, this is cool, just want to precise that you can also use a debounce function to do that.
As a note I had to use the[ `callback()` syntax](https://github.com/JedWatson/react-select#async-options) and not the "Async options with Promises" as the Lodash `debounce()` is returning `undefined` on Promises (or maybe I was doing something wrong).

alvinyao:
The old scenario：
* user input query characters 'abcd', there will be 4 remote query connections generated with params 'a', 'ab', 'abc' and 'abcd' whatever we do.

With the debounce function, the new scenario will be:
* user input query characters 'abcd' once only, there will be only 1 remote query connection generated with param 'abcd';
* In case we set `delay=500`, user input query character 'a', pause a while and continue to input 'bcd', then there will be 2 remote query connections generated with params 'a' and 'abcd'.

I think it's no need to let user to customize the debounce function. Also, I found it difficult to use [callback() syntax](https://github.com/JedWatson/react-select#async-options) with debounce function, which is different from setTimeount.

upq:
Any news with this PR ,  I think this features is really critical.. and shouldve been a core feature.

andrewmclagan:
Would be awesome to get this merged!!

andrewmclagan:
Any movement on this?

bvaughn:
👍 for @andrewmclagan suggestion. This sort of functionality is definitely useful but I think it's probably best managed by application code. Adding it to the library adds complexity for all of the users who _don't_ want/need the functionality.

For example, one way you could implement this:
```js
let lastPromise, lastPromiseTime

function loadOptions (inputValue) {
  if (Date.now() - lastPromiseTime < 5e3) { // 5 second delay between loads
    return lastPromise
  } else {
    lastPromiseTime = Date.now()
    lastPromise = fetch(...) // Load options from your API
    return lastPromise
  }
}
```

Note that the above would result in multiple calls to `loadOptions` only loading options for the first call. Maybe this is what you want, maybe not. Alternately you could create a queue of promises and resolve them after the timeout. etc

bkyryliuk:
@bvaughn - thanks for the suggestion.
I am fairly new in the react world. Do you have any suggestion how to implement the delay externally that will handle a case when user entered couple symbols, request issued and then user added couple more, but the fetch(...) wasn't called as the delay wasn't met ?

bvaughn:
At work and don't have time to really think about this now. Maybe something like?

```js
let lastPromise, lastPromiseTime, lastInputValue

function loadOptions (inputValue) {
  lastInputValue = inputValue

  if (Date.now() - lastPromiseTime < 5e3) { // 5 second delay between loads
    return lastPromise
  } else {
    lastPromiseTime = Date.now()
    lastPromise = fetch(...) // Load options from your API

    lastPromise.then(
      () => {
        if (inputValue !== lastInputValue) {
          // Newer input was ignored due to debounce
          // Trigger another load to pull it in
          loadOptions(loadOptions)
        }
      }
    )

    return lastPromise
  }
}
```

bmcmahen:
In my experience, it's easiest to use something like [debounce-promise](bjoerge/debounce-promise) to wrap your `loadOptions` function.

PR #961, opened 2016-05-16
Update Select.js

damienleroux:
update `lib/select.js` with `src/select.js` code. otherwise, using react-select as node module package give an error on the targeted line. Thank you


JedWatson:
Cheers @damienleroux

PR #969, opened 2016-05-18
Adds unit (px) to minWidth attribute

### LINKED ISSUE #820: Warning: a `input` tag was passed a numeric string value for CSS property (React v15.0)
I trying to upgrade my application to [React v15.0](https://facebook.github.io/react/blog/2016/03/07/react-v15-rc1.html). After the upgrade, I saw a message on the console:

```
Warning: a `input` tag (owner: `AutosizeInput`) was passed a numeric string value for CSS property
`width` (value: `5`) which will be treated as a unitless number in a future version of React.
```

Perhaps, could you correct it now?

ianwitherow:
Fixes #820


JedWatson:
Thanks @ianwitherow!

PR #973, opened 2016-05-19
Update GithubUsers.js

adaptunit:
chain call need 'return'


jooj123:
`.then((response) => response.json())`
is just shorthand syntax with arrow functions. See (shorter functions section): 
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions

bvaughn:
@jooj123 's correct. This is just the shorter syntax. 😄

PR #976, opened 2016-05-20
style prop warning - can be either object or array

andreigec:
had warnings with the style prop type


gwyneplaine:
@andreigec, just wanted to check in with this PR and clarify its use, as React still doesn't support an array in style props as of React 16

JedWatson:
Looking further into this, react-dom doesn't support arrays for styles - see facebook/react#2196

So I don't think it's a good idea to support the propType `Array` here

Issue #978, opened 2016-05-20
Error if `onChange` causes the component to unmount.


Xesued:
If an `onChange` callback cause the component to unmount, the `Select`'s `selectValue` tries to setState after, causing a React Error.

Xesued:
I believe that swapping the `this.setState(...)` and `this.setValue(value)`  (https://github.com/JedWatson/react-select/blob/master/src/Select.js#L499)  should solve the issue.

bvaughn:
@Xesued,

Can you provide a reproduction test-case for this problem? I'm having some difficulty reproducing it the error you're describing with 1.0 beta 13 and React 15.

Xesued:
https://plnkr.co/edit/nXHsCdRyUfxa8OR48bmd?p=info

If you change the results, you will notice an error thrown in the console.

PR #979, opened 2016-05-20
Fixed React error if onChange callback causes a root component to unm…

Xesued:
…ount.

@bruderstein I'm not sure how you would unit test this, I'm still new to React.  So, looking for guidance here.  Fixes https://github.com/JedWatson/react-select/issues/978


JedWatson:
Good call @Xesued, thanks!

PR #981, opened 2016-05-22
extend to support usage as input field

ionelh:
I needed something like React-Select, except I wanted it to behave more like an input field. I wanted to have more control on when suggestions open, like sending the minimum number of chars in the suggestion need before showing up. Also, when the field is populated with one or more values, the enter key press calls a handler provided in the props, which can do stuff like search operations.


dentuzhik:
That would be quite useful for me as well.

Issue #983, opened 2016-05-23
Document need to be updated to require style


rahulchavan30:
Document needs to be updated to require "require('react-select/dist/react-select.css')", 
without using this its not possible and component rendered is not what is expected . 

i included this and everything worked perfectly fine thanks for all the work you have done on this component

jkirkpatrick:
For ES6+:
`import 'react-select/dist/react-select.css'; `

bvaughn:
FWIW there are 2 ways for you to include react-select styles in your project.

#### Use a CDN

In your index.html file load the styles directly from NPM CDN:
```html
<link rel="stylesheet" href="https://npmcdn.com/react-select@1.0.0-beta13/dist/react-select.css">
```

#### Use a bundler

(Assuming you've `npm install`ed react-select...) Somewhere in your entry file either _require_ or _include_ the styles:
```js
// either...
require('react-select/dist/react-select.css');
// or...
import 'react-select/dist/react-select.css';
```

I'll update the documentation to reflect this.

bvaughn:
I've updated the [installation instructions](https://github.com/JedWatson/react-select#installation) in the main README to show how to import styles. Thanks for the suggestion!

PR #985, opened 2016-05-23
#983 Readme includes require CSS

glundgren93:
This should fix issue #983.


gwyneplaine:
As noted in #983, a fix for this has already been pushed to master, but thank you for this!

PR #988, opened 2016-05-25
Add support for assistive technologies

### LINKED ISSUE #353: add support for assistive technologies
Hello,

key support is good, but It would be great to improve accessibility with meaningful tags (ul/li) or aria attributes.

king regards

bruderstein:
This adds support for aria. Some minor changes to the markup have been
made in order to get a better experience with a screen reader.

This should be regarded as a first step, and there will almost certainly
be more things we can do to improve this further, and fixes for other
screenreaders etc.  

There's some tidy-up I'd like to still do, particularly with the `focusedOption` / `focusedOptionIndex` tracking, as that can probably be simplified.  But I wanted to get this up for review, so we can see what else needs doing. 

I've converted some tests that started failing due to the changes to unexpected-react, as there is an extra `&nbsp;` in the label, hidden except from screenreaders - this made the tests using the label tricky with just the DOM. (I'm very biased when it comes to this lib, so completely understand if we'd rather just DOM testing :smile: - can be changed with some effort, but I think the tests are actually more readable with it - but bias and full disclosure blah blah :wink:)

I've tried this out in ChromeVox and Orca on Linux. I'd be immensely grateful for feedback from users of other screenreaders (or any regular screenreader users).

I recorded a (fairly poor sound quality) video, to show the difference - https://www.youtube.com/watch?v=Qzr1Y9CFVDk

Fixes #353

### DIFF
--- a/less/control.less
+++ b/less/control.less
@@ -112,8 +112,8 @@
 	white-space: nowrap;
 }
 
-.has-value.Select--single > .Select-control > .Select-value,
-.has-value.is-pseudo-focused.Select--single > .Select-control > .Select-value {
+.has-value.Select--single > .Select-control .Select-value,
+.has-value.is-pseudo-focused.Select--single > .Select-control .Select-value {
 	.Select-value-label {
 		color: @select-text-color;
 	}
@@ -240,8 +240,17 @@
 	border-top-color: @select-arrow-color-hover;
 }
 
-
-
+.Select--multi .Select-multi-value-wrapper {
+  display: inline-block;
+}
+.Select .Select-aria-only {
+  display: inline-block;
+  height: 1px;
+  width: 1px;
+  margin: -1px;
+  clip: rect(0,0,0,0);
+  overflow: hidden;
+}
 
 // Animation
 // ------------------------------
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "sinon": "^1.17.3",
     "unexpected": "^10.13.2",
     "unexpected-dom": "^3.1.0",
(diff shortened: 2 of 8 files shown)


JedWatson:
This looks fantastic, @bruderstein, thanks!

Brilliant work as always.

reintroducing:
@bruderstein Thanks for this Dave, really appreciate it.  I'm using react-a11y in my project during dev to help with the a11y efforts and I'm seeing the following warning in the console:

`AutosizeInput You have an unlabeled element or control. Add `aria-label` or `aria-labelledby` attribute, or put some text in the element.`

This happens every single time that I open the dropdown, roll over an item, and select an item.  Is there a way to prevent this?  Am I missing an attribute being passed somewhere which I'm unaware of?

bruderstein:
@reintroducing I think that's actually an issue with the [AutosizeInput](https://github.com/JedWatson/react-input-autosize) project. I've got very little time at the moment, but maybe you could take a look - possibly they just need to pass an `aria-label` or `aria-labelledby` through to the actual input element?

reintroducing:
@bruderstein Good call, I went ahead and did that.  Brain farted and didn't realize it was a separate component.

jasonh42:
@JedWatson @bruderstein The addition of "Select-multi-value-wrapper" seems to have broken some styles in beta14. E.g. https://github.com/JedWatson/react-select/blob/v1.0.0-beta14/dist/react-select.css#L88 is no longer valid, because the structure now looks like .Select > .Select-control > .Select-multi-value-wrapper > Select-value ....

Issue #967, opened 2016-05-18
Text in input not selectable


Reggino:
Thanks for this!

When typing text in the text-input to search for an option, it's not possible to select the search-text using the mouse. One has to remove the input using the backspace key, which is a bit inconvenient. 

Related to #916

daletune:
Any plans for this by maintainers?

MarkLeMerise:
Not straightforward, but you can use `Ctrl + A` to select all and then use the arrows to navigate within the word, so at least it's a workaround until a fix.

fleetofthemalden:
@JedWatson Any plans to support this?

Without it we need to choose another implementation for our org.

JedWatson:
This has been fixed

IamKritika:
@JedWatson this is not working. I was referring to the examples in the documentation. Can you please look into this?

pk936:
I am facing the same issue.

vertic4l:
Nice maintenance on this one!

Issue #916, opened 2016-04-27
On typing input , unable to move cursor position via mouse


aqumus:
Just while searching options , when we type on input field , not able to move cursor position with mouse click to specific position of input field.
Any workaround for this issue ?

JedWatson:
This has been fixed

PR #989, opened 2016-05-26
Fix for #916 and #967

Reggino:
Any input in the search-box now is selectable using the mouse. This is convenient for Desktop-users.


Reggino:
I don't think the failing test is related to the change?

ryanzec:
It is not, I am going to be pushing a pull request to fix allowCreate for 1.0.0-beta and I fixed that other test too.

JedWatson:
This has been fixed

PR #992, opened 2016-05-27
Promise bug 984

lobsteropteryx:
This feels a bit hacky to me, but I think it works.  Long term, it might be better to throw if `thenPromise` gets something that's not a promise?

EDIT:  I believe there is currently a single, failing test on master, which this change doesn't address.


JedWatson:
Took me a bit to get my head around what you're doing here but now it makes sense. Looks like a good solution. Thanks @lobsteropteryx!

Lacerda:
I'm sorry, will this be published in a new version any time soon? :) 
I don't think `v1.0.0-beta13` has this fix or am I missing something?

PR #995, opened 2016-05-30
Update README with autofocus prop

choonkending:
**Reason for Change**

Since [beta 3](https://github.com/JedWatson/react-select/blob/master/HISTORY.md#v100-beta3--2015-11-08), we have had [autofocus](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L37) as a prop
```javascript
...
		autofocus: React.PropTypes.bool,            // autofocus the component on mount
...

```

Just exposing that in the README.md.


JedWatson:
Thanks!

PR #1000, opened 2016-06-01
Don't render backspaceToRemoveMessage if backspaceRemoves is set to false

ryanzec:
Currently if the backspaceRemoves is set to false, the backspaceToRemoveMessage it still displayed which is bad (since backspace won't remove the last item).  this just adds an additional check on the backspaceRemoves to wether it should render that message.


JedWatson:
Good catch @ryanzec. Thanks!

PR #1002, opened 2016-06-02
Fixed bug with getFocusableOptionIndex that prevents selected option …

jooj123:
This fixes a bug in the getFocusableOptionIndex func where on some circumstances it would not return the appropriate focusable option.
The current indexOf check wont work for arrays of objects - it will only work if the object references are the same, example here shows it wont work correctly: http://jsfiddle.net/7B7dQ/1/).

This could be reproduced in the example page by replacing:

```
var options = STATES[this.state.country];
```

in States.js `render()` with:

```
var options = [
    { value: 'australian-capital-territory', label: 'Australian Capital Territory', className: 'State-ACT' },
    { value: 'new-south-wales', label: 'New South Wales', className: 'State-NSW' },
    { value: 'victoria', label: 'Victoria', className: 'State-Vic' },
    { value: 'queensland', label: 'Queensland', className: 'State-Qld' },
    { value: 'western-australia', label: 'Western Australia', className: 'State-WA' },
    { value: 'south-australia', label: 'South Australia', className: 'State-SA' },
    { value: 'tasmania', label: 'Tasmania', className: 'State-Tas' },
    { value: 'northern-territory', label: 'Northern Territory', className: 'State-NT' },
];
```

If you select (click) option 'Northern Territory' for example and then re-open the selector - it always focuses on the first option.
This fails because its a new reference on each render.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -920,7 +920,7 @@ const Select = React.createClass({
 	},
 
 	getFocusableOptionIndex (selectedOption) {
-		let { valueKey, multi } = this.props;
+		let { valueKey } = this.props;
 		let options = this._visibleOptions;
 		if (!options.length) return null;
 
--- a/src/Select.js
+++ b/src/Select.js
@@ -920,12 +920,16 @@ const Select = React.createClass({
 	},
 
 	getFocusableOptionIndex (selectedOption) {
-		var options = this._visibleOptions;
+		let { valueKey, multi } = this.props;
+		let options = this._visibleOptions;
 		if (!options.length) return null;
 
 		let focusedOption = this.state.focusedOption || selectedOption;
 		if (focusedOption && !focusedOption.disabled) {
-			const focusedOptionIndex = options.indexOf(focusedOption);
+			const focusedOptionIndex = options
+				.map((option) => option.hasOwnProperty(valueKey) ? option[valueKey] : option)
+				.indexOf(focusedOption.hasOwnProperty(valueKey) ? focusedOption[valueKey] : focusedOption);
+
 			if (focusedOptionIndex !== -1) {
 				return focusedOptionIndex;
 			}
@@ -934,6 +938,7 @@ const Select = React.createClass({
 		for (var i = 0; i < options.length; i++) {
 			if (!options[i].disabled) return i;
 		}
+
 		return null;
 	},
 


jooj123:
I was unable to reopen my PR from https://github.com/JedWatson/react-select/pull/797 after rebasing - so this has been opened as another PR

src/Select.js
```diff
@@ -879,9 +880,17 @@ const Select = React.createClass({
 		}
 
 		for (var i = 0; i < options.length; i++) {
-			if (!options[i].disabled) return i;
+			// if there is no selected option just use first value in options
+			if (!options[i].disabled && !focusedOption) {
+				return i;
```
reubenmiller: I assume this should actually be 
`return options[i]`

src/Select.js
```diff
@@ -879,9 +880,17 @@ const Select = React.createClass({
...
+			} else if (!options[i].disabled && options[i].value === focusedOption.value) {
+				return i;
```
reubenmiller: and here (same comment as above):
`return options[i]`

src/Select.js
```diff
@@ -879,9 +880,17 @@ const Select = React.createClass({
 		}
 
 		for (var i = 0; i < options.length; i++) {
-			if (!options[i].disabled) return i;
+			// if there is no selected option just use first value in options
+			if (!options[i].disabled && !focusedOption) {
+				return i;
+			} else if (!options[i].disabled && options[i].value === focusedOption.value) {
```
reubenmiller: Instead of using the hard coded 'value' prop. Use the valueKey prop. That way it works when a non-standard valueKey is used.
`} else if (!options[i].disabled && options[i][this.props.valueKey] === focusedOption[this.props.valueKey]) {`

src/Select.js
```diff
@@ -879,9 +880,17 @@ const Select = React.createClass({
...
+				return i;
+			} else if (!options[i].disabled && !selected) {
+				// get first option thats not disabled
+				selected = i;
```
reubenmiller: should be:
`return selected = options[i];`

This way it will be consistent with the other return statements in the for loop.

src/Select.js
reubenmiller: Please use the reference to the `this.props.valueKey` instead of the `.value`.

(Sorry for the previous noise with the comments, I was comparing it against the 1.0.0.beta13 build and not the master.

jooj123:
@JedWatson is there anything stopping this one from being merged?

gwyneplaine:
thanks for this @jooj123, this issue has since been resolved in subsequent PRs.

Issue #1001, opened 2016-06-02
Add support for page up, page down, end, and home keys for long lists.


blackjk3:
Add support for page up, page down, end, and home keys for long lists.

- Page up/down would default to jump 5 items.
- Home and End key behavior would be standard jump to start/end of the list.

JedWatson:
Implemented #1003

PR #1003, opened 2016-06-02
Adding Home, End, Page Up/Down support. #1001

blackjk3:


JedWatson:
This is really nice @blackjk3, thanks!

PR #1008, opened 2016-06-03
Prevent getFocusableOptionIndex from returning a disabled option.

### LINKED ISSUE #1006: Disabled options shouldn't be focusable
We're using react-select as a multi-select field with a custom `filterOptions` function so that when options are selected, they are disabled rather than filtered out of the available list.

The problem we're experiencing is that once an option has been focused and then selected, it is allowed to be focused again and re-selected through tab-completion.

This Plunker demonstrates our usage and the issue we are experiencing:  
https://plnkr.co/edit/JLobYn22fBTK7Lnk8azt?p=preview

In the above example, "One" is allowed to be focused and added multiple times.

This behavior looks like it can be prevented by ensuring that `getFocusableOptionIndex` never returns a disabled option.  

I'm going to submit a quick PR for this for your review.

brianspowers:
This PR prevents `getFocusableOptionIndex` from ever returning the index of a disabled option.

This resolves #1006 and prevents disabled options from ever receiving focus and becoming selectable by tab-completion.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -871,7 +871,7 @@ const Select = React.createClass({
 		if (!options.length) return null;
 
 		let focusedOption = this.state.focusedOption || selectedOption;
-		if (focusedOption) {
+		if (focusedOption && !focusedOption.disabled) {
 			const focusedOptionIndex = options.indexOf(focusedOption);
 			if (focusedOptionIndex !== -1) {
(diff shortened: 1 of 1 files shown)


JedWatson:
Neat fix @brianspowers, thanks!

PR #1014, opened 2016-06-07
Update the Readme.md

EduardoCHenrique:
just to say that this function receives the option like onChange for example .


JedWatson:
Thanks @EduardoCHenrique

PR #1021, opened 2016-06-10
Close opened menu if disabled becomes true

### LINKED ISSUE #762: Disabling the component doesn't close the menu
Passing the "disabled={true}", doesn't close the menu.
It means that if you update then the component with "disabled={false}", the component is bugged: it doesn't close properly because the state.isOpen is still false (even though the menu is rendered).

So the closeMenu() method at https://github.com/JedWatson/react-select/blob/master/src/Select.js#L247 isn't invoked.

jsnmoon:
This fixes #762 by invoking `this.closeMenu()` when `disabled` is changed. I also added a corresponding test. Cheers!

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -212,6 +212,7 @@ const Select = React.createClass({
 		}
 		if (prevProps.disabled !== this.props.disabled) {
 			this.setState({ isFocused: false }); // eslint-disable-line react/no-did-update-set-state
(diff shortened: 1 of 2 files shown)


JedWatson:
Looks great @jsnmoon, thanks!

NeutralMike:
Bug is open now.

PR #1032, opened 2016-06-16
Validate input change

ndreckshage:
Have way of validating input change.

Use case is in my app I'm supporting a card reader that takes in raw input that I need to parse. If I detect raw input vs user keypress, I want to skip entry.

I looked into doing this with inputRenderer but that seems pretty limited as it doesn't pass down props, state, etc.


bradbirnbaum:
+1 - we could use this as well

JedWatson:
Thanks for the PR @ndreckshage, sorry it took so long to review.

I'd prefer to approach this some other way - you mentioned that inputRenderer is limited, maybe we could extend that instead?

Going to close this, but let me know if it's still relevant and we can work out how to extend it without adding another prop.

PR #1033, opened 2016-06-17
Update LICENSE to a YEAR-present format

buildbreakdo:
Picked this up from the Facebook React Repo so you don't have to update the year.. well every year! That gets old.

See here:
https://github.com/facebook/react/blob/master/LICENSE

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2016 Jed Watson
+Copyright (c) 2016-present Jed Watson
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


JedWatson:
I'm a bit confused about this tbh. I'm sure Facebook have good advice, but I haven't seen it done often and couldn't find anything about the validity of this with a quick google. Do you have any other background about it? Is there anything else that needs to be done to ensure the validity of this format?

My understanding is that copyright should be updated when you make changes, and otherwise won't perpetually be valid as of the "present", so I usually update copyright on projects in a new year when I actually make changes to them...

buildbreakdo:
Well this is interesting.. seems the date should not be updated but added to.. maybe "-present" is a version of added to? Assume something as serious as changing wording in a license was vetted by FB legal.

Maybe @rickbeerendonk would be kind enough to confirm? :)

http://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

> The copyright notice on a work establishes a claim to copyright. The date on the notice establishes how far back the claim is made. This means if you update the date, you are no longer claiming the copyright for the original date and that means if somebody has copied the work in the meantime and they claim its theirs on the ground that their publishing the copy was before your claim, then it will be difficult to establish who is the originator of the work.
> 
> Therefore, if the claim is based on common law copyright (not formally registered), then the date should be the date of first publication. If the claim is a registered copyright, then the date should be the date claimed in the registration. In cases where the work was substantially revised you may establish a new copyright claim to the revised work by adding another copyright notice with a newer date or by adding an additional date to the existing notice as in "© 2000, 2010". Again, the added date establishes how far back the claim is made on the revision.
> 

From the horses mouth (least in the US) http://www.copyright.gov/circs/circ03.pdf:

> Use of the notice informs the public that a work is protected by copyright,
> identifies the copyright owner, and **shows the year of first publication**.

JedWatson:
Trying to clean up all open PRs; going to close this one, because it doesn't need to be updated yearly, but I do like the aspect that it's updated in each year that some nontrivial level of work is done on react-select. Thanks @buildbreakdo

PR #999, opened 2016-06-01
Implemented allowCreate and more flexible add item on comma functionality

ryanzec:
I took the pull request #660 (which relates to #658 and #586) and re-did it as this is a huge show stopper for me on using this library and that implementation seemed a bit broken (and the 0.9.1 does not have some of the functionality that I also need the 1.0.0-beta does).  

This adds back in the allowCreate functionality with the tests reworked to pass (I also fixed an existing failing test as it seemed to be using the old `setPropselectorChild` method instead of `setPropsForChild`)

**UPDATE:**

I added another commit here to add a more flexible solution to the old add item on comma when `allowCreate` and `mutli` are enabled.  My solution allows you to specific the keyCode to use to add a new item (set with the `addItemOnKeyCode` prop).  While comma is pretty common, it is not always the best option and making it flexible like this it not that much difference in the implementation.  This also allows you to turn off this feature which was not possible in the older implementation which i think is not completely uncommon.  By default the `addItemOnKeyCode` prop is set to null so that the user need to explicitly enable this functionality.

Also, I have no idea what the @coveralls messages mean.  I did run coverage reports before and after and the coverage did not go down after my src / test updates (they actually went up on 1 or 2 src files but only marginally).

If I could get an update if this will be an acceptable PR (or if any changes are needed to get this merged) and included into a npm tagged beta release relatively soon that would be great.  Without these changes, using this library is a show stopper.  Since 0.9.1 is also missing functionality the 1.x beta version has, using that version is also a show stopper.  If these features are not going to be able to be consumable through npm relatively soon, I am going to have to publish my own version to be able to move forward with this library which I would prefer not to do.

PR #1042, opened 2016-06-23
Update Readme for clarity on 1.0.0-beta

gertig:
Clarify that allowCreate is not implemented in 1.0.0-beta


gertig:
This is to help with the confusion of those who are updating from 0.9 to 1.0.0-beta and expecting `allowCreate` to work. Once #999 is merged this should be updated to reflect that change.

JedWatson:
Thanks @gertig - appreciated.

PR #1047, opened 2016-06-27
fixes issue with an outline appearing on the auto sized input

ryanzec:
So with the latest code in master, if searchable is set to false, this happen when you click in the input looking element:

screen shot 2016-06-27 at 7 14 23 am

This is basically just the default browsers outline for the input element so I added some css to prevent this issue from happening.


JedWatson:
Thanks @ryanzec

Issue #1019, opened 2016-06-09
Typing out values instead of just selecting them causing space between input and menu


ryanzec:
So it was mentioned in the #999 PR about an issue with a space between the input and menu (and this issue is existing, not related to that PR) and I want to give some more insight I noticed about this issue.

First I want to point out that I testing this on Chrome, Safari, FireFox, and IE 10 and oddly enough, I only noticed this issue in Chrome.

This issue seems to happen when the user types the value more instead of just selecting a value.  Here is an existing example selecting 3 items and everything works fine:

![working](https://cloud.githubusercontent.com/assets/444206/15928943/f8bd3eec-2e17-11e6-8419-fdb7de2620bb.gif)

Here is me selecting the same 3 items for the example except I am typing out the value before selecting them:

![broken](https://cloud.githubusercontent.com/assets/444206/15928950/fe1ec266-2e17-11e6-86bb-2e66bd11ac1c.gif)

For whatever reason when I type the value out, the main element (`.Select`) seems to think it is taller than it should be as soon as some causing the component to rendering again, it fixes itself.

ryanzec:
Ok, so I did some investigation and that only thing that made sense to me was that the value of the input and the tag was appearing in the input at the same time causing the input to briefly become multiple lines and for whatever reason, Chrome was not re-sizing the height of the element. once the input value was cleared out  As an experiment, I change the follow code:

```javascript
selectValue (value) {
    this.hasScrolledToOption = false;
    if (this.props.multi) {
        this.addValue(value);
        this.setState({
            inputValue: '',
            focusedIndex: null
        });
    } else {
        this.setState({
            isOpen: false,
            inputValue: '',
            isPseudoFocused: this.state.isFocused,
        });
        this.setValue(value);
    }
```

to

```javascript
selectValue (value) {
    this.hasScrolledToOption = false;
    if (this.props.multi) {
        this.setState({
            inputValue: '',
            focusedIndex: null
        }, () => {
            this.addValue(value);
        });
    } else {
        this.setState({
            isOpen: false,
            inputValue: '',
            isPseudoFocused: this.state.isFocused,
        }, () => {
            this.setValue(value);
        });
    }
}
```

In this code, the tag is not added until the input value is cleared and that seemed to fix the issue. I tested this code in all the browsers I previously did and everything seems to work perfectly. I also re-ran the tests and everything passes there too (except the one test that is still using the old `wrapper.setPropselectorChild` that does not exist).

[...]

PR #1048, opened 2016-06-27
Weird spacing fix

ryanzec:
This fixes the issue #1019 about weird extra spacing below the input that was originally reported in PR #999 (though it was an existing issue and not because of changes with the PR which is why I have this separate PR).


ryanzec:
oh, this branch is bad, let me fix it

JedWatson:
Thanks!

PR #1049, opened 2016-06-27
fixes issue with an outline apear on the auto sized input

ryanzec:
Apparently is was the LESS version of the code that needed to be updated (related to #1047).


JedWatson:
It was - I should have caught that when merging your PR. Thanks @ryanzec

PR #1054, opened 2016-06-29
Fix async loadOptions bug.

chenzhihao:
We should not return a promise here.

The code:
```javascript
return inputPromise.then(() => {
  return input;
})
```
is still a promise.

The returned value will be used at handleInputChange() of [Select.js()](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L384);
And this value will be used to 'setState(...)' as 'inputValue' which will finally become '[object Promise]' string as 'inputValue'.


dcurletti:
This should get pulled in ASAP.

bvaughn:
This issue has been resolved via PR #1226 by way of an `Async` refactor.

PR #1064, opened 2016-07-02
Use the new version of react-input-autosize

artemT82:


bdj:
This doesn't actually remove the minWidth warning, but the current master on react-input-autosize does. The version there needs to be bumped and then this can be updated.

JedWatson:
I've updated this now there's a new version of `react-input-autosize` with 15.2 support. Thanks @artemT82!

PR #1066, opened 2016-07-04
Stop propagation in pressing escape key

jmblog:
I use `react-select` on `react-modal`, but when pressing the escape key to close the `react-select` options menu, `react-modal` is also closed unexpectedly. To resolve this issue, I'd like to stop propagation in pressing escape key.

### DIFF
--- a/lib/Select.js
+++ b/lib/Select.js
@@ -450,6 +450,7 @@ var Select = _react2['default'].createClass({
 				} else if (this.props.clearable && this.props.escapeClearsValue) {
 					this.clearValue(event);
 				}
+				event.stopPropagation();
 				break;
 			case 38:
 				// up


klaemo:
I have the same problem and this resolves it. I agree that it's sensible to stop propagation.

JedWatson:
:+1: Makes sense!

JedWatson:
Actually just thinking - wouldn't it make sense to stop propagation if either of the two conditions here are true? So if `esc` has no effect on react-select it propagates like normal... will push a commit for you to review.

(also, changes need to be made to `src`, not `lib`)

jmblog:
@JedWatson It makes sense! It's better than my changes. :+1:
And sorry for changing `lib`. I meant to make changes to `src`. :sweat:

JedWatson:
@jmblog no worries, thanks for the PR!

cnrudd:
I notice that in v2 of the react-select component, this code  (call event.stopPropagation() after the Escape key has been processed) has been dropped.  

v1: https://github.com/JedWatson/react-select/blob/v1.x/src/Select.js#L492
v2: https://github.com/JedWatson/react-select/blob/master/src/Select.js#L1145

Was this intentional?
If so, what is the recommended way in v2 (besides using the onKeyDown prop) to get the behavior discussed in this ticket?

PR #1068, opened 2016-07-04
Fix min width

### LINKED ISSUE #1063: minWidth on <input> tag is invalid
Hi,

There is a breaking change in your ReactInputAutosize library in React 15.2.0 that reports the above error.

For now, I have just downgraded back to 15.1, but this may be something you wish to fix (correcting it in react-input-autosize and bumping the version in this one).

To repro, just upgrade react and view a Select.

Lacerda:
Fixes #1063


caedmon:
The minWidth prop is valid on the <Input> tag.  The problem is that react-select is using an older version of react-input-autosize which passes this prop on to the <input> tag.  If react-select is updated to use react-input-autosize ^1.0.0 this will be fixed.  See #1064

bdj:
react-input-autosize 1.0.0 doesn't actually fix the minWidth warning, but the version in master does.

caedmon:
Thanks for the correction @bdj. I guess react-input-autosize needs a version bump too.  I'm new at this npm business.

PR #1078, opened 2016-07-09
update required prop based on nextProps on update. closes #1077

### LINKED ISSUE #1077: Required attribute not updated properly on the typeahead input when the value and options are changed programmatically
This is similar to [some](https://github.com/JedWatson/react-select/issues/297) [other issues](https://github.com/JedWatson/react-select/issues/806) in a way: the `required` attribute on a `<Select />` gets improperly set on the underlying typeahead input when both the value and options are changed programmatically.

The particular scenario where I discovered this is shown in this jsbin: http://jsbin.com/xeluwo/edit?js,output

The gist is that you might have two `react-select` components where the first one changes the value and options of the second.

mshwery:
This resolves the issue where the required prop (and the `valueArray`) retrieved during `componentWillReceiveProps` failed to reflect the values in `nextProps` due to relying on `this.props` properties.

I wrote a test case that failed before my change, and now it passes as expected.

The basic premise is that if you use the `required` attribute while _updating_ props like `value`, `options`, `valueKey`, it may get improperly set on the typeahead input. For instance, if you change the `value` and `options` to a completely different set the method `this.getValueArray` fails to find the value within the old set of options and despite rendering the appropriate selection text from the nextProps on `render`, the input gets the `required` attribute applied during `componentWillReceiveProps`'s `handledRequired` call.

Not sure what style of documentation you prefer, but I figured I would also add some jsdoc comments for these two methods that I touched.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -169,7 +169,7 @@ const Select = React.createClass({
 	},
 
 	componentWillReceiveProps(nextProps) {
-		const valueArray = this.getValueArray(nextProps.value);
+		const valueArray = this.getValueArray(nextProps.value, nextProps);
 
 		if (nextProps.required) {
 			this.setState({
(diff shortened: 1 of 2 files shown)


JedWatson:
This looks great @mshwery, thanks!

mshwery:
Thanks @JedWatson. When do you think this will make a new release?

PR #1081, opened 2016-07-11
Fix the "Unknown prop `inputClassName` on <div> tag" warning

mxstbr:
By removing the `inputClassName` from the props we pass to the div we
get rid of the warning.


JedWatson:
Sweet, thanks @mxstbr!

PR #1082, opened 2016-07-12
On focus check whether input ref is a real input or an input component

gpoole:
Fixes an issue where configuring react-select with `autosize` off causes an error to be emitted when clicking on the already focused input:

```
Uncaught TypeError: this.refs.input.getInput is not a function
```

It looks like the change in #946 was based on `this.refs.input` being an `Input` instance, but [if `autosize` is disabled it's just a plain HTML input](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L785).

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -292,8 +292,14 @@ const Select = React.createClass({
 			// Call focus() again here to be safe.
 			this.focus();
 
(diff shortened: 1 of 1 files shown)


JedWatson:
Great fix @gpoole, thanks!

PR #1091, opened 2016-07-13
Adjust input clear call

### LINKED ISSUE #1071: this.refs.input.getInput is not a function
commit #eefa843 add this.refs.input.getInput().value = '';
causing Uncaught TypeError: this.refs.input.getInput is not a function

pbrant:
This fixes #1071. I couldn't find any references to .getInput(). Removing it works for me on recent versions of FF and Chrome.


JedWatson:
Thanks @pbrant!

PR #1101, opened 2016-07-18
Add onCloseResetsInput option w/tests

frankievx:
Added a new property to Select similar to the functionality of onBlurResetsInput, to work with closing the menu when clicking on the arrow. I have added the option to the readme, implemented the feature, and written tests to cover the new code that I have added. There has been no change in the coverage and I have followed the guidelines for contribution as closely as possible.


JedWatson:
Thanks @frankievx!

frankievx:
No problem, it's a really good component. I'll see if I can help out some more.

freedom19920918:
May i ask you some question? I always use this component with mobx-react-form.But there has a problem that the close button does not work(the input value does not transform to empty) after I submit the form and then edit it again. And there's no problem when I first edit the Select component

PR #1105, opened 2016-07-19
Added an optional instance id property

JevinAnderson:
This will allow synchronization of component id’s on both the server
and the client.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -144,7 +144,6 @@ const Select = React.createClass({
 
 	getInitialState () {
 		return {
-			instanceId: this.props.instanceId || ++instanceId,
 			inputValue: '',
 			isFocused: false,
 			isLoading: false,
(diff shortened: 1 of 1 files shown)


JevinAnderson:
Should close #1105

JedWatson:
Good solution!

I don't think you need to add the instanceId to the state though. Defaulting to this in the `componentWillMount` method should be sufficient?

```js
this._instancePrefix = 'react-select-' + (this.props.instanceId || ++instanceId) + '-';
```

Did you link to the wrong "should close" issue number btw?

JevinAnderson:
I've updated my solution @JedWatson. Thanks for the suggestion.

etburke:
We're also running into this. Thanks for working out a solution!

JevinAnderson:
@bruderstein @JedWatson 

Any word on this?

JedWatson:
Thanks for that @JevinAnderson, looks good now! Sorry for the delayed merge.

mverderese:
@JedWatson What is your workflow for rebuilding the dist files? I noticed that master is updated with this change, but if I do an `npm install` from master, the dist files are still outdate. I'd like to avoid forking if possible.

Thanks!

JedWatson:
@mverderese I'm about to publish an rc with this fix in it, it should be fine if you update the dependency in your project to that.

gabrielflorit:
I'm still getting the `Warning: React attempted to reuse markup in a container...` warning. I'm on rc2 and I'm passing an `instanceId` to my component (like I believe @JevinAnderson suggested), e.g.

`<Select instanceId='gabrielflorit' ... />`

Am I doing something wrong?

PR #1113, opened 2016-07-21
Apply less changes to scss versions

pbrant:
A couple of small changes from 1d012fc0 and e984b9c3 that didn't make it into the scss version


JedWatson:
Thanks for catching that @pbrant!

PR #1115, opened 2016-07-22
replace `blacklist` dependency by using the spread operator

jochenberger:
See #1081

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
     "url": "https://github.com/JedWatson/react-select.git"
   },
   "dependencies": {
-    "blacklist": "^1.1.2",
     "classnames": "^2.2.4",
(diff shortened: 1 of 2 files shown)


JedWatson:
Thanks @jochenberger! that's nicer. cc @mxstbr

oakley808:
FWIW, it looks like blacklist is still required in /master/lib/Select.js

https://github.com/JedWatson/react-select/blob/e19bce383a8fd1694278de47b6d00a608ea99f2d/lib/Select.js#L29

JedWatson:
@oakley808 I haven't updated the build folders after merging this PR, I tend to only do that when we're getting close to the next npm release because it reduces noise in the git history.

PR #1119, opened 2016-07-25
update npm badge

kentcdodds:
Using shields gives an svg which looks much better (especially on retina screens):

before:

screen shot 2016-07-25 at 2 07 45 pm

after:

screen shot 2016-07-25 at 2 09 42 pm


JedWatson:
:+1: thanks @kentcdodds!

PR #1124, opened 2016-07-28
Use preferred callback refs

gojohnnygo:
This PR implements preferred callback refs (vs string refs) per React [Refs documentation](https://facebook.github.io/react/docs/more-about-refs.html#the-ref-string-attribute).

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -88,7 +88,7 @@ const Async = React.createClass({
 		}
 	},
 	focus () {
-		this.refs.select.focus();
+		this.select.focus();
 	},
 	resetState () {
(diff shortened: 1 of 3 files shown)


gojohnnygo:
If you like at the individual files btwn Overalls [master build](https://coveralls.io/github/JedWatson/react-select) and [my branch](https://coveralls.io/builds/7197355), coverage is the same. Though, somehow my overall coverage dropped 1.8%? I'm happy to get coverage back in the 90s, but I'm not sure how.

Any suggestions?

JedWatson:
Thanks @gojohnnygo, this looks good!

PR #1137, opened 2016-08-02
adding a nvmrc file and adding coverage to eslintignore

uxtx:
I've taken to using an nvmrc file in my node projects to peg the node version, and thought it might be helpful here. This allows autoswitching to be much easier, especially when used in conjunction with this plugin https://github.com/creationix/nvm#zsh

I also adding coverage to exclusion in eslintignore - if you happened to run `npm run cover` prior to `npm test`, you'd get linting errors from the coverage report.


uxtx:
ugh. it looks like there's an issue with tests running in the latest version of react (15.3.0 was just released 4 days ago and now tests are failing on all branches. I just checked using `~15.2.0` and tests pass.

uxtx:
for the time being, i'm locking version to 15.2. Going to look at why simulate events are failing in a separate branch.

gwyneplaine:
thanks for this @uxtx, merged

PR #1143, opened 2016-08-03
refactor to use class syntax

pr1ntr:
I converted all the createClass stuff to use class syntax.

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component, PropTypes } from 'react';
 
 import Select from './Select';
 import stripDiacritics from './utils/stripDiacritics';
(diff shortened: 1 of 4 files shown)


pr1ntr:
I dunno... they fail on master anyway.

uxtx:
i see the same error - it's due to a breaking change in react 15.3.0, i believe.

uxtx:
i've got an issue open here: https://github.com/facebook/react/issues/7418

pr1ntr:
fixed this PR. reset all the built files

bvaughn:
Curious...what's the motivation behind this PR?

I'm also a fan of the class syntax for stateful components but in this case, "if it ain't broke..."

pr1ntr:
I have another PR with implementing React DND, still WIP. But it makes working on this lib much easier IMO.

bvaughn:
I see. I agree that the class syntax is easier to read, but I don't think the risk/value is there for a refactor just for stylistic preference.

FWIW though, `Async` is now class-based. Maybe gradually the others will be updated as well. 😄

In the meanwhile I'm going to close this PR.

PR #1146, opened 2016-08-05
fix value check in renderClear

hanwencheng:
I assume that it is common case that some one use options array with value started from 0:

Then we need to render clear button when the 0 value is chosen.
e.g. 

```
[
  {value:0,label:'Aachen [Aix-la-Chapelle]',state:'NW'},
  {value:1,label:'Aalen',state:'BW'},
]
```

### DIFF
--- a/lib/Select.js
+++ b/lib/Select.js
@@ -866,7 +866,7 @@ var Select = _react2['default'].createClass({
 	},
 
 	renderClear: function renderClear() {
-		if (!this.props.clearable || !this.props.value || this.props.multi && !this.props.value.length || this.props.disabled || this.props.isLoading) return;
+		if (!this.props.clearable || !this.props.value && this.props.value !== 0 || this.props.multi && !this.props.value.length || this.props.disabled || this.props.isLoading) return;
 		return _react2['default'].createElement(
 			'span',
 			{ className: 'Select-clear-zone', title: this.props.multi ? this.props.clearAllText : this.props.clearValueText,


JedWatson:
Thanks @hanwencheng!

JedWatson:
I just realised you've applied this change to `lib`, I've patched it in `src` as well

hanwencheng:
Thanks, @JedWatson  I just find that the lib file has not been changed.

kevinzwhuang:
@JedWatson @hanwencheng 

Just looking back at this PR, it doesn't seem to have changed anything nor is it necessary. The behavior without this works normally event with a value of `0`.

Let's take a look at what actually changed (this exists in master) -
```diff
-		!this.props.value
+		(!this.props.value || this.props.value === 0)
```
Source from master: https://github.com/JedWatson/react-select/blob/master/src/Select.js#L892

 means that we're checking the whole **object** to see if it is 0, such that the original example is checked like so: 
`{value:0,label:'Aachen [Aix-la-Chapelle]',state:'NW'} === 0`

The `this.props.value === 0` check does not achieve its original goal - in fact, this check already worked previously on its own. `!this.props.value` should never evaluate to false if such an object existed.

The original check for `!this.props.value` was sufficient on its own.

hanwencheng:
@kevinzwhuang Hi, it is not the [origin commit](https://github.com/JedWatson/react-select/pull/1146/commits/adc82bd495a493b5f99968953fe20dcbf819f713)

TrevorBurnham:
PR to actually fix this: https://github.com/JedWatson/react-select/pull/1395

PR #1147, opened 2016-08-06
Implement Sorting w/ React-dnd

pr1ntr:
This is my first pass implementing sorting into react-select. It's pretty much a spec PR since its probably not ready for primetime. I just wanted to share and get your thoughts.

This is already based on a previous PR to convert things to ES6 Classes


examples/dist/app.js
```diff
@@ -1503,21 +1510,35 @@ var cachedSetTimeout;
...
+        cachedClearTimeout = clearTimeout;
+    } catch (e) {
+        cachedClearTimeout = function () {
+            throw new Error('clearTimeout is not defined');
+        }
     }
-  }
 } ())
+function runTimeout(fun) {
+    if (cachedSetTimeout === setTimeout) {
```
wmertens: Why do you do this? If they are the same, they are the same, and if they are different, you want the cached one…

src/Value.js
```diff
@@ -24,9 +27,12 @@ class Value extends Component {
 			this.props.onClick(this.props.value, event);
 			return;
 		}
-		if (this.props.value.href) {
+
+		console.log(this.props.draggable);
+		if (this.props.value.href || this.props.draggable) {
```
wmertens: Would it not be better to give items drag handles? Maybe make it optional?

src/dnd/ValueSortItem.js
```diff
@@ -39,26 +39,12 @@ class ValueSortItem extends Component {
 		super(...arguments);
 	}
 
-	getPreview(url) {
```
wmertens: ? Is this not being used?

src/Async.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {Component} from 'react';
```
wmertens: You can put `PropTypes` in there as well…

src/Async.js
```diff
@@ -159,9 +149,36 @@ const Async = React.createClass({
...
-				/>
+			/>
 		);
 	}
-});
+}
+
+Async.propTypes = {
+	cache: React.PropTypes.any,                     // object to use to cache results, can be null to disable cache
+		ignoreAccents: React.PropTypes.bool,            // whether to strip diacritics when filtering (shared with Select)
```
wmertens: nit: double indentation

src/Async.js
```diff
@@ -41,75 +41,64 @@ const stringOrNode = React.PropTypes.oneOfType([
...
-			searchPromptText: 'Type to search',
-		};
-	},
-	getInitialState () {
-		return {
+
+class Async extends Component {
+
+	constructor() {
+		super(...arguments);
```
wmertens: Hmmm, according to https://facebook.github.io/react/docs/reusable-components.html#es6-classes, `constructor(props){ super(props) }` is sufficient…

wmertens:
Nice! Is it fully working? Anything left to do?

I wonder if the items should have drag handles or not. Might take up a lot of space on phones, but then you'd be able to click on draggable items (although I'm not sure what that is used for)

Generic PR comments
==
* don't leave console.log or commented code in your commits
* don't commit built files
* make sure eslint passes, that should have caught some issues

It would be great if you could split and squash your commits so you have only the ES6 commit and the DND commit after that. Makes reviewing much easier.

pr1ntr:
Thank @wmertens, I'll make the changes. I wasn't sure what the deal was with published files. they are being tracked in git. if they are generated i don't know what good that does.

Personally I don't think npm packages should have the need for compiled files in them since they are able to build on `npm install`. My suggestion is to create a bower package of compiled files since that the concat/<script> tag workflow usually lends itself better to bower projects. 

is the cover/ folder supposed to be tracked?

pr1ntr:
damn. I really messed this whole commit chain up with committing the build files. realized i rebased wrong

wmertens:
You still have that cachedTimeout and the build files in that commit?

Also, I think the eslint sorting rule was https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-prop-types.md

wmertens:
Also, npm install shouldn't need to install extra dependencies just to build. I think Jed added the build files in git so you can easily run from git commits…

pr1ntr:
I didn't really mess with the examples except the code regarding mousedown. I think more work is needed. this doesn't work on mobile due to using html5-dnd-backend. I'll have to convert to touch backend but then it requires dragpreview stuff which I need to add. 

The other feature is to detect whether to sort in front or behind where you hover. Lets just keep it open for now.

pr1ntr:
I don't know when i'll be able to work on this. I think this is a good base to get started but there is a lot of work to be done. Getting this to work on mobile with react-dnd is kind of a pain since the only backend that can work is yahoo's and some of my attempts created some problems with positioning relative to where the component is on the page. I would love for some1 else to contribute to this however.  This feature (or one like it) is necessary for Keystone to be taken seriously (sorry if that sounds harsh). Sorting relationships is a must for any CMS. (ie Join Tables). 


Regarding the built files, that is odd. I fixed all this in the in the ES6 conversion commit that you closed for being "unnecessary". However it was necessary for me to do this, and I don't know why modernizing the lib is a bad thing.

examples/dist/app.js
pr1ntr: I didn't do any of this. I left whatever logic was there intact. All I did is wrap the components with DND for multi-select

agirton:
Hi @pr1ntr thank you for the PR. But we are going a separate direction. Drag and drop should be handle in user land and with the great [react-sortable-hoc](https://github.com/clauderic/react-sortable-hoc), we only need to make one small change to make that possible.

elyobo:
@agirton could you provide some more details about how you would use this with `react-sortable-hoc` and the one small change that would make it possible (i.e. has it been done)?  I'd like to have a multi select where the selectable options are sortable.

PR #1155, opened 2016-08-12
react 15 compatibility fix

copas2:
uncontrolled to controlled hidden input state change fix


StevenLangbroek:
Can you add the following:

- What was the faulty behaviour?
- How does this PR fix that behaviour?

Is it possible to write a test for this?

JedWatson:
@copas I looked into this, and have implemented a different fix by ensuring `stringifyValue` always returns a string. Thanks for the PR.

PR #1167, opened 2016-08-22
Allow use of react-input-autosize 1.0+

saylynt:


JedWatson:
Thanks @saylynt - I've actually updated it to `^1.1.0` to match the package.json

Issue #1171, opened 2016-08-25
Add css class to the Select-control dynamically


dalkir:
Sometime you'd need to add a class to div.Select-control when you are validating the form like example, change the background to red when the select field is wrong. I'd like to help with this.

JedWatson:
Couldn't you do this by wrapping the `<Select>` component inside an `invalid` class and targeting it using css like:

```css
.invalid .Select-control { background: red; }
```

This component already has a crazy number of props, I'd prefer not to add more unless they're actually critical.

bvaughn:
👍  for Jed's suggested workaround. It's reasonable easy using the existing react-select Api.

Going to close this issue for now. 😄 Cheers!

PR #1173, opened 2016-08-26
Add prop to add extra className to select Control

dalkir:
This PR is in reponse to the issue [#1171](https://github.com/JedWatson/react-select/issues/1171).

It just add a prop for handling dynamically css classes in the select control.


gwyneplaine:
@dalkir closed as per the following response https://github.com/JedWatson/react-select/issues/1171#issuecomment-243421071

PR #1175, opened 2016-08-27
Test Result Dashboard added

Nishchit14:
VigoReport link up for public dashboard (https://vigoreport.io/p/4S6PdVz)
[vigoreport](https://vigoreport.io/p/4S6PdVz)

It will gives better ideas on passing & failing test cases.


JedWatson:
Thanks for the PR @Nishchit14 

We're now ensuring all tests pass with a pre-commit hook and before PRs get merged in Github so they should all always be green... with that in mind, I'm going to keep things simple and not add another service to our setup.

Issue #1170, opened 2016-08-24
Inconsistent multiselect onChange value when empty


jakezatecky:
The value parameter supplied to the `onChange` function when clearing out multiple selections is inconsistent. If you clear out selections using the right-most "x", the value `null` is returned. Otherwise, if you clear out the selections using the "x" on the individual items, eventually an empty string (`''`) is returned. Preferably, one or the other would be returned consistently.

``` jsx
const options = [
    { value: 'one', label: 'One' },
    { value: 'two', label: 'Two' }
];

const logChange = (val) => {
    console.log(typeof val, val);
};

ReactDOM.render(
  <Select
    multi
    joinValues
    options={options}
    onChange={logChange}
    simpleValue
    value="one"
  />,
  document.getElementById('example')
);
```

See test case below. Clear the selection using the right-most "x" and then clear the selection of the left-most blue "x". Two different values are returned.

https://plnkr.co/edit/UJ6oNaqRa5FOVHCGSani?p=preview

hbarudin:
I have this as a comment in my code, but it's helpful as reference here:
```
    // parameter supplied to onChange is:
    //    - an array of option objects if multi = true
    //    - a single option object if multi = false
    //    - Select.props.resetValue (`null` by default, but can
    //      be set as a prop) if Select was cleared via the "X" button
    //    - an empty array if last token was deleted
```

So many options!

mwilliamson-healx:
We have a wrapper around `Select` in our codebase, so I've just added something to the effect of the below to work around this:

```javascript
    var {resetValue, ...otherProps} = this.props;
    <ReactSelect
        {...otherProps}
        resetValue={resetValue === undefined && otherProps.multi ? [] : resetValue}
    />
```

JedWatson:
@jakezatecky thanks for the clear report and plunker test case!

I've just merged #1177, would you mind confirming that resolves this issue?

cheers.

jakezatecky:
I built the master branch and verified that the array/string value is now being consistently returned.

JedWatson:
Great! Thanks for letting me know @jakezatecky

PR #1177, opened 2016-08-28
 Default reset value to [] when multi=true

mwilliamson:
If `resetValue` is not set, then `onChange` is called with `[]` instead of `null` when `multi=true`, which should help with #1170.

This also fixes the other tests that clear by clicking `.Select-clear`.


JedWatson:
Nice fix, thanks @mwilliamson!

PR #1180, opened 2016-09-01
add default value/label for Select.Async

yonaichin:
add default value/label props support for Select.Async

in `src/Async.js`
```
render () {
        let { noResultsText, selectedValue, selectedLabel } = this.props; // add selectedValue , selectedLabel
        let { isLoading, options } = this.state;
        if (this.props.isLoading) isLoading = true;
        let placeholder = isLoading ? this.props.loadingPlaceholder : this.props.placeholder;
        if (isLoading) {
                noResultsText = this.props.searchingText;
        } else if (!options.length && this._lastInput.length < this.props.minimumInput) {
                noResultsText = this.props.searchPromptText;
        }
        // set default value for Select Component
        if (selectedValue !== '' && selectedLabel !== '') {
          options = [{
            value: selectedValue,
            label: selectedLabel
          }];
        }
        return (
                <Select
                        {...this.props}
                        ref="select"
                        isLoading={isLoading}
                        noResultsText={noResultsText}
                        onInputChange={this.loadOptions}
                        options={options}
                        placeholder={placeholder}
                        />
        );
}
```

with this workaround, you'll be able to set default value like this

```

/*
 * assuming the API returns something like this:
 *   const json = [
 *     { value: 'one', label: 'One' },
 *     { value: 'two', label: 'Two' }
 *   ]
 */

const getOptions = (input) => {
  return fetch(`/users/${input}.json`)
    .then((response) => {
      return response.json();
    }).then((json) => {
      return { options: json };
    });
}

<Select.Async
    name="form-field-name"
    value="one"
    loadOptions={getOptions}
    selectedLabel="default label display on Select.Async"
    selectedValue ="default_value"
/>
```


bvaughn:
As of PR #1226, `Select.Async` now supports default `value` and `options` properties. Closing this PR as it's already been fixed as the result of another refactor.

PR #1183, opened 2016-09-01
All but one test passing

mikehazell:
React TestUtils does not set `{ button: 0} ` when simulating events which react-select is checking for in lots of places.

I have added this where needed in the tests.

One failing test remains: Select with props onBlur should focus on the input when the menu is active.


mikehazell:
Related to this: https://github.com/facebook/react/issues/7418

> The code we have in React now is more correct than it used to be and so the recommendation is going to be that you fix the tests in react-select. Feel free to point them here if you need something to back up your claim (there or in any other projects suffering from this).

JedWatson:
Thanks @mikehazell!

PR #1186, opened 2016-09-02
Rename autosize "Input" references to "AutosizeInput"

Aeon:
Frequently getting this warning:

> warning.js?85a7:36Warning: Unknown prop `minWidth` on <input> tag. Remove this prop from the element. 

which means that somehow the plain html input is being used instead of AutosizeInput. Also, having two references that differ only in case is just begging for user error.


JedWatson:
I can't see how a plain html input would be being used in that code, but other than that this seems like a sensible change :)

Thanks @Aeon

Issue #435, opened 2015-09-09
Option Menu doesn't close when tapping / clicking away on iOS Safari


jwarning:
When you try to click out from the dropdown menu on an iPad for example the options menu doesn't automatically close. This issue may possibly explain the cause: https://github.com/facebook/react/issues/134. React 0.14 supposedly may fix this issue, but it may be worth noting for now anyway.

jooj123:
This is fixed in this PR: 
https://github.com/JedWatson/react-select/issues/435

PR #1189, opened 2016-09-03
Handle touch outside to dismiss menu

jooj123:
This is so that the menu closes after you touch outside of select on touch devices

Fixes: https://github.com/JedWatson/react-select/issues/435

Note: failed test has nothing to do with this PR


jooj123:
Just added a test for this as well

JedWatson:
Thanks @jooj123, looks good!

PR #1191, opened 2016-09-04
Added arrowRenderer prop and updated examples.

### LINKED ISSUE #1176: Allow using other icons not only arrow
I need to use a magnifier instead of arrow at the right of select-input.
So i hid through css select `Select-arrow` then spend time to align
bootstrap icon instead of arrow ( fa fa-search fa-lg` ).
It would be very useful if we can pass own class instead of  `Select-arrow`.

I want to start working on it, if it is possible.
Thank you

bvaughn:
Also added an example showing how to highlight matching search text using a custom optionRenderer.

Resolves #1176

Added via PR #1191.

```js
function arrowRenderer () {
  return (
    <span>+</span>
  );
}

function render () {
  return (
    <Select
      arrowRenderer={arrowRenderer}
      {...otherProps}
    />
  );
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,6 +316,7 @@ function onInputKeyDown(event) {
 	Property	|	Type		|	Default		|	Description
 :-----------------------|:--------------|:--------------|:--------------------------------
 	addLabelText	|	string	|	'Add "{label}"?'	|	text to display when `allowCreate` is true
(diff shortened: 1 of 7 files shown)


o1dnik:
For some reason the Select component doesn't see custom `arrowRenderer` prop and renders only default arrow. How do I debug this ?

o1dnik:
Sorry, my fault - had to upgrade to version `1.0.0-rc.2`

Issue #939, opened 2016-05-04
Old options are shown after selecting one in multi mode


mikecousins:
In beta12 after selecting an option the drop-down was cleared, in beta13 an old list of options stays dropped down. This doesn't play well with tab selection because if you now try to tab out it will select the first option and now you've selected two when you meant to just select one. Is there a way to set a flag to revert to the old behaviour or is this a bug?

mikecousins:
Is this not going to get fixed? It still exists in beta14 and it is super horrible.

melgar:
It was an intentional change,  #465 
I would also like the ability for multi selects to close after a value is selected.  I will work on a pull request to make it an option.

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you feel this issue / pull request is still relevant and you'd like us to review it, please leave a comment and we'll do our best to get back to you.

PR #1194, opened 2016-09-06
Allow closing multi select on select

melgar:
#939 Add property multiCloseOnSelect to close mutli select after the user selects a value.  The default value of false leaves the behavior of leaving the multi select open unchanged.


AlexRiina:
Is this the same as `autoblur`?

melgar:
If you set autoBlur to true on a multi select, after you make a selection focus is lost but the drop down remains open.  Perhaps on blur the select drop down should close.

JedWatson:
Thanks for the original implementation here. I've implemented a new prop called `closeOnSelect` that works for both single and multi-selects

PR #1195, opened 2016-09-07
react-select version screwed up

juliankrispel:
it was `1.0.0-rc.1` should be `1.0.0-rc`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-select",
-  "version": "1.0.0-rc.1",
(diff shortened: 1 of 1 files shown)


juliankrispel:
cc @JedWatson

bvaughn:
Why do you feel it should be one vs the other? This version numbering allows us to release multiple release candidates (eg rc.1, rc.2, etc) in a way that NPM will handle gracefully.

Regardless, the release has been published to NPM and so it cannot be changed now. 😄

Issue #1138, opened 2016-08-03
Setting initial/default label for Select.Async


yonaichin:
As title states, setting initial value for Select.Async is possible but failed to show the label due to the empty callback options from loadOptions.

Is there any workaround ?
I tried #861 but not quite fit the scenario.
Thanks in advance.

tingwei628:
Here is my workaround...
I marked on Select  and depend on the state "isFirstTime"
```js
  constructor() {
    this.state = {
      isFirstTime: true,
    }
  }
render() {
return(<div>
<Select.Async
  loadOptions={(input, callback) => {
    if (this.state.isFirstime) {
      callback(null, options: [ { ... } ]); // your default options
      this.setState({
         isFirstTime: false,
      });
      rerurn;
    }
    callback(null, options: [ {  ...  } ]); // other options...
 }}/></div>);
 }
```

yonaichin:
Thanks for the reply. 
I actually have this problem fixed with my workaround and send a PR.
For your refrences, https://github.com/JedWatson/react-select/pull/1180

dstuecken:
I actually created another PR that will fix this issue as well: The ability to pass an options array to Select.Async.

bvaughn:
FYI I am working on a refactor for `Async` that should address this issue. (See PR #1226)

bvaughn:
Resolved via PR #1226.

Look for an RC2 to come out (hopefully) sometime today with this change.

zhaoyi0113:
How does the initial value work? Is there an example for that? I tried `defaultOptions`, `defaultValues` but none of them work.

zhaoyi0113:
I see why it doesn't work for me. The `defaultValues` in my component is also an async request. How can I make it work with async data?

ngohieutp:
> I see why it doesn't work for me. The `defaultValues` in my component is also an async request. How can I make it work with async data?

As 2020, is there any solution?

PR #1199, opened 2016-09-08
Allowing to pass options to Select.Async

dstuecken:
With this PR you are able to pass default options to Select.Async. This allows you to set a default value without calling the Async promise.

See Issue #1138

Example:
```js
     const options = [{
        name: "My selected value",
        id: "1",
      }];

    const asyncOptions = {
      options,
      value: "1",
      // loadOptions: fetchFn,
    };

    return <Select.Async { ...asyncOptions } />;
```

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -55,6 +55,7 @@ const Async = React.createClass({
 		placeholder: stringOrNode,                      // field placeholder, displayed when there's no value (shared with Select)
 		searchPromptText: stringOrNode,       // label to prompt for search input
 		searchingText: React.PropTypes.string,          // message to display while options are loading
+		options: React.PropTypes.array,                 // initial options, usefull if you want to do a preselection and your minimumInput is > 0
 	},
 	getDefaultProps () {
 		return {
@@ -85,6 +86,14 @@ const Async = React.createClass({
 			this.setState({
 				cache: initCache(nextProps.cache),
 			});
+		}
+
+		if (nextProps.options !== this.props.options) {
+			if (nextProps.options && nextProps.options.length > 0) {
+				this.setState({
+					options: nextProps.options,
+				});
+			}
 		}
 	},
 	focus () {


bvaughn:
I'm concerned about the idea of supporting both `options` and `loadOptions` in `Async`. It seems ambiguous. (What's expected if both change?) And adding more forking logic seems undesirable.

Couldn't you achieve the same result by just returning default options the first time `loadOptions` was called? Something like...

```jsx
let defaultOptions = [/*...*/]
let initialized = false

function loadOptions () {
  if (!initialized) {
    initialized = true
    return Promise.resolve(defaultOptions)
  } else {
    return fetch(/*...*/)
  }
}
```

dstuecken:
You could, but i find it more confusing to have an options parameter that is omitted if you pass options to it and always do this kind of special handling you suggested. 

Imagine a search form with over ten Async selects that you want to preselect. I would want to implement this 10 lines of code for each of them and handle initialization state for all of them.
I'd rather just want to set the selected option and give it a name without any workarounds and hassle - and not to forget, without fetching a range of data in 10 separate requests.

bvaughn:
> You could, but i find it more confusing to have an options parameter that is omitted if you pass options to it and always do this kind of special handling you suggested.

That's totally fair. Your primary concern is maintainable application code and my primary concern (at least in this context) is maintainable library code. Those concerns won't always align. 😄

> Imagine a search form with over ten Async selects that you want to preselect. I wouldn't want to implement this 10 lines of code for each of them and handle initialization state for all of them.
I'd rather just want to set the selected option and give it a name without any workarounds and hassle - and not to forget, without fetching a range of data in 10 separate requests.

This sounds like a situation where you could create your own decorator function or HOC to DRY up application code. For example:

```jsx
import React, { Component, PropTypes } from 'react'
import { Async } from 'react-select'

class AsyncDecorator extends Component {
  static propTypes = {
    options: PropTypes.array.isRequired
  };

  loadOptionsWrapper (params) {
    const { loadOptions, options } = this.props

    if (!this.initialized) {
      this.initialized = true

      return Promise.resolve(options)
    } else {
      return loadOptions(params)
    }
  }

  render () {
    return (
      <Async
        {...this.props}
        loadOptions={::this.loadOptionsWrapper}
      />
    )
  }
}
```

bvaughn:
In general, I am wary of implementing more than 1 way to achieve the same result within library code because of the reasons I mentioned above:
* When the 2 routes conflict, which one wins? (In many cases it's ambiguous and can be a source of bugs.)
* Forks increase library maintenance load.

dstuecken:
Yea sure, you can do many other things with HOC that are actually implemented in the library code, too :) I am not disagreeing with your points, they are all legitimate. 
But don't you think it is more clear to have the same options parameter in either Select and Select.Async, especially new users will find that a hard nut to crack. I mean i was shocked at first time i realized options isn't doing anything in Select.Async :) And this use case is not that special. Who wants a select that is not able to preselect "out-of-the-box"? :)

bvaughn:
I think your concerns are legitimate as well. 😄

I'll give this a little time to sink in and come back to it (unless Jed or another contributor beats me to it).

bvaughn:
> Who wants a select that is not able to preselect "out-of-the-box"?

To be fair, this is only applicable to the special, _async_ variation of `Select`. 😁

I'm not convinced that the `Select.Async` component is the simplest way of achieving async behavior. Seems to me like using a stateful/managed component that listens to `onInputChange`- refreshes its own `options`- and then re-renders with an updated `options` list passed to `Select` seems more straight forward. But I didn't write `Select.Async` and I haven't spent much time thinking about it so I might be overlooking something.

For example, something like:
```jsx
class AsyncSelect extends Component {
  constructor (props, context) {
    super(props, context)

    this.state = {
      options: props.options || []
    }

    this._onInputChange = this._onInputChange.bind(this)
  }

  componentWillUpdate (nextProps, nextState) {
    if (this.props.options !== nextProps.options) {
      this.setState({
        options: nextProps.options
      })
    }
  }

  render () {
    return (
      <Select
        {...this.props}
        onInputChange={this._onInputChange}
        options={this.state.options}
      />
    );
  }

  _onInputChange (inputValue) {
    const promise = this.props.loadOptions(inputValue)

    this._promise = promise

    promise.then((options) => {
      // Ignore all but the most recent request
      if (promise === this._promise) {
        this.setState({ options })
      }
    })

    return inputValue
  }
}
```

bvaughn:
Okay. Have thought through this a bit more now and have a better understanding about how `Async` works and why.

I believe I'll try my hand at refactoring `Async`. I'll come back to this PR afterward, but I'm pretty sure I can address the use-case you mentioned. :)

bvaughn:
Closing in favor of PR #1226. I refactored `Async` and resolved a few other related issues along the way. Thanks again for the contribution and discussion!

dstuecken:
Hey Brian,

How did you resolve this issue btw?

bvaughn:
#1226

dstuecken:
Ok checked the code, passing options seems to work now with Select.Async 👍 Thanks

Issue #1204, opened 2016-09-10
Disable focusable when disabled


nhducit:
When  React-Select is disabled, it should not focusable and open option menu

https://plnkr.co/edit/26ZywmCu6AwVnHRBmTNR?p=preview

JedWatson:
Fixed by #1205

PR #1205, opened 2016-09-10
Do not focus and open menu when disabled

nhducit:
resolve issue https://github.com/JedWatson/react-select/issues/1204


nhducit:
resolve https://github.com/JedWatson/react-select/issues/1204

JedWatson:
Thanks @nhducit!

Issue #1202, opened 2016-09-09
Override "Create option "{label}"" in Creatable


rubencodes:
Is there an equivalent to the `addLabelText` option in the newest version of this package? It seems right now it defaults to "Create option "{label}"", but I would like to be able to customize this.

mpigsley:
@rubencodes It looks like you can override it with the `promptTextCreator` prop that has the signature `(label: String): String`.

rubencodes:
Wonderful! I'll try this. I didn't see it in the documentation; should I
make a pull request adding it?
-- 
Ruben Martinez Jr

mpigsley:
Yeah, I had to dive into the source to find it. Makes sense that it would be in the documentation 👍

JedWatson:
Closed by #1206

PR #1206, opened 2016-09-10
Document promptTextCreator property of Creatable.

rubencodes:
Documents ability to override default prompt label for the `Creatable` component.

https://github.com/JedWatson/react-select/issues/1202#issuecomment-246130181


JedWatson:
Nice, thanks @rubencodes!

Issue #1200, opened 2016-09-09
Combine Creatable and Async


traumatic:
Hello, I was wondering whether someone extended the Select.Async component with the behaviour of the Select.Creatable component? The use case is that a user may select predefined values from the given list (async) and in addition want to create new values.

Would be great if someone can refer to an existing project.

Thanks!

bvaughn:
I'm going to go with a child-function patter for this so that `Creatable`, `Async` and `VirtualizedSelect` (from [react-virtualized-select](https://github.com/bvaughn/react-virtualized-select)) can all be composed. So it will look something like this:

```js
return (
  <Creatable {...outerProps>
    {(childProps) => <Async {...childProps} />}
  </Creatable>
)
```

bvaughn:
PR #1207 adds partial support for this feature. You can compose an `Async` in a `Creatable` or vice versa. Unfortunately as the PR points out- the 2 don't seem to play nicely together. Additional time will need to be spent ironing out the details. (I'm traveling this week with limited time and Internet access and I haven't had the chance to dig into it yet.)

bvaughn:
PR #1210 (potentially) finishes this work out with the introduction of a new HOC, `AsyncCreatable`. It ties `Async` and `Creatable` together in a way that's simple to use externally. For example:

```jsx
import React from 'react';
import Select from 'react-select';

function render (props) {
  return (
    <Select.AsyncCreatable {...props} />
  );
}
```

traumatic:
Hi Brian, great work!! I'll close this issue and we'll continue discussion around pr [#1200](https://github.com/JedWatson/react-select/pull/1210).

bvaughn:
PR #1210 😉

If you have the chance to test that branch, please do. I'd welcome more sets of eyes on it.

PR #1207, opened 2016-09-11
Creatable and Async components can compose each other

bvaughn:
Added child-function support to `Async` and `Creatable` components so that they can compose each other (or future HOCs).

While this PR adds _support_ for composing- unfortunately the 2 don't seem to play nicely together as can be seen by applying the following diff and playing with the "_Github users_" example.
```diff
diff --git a/examples/src/components/GithubUsers.js b/examples/src/components/GithubUsers.js
index a206477..b778a21 100644
--- a/examples/src/components/GithubUsers.js
+++ b/examples/src/components/GithubUsers.js
@@ -44,7 +44,9 @@ const GithubUsers = React.createClass({
 		return (
 			<div className="section">
 				<h3 className="section-heading">{this.props.label}</h3>
-				<Select.Async multi={this.state.multi} value={this.state.value} onChange={this.onChange} onValueClick={this.gotoUser} valueKey="id" labelKey="login" loadOptions={this.getUsers} minimumInput={1} backspaceRemoves={false} />
+				<Select.Async multi={this.state.multi} value={this.state.value} onChange={this.onChange} onValueClick={this.gotoUser} valueKey="id" labelKey="login" loadOptions={this.getUsers} minimumInput={1} backspaceRemoves={false}>
+					{(props) => <Select.Creatable {...props} />}
+				</Select.Async>
 				<div className="checkbox-list">
 					<label className="checkbox">
 						<input type="radio" className="checkbox-control" checked={this.state.multi} onChange={this.switchToMulti}/>
```

Resolves issue #1200

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,9 +216,11 @@ function render (selectProps) {
 
 Property | Type | Description
 :---|:---|:---
+`children` | function | Child function responsible for creating the inner Select component. This component can be used to compose HOCs (eg Creatable and Async). Expected signature: `(props: Object): PropTypes.element` |
 `isOptionUnique` | function | Searches for any matching option within the set of options. This function prevents duplicate options from being created. By default this is a basic, case-sensitive comparison of label and value. Expected signature: `({ option: Object, options: Array, labelKey: string, valueKey: string }): boolean` |
 `isValidNewOption` | function | Determines if the current input text represents a valid option. By default any non-empty string will be considered valid. Expected signature: `({ label: string }): boolean` |
 `newOptionCreator` | function | Factory to create new option. Expected signature: `({ label: string, labelKey: string, valueKey: string }): Object` |
+`promptTextCreator` | function | Creates prompt/placeholder option text. Expected signature: `(filterText: string): string`
 `shouldKeyDownEventCreateNewOption` | function | Decides if a keyDown event (eg its `keyCode`) should result in the creation of a new option. ENTER, TAB and comma keys create new options by dfeault. Expected signature: `({ keyCode: number }): boolean` |
 
 ### Filtering options
--- a/src/Async.js
+++ b/src/Async.js
@@ -44,6 +44,7 @@ const stringOrNode = React.PropTypes.oneOfType([
 const Async = React.createClass({
 	propTypes: {
 		cache: React.PropTypes.any,                     // object to use to cache results, can be null to disable cache
+		children: React.PropTypes.func,									// Child function responsible for creating the inner Select component; (props: Object): PropTypes.element
 		ignoreAccents: React.PropTypes.bool,            // whether to strip diacritics when filtering (shared with Select)
 		ignoreCase: React.PropTypes.bool,               // whether to perform case-insensitive filtering (shared with Select)
 		isLoading: React.PropTypes.bool,                // overrides the isLoading state when set to true
@@ -53,7 +54,7 @@ const Async = React.createClass({
 		noResultsText: stringOrNode,                    // placeholder displayed when there are no matching search results (shared with Select)
 		onInputChange: React.PropTypes.func,            // onInputChange handler: function (inputValue) {}
 		placeholder: stringOrNode,                      // field placeholder, displayed when there's no value (shared with Select)
-		searchPromptText: stringOrNode,       // label to prompt for search input
+		searchPromptText: stringOrNode,   					    // label to prompt for search input
 		searchingText: React.PropTypes.string,          // message to display while options are loading
 	},
 	getDefaultProps () {
@@ -141,7 +142,11 @@ const Async = React.createClass({
 		}) : input;
 	},
 	render () {
-		let { noResultsText } = this.props;
+		let {
+			children = defaultChildren,
+			noResultsText,
+			...restProps
+		} = this.props;
 		let { isLoading, options } = this.state;
 		if (this.props.isLoading) isLoading = true;
 		let placeholder = isLoading ? this.props.loadingPlaceholder : this.props.placeholder;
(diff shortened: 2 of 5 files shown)


JedWatson:
Nice, looks good! Thanks @bvaughn

bvaughn:
Got a few more minutes to think about this this afternoon and I'm no longer sure that arbitrary HOC composition is doable (at least not easily) with 1.x.

For starters, there are several places in `Creatable` (and one in `Async`) where the components expect to a `this.select` ref to point to an instance of `Select`. That can be worked around by this (kind of nasty) patch:
``` diff
diff --git a/src/Async.js b/src/Async.js
index 6840fc3..15e26de 100644
--- a/src/Async.js
+++ b/src/Async.js
@@ -89,7 +89,7 @@ const Async = React.createClass({
 		}
 	},
 	focus () {
-		this.select.focus();
+		this.getInnerSelect().focus();
 	},
 	resetState () {
 		this._currentRequestId = -1;
@@ -98,6 +98,17 @@ const Async = React.createClass({
 			options: [],
 		});
 	},
+	getInnerSelect () {
+		let select = this.select
+
+		while (select) {
+			if (select instanceof Select) {
+				return select
+			} else {
+				select = select.select
+			}
+		}
+	},
 	getResponseHandler (input) {
 		let _requestId = this._currentRequestId = requestId++;
 		return (err, data) => {
diff --git a/src/Creatable.js b/src/Creatable.js
index bdaadf3..4d9053d 100644
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -62,9 +62,12 @@ const Creatable = React.createClass({
 
 	createNewOption () {
 		const { isValidNewOption, newOptionCreator, shouldKeyDownEventCreateNewOption } = this.props;
-		const { labelKey, options, valueKey } = this.select.props;
 
-		const inputValue = this.select.getInputValue();
+		const select = this.getInnerSelect()
+
+		const { labelKey, options, valueKey } = select.props;
+
+		const inputValue = select.getInputValue()
 
 		if (isValidNewOption({ label: inputValue })) {
 			const option = newOptionCreator({ label: inputValue, labelKey, valueKey });
@@ -74,7 +77,7 @@ const Creatable = React.createClass({
 			if (isOptionUnique) {
 				options.unshift(option);
 
-				this.select.selectValue(option);
+				select.selectValue(option);
 			}
 		}
 	},
@@ -84,13 +87,14 @@ const Creatable = React.createClass({
 
 		const filteredOptions = filterOptions(...params);
 
-		const inputValue = this.select
-			? this.select.getInputValue()
+		const select = this.getInnerSelect()
+		const inputValue = select
+			? select.getInputValue()
 			: '';
 
 		if (isValidNewOption({ label: inputValue })) {
 			const { newOptionCreator } = this.props;
-			const { labelKey, options, valueKey } = this.select.props;
+			const { labelKey, options, valueKey } = select.props;
 
 			const option = newOptionCreator({ label: inputValue, labelKey, valueKey });
 
@@ -110,18 +114,31 @@ const Creatable = React.createClass({
 		return filteredOptions;
 	},
 
+	getInnerSelect () {
+		let select = this.select
+
+		while (select) {
+			if (select instanceof Select) {
+				return select
+			} else {
+				select = select.select
+			}
+		}
+	},
+
 	isOptionUnique ({
 		option,
 		options
 	}) {
-		if (!this.select) {
+		const select = this.getInnerSelect()
+		if (!select) {
 			return false;
 		}
 
 		const { isOptionUnique } = this.props;
-		const { labelKey, valueKey } = this.select.props;
+		const { labelKey, valueKey } = select.props;
 
-		options = options || this.select.filterOptions();
+		options = options || select.filterOptions();
 
 		return isOptionUnique({
 			labelKey,
@@ -142,7 +159,8 @@ const Creatable = React.createClass({
 
 	onInputKeyDown (event) {
 		const { shouldKeyDownEventCreateNewOption } = this.props;
-		const focusedOption = this.select.getFocusedOption();
+		const select = this.getInnerSelect()
+		const focusedOption = select.getFocusedOption();
 
 		if (
 			focusedOption &&
@@ -160,7 +178,7 @@ const Creatable = React.createClass({
 		if (option === this._createPlaceholderOption) {
 			this.createNewOption();
 		} else {
-			this.select.selectValue(option);
+			select.selectValue(option);
 		}
 	},
```

After this change though, `Async`'s async options are still clobbering `Creatable`'s prepending of the special "_create option..._" placeholder (which means the option-creation doesn't really work).

I think we should consider reverting this PR for now.

bvaughn:
Scratch the awkward diff above. We could avoid it by nesting HOCs like so:
```js
<Select.Async multi={this.state.multi} value={this.state.value} onChange={this.onChange} onValueClick={this.gotoUser} valueKey="id" labelKey="login" loadOptions={this.getUsers} minimumInput={1} backspaceRemoves={false}>
	{(asyncProps) => (
		<Select.Creatable>
			{(creatableProps) => (
				<Select
					{...asyncProps}
					{...creatableProps}
					ref={(ref) => {
						asyncProps.ref(ref)
						creatableProps.ref(ref)
					}}
				/>
			)}
		</Select.Creatable>
	)}
</Select.Async>
```

Unfortunately the options-clobbering is still an issue. Hmm...

JedWatson:
Maybe we could do this better by creating a `AsyncCreatable` HOC that merges logic from `Async` and `Creatable`?

I haven't had a chance to look at what that would look like, not sure if we could do it neatly or if it would just lead to a bunch of code duplication. Depends on how cleanly the functions overlap, I think?

bvaughn:
Maybe. That feels like a sad path to start down though because it doesn't scale well. 😞

I'll poke at it some more. Maybe there's still a less extreme option.

bvaughn:
I think the fundamental problem is with [these lines](https://github.com/JedWatson/react-select/blob/f723189ed547a868cb96ad321d7918b6636fd872/src/Select.js#L430-L446). They are clearing the input value as a result of `Async.onInputChange` which is in turn breaking `Creatable` (since it thinks there's no input text).

Not yet sure why the state is being cleared.

bvaughn:
Progress!

bvaughn:
Let's continue discussion over on PR #1210 (demo and updated code)

PR #1221, opened 2016-09-15
Add "select all text" functionality Shift+Home|Del

piecyk:
Use-case:

I've written a long text in the Select input-box
I want to clear all text: I press Shift+Home to select all text, and then Delete to clear it.
^ It doesn't work.


JedWatson:
Good call @piecyk, thanks!

PR #1224, opened 2016-09-16
Fixed #546. Added support for boolean values.

### LINKED ISSUE #546: true/false values is not working
``` javascript
<Select value={true} options=[{value: true, label: "Yes"}, {value: false, label: "No"}] />
<Select value={false} options=[{value: true, label: "Yes"}, {value: false, label: "No"}] />
```

In both cases the option isn't selected. Using string like `"true"` and `"false"` make it works.

Isn't allowed to use boolean values?

Aaronius:
Fixes #546. This adds support for boolean values but suffers one issue that also exists with numeric values: If you go to the [live demo](http://jedwatson.github.io/react-select/) and in the Numeric Values example you select "Multi-select" you'll notice you can't select values. This is due to the select having the `simpleValue` prop and [this line](https://github.com/JedWatson/react-select/blob/17d76a4161096cb88fb04a6e9e6529580e67b305/src/Select.js#L582) which converts the numeric values into a string. The same thing happens with booleans and it fouls up the selection process for multi-select when `simpleValue` is enabled.

### DIFF
--- a/examples/src/app.js
+++ b/examples/src/app.js
@@ -11,6 +11,7 @@ import CustomComponents from './components/CustomComponents';
 import CustomRender from './components/CustomRender';
 import Multiselect from './components/Multiselect';
 import NumericSelect from './components/NumericSelect';
+import BooleanSelect from './components/BooleanSelect';
 import Virtualized from './components/Virtualized';
 import States from './components/States';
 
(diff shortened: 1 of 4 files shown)


JedWatson:
Nice work @Aaronius!

Sounds like we need to do some work around how simpleValue works with numbers and booleans, but in the meantime this change is good 👍 

Thanks!

Issue #1220, opened 2016-09-14
Using Asyc with minimunInput > 0, all selected items get removed from the text area when you clear the search inputValue.


yanmaoyuan:
### Thanks for using react-select!

If you are reporting an error please include a test case that demonstrates the issue you're reporting!
This is very helpful to maintainers in order to help us see the issue you're seeing.

Here is a Plunker you can fork that has react-select loaded and supports JSX syntax:
https://plnkr.co/edit/HTmtER9AMNcPoWhXV707?p=preview

You may also find the [online Babel tool](https://babeljs.io/repl/) quite helpful if you wish to use ES6/ES7 syntax not yet supported by the browser you are using.

yanmaoyuan:
Steps to produce:
1. Use Asyc fetching options
2. set minimum input to 2 (anything greater than 0 will do)
3. enter some input value into the search box, and select a few items from the dropdown
4. backspace to remove your search input value

expected: the selected items remains
actual: the selected items are removed as soon as inputValue.length < minimumInput

bvaughn:
FYI I am working on a refactor for `Async` that will invalidate this issue. (See PR #1226)

I plan to deprecate the `minimunInput` property. I believe it causes more complication than it's worth and is better handled by application logic. For example, you can enforce a similar behavior in a more customizable way in your application code like so:

```js
const MINIMUM_INPUT_LENGTH = 2

let previousLoadPromise

function loadOptions (inputValue) {
  if (inputValue && inputValue.length > MINIMUM_INPUT_LENGTH) {
    previousLoadPromise = fetch(...) // load options based on inputValue
  if (!previouslyLoadedOptions) {
    previousLoadPromise = fetch(...) // load initial / default options
  }

  // else return the previously-loaded options
  // this way, input text that's shorter than 2 chars won't reset options

  return previousLoadPromise
}
```

bvaughn:
Resolved via PR #1226.

Look for an RC2 to come out (hopefully) sometime today with this change.

Issue #1216, opened 2016-09-13
Implement autoLoad={false} in Select.Async 


f0urfingeredfish:
In the current documentation ```autoLoad={false}``` is listed as an option to stop the Async Select from loading when the component mounts. 
https://github.com/JedWatson/react-select#async-options

This isn't currently implemented. 
Look at:
https://github.com/JedWatson/react-select/blob/master/src/Async.js#L82

FIX:
Add autoLoad as an option with a default of true. 
Check autoLoad on componentDidMount to determine if it should fetch.

bvaughn:
FYI I am working on a refactor for `Async` that should address this issue. (See PR #1226)

bvaughn:
Resolved via PR #1226.

Look for an RC2 to come out (hopefully) sometime today with this change.

Issue #1181, opened 2016-09-01
Github users (Async with fetch.js)


tidjungs:
Github users (Async with fetch.js) witch single value crash
i cant type backspace when finish search

bvaughn:
@tidjungs Can you provide more information about how it "crashes" ? The demo does not crash for me (in single or multi mode).

I do notice that multi select mode doesn't delete an entry when you type backspace like a non-async `Select` does. I don't think that's intentional.

shashikanthgadgay:
@bvaughn Backspace doesn't work when you search for something for which results aren't available. It throws a undefined error at [Async.js](https://github.com/JedWatson/react-select/blob/master/src/Async.js#L155)

bvaughn:
FYI I am working on a refactor for `Async` that should address this issue. (See PR #1226)

bvaughn:
Resolved via PR #1226.

Look for an RC2 to come out (hopefully) sometime today with this change.

Issue #1166, opened 2016-08-22
How to prefill select value when it is Async with fetch.js?


ananddodia:
@JedWatson - I Need your help. please reply to below issue I am facing.
If any one else knows solution to this problem please help me.
I have made Async select with fetch.js. How can I make it prefilled when its in edit form scenario?
below is my code
`<Select.Async name={name} value={this.state.city} className={className} id={id} ref={ref} loadOptions={this.getCity} placeholder={placeholder} onChange={this.updateValue} autosize={false} clearable={false} minimumInput={1} backspaceRemoves={false} />`

value and label is already available from parent component.

bvaughn:
FYI I am working on a refactor for `Async` that will invalidate this issue. (See PR #1226)

Once it's done, setting a value (as you've shown above) will work for initializing `Async`.

bvaughn:
Resolved via PR #1226.

Look for an RC2 to come out (hopefully) sometime today with this change.

PR #1226, opened 2016-09-18
Refactored Async.

bvaughn:
This PR resolves the following issues #1138, #1141, #1166, #1181, #1216, and #1220.

The following props have been removed from `Async`:
* `minimumInput`: This property belongs in user-land. It conflicts with the `autoload` property and makes for complicated edge-cases (eg load initial options when component is mounted but don't load again until input value reaches a certain length).
* `searchingText`: This property was not being used and seemed redundant with `loadingPlaceholder`.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,6 +8,7 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 indent_style = tab
+translate_tabs_to_spaces = false
 
 [*.json]
 indent_style = space
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,7 +19,6 @@ Multiple values are now submitted in multiple form fields, which results in an a
 ## New Select.Async Component
 
 `loadingPlaceholder` prop
-`autoload` changed to `minimumInput` and now controls the minimum input to load options
 `cacheAsyncResults` -> `cache` (new external cache support) - defaults to true
 
 ## Fixes & Other Changes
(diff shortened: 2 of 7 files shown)


TheSharpieOne:
Would have been nice to factor in pagination, it is on the [list for v1.0.0](https://github.com/JedWatson/react-select/issues/568#issuecomment-153785740)

bvaughn:
Ah, I didn't write the list you're referring to @TheSharpieOne. (I actually wasn't even a contributor yet at that point.) In fact this is the first time I've even seen the list.

My 2 cents would be that pagination within a drop-down component seems weird. I don't know what Jed had/has in mind WRT that feature though. Maybe he has better ideas than me. 😄

Anyway this refactor fixed 6 issues (and I'm sure there's more to come on this component). Don't be greedy. 😆 😉

TheSharpieOne:
@bvaughn, no big deal. I'm creating a PR for it now. It wasn't too much work and didn't really require much effort thanks to this refactor you did (that and the `onMenuScrollToBottom` prop which I didn't know was already there for the `<Select>`).

jochenberger:
Since it is now the user's responsibility to implement `minimumInput`, I think there should be an example in the docs.

bvaughn:
`minimumInput` was just a string-length check with a short-circuit that returned an empty `options` array if the string was too short. Didn't seem necessary to add to the docs (since the bigger the docs are, the more difficult it typically is to find things). If you disagree, feel free to submit a PR with the docs addition for consideration. 😄

jochenberger:
In fact, I do disagree. For example, the fact that you need to call the callback also if you don't want any results to be shown, is something that should be pointed out IMO, otherwise people are going to have to figure it out the hard way (like I did).

bvaughn:
As I said, feel free to submit a PR with the docs addition for consideration. Docs PRs are the easiest to contribute _and_ the easiest to review (and so much more likely to be accepted).

Issue #810, opened 2016-03-02
positioning the component in containers that have `overflow: scroll;`


ksmth:
I recently ran into this issue when using `react-select` components in a modal. For mobile the contents of the modal are scrollable. In this particular case, the `Select` was at the bottom of the content. Activating the `Select` made the container overflow and scrollable.

Maybe the `container` notion of http://react-bootstrap.github.io/react-overlays/examples/ might make sense?
Alternatively, maybe make use of https://github.com/souporserious/react-tether?

geon:
Really important feature.

There are plenty of situations where you you don't want the drop-down to cause overflow. Actually, it should be the default behavior, since that's how the native select box work.

oluckyman:
one more use case when dropdown should flow above all elements and should not be restricted by container’s size
 <img width="635" alt="ss 2016-03-10 at 4 13 28 pm" src="https://cloud.githubusercontent.com/assets/642673/13670443/1334acda-e6db-11e5-828e-29ff357f273f.png">

khmelevskii:
+1

juan0087:
@oluckyman Having the same issue here! If I have other elements below the Select, I have no issues since there is plenty of space and I also reduce the height of the dropdown, but when the select is closer to the bottom of the page it pushed the main container.

Were you able to find a solution for this? 

![dropdown position](https://cloud.githubusercontent.com/assets/13401194/14901255/c642e8fe-0d51-11e6-8955-f16ed6d2dc2b.png)

maxmatthews:
Has anyone had any luck with this? I can't use a react-select in a modal without it causing scrolling issues. If I try and hijack `.Select-menu-outer` and set it to `position: fixed` I get weird scrolling issues. Someone suggested using react-tether, but it doesn't look like that's an easy implementation.

juan0087:
@maxmatthews you can try the following, it worked for me.

1 - Comment out the rule `/*top:100px*/` from the original css class `.Select-menu-outer`
2 - Add the following custom css to your stylesheet:
```
.menu-outer-top .Select-menu-outer {
	bottom: 35px!important;
	border-bottom-right-radius: 0px!important;
	border-bottom-left-radius: 0px!important;
	border-top-right-radius: 4px!important;
	border-top-left-radius: 4px!important;
}
```
3 - Add the class .menu-outer-top to the Select to manually change to position of the dropdown (See screenshot below)

                    <Select
                      name="form-field-Country"
                      value="one"
[...]

PR #1232, opened 2016-09-20
Fixed #284. Added support for direction control.

### LINKED ISSUE #284: Control opening direction up/down
Hi! 
Is there a way to control the opening direction programmatically?
kinda like the native select tag behavior, opens up or down (when there is not enough space for the dropdown to open up ).
thanks!

Kingwl:
Fixes #284 . This adds support for select direction of menu

### DIFF
--- a/examples/src/app.js
+++ b/examples/src/app.js
@@ -14,6 +14,7 @@ import NumericSelect from './components/NumericSelect';
 import BooleanSelect from './components/BooleanSelect';
 import Virtualized from './components/Virtualized';
 import States from './components/States';
+import SelectDirection from './components/SelectDirection';
 
 ReactDOM.render(
 	<div>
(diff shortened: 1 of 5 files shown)


rvignacio:
Any chance we get this merged?

JedWatson:
Thanks for the PR @Kingwl and sorry it's taken so long to review.

Makes sense as a prop, but I'm wondering if/how this should interact with auto-detecting space around the control instead if we were to implement that?

What are your thoughts?

adjohnson916:
@JedWatson It'd be great if this feature made it into a release soon.

burtyish:
> Makes sense as a prop, but I'm wondering if/how this should interact with auto-detecting space around the control instead if we were to implement that?

@JedWatson FYI, #810 mentions a solution which provides automatically determining the direction based on available space via usage of `tether.js`.

PR #1241, opened 2016-09-22
fix(async): create new cache object for each instance by default

### LINKED ISSUE #1236: Async shared cache (probably not wanted)
#1226 introduced a new ~~bug~~ undocumented feature 😄 , shared cache. This is probably not wanted as it causes issues. The [object which is used as a default prop](https://github.com/JedWatson/react-select/pull/1226/files#diff-149b30fde0306e461dc519d4853f1e50R26) is used by all of the async components by default. Because of this, and how objects and references work, everyone shares the same cache unless they specifically supply their own cache object. This of course causes issues when two async component which utilize two different API endpoints both have the same cached results for a given input. This can be seen in the updated docs that were part of the PR that is mentioned above. The "official" docs website isn't updated yet. Simply enter a value into the async field (it will fetch it), then remove the value (making the input '') this will trigger `loadOptions` for the cache of `''` which matches the cache for the contributors and it will load the contributors as options. (Note: that because the shape/properties are different between the two selects, the async will show blank options. Looking at the data, it is in fact the contributors).

So you don't have to deploy the new docs or run them locally, here is this: http://www.webpackbin.com/EkyrD3shW

TheSharpieOne:
fix #1236
Previously, the same object was being used by for async cache by
default when no cache was passed to the component. This same object
was used for every instance.
Now, an object is used to determine if the default is being used
and will create a new object for each instance in which no prop
was passed.

If the user wants to have shared cache, they will need to opt-in by providing the same object to the component via the `cache` prop.

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -21,9 +21,11 @@ const propTypes = {
 	]),
 };
 
+const defaultCache = {};
+
 const defaultProps = {
 	autoload: true,
-	cache: {},
+	cache: defaultCache,
 	children: defaultChildren,
 	ignoreAccents: true,
 	ignoreCase: true,
@@ -42,6 +44,10 @@ export default class Async extends Component {
 		};
 
 		this._onInputChange = this._onInputChange.bind(this);
+	}
+
+	componentWillMount () {
+		this.cache = this.props.cache === defaultCache ? {} : this.props.cache;
 	}
 
 	componentDidMount () {
@@ -64,7 +70,8 @@ export default class Async extends Component {
 	}
 
 	loadOptions (inputValue) {
-		const { cache, loadOptions } = this.props;
+		const { loadOptions } = this.props;
+		const cache = this.cache;
 
 		if (
 			cache &&
--- a/test/Async-test.js
+++ b/test/Async-test.js
@@ -117,6 +117,14 @@ describe('Async', () => {
 			typeSearchText('a');
 			return expect(loadOptions, 'was called times', 1);
 		});
+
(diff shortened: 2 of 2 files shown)


TheSharpieOne:
@bvaughn, this is related to a change you made recently with #1226. I'm not sure if the shared cache thing was intentional or not. If it was, you can just close this. If it was not intentional, please take a look and hopefully get this in before rc3 (or v1 if there isn't another rc planned).

somsub:
We are having the same issue and need this change above? Is there a PR created for it ? Is there a timeframe on this being available? 

Thank @TheSharpieOne for fixing this

Seldaek:
@somsub as a workaround for now you can just specify your own cache on every instance:

```
var yourCache = {};
<Select.Async [...]
            cache={yourCache}
/>
```

Just make sure you somehow define that variable smartly enough that you do get a cache still, but not so broadly that the cache is shared across instances.

bvaughn:
D'oh! Sorry. This fell through the cracks on my end.

bvaughn:
Merged with a minor change to move cache prop setting from `componentWillMount` to the constructor.

somsub:
Thanks for merging @bvaughn. When will this be released for consumption from npm ?

bvaughn:
Not sure. I'm at Connect Tech right now (presenting) so I won't be making a release any time over the next few days> Jed or someone else might.

cpruijsen:
@bvaughn is a release planned? Not having much luck building from sources and requiring locally to use the patched version.

PR #1243, opened 2016-09-22
Update aria-only class to have absolute positioning

JHilker:
When the component is restyled using flex, the width of the aria-only box causes the width of the `Select-multi-value-wrapper` to fluctuate when clicking into the input, and starting to type. Making this `position: absolute` removes it from the flow of flex, and thus solves this issue.

This seems to be a standard way to do this, while keeping it accessible. See:
http://a11yproject.com/posts/how-to-hide-content/
http://itstiredinhere.com/accessibility/

Error case:
image

### DIFF
--- a/less/control.less
+++ b/less/control.less
@@ -249,6 +249,7 @@
   display: inline-block;
 }
 .Select .Select-aria-only {
+  position: absolute;
   display: inline-block;
   height: 1px;
   width: 1px;


gwyneplaine:
@JHilker I can't seem to reproduce this on the examples page https://react-select.netlify.com 
can i confirm that you can still reproduce this issue?

Issue #1247, opened 2016-09-23
Creatable component doesn't allow input key down handling.


IvanLeonenko:
First of all, thanks for a nice component.

Creatable component has own handler of `onInputKeyDown`, which doesn't check if there's user's handler provided. So setting a handler for `onInputKeyDown` has no effect.

Doesn't work:

    <Creatable
        onInputKeyDown={this.inputKeyDownHandler}
    />

I think, Creatable should call user's handler as well, otherwise it breaks defined API.

Going to create pull request on that.

JedWatson:
Fixed by #1248

andfelzapata:
@JedWatson Very nice Component, thank you ! I have this issue with 1.0.0-rc.2. Is there any chance this update could be merged to it  ?

PR #1248, opened 2016-09-23
Fix for Creatable doesn't allow input key down handling. Issue #1247

IvanLeonenko:
Issue  https://github.com/JedWatson/react-select/issues/1247
Creatable component has own handler of onInputKeyDown, which doesn't check if there's user's handler provided. So setting a handler for onInputKeyDown, has no effect.

Doesn't work:
`<Creatable
    onInputKeyDown={this.inputKeyDownHandler}
/>`

Creatable should call user's handler as well, otherwise it breaks defined API.

### DIFF
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -157,7 +157,7 @@ const Creatable = React.createClass({
 	},
 
 	onInputKeyDown (event) {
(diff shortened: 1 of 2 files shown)


JedWatson:
:+1: Good fix @IvanLeonenko, thanks!

PR #1249, opened 2016-09-24
Adding back ref that was removed in rc2

jooj123:
Seems like ref to `Select` was removed from `Async` in rc2 - adding it back
I have needed it for my use case for a while.

I know this only caters for the 1 select - can modify to cater for all children if necessary.

I can kind of get around this by doing something like:
```
<Select.Async
    {...this.props}
 >
    {(asyncProps) => (
        <Select
            {...asyncProps}
            ref={(ref) => (this.selectRef = ref)}
        />
     )}
</Select.Async>
```

But I feel this seems now is a clunky way to setting a ref for the default `Select`


JedWatson:
This looks good to me @jooj123, thanks!

PR #1260, opened 2016-09-29
Allow react nodes to be passed as loadingPlaceholder

DanielHeath:
Everywhere we use loadingPlaceholder already supports `node` or `string`, 
so no other changes should be required if I'm reading correctly.


JedWatson:
:+1: Thanks @DanielHeath!

PR #1265, opened 2016-09-30
IE8 compatibility fixes

strayiker:
Use attachEvent/detachEvent when ie8.
Remove units from minWidth attr, because ie8 throws the exception "Invalid argument" when try to set width = 5pxpx


JedWatson:
Good fixes @strayiker, thanks!

Issue #1267, opened 2016-09-30
Options menu appear outside the viewport for some layouts


IvanLeonenko:
Default behaviour for when options menu goes beyond the visible area is scrolling down to show hidden part of options menu. Current version uses `window.scrollBy` to scroll and show menu.

However, it doesn't always work. Sometimes you need to scroll particular parent in DOM hierarchy to make it actually scroll. Like in my case - I have sidebar div and a page div inside the `flex row` parent, in this case window.scrollBy will not work but scrolling page div does the trick.

Here is the Plunker of the problem:
https://plnkr.co/edit/8FDDskB6qpoZWaNiaxvR?p=preview

Going to fix it by introducing new property of scrollable parent container id you want to scroll.

hshansen7:
What is the status on this PR?

IvanLeonenko:
@hshansen7 Although change is pretty straightforward #1268 still needs to be reviewed and merged. I think it could be merged, what do you think @JedWatson? Coverage decreased a bit, but it's not possible to get proper(rendered) size of the container, so I didn't add any tests.

aaronincincy:
I've solved this problem temporarily by adding the below code to an onOpen handler, but I would greatly prefer not having to rely on internal implementation to make this work.  Is there any plan to move this forward?  Can my code be of any use?

    onOpen = () => {
      window.requestAnimationFrame(() => {
        if (this.select.props.scrollMenuIntoView && this.select.menuContainer && this.select.menu) {
          const container = getScrollParent(this.select.menuContainer);
          const menuRect = this.select.menu.getBoundingClientRect();
          const containerRect = container.getBoundingClientRect();
          const scrollBottom = containerRect.top + containerRect.height;

          if (menuRect.bottom > scrollBottom) {
            container.scrollTop += menuRect.bottom - scrollBottom;
          }
        }
      })
    }

The definition of `getScrollParent` is something I found on StackOverflow I think:

    function getScrollParent(element, includeHidden) {
      var style = getComputedStyle(element);
      var excludeStaticParent = style.position === "absolute";
      var overflowRegex = includeHidden ? /(auto|scroll|hidden)/ : /(auto|scroll)/;

      if (style.position === "fixed") return document.body;
      for (var parent = element; (parent = parent.parentElement);) {
        style = getComputedStyle(parent);
        if (excludeStaticParent && style.position === "static") {
          continue;
        }
[...]

PR #1268, opened 2016-09-30
Custom scrollable container, fix for the #1267

IvanLeonenko:
Fixes issue when options menu is hidden due to parent is not scrolled.
Related issue #1267.

Couldn't write tests for that one, so coverage is going to drop a bit. If I render a parent div and a Select, then getting measurements like scrollTop or height of DOM element returns just undefined. May be test framework doesn't provide such data for some reason. Anyway if you have any ideas how to test that code, I will add unit tests for that.


JedWatson:
Hi @IvanLeonenko,

Thanks for the PR, and sorry it's taken so long to get to. I've reviewed this approach and the discussion on #1267, and I think that we're better off implementing automatic detection of the scroll parents rather than relying on IDs - so I'm going to close this, in favor of implementing that approach instead.

PR #1281, opened 2016-10-04
Documentation to backspaceToRemoveMessage Property

jtulk:
I don't see a better place to enable/disable this function, so adding some detail about how to clear this prompt if not wanted.


JedWatson:
Cheers @jtulk

PR #1289, opened 2016-10-09
Clear Async options after making a selection

DanielHeath:
Currently, when you make a selection from an async
dropdown, it keeps the list open but clears the input.

This looks inconsistent and doesn't match the common UX
for async search (looking up an item from a long list).

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -72,6 +72,10 @@ export default class Async extends Component {
 		});
 	}
 
+	clearOptions() {
+		this.setState({ options: [] });
+	}
+
 	loadOptions (inputValue) {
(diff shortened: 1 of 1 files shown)


JedWatson:
I think this is a good behavioural change, would you mind resolving the conflicts though? I think it's confused with another PR of yours I just merged.

DanielHeath:
Done :)

JedWatson:
Great! Thanks again @DanielHeath

JedWatson:
I've changed the way this works slightly; it's now implemented by default in Select itself; but you can leave the suggestions open by setting `closeOnSelect={false}` and `onSelectResetsInput={false}`

PR #1290, opened 2016-10-09
Tell git not to merge built files

DanielHeath:
Currently when merging / rebasing, git tries to merge compiled assets (and practically always gets it wrong).

This change tells git not to even try to merge assets - instead, the developer should run 'npm run build' when a merge conflict occurs and then commit the results.


JedWatson:
Great call, cheers

PR #1292, opened 2016-10-09
Only remove options if a loading placeholder is available

DanielHeath:
Currently, when `isLoading` is set to true, the options array
is set to an empty array and we show the loading placeholder instead.

Our app has an async service which responds in a couple of ms,
resulting in the autocomplete list flickering in and out of existence
as you type.

This change allows users to opt out of the 'clear options while loading'
behavior by not providing a loading placeholder.


JedWatson:
Nice change, that makes sense :+1:

PR #1293, opened 2016-10-09
Allow users to specify noResultsText

DanielHeath:
Currently if there are no search results the message displayed
is the same one used when a user hasn't started typing.

This change enables users of Async to provide different messaging when
a user has not started typing vs when no records matched a search.


JedWatson:
Great, thanks @DanielHeath

PR #1295, opened 2016-10-11
Updates onChange handler signature

lparvinsmith:
Hi! I noticed the signature of the `onChange` prop has changed from what is described in the docs, and it appears `joinValues` and `simpleValue` props are being deprecated.


JedWatson:
Thanks for the PR @lparvinsmith, I've checked and it's now correct for the 1.0 release.

Issue #1287, opened 2016-10-07
Firefox displays items in two rows


heaks:
Firefox shows items like this
![screen shot 2016-10-07 at 5 01 47 pm](https://cloud.githubusercontent.com/assets/17325467/19192772/05ef51bc-8cb0-11e6-983f-d8df79f07aef.png)
And Google Chrome shows them in one row
![screen shot 2016-10-07 at 5 01 58 pm](https://cloud.githubusercontent.com/assets/17325467/19192803/21f7c1a0-8cb0-11e6-90a2-edb18c835d24.png)
As we can see "Cookies and cream" has 2 rows. Can it be fixed somehow ?

MilllerTime:
Looks like this was already reported, with an explanation of why the wrapping happens: https://github.com/JedWatson/react-select/issues/1272

rafaelquintanilha:
Adding the following to `dist/react-select.css` did the trick for me.

```
span.Select-value-label {
  white-space: nowrap;
}
```

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you feel this issue / pull request is still relevant and you'd like us to review it, please leave a comment and we'll do our best to get back to you.

PR #1300, opened 2016-10-11
Fix issues #1272 and #1287

DanielHeath:


DanielHeath:
https://github.com/JedWatson/react-select/issues/1272

https://github.com/JedWatson/react-select/issues/1287

JedWatson:
This looks good, thanks @DanielHeath - I'm happy to merge it.

@jossmac can you make sure this change doesn't mess with anything you're aware of? I'm not sure why we'd avoid `float: left` but you may know something I don't.

less/control.less
```diff
@@ -255,6 +255,7 @@
   margin: -1px;
   clip: rect(0,0,0,0);
   overflow: hidden;
+  float: left;
```
jossmac: @JedWatson should be fine. `position: absolute` would also work

JedWatson:
Awesome. Thanks!

PR #1306, opened 2016-10-15
Hide create option after closing menu

### LINKED ISSUE #1256: Remove Creatable's new option on close
Hey all,

The new `Creatable` component works really well! Thank you.

On thing I am trying to figure out is how I can make sure that the "Create option XYZ" option gets removed when the select closes. Some users are confused that they are presented with that option, even though there is nothing typed in the field.

Looking at the list of props, I don't see an option for this. Am I missing something?

Thanks,

Alexis

andreme:
Always pass changes to the input value back to the Creatable.

There is one change in focusAdjacentOption where I wasn't sure whether handleInputValueChange should also be called?

This should fix #1256

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -395,13 +395,12 @@ const Select = React.createClass({
 			this.setState({
 				isOpen: false,
 				isPseudoFocused: this.state.isFocused && !this.props.multi,
-				inputValue: ''
+				inputValue: this.handleInputValueChange('')
 			});
 		}	else {
 			this.setState({
 				isOpen: false,
-				isPseudoFocused: this.state.isFocused && !this.props.multi,
-				inputValue: this.state.inputValue
+				isPseudoFocused: this.state.isFocused && !this.props.multi
 			});
 		}
 		this.hasScrolledToOption = false;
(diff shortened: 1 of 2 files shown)


JedWatson:
This looks good to me, @bvaughn let me know if you can see any issues merging this?

bvaughn:
Nice fix, @andreme. 😄 

At a quick glance, the change seems good to me @JedWatson. Cannot easily smoke-test it b'c I am without a laptop this weekend (in between jobs) but I think the approach looks solid.

andreme:
@JedWatson @bvaughn Any chance for a merge soon?

Praiodan:
Would be a useful fix for us, too.

xthiago:
@JedWatson @bvaughn Could you merge this PR?

g3ortega:
I'm needing this, any news?

agirton:
Tested this locally and works as expected. Thank you for your patience!

xthiago:
Thank you @agirton and @andreme! \o/

Will this feature be released in the next release candidate of version 1.0?

agirton:
@CShigaki did you build the examples or just clone and run install?

lorenzos:
I think this has fixed also #1575. On the examples I'm not able to reproduce #1575 anymore. If I checkout before this PR's merge with `git checkout d8a34ec`, I see #1575 again.

PR #1310, opened 2016-10-16
handle options with value 0 properly

### LINKED ISSUE #1284: Clearable icon doesn't show on value = 0
On using the following option, I'm not able to see clearable cross icon on selecting False but it shows in case I select True. 

``` javascript
const ENABLED_FLAG_OPTIONS = [
  { label: 'True', value: 1 },
  { label: 'False', value: 0 },
];

<Select
     value={enabledFlagFilter}
     options={ENABLED_FLAG_OPTIONS}
     onChange={selectedValue => this.setState({ enabledFlagFilter: selectedValue && selectedValue.value })}
/>
```

andreme:
fixes #1284

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -553,7 +553,7 @@ const Select = React.createClass({
 	 * @param	{Object}		nextProps	- optionally specify the nextProps so the returned array uses the latest configuration
 	 * @returns	{Array}	the value of the select represented in an array
 	 */
-	getValueArray (value, nextProps) {
+	getValueArray (value, nextProps = undefined) {
 		/** support optionally passing in the `nextProps` so `componentWillReceiveProps` updates will function as expected */
 		const props = typeof nextProps === 'object' ? nextProps : this.props;
 		if (props.multi) {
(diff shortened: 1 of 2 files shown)


msaglietto:
Nice it looks good to me, I agree that 0 it a possible value should not be consider as empty 👍

src/Select.js
```diff
@@ -553,7 +553,7 @@ const Select = React.createClass({
 	 * @param	{Object}		nextProps	- optionally specify the nextProps so the returned array uses the latest configuration
 	 * @returns	{Array}	the value of the select represented in an array
 	 */
-	getValueArray (value, nextProps) {
+	getValueArray (value, nextProps = undefined) {
```
andreme: Yes you're right, but my IDE was warning me that this function was sometimes called with only 1 argument. I thought making it explicit wouldn't hurt.

malucky: This param being optional should be indicated in the jsdoc http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values, which should also stop the IDE warning

grumd:
Is this going to be merged any time soon?

hillerstorm:
No merged PRs since mid-november? Would really like this one to land :/

bmv437:
It would be great to get this merged 👍

gwyneplaine:
Thanks for this @andreme, this has been fixed elsewhere in https://github.com/JedWatson/react-select/pull/1395

PR #1329, opened 2016-10-26
Add `keepInputCase` prop - Closes #1106

### LINKED ISSUE #1106: Async select changes the case of input when ignoreCase is false
Here is a plunkr that demonstrates the issue https://plnkr.co/edit/i947CSpI44PuVxx7wsT8?p=preview

Begin typing 'O', then 'n'   After the options are loaded, the input is altered to now contain 'on' instead of 'On'.   Though filtering is case insensitive, we would like to not alter the case of the string the user enters.

This is a change in behavior from 1.0.0-beta13 to 1.0.0-beta14.

Great job by the way, I'm loving react-select 👍

kauffecup:
Closes #1106 

Add `keepInputCase` prop that leaves the casing of what the user types in the input, but still allows the control to ignore the case when filtering the results.

For example, with `keepInputCase=true` and `ignoreCase=true`, we get the following:

image

but still get:

image

The one caveat is it requires the async method to be case insensitive. But this way it doesn't change the input to be lowercased which, when using with a creatable, can prevent the user from typing it in correctly.

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -81,16 +81,16 @@ export default class Async extends Component {
 		this.setState({ options: [] });
 	}
 
-	loadOptions (inputValue) {
+	loadOptions (inputValue, normalizedInputValue=inputValue) {
 		const { loadOptions } = this.props;
 		const cache = this._cache;
 
 		if (
 			cache &&
-			cache.hasOwnProperty(inputValue)
+			cache.hasOwnProperty(normalizedInputValue)
 		) {
 			this.setState({
-				options: cache[inputValue]
+				options: cache[normalizedInputValue]
 			});
 
 			return;
@@ -103,7 +103,7 @@ export default class Async extends Component {
 				const options = data && data.options || [];
 
 				if (cache) {
-					cache[inputValue] = options;
+					cache[normalizedInputValue] = options;
 				}
 
 				this.setState({
@@ -116,7 +116,7 @@ export default class Async extends Component {
 		// Ignore all but the most recent request
 		this._callback = callback;
 
-		const promise = loadOptions(inputValue, callback);
+		const promise = loadOptions(normalizedInputValue, callback);
 		if (promise) {
 			promise.then(
 				(data) => callback(null, data),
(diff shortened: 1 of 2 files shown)


TheSharpieOne:
This extra prop is really not need. There is an error in the logic which should be fixed. Currently, the [`_onInputChange`](https://github.com/JedWatson/react-select/blob/master/src/Async.js#L139-L154) method modifies the inputValue which it passes to both the `onChange` prop and the `loadOptions` method, [returning the result of the `loadOptions`](https://github.com/JedWatson/react-select/blob/master/src/Async.js#L154) method, which is just the manipulated [value that was passed it](https://github.com/JedWatson/react-select/blob/master/src/Async.js#L136) to (when no cache is found).

Really, IMHO, `_onInputChange` should pass the original inputValue to the user-defined `onChange` prop and return it (or nothing) to [the calling function](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L448-L452).
If user-defined `onChange` prop is not there or returns nothing, it will leave the input as-is (as the end-user typed). If it does return something, use the result of the user-defined `onChange` prop, allowing the user (in user-land) to manipulate the inputValue which is eventually displayed back to the user.

Either way, `ignoreCase` affecting the inputValue displayed back to the user is not a feature which should be preserved, but a defect which should be eliminated. After all, `ignoreCase` is supposed to be for searching, nothing more.

kauffecup:
@TheSharpieOne, is @bardediger's commit what you're looking for? if so i can cherry pick it onto my branch or close this PR and he can open a new one?

TheSharpieOne:
His commit looks good to me

kauffecup:
ok! this PR now contains only his commit!

caguthrie:
Can we please merge this PR?  Happy to help if there is anything we are waiting on here but I don't think there is

mikecx:
👍 can we get this merged in soon? Having to keep another fork up to date for this kinda sucks.

BossGrand:
Whats the status on this?

jperl:
We are running into this issue as well. Is there any way we can help get this merged?

davidemerritt:
I would love to see this merged in as well.

agirton:
I'm in agreement with @TheSharpieOne, the `_onInputChange` method should not be mutating the input, this should be done by the filter. I don't think this is the right way to go. I will open an issue and then someone can grab :smile:

zwright:
Here's another vote for getting this merged.

sponrad:
> the _onInputChange method should not be mutating the input

@agirton Agree!

We are also having this issue. Could be as simple as this change: https://github.com/sponrad/react-select/commit/029e56d6dd088a7b5c7dab09d40c117543e29426

gwyneplaine:
@kauffecup  thanks for this, a separate body of work has resolved the root issue behind this PR in the manner that @agirton suggested. You can see it here #1858

PR #1347, opened 2016-11-03
Delete key removes an item when there is no input

forum-is:
Thanks for creating react-select!

Would you consider merging our pull request? It extends the keydown handler to provide the same behaviour as the backspace key for the delete key, which is commonly used in Germany.


JedWatson:
@forum-is looks great, thanks! happy to add this.

PR #1353, opened 2016-11-04
Update onChange callback example

etherealite:
Currently, in this example the onChange callback param name of `val` implies that it it's type will be a string of `'one'` or `'two'`.  This is not the case, updated to reflect this fact.


JedWatson:
That's actually the case with the `simpleValue` prop... I'll update this in a way that calls it out better.

Thanks @etherealite

Issue #1348, opened 2016-11-03
Disable forced focus after remove via prop


forum-is:
We have wrapped a multi-value react-select in a component for creating and removing connections between certain business objects. Our wrapping component is connected to a redux store.

We noticed that whenever we remove an item from the values via the little **x**, Select.state.**isFocused** will be set **true** and stay that way, even if we focus other controls on our page, resulting in a persistent blue highlight around the Select box. This will only disappear when once more clicking the Select-input and clicking away again. 

Looking into the code we noticed that the **removeValue**-function of **Select** will always call this.focus() **after** setting the new values (which triggers **onChange**), meaning that a call of this.refs('select').**blurInput()** in onChange will stand no chance of overriding the behaviour as the control will be immediately focused again.

As a fix we would like to propose a prop **blurOnRemove**, which when **true** would prevent the call to this.focus() in removeValue and (if necessary) would instead call this.blurInput(). We would happily provide a respective pull request.

tosho-ait:
+1, It would be very useful to have the option to diable the focus() when using the clear "button". My issue with it is that it sometimes opens the options menu after clearing the selected value but in other cases closes it (if it is already open).  This happens when I use custom menuRenderer and valueRenderer functions ...

TrevorBurnham:
The focus-on-remove behavior has another problem: It can force scrolling, which is disconcerting. The user wouldn't expect scroll to occur as a result of clicking "Remove" on an item.

![scroll-on-mousedown](https://cloud.githubusercontent.com/assets/224895/21826541/e81e79a6-d755-11e6-92c4-602f1eb38532.gif)

psyanite:
I noticed this behaviour and was also trying to look at removing it, can I write custom CSS to remove this? Anyone know?

flexdinesh:
Closing this issue now. Chances are this might already be fixed or non-existent in v2. More [here](https://github.com/JedWatson/react-select/pull/1354#issuecomment-602155370).

PR #1354, opened 2016-11-04
Disable forced focus after remove via prop

forum-is:
The PR refers to issue [1348](https://github.com/JedWatson/react-select/issues/1348).

It adds an optional prop which allows to replace the forced focus after removing a value with a forced blur. This resolves issues with falsely persistent isFocused-states that may occur when wrapping a react-select. It was tested manually as automated focus tests would not work.


flexdinesh:
Hey @forum-is [v1 of react-select is no longer maintained](https://github.com/JedWatson/react-select/pull/3786). 

PR #1360, opened 2016-11-09
Expose select to async creatable

DanielHeath:
Both `Async` and `Creatable` keep a reference to the underlying `Select` (which can be accessed by consumers), but `AsyncCreatable` does not. 

This patch unifies this behavior.


gwyneplaine:
thanks for this @DanielHeath  I've taken out the main chunk of this PR and ported it over to this one https://github.com/JedWatson/react-select/pull/2084 

the other fixes you've made have already been resolved in master.

PR #1350, opened 2016-11-03
creatable-runs-onInputChange-prop: Fix Creatable to use onInputChange…

wmcmeans:
… function passed as prop

PR #1363, opened 2016-11-11
Fix onInputChange lost

leesiongchan:
This is to preserve the functionality of `onInputChange` when using `Creatable`.


JedWatson:
Good catch @leesiongchan, thanks!

TkDodo:
@JedWatson Why did you merge this instead of https://github.com/JedWatson/react-select/pull/1350 ? The fix is a bit better (typecheck) and there are tests for it

PR #1364, opened 2016-11-11
Add onNewOptionClick handler

leesiongchan:
This is useful when we want to customize the click action of the new option instead of creating the option immediately.


JedWatson:
Nice addition @leesiongchan, thanks!

Issue #1279, opened 2016-10-04
AsyncCreatable does not add value to options


lPadier:
In the following code, I mount an AsyncCreatable Select and when I add a value to it, `baz` for instance, it is not added to the dropdown, but I am still prompted to `Create option "baz"`
http://codepen.io/lPadier/pen/kkxQGp?editors=0011

Stas-Buzunko:
Hi, I have the same issue with Creatable Select, have solved the problem somehow?

midknightmare:
I am seeing the same behavior as well.  AsyncCreatable triggers the loadOptions function, but the returned values are never displayed in the dropdown.  Only the "Create option" section is shown.

unity:
I dug into this, and found the issue seems to be the way props `asyncProps` and `creatableProps` are merged in https://github.com/JedWatson/react-select/blob/master/src/AsyncCreatable.js#L14-L15

Having a value set to undefined results in a different merging strategy than having no value set. Javascript finesse it seems.

```js
const asyncProps = {
  options: [{value: 'foo', label:'bar'}, {value: 'baz', label:'bat'}]
}
const creatableProps = {
  options: undefined
}

const merged = {...asyncProps, ...creatableProps}
console.log(merged)
```

I'll start working on a fix.

PR #1374, opened 2016-11-18
fix AsyncCreatable options parsing

unity:
fixes https://github.com/JedWatson/react-select/issues/1279

### DIFF
--- a/src/AsyncCreatable.js
+++ b/src/AsyncCreatable.js
@@ -1,6 +1,15 @@
 import React from 'react';
 import Select from './Select';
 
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @unity!

bvaughn:
Curious why an inline `reduce` helper was added instead of just letting Babel do its thing with the pre-existing spread syntax? The helper added has the effect of reversing the order in which the props get applied (see the discussion on #1547). Not sure if this was intentional or accidental.

Issue #1053, opened 2016-06-28
Scrolling past the bottom of list scrolls the page.


MCBrandenburg:
When at the bottom of the list, continued downward scrolling scrolls the page.

mircea-dinoiu:
This issue is really annoying :(. There's no elegant way I could get around it.
For anyone else who is upset with this, I have a temporary workaround for it, but I'm not sure it works in all the scenarios:

```
const onWheel = (event) => {
        if (propagateWheelEvent === false) {
            const selectMenu = event.currentTarget.querySelector('.Select-menu');
            const height = selectMenu.clientHeight;
            const scrollHeight = selectMenu.scrollHeight;
            const deltaY = -event.deltaY;
            const scrollTop = selectMenu.scrollTop;

            if ((scrollTop === (scrollHeight - height) && deltaY < 0) || (scrollTop === 0 && deltaY > 0)) {
                event.preventDefault();
            }
        }
    };

return (
    <div onWheel={onWheel}>
        <Select {...props}/>
    </div>
);
```

Hope this helps somebody

jpriollaud:
> This issue is really annoying :(. There's no elegant way I could get around it.
> For anyone else who is upset with this, I have a temporary workaround for it, but I'm not sure it works in all the scenarios:
> 
> ```
> const onWheel = (event) => {
>             const selectMenu = event.currentTarget.querySelector('.Select-menu');
> 
>             if (selectMenu != null) {
>                 const height = selectMenu.clientHeight;
>                 const scrollHeight = selectMenu.scrollHeight;
>                 const deltaY = -event.deltaY;
>                 const scrollTop = selectMenu.scrollTop;
> 
>                 if ((scrollTop === (scrollHeight - height) && deltaY < 0) || (scrollTop === 0 && deltaY > 0)) {
>                     event.preventDefault();
>                 }
>             }
>     };
> 
> return (
>     <div onWheel={onWheel}>
>         <Select {...props}/>
>     </div>
> );
> ```
> 
> Hope this helps somebody

So close!  But where would one add { passive: false } to prevent Chrome from breaking it? :
_Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/6662647093133312_

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our efforts towards the current major version.

[...]

PR #1376, opened 2016-11-18
Add ability to prevent mouse wheel propagation on select menu

ghost:
This is an intention of fixing https://github.com/JedWatson/react-select/issues/1053

When you're scrolling the menu with the mouse wheel and it reaches the top/bottom line, the scrolling moves to the page.
I've added an option to prevent this kind of behaviour.

Let me know what you think


manatarms:
This would be awesome to have. Any updates on this?

bladey:
Hi there,

Thanks for this pull request.

In an effort to sustain the `react-select` project going forward, we're cleaning up and closing old and unmergeable pull requests.

However - if you feel this pull request is still relevant and you'd like us to review it, we'd appreciate if you updated this pull request use the latest version (`master`) and request us to review it.

PR #773, opened 2016-02-16
Fix backspace handling for non-multi select controls. Fixes #638

MrLeebo:
**Merging this PR will resolve the following issues: #638, #746, #1596, #1680, #1215**

REPRO STEPS:

1. `npm start`
2. `open http://localhost:8000`
3. Scroll down to "Contributors (Async)". This issue impacts components with the following props: `<Select multi={false} simpleValue={false} />`
4. Check "Single Value"
5. Change the dropdown selection
6. Press backspace

EXPECTED:

screen shot 2017-06-13 at 11 56 08 am

ACTUAL:

screen shot 2017-06-13 at 11 54 26 am

RESOLUTION:

Non-multi select controls return an empty array in the onChange event instead of the expected `null` value when you press backspace to clear the selection.

This is @slestang's implementation from #638 and I've added a unit test.

PR #1381, opened 2016-11-22
Fix for issue 1215

richardhills:
Only attempt to popValue on backspace being pressed if we're in multi mode.


mshwery:
lgtm – any reason to not get this merged in and released?

gwyneplaine:
@richardhills thanks for this, closing this now as https://github.com/JedWatson/react-select/pull/773/files resolves this while still allowing clearing on backspace for non multi selects (which is desired behaviour).

PR #1388, opened 2016-11-25
fix error on empty options

frescoraja:
i think it's feasible to assume one may want to start out with an empty options list. at the very least, i don't think it should throw an error if in that scenario.


gwyneplaine:
thanks for this PR @frescoraja, I am as of yet unable to replicate the error case. Could you provide us with more information in regards to how to best replicate this? Please also provide the version of react-select to replicate this on (it may very well be a problem solved in a later release)

closed for now

PR #1391, opened 2016-11-26
Added scrollToIndex prop

akopichin:
I made this change for these case such year select from long list of years, to scroll position to the current year in the list, but dont select this option. It's not allowed by law in some cases to prefill the forms with values. This option work only if no selected for now.

In the first example if you remove value you can see it scrolled to 4th options. (I think no need separate example for it, or.. ?)


examples/src/components/States.js
```diff
@@ -51,7 +51,7 @@ var StatesField = React.createClass({
 		return (
 			<div className="section">
 				<h3 className="section-heading">{this.props.label}</h3>
-				<Select ref="stateSelect" autofocus options={options} simpleValue clearable={this.state.clearable} name="selected-state" disabled={this.state.disabled} value={this.state.selectValue} onChange={this.updateValue} searchable={this.state.searchable} />
+				<Select ref="stateSelect" scrollToIndex={4} autofocus options={options} simpleValue clearable={this.state.clearable} name="selected-state" disabled={this.state.disabled} value={this.state.selectValue} onChange={this.updateValue} searchable={this.state.searchable} />
```
freask: чоОООО

JedWatson:
This is a good idea but I think that the implementation will "lock" the scroll position on re-renders, because it's always applying the scroll in `componentDidUpdate`

We'd need a more robust implementation that only performs the scroll when the prop changes, or when the menu opens for the first time with the prop set.

You're also setting `scrollToItem` in state, but not using it.

I'm going to close this because it's got conflicts and the issues above, but if you're still interested in implementing it please reopen with the updates.

Thanks!

PR #1395, opened 2016-11-29
Fix #1394

### LINKED ISSUE #1394: Unable to clear selection if value is boolean false
I use a Select with 3 possible values: `true`, `false` and `null` (`null` being the default)
The select is clearable.
When I select the `true` option, no problem to clear the select.
However, if I select the `false` option, the clear button is not displayed.

See https://plnkr.co/edit/rZPhbvO3ajW2MxIoJYmm?p=preview

sgaestel:
This way, the X clear button won't be displayed only for `null` or `undefined` values, allowing to use `0` or `false` as clearable values in select options.


src/Select.js
```diff
@@ -889,7 +889,7 @@ const Select = React.createClass({
 	},
 
 	renderClear () {
-		if (!this.props.clearable || (!this.props.value || this.props.value === 0) || (this.props.multi && !this.props.value.length) || this.props.disabled || this.props.isLoading) return;
+		if (!this.props.clearable || this.props.value == null || this.props.value === 0 || this.props.multi && !this.props.value.length || this.props.disabled || this.props.isLoading) return;
```
atatakobry: and what about `this.props.value === 0`?
sometimes we have to use zero value and clear button is not displayed (

sgaestel: Yes, I guess we could remove this condition, and not display the clear button only when `this.props.value` is `null` or `undefined`

sgaestel: src and tests updated :wink:

sgaestel:
@JedWatson Any chance for this PR to be merged soon ?

src/Select.js
```diff
@@ -901,7 +901,8 @@ const Select = React.createClass({
 	},
 
 	renderClear () {
-		if (!this.props.clearable || (!this.props.value || this.props.value === 0) || (this.props.multi && !this.props.value.length) || this.props.disabled || this.props.isLoading) return;
+
+		if (!this.props.clearable || this.props.value == null || this.props.multi && !this.props.value.length || this.props.disabled || this.props.isLoading) return;
```
JedWatson: I'm not keen on the loose equality operator - I know how it works, but given the number of people who use / contribute to this project, without an explicit comment it could be misinterpreted.

How about we replace this with a more explicit `value === undefined || value === null` so that the intended behaviour is explicit?

test/Select-test.js
```diff
@@ -691,6 +691,11 @@ describe('Select', () => {
 			expect(onChange, 'was called with', 0);
 		});
 
+        it('displays the X button for 0 value', () => {
```
JedWatson: Thanks for also adding tests for this 😁

src/Select.js
sgaestel: :white_check_mark: Done
Replaced as proposed by `this.props.value === undefined || this.props.value === null`

test/Select-test.js
sgaestel: You're welcome :wink:

src/Select.js
```diff
@@ -901,7 +901,8 @@ const Select = React.createClass({
 	},
 
 	renderClear () {
-		if (!this.props.clearable || (!this.props.value || this.props.value === 0) || (this.props.multi && !this.props.value.length) || this.props.disabled || this.props.isLoading) return;
+
+		if (!this.props.clearable || this.props.value === undefined || this.props.value === null || this.props.multi && !this.props.value.length || this.props.disabled || this.props.isLoading) return;
```
TrevorBurnham: One small tweak: You could write `this.props.value == null` instead of `this.props.value === undefined || this.props.value === null`.

```
> undefined == null
true
```

TrevorBurnham:
I'd love to see this merged. You can see this bug right on http://jedwatson.github.io/react-select/:

falsey-clear

agirton:
Hi @sgaestel this looks good to me, thanks again for your patience and the tests!

PR #1406, opened 2016-12-01
try to focus the next available option after a selection.

nraynaud:
When the menu is displayed into a "scrollable" div, we like to have the scroll following the focused element so that keyboard navigation shows the current element. 

The problem is that when selecting an option in a multi-option component, the current system focuses the first option of the list, forcing users to re-navigate to their last position.

This PR focuses the option immediately adjacent to the last selected one, to avoid scrolling issues.


jfernandogt:
+1

r1g0:
+1 Wanting this too

JedWatson:
Thanks @nraynaud

PR #1408, opened 2016-12-05
Provide `isOpen` to arrowRenderer

khankuan:
It is useful for arrowRenderer to determine if menu is opened for rendering purposes.


JedWatson:
Good addition, thanks @khankuan!

Issue #1349, opened 2016-11-03
Creatable does not invoke onInputChange prop


wmcmeans:
The Creatable class does not run `onInputChange` when designated as a prop.

stinoga:
I noticed something similar with `AsyncCreatable`. Since `onInputChange` is called as a prop directly in it's source, you can't overwrite it when calling a parent `Select`. I also noticed that `onInputChange` isn't invoked when blurring a select with an input value, even when `onBlurResetsInput` is set to true. Seems like it should, because refocusing the input then gives you a menu with the last input you typed.

TheTedAdams:
I was trying to use this event to catch when a user pastes in a comma separated string of values. Pretty unfortunate that it doesn't work as documented...

mhubenthal:
I am seeing this issue as well for the `Creatable` component. The bug lies on these lines in that component:
```javascript
onInputChange (input) {
	const { onInputChange } = this.props;

	if (onInputChange) {
		onInputChange(input);
	}

	// This value may be needed in between Select mounts (when this.select is null)
	this.inputValue = input;
}
```
Even though `onInputChange()` is called, the value is not *returned*. When I just use the `Select` component I am able to apply a masking function, whereas when using the `Creatable` component, that masking function value is never returned from the `Creatable` `onInputChange` call to the `Select` component. The function passed in as a prop *is* called, but the value is never returned up the call stack.

The correct code would look like this:
```javascript
onInputChange (input) {
	const { onInputChange } = this.props;

	// This value may be needed in between Select mounts (when this.select is null)
	this.inputValue = input;

	if (onInputChange) {
		return onInputChange(input);
	}
}
```
cc: @JedWatson

cscholze:
Do we have any update on this? Creatable is still unable to modify text with the `onInputChange` prop.

JedWatson:
I believe this can be closed now because #2057 has been merged, could everyone please confirm that it solves the problem correctly for their use-case? and continue discussion on #2056 if not.

Thanks!

Issue #1283, opened 2016-10-06
Select.Async does not call onInputChange


bmamouri:
I am trying to implement this [demo](https://github.com/JedWatson/react-select/blob/master/examples/src/components/CustomRender.js) using Select.Async.

The event `onInputChange` does not call in the `Select.Async` component, but it works fine in the `Select` component itself.

kr0:
This is related to [https://github.com/JedWatson/react-select/pull/1231](this) . There is a simple fix it seems. This change hasn't been pulled.

kr0:
I added the changes manually for right now and it works. In the Async component, there is no call to this.props.onInputChange .

kr0:
Plot thickens.. When I import using var react-select = require("react-select") I get definitions from the react-select/lib folder. These don't reflect the src folder. This is a problem because it's in the src folder that the onInputChange issue was resolved.

kr0:
I opted to use 1.0.0-rc.1 instead of rc.2

aminmc:
This seems to be still an issue in rc.2 downgrading to rc.1 works.

mhubenthal:
I am seeing this issue as well for the `Creatable` component. The bug lies on these lines in that component:
```javascript
onInputChange (input) {
	const { onInputChange } = this.props;

	if (onInputChange) {
		onInputChange(input);
	}

	// This value may be needed in between Select mounts (when this.select is null)
	this.inputValue = input;
}
```
Even though `onInputChange()` is called, the value is not *returned*. When I just use the `Select` component I am able to apply a masking function, whereas when using the `Creatable` component, that masking function value is never returned from the `Creatable` `onInputChange` call to the `Select` component. The function passed in as a prop *is* called, but the value is never returned up the call stack.

The correct code would look like this:
```javascript
onInputChange (input) {
	const { onInputChange } = this.props;

	// This value may be needed in between Select mounts (when this.select is null)
	this.inputValue = input;

	if (onInputChange) {
		return onInputChange(input);
	}
}
```
cc: @JedWatson

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you feel this issue / pull request is still relevant and you'd like us to review it, please leave a comment and we'll do our best to get back to you.

PR #1411, opened 2016-12-05
fix onInputChange

svsool:
fix for #1363, value returned from onInputChange will used as nextState in Select.js see https://github.com/JedWatson/react-select/blob/master/src/Select.js#L450, otherwise value normalisation not working

### DIFF
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -170,11 +170,13 @@ const Creatable = React.createClass({
 		const { onInputChange } = this.props;
 
 		if (onInputChange) {
-			onInputChange(input);
+			input = onInputChange(input);
 		}
 
 		// This value may be needed in between Select mounts (when this.select is null)
 		this.inputValue = input;
+
+		return input;
 	},
 
 	onInputKeyDown (event) {


mhubenthal:
@JedWatson can we get this in? `onInputChange` does not work correctly for `Creatable` or `AsyncCreatable` right now.

Even though `onInputChange()` is called, the value is not *returned*. When I just use the `Select` component I am able to apply a masking function, whereas when using the `Creatable` component, that masking function value is never returned from the `Creatable` `onInputChange` call to the `Select` component. The function passed in as a prop *is* called, but the value is never returned up the call stack.

mhubenthal:
@JedWatson I believe all of these issues are related to this bug: https://github.com/JedWatson/react-select/issues/1506, https://github.com/JedWatson/react-select/issues/1349, https://github.com/JedWatson/react-select/issues/1283,

mhubenthal:
@svsool can you add a test that would catch this bug? It might be helpful to demonstrate how it is currently broken and potentially get the fix in sooner...

cc: @JedWatson

mhubenthal:
@JedWatson looks like #1363 was merged, and either created this bug, or did not fully solve an existing bug.

JedWatson:
I believe this can be closed now because #2057 has been merged, could everyone please confirm that it solves the problem correctly for their use-case? and continue discussion on #2056 if not.

Thanks!

svsool:
Yes, this problem solved in #2057, thanks

PR #1414, opened 2016-12-06
Focusing options using the arrow key calls a handler function prop

### LINKED ISSUE #1413: onFocus prop does not work when focusing with arrow keys
[`focusAdjacentOption`](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L703-L760) handles focusing options using the arrow keys in the UI. However, there doesn't seem to be an easy way for a parent component to get the focused option when it is focused using arrow keys (since no callbacks/props fire in that function).

mikestone14:
This PR allows passing a function prop called `onArrowKeyFocus` which is called when a new option is focused using the up and down arrow keys.

closes #1413

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -80,6 +80,7 @@ const Select = React.createClass({
 		multi: React.PropTypes.bool,                // multi-value input
 		name: React.PropTypes.string,               // generates a hidden <input /> tag with this field name for html forms
 		noResultsText: stringOrNode,                // placeholder displayed when there are no matching search results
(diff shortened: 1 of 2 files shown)


julianljk:
Is this PR being merged?

JedWatson:
Thanks for the PR @mikestone14, can you (or anyone) clarify the use-case for this? the information would get out of sync if the mouse was subsequently used to "focus" another option.

I'm going to close this for lack of clarity around the requirement and because of conflicts, but if you're still interested in contributing it please do let me know more of the rationale and we can look at adding support for this feature.

PR #1416, opened 2016-12-07
Allow user-create option to locate at the bottom of the list.

blutarche:
In my use-case I need the user-create option to be at the very bottom of the list (so the user can just press enter to select the first element on the list, kinda like autocomplete I guess).
Along with `isCreateAtBottom` prop to determine this.


Stenerson:
This has gotten a little stale. Recommend closing, in favor of #1436

gwyneplaine:
Thanks for this @blutarche, closing this in favor of https://github.com/JedWatson/react-select/pull/1436 which will be pulled in after 1.0 has been released.

Issue #1419, opened 2016-12-08
No results found click handler or special component


tnrich:
Hey there,

Thank you for this most excellent tool.

I'm trying to implement something like what you can see in the following image: 
![image](https://cloud.githubusercontent.com/assets/2730609/20995521/804e2e3e-bcac-11e6-83fe-440fdffd9fd0.png)

Ideally I'd even be able to have a custom component in there so I could add a little + button or something of the sort. 

I didn't see anything quite like this yet in the docs. Does this exist already? And if not, can this use case be added? It seems like it should be fairly simple given that "noResultsText" is already an accepted prop.

Thanks!

tnrich:
Well, whaddya know! It looks like this already works. So maybe just tweaking the docs is all that's needed. 

Here's a pic of my custom component for proof:
![image](https://cloud.githubusercontent.com/assets/2730609/20995672/6b4c2fbc-bcad-11e6-8b69-941595f23652.png)

I'll put a PR in to update the docs. Thanks!

JedWatson:
> Thank you for this most excellent tool.

You're welcome 😄

PR #1420, opened 2016-12-08
tiny update to readme fixing noResultsText info

tnrich:
Connects #1419
readme now reflect fact that noResultsText can be a string or a react component


JedWatson:
Thanks @tnrich!

Closes #1419

PR #1421, opened 2016-12-08
Allow to set default input value through props

takashi:
In `Select.*` components. they has configuration that named `inputProps` . 

We can set extra attributes to `input` element inside of `Select.*` components.

But atm, we cannot set `value` of `input` element because of `Object.assign` overrides the second argument value with third argument value. [ref](https://github.com/JedWatson/react-select/blob/ef3a4687bc67684809c3a2648143a1c325ee987e/src/Select.js#L838-L854)

I would like to set the default value of input like the case that the mount component will change with the state value like 

```js
const AsyncComponent = this.state.isCreatable
      ? Select.AsyncCreatable
      : Select.Async
```

In this case, we cannot hold the input value that I type before the component change.

This PR will fix the problem above.


JedWatson:
Thanks for the PR @takashi 

For me this confuses the boundary between what is "controlled" and "uncontrolled" with regard to state and props... would need some more thought because our handling of the input and updating it is quite nuanced. We'll probably open this up more in a future version but I'm going to close this for now.

PR #1425, opened 2016-12-11
Add menuScroll prop to allow customizing menu scroll behaviour

serkanozer:
Allow customizing scroll behaviour of menu node by letting users specify a `menuScroller` handler with  a `menuNode` parameter.  

Current default is scrolling to top of last focused (selected) option. This perfectly fits for single select case and 99% of the time customizations won't be needed. Still, I think scrolling behaviour should  be overridable and it will especially come in handy when `multi` is `true`


JedWatson:
Thanks for the PR @serkanozer 

I'm trying to limit our props a bit and don't really understand the value of being able to customise this - when would you actually want to change how the scroll behaves, and in which way?

It's a nice, neat implementation but I'm worried about adding yet another thing to test and maintain without a clear purpose.

I also believe you'd need to pass the `focusedOptionNode` into the `menuScroller` handler, otherwise a custom implementation wouldn't have access to the `this.focused` ref..?

Going to close this for now but if you've got a specific reason to want this feel free to reopen and we can discuss it further.

Issue #1399, opened 2016-11-30
Accessibility Enhancements - Aria tags, Space/Enter keys


strohr1:
The select box currently does not allow Enter/Space keys to make a selection from the drop down. This is to enable users that are navigating with keyboard to make selections.
The select box should only use supported Aria attributes. It currently fails Chrome extension's Accessibility Audit.

We would like to contribute a fix for these issues, which will allow our use of this component to be ADA compliant.

Note: We have made these changes internally, but would like to contribute back to this library.

Please let us know if you want more details. Thanks.

postrad:
This is exactly the same issue I am currently dealing with, and it is precisely why I was looking through issues. This would be a big improvement for the overall Accessibility compliance of this component.

adrexia:
Yeah, us too.

egoens:
@strohr1 Were these changes made to a public repo? I'd love to see what you've forked and possibly use in the interim. Assuming that the a11y branch gets merged at some point.

Thanks for all the awesome guys. This is great.

strohr1:
@egoens there is a commit against the PR referenced above.
Here is the link for it anyways: https://github.com/JedWatson/react-select/pull/1428

egoens:
🙌  @strohr1 Perfect. Wasn't sure if there was anything additional that you might have added, that wasn't a part of this PR. Thanks again for the awesome work.

maureenwaggl:
Will this be supported in V2? Specifically, I'm using a custom DropdownIndicator component, and I'm getting an a11y error for the [button-role-space](https://github.com/reactjs/react-a11y/blob/master/docs/rules/button-role-space.md) rule:

```
Error: DropdownIndicator You have `role="button"` but did not define an `onKeyDown` handler. 
Add it, and have the "Space" key do the same thing as an `onClick` handler. 
See 'https://www.w3.org/WAI/GL/wiki/Making_actions_keyboard_accessible_by_using_keyboard_event_handlers_with_WAI-ARIA_controls' for more info.
```

Thanks!

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue  is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

[...]

PR #1428, opened 2016-12-12
Accessibility Enhancements - Aria tags, Space/Enter keys

sushmabadam:
This PR contains the fixes for the issue #1399 Accessibility Enhancements - Aria tags, Space/Enter keys

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -485,7 +485,10 @@ const Select = React.createClass({
 				this.selectFocusedOption();
 			return;
 			case 13: // enter
-				if (!this.state.isOpen) return;
+				if (!this.state.isOpen) {
+					this.focusNextOption();
+					return;
+				}
 				event.stopPropagation();
 				this.selectFocusedOption();
 			break;
@@ -497,6 +500,17 @@ const Select = React.createClass({
 					this.clearValue(event);
 					event.stopPropagation();
 				}
+			break;
+			case 32: // space
+				if (!this.props.searchable) {
+					event.preventDefault();
+				}
+				if (!this.state.isOpen) {
+					this.focusNextOption();
+					return;
+				}
+				event.stopPropagation();
+				this.selectFocusedOption();
 			break;
 			case 38: // up
 				this.focusPreviousOption();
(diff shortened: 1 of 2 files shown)


cschoder:
Is the only thing preventing this merge the test coverage?

src/Select.js
```diff
@@ -866,12 +880,14 @@ const Select = React.createClass({
 					aria-expanded={isOpen}
 					aria-owns={isOpen ? this._instancePrefix + '-list' : this._instancePrefix + '-value'}
 					aria-activedescendant={isOpen ? this._instancePrefix + '-option-' + focusedOptionIndex : this._instancePrefix + '-value'}
+					aria-labelledby={this.props['aria-labelledby']}
```
agirton: Nit: Looks like there is a space before the dash, please remove.

src/Select.js
```diff
@@ -866,12 +880,14 @@ const Select = React.createClass({
...
+					aria-label={this.props['aria-label']}
```
agirton: Please update this also.

src/Select.js
```diff
@@ -866,12 +880,14 @@ const Select = React.createClass({
...
 					className={className}
 					tabIndex={this.props.tabIndex || 0}
 					onBlur={this.handleInputBlur}
 					onFocus={this.handleInputFocus}
 					ref={ref => this.input = ref}
-					aria-readonly={'' + !!this.props.disabled}
+					aria-disabled={'' + !!this.props.disabled}
```
agirton: One more :smile:

sattaman:
It would be great to get this merged!

dsumer:
bump :) .. would be really nice to see those changes in the next version of react-select!

gwyneplaine:
Thanks for this @sushmabadam  I've cleaned this up and merged it into master in the following PR https://github.com/JedWatson/react-select/pull/2092

src/Select.js
```diff
@@ -498,6 +501,17 @@ const Select = React.createClass({
...
+			case 32: // space
+				if (!this.props.searchable) {
+					event.preventDefault();
+				}
+				if (!this.state.isOpen) {
+					this.focusNextOption();
+					return;
+				}
+				event.stopPropagation();
+				this.selectFocusedOption();
```
buob: Hey guys, sorry to drudge up an old PR discussion, but I'm noticing some quirky behavior since this code.  Bisecting for bug #2156 led me to PR #2092 and therefore here.

It would seem this didn't intentionally cause the functionality I'm noticing on multiselect (selecting options on space or creating options on space for `Creatable`), so I'm trying to figure out how I might fix this bug while keeping the intentions of this change.  

What use case are we trying to add here?  As an example of a use case we don't want (I'm assuming) is if you're on the example page in the first example of states and click "United States," then start typing to New York, when you hit space after `New` it will select `New Hampshire` instead of continuing to filter down.

Tagging @gwyneplaine as the merger.

gwyneplaine: Hey @buob  thanks for catching this, the intention was to mimic native select behaviour, to enable selection of a focused option on SPACE. this is definitely an unintended side-effect. 

Let me have a think about how to best solve for both these cases (it may very well be removing this behaviour when props.searchable is true). Happy to hear your thoughts on this as well

dlinch: I went hunting, and this seems to be the most relevant. I'm still seeing a space to select items here, but my search is supposed to support space characters for my options.

Was there ever an option added to disable the space select on a multiselect search?

PR #1430, opened 2016-12-14
[Creatable] Handle null children prop gracefully

jscinoz:
Previously ```Creatable``` used ES6 destructuring default values to set ```children``` to ```defaultChildren```. 

If a null ```children``` prop is passed, the null value will be used, rather than the default value, as Babel compiles destructuring defaults to ```value === undefined ? defaultValue : value```. This will result in an exception when ```children``` is invoked as a function a few lines later.

This patch simply amends Creatable's ```render()``` method to use ```defaultChildren``` if ```children``` is any kind of falsey value.

### DIFF
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -205,11 +205,20 @@ const Creatable = React.createClass({
 
 	render () {
 		const {
-			children = defaultChildren,
 			newOptionCreator,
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @jscinoz

Issue #1439, opened 2016-12-17
No options are found on mouse down  after 'onBlur' for Async Select.


vapon:
The bug could be reproduced on the standard demo page.

- Navigate to Contributors (Async) and choose 'Single Value'
- Try to type something that doesn't fit any name (like 'aaa')
- Click somewhere outside the select 
- Put the cursor back on the input field 

Result: the initial input was wiped out (onBlurResetsInput=true by default) but options are missing!

![step-1](https://cloud.githubusercontent.com/assets/18501699/21286790/8b409f42-c46e-11e6-8a9b-5b7eedae8ba1.png)
![step-2](https://cloud.githubusercontent.com/assets/18501699/21286791/8d8ab4f4-c46e-11e6-9ddd-716eadfd6d63.png)
![step-3](https://cloud.githubusercontent.com/assets/18501699/21286793/901496ea-c46e-11e6-8f71-8f546d4bdaee.png)

I was able to fix it in my forked repo of react-select.
Test was added as well.

vapon:
https://github.com/JedWatson/react-select/pull/1445 - pull request

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you feel this issue / pull request is still relevant and you'd like us to review it, please leave a comment and we'll do our best to get back to you.

PR #1445, opened 2016-12-21
No options are found on mouse down after 'onBlur' for Async Select. #1439

vapon:
Hi this is a fix for the issue I've created.


gwyneplaine:
thanks for this @vapon, a fix for the root problem addressed in this PR has been merged into master

PR #1448, opened 2016-12-21
Fixes #1447 - Uncaught TypeError

### LINKED ISSUE #1447: Cannot read property 'option' of undefined
# Problem
If you try to use arrow keys in a Select with zero options after tabbing to it it throws `TypeError Cannot read property 'option' of undefined`.

## Demo
https://plnkr.co/edit/CYKgXtDCN8N3F16StNTq?p=preview
Steps to reproduce:
1. Focus the top input element
2. Tab to the Select element
3. Push up or down arrow key

thezanke:
Arrowing up or down in an unopened Select with 0 options caused TypeError. See: #1447

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -709,7 +709,7 @@ const Select = React.createClass({
 			this.setState({
 				isOpen: true,
 				inputValue: '',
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @thezanke!

PR #1451, opened 2016-12-27
Document the undocumented `ignoreAccents` property

dmatteo:
Just adding some documentation for `ignoreAccents`, after having scrambled a bit to understand why the hell my accents were not sent over the wire :)


JedWatson:
Thanks @dmatteo!

PR #1452, opened 2016-12-28
Add config option className dynamically feature

geoffgu:
i want to define option className dynamically, but i cannot find this feature in doc, util i found the following code in Option.js
`var className = classNames(this.props.className, option.className);`
it can meet my requirement, so you can close my pull request, but i suggest that you can add option object's property introduction to the doc. like the following code.
`{options: [{label: 'xxx', value: 'xxx', className: 'xxx', style: 'xxx'}, {}, ...]}`
                                                                                                                                                            thanks


JedWatson:
Thanks @geoffgu, I've added that to the docs. Let me know if you think it could be clearer.

seanlindo:
Thank you @JedWatson!

Issue #489, opened 2015-09-30
Pass current value in `onBlur` event


JedWatson:
As discussed in erikras/redux-form#82

When you listen to the `onBlur` event on input fields, `e.target.value` will contain the value of the input which can be useful. This behaviour doesn't work in `react-select` because the DOM node being blurred is the search input, not the field itself.

We could hack around this and set `e.target.value` before firing the event handler - I'm not absolutely positive this would have no unintended side effects, but it seems safe.

Anyone want to weigh in on this?

oluckyman:
Related discussion was here #186

Agamennon:
@JedWatson i might have a related issue here,  if the user types in a search, but does not select it,  it just completely types in the option then blurs the field, if what the user input completely matches on of the options, the select component should select that option and fire a change event.  (at this time it will fire if the user tab's out, but not if the user clicks out from the component)

gmethvin:
@JedWatson I have a similar issue. In my case, if the user has typed something but deselects the input, I'd like it to add that value to my multiselect. If a user presses tab, I would like it to also add the value and keep the input selected. For an example, check out how the compose view in Gmail works when you type someone's email in the "to" field.

I think there should be some configurability with respect to what happens when a user fails to select an option.

motoedie:
@JedWatson I use multiselect with redux-form and it expects data which is stored and propagated back to form. Could it be possible to add calling eg. onBlurValues?

I use wrapper component between redux-form and select component so I could map onBlur to onBlurValues.

wmertens:
So basically, `onBlur` should behave like any input and provide the current value. Can't think of any reason overwriting `event.target.value` is a problem, the events are syntetic events created by React to begin with.

Configurability of behavior could be done by adding methods to the Select class that are available via a ref to the component?

ntucker:
any new ideas here?

theskillwithin:
causes problems with redux-form

kumar91mahesh:
@JedWatson  i am getting blank value in mobile browser in onBlur method. I am getting value of event.target.value in side onBlur method. 
can anyone tell me the solution ?

jossmac:
Assuming resolved in v2. Bump if needs to be reopened.

Issue #186, opened 2015-05-12
Getting value onBlur


anthonator:
Is it possible to get the value of select on blur? Unlike `onChange` `onBlur` returns an event. So when I do something like `e.target.value` I always get an empty string.

dcousens:
@anthonator I'm not familiar, but with a typical `onBlur` event for a text box,  is the `e.target.value` the text from the text box?  We should probably mimic typical components in this respect.

anthonator:
Yes, `e.target.value` is the value of the input.

at0g:
:+1: for adding this.

would be nice if the first argument passed to the onChange callback was a similar event rather than just the value.

P.S This issue was killing me today in conjunction with #188, as I want to trigger validation after the initial blur event, then on subsequent change events.

If anyone else is in a similar situation, I ended up resolving it with this wrapper component.
The faux event objects are naive, but I only need `e.type` and `e.target.value`

```
'use strict';

var React = require('react');
var Select = require('react-select');

module.exports = React.createClass({

    displayName: 'SelectWrapper',

    getDefaultProps () {
        return {
            value: null
        }
    },

    getInitialState () {
        return {
            value: this.props.value
        }
    },

    onBlur () {
        let blurHandler = this.props.onBlur;
        if (blurHandler) {
            blurHandler({
                type: 'blur',
                target: {
                    value: this.state.value
                }
            })
        }
    },

    onChange (value) {

        let changeHandler = this.props.onChange;
        if (changeHandler) {
            changeHandler({
                type: 'change',
                target: {
                    value: value
                }
            });
        }

        this.setState({ value: value });
    },

    render () {
        let { onBlur, onChange, value, ...props } = this.props;

        return (
            <Select onBlur={ this.onBlur } onChange={ this.onChange } { ...props } { ...this.state } />
        );
    }

});
```

dh376:
+1

arun678:
> 👍 for adding this.
> 
> would be nice if the first argument passed to the onChange callback was a similar event rather than just the value.
> 
> P.S This issue was killing me today in conjunction with #188, as I want to trigger validation after the initial blur event, then on subsequent change events.
> 
> If anyone else is in a similar situation, I ended up resolving it with this wrapper component.
> The faux event objects are naive, but I only need `e.type` and `e.target.value`
> 
> ```
> 'use strict';
> 
> var React = require('react');
> var Select = require('react-select');
> 
> module.exports = React.createClass({
> 
[...]

PR #1456, opened 2017-01-02
Adding new option to select the currently focused option on blur

craigatk:
In the current application where I'm using react-select, I'm looking for a way to automatically select the current option when the user clicks out of the select box without hitting 'enter' or 'tab' with a multi-select and add the currently-focused option to the list of selected options. I'm currently using a wrapper element similar to the one in https://github.com/JedWatson/react-select/issues/186 to manually select the value. Moving this functionality into the select component will remove the need for this type of workaround and hopefully resolve at least part of the cause of issues  https://github.com/JedWatson/react-select/issues/186 and https://github.com/JedWatson/react-select/issues/489

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ bower_components
 
 # Editor artifacts
 .idea
+*.iml
 
 # Other
 .DS_Store
--- a/src/Select.js
+++ b/src/Select.js
@@ -82,6 +82,7 @@ const Select = React.createClass({
 		noResultsText: stringOrNode,                // placeholder displayed when there are no matching search results
 		onBlur: React.PropTypes.func,               // onBlur handler: function (event) {}
 		onBlurResetsInput: React.PropTypes.bool,    // whether input is cleared on blur
+		onBlurSelectsFocusedOption: React.PropTypes.bool, // whether to select the currently focused option on blur
 		onChange: React.PropTypes.func,             // onChange handler: function (newValue) {}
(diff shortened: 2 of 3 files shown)


zacharytamas:
We need this functionality as well. Is there any movement on it?

stephen-thomas:
We also need this functionality. Would really improve UX for our users.

peterchappy:
@JedWatson Any movement on this?

jossmac:
Hi @craigatk,

Sorry mate, this is not something that belongs in core; it's not the expected behaviour and can be achieved in consumer source by extending the class.

timurridjanovic:
why did you close this? It makes no sense that you are not providing the value on the input onBlur... Absolutely no sense..

mponizil:
@jossmac any tips for implementing this in user-land? Not having luck with the following:

```
const creatable = useRef(null);
const handleBlur = () => {
        if (creatable.current.select.state.focusedOption) {
            creatable.current.select.selectOption(creatable.current.select.state.focusedOption);
        }
}

... 

return (
    <Creatable ref={creatable} onBlur={handleBlur} ... />
)
```

jossmac:
@gwyneplaine perhaps we could pass some pertinent data to `onBlur`?

```jsx
<Select onBlur={(event, { focusedOption, inputValue, selectedOption }) => {
  // Some consumer logic
}} />
```

PR #1468, opened 2017-01-06
Adding the option to keep the dropdown menu open

athese6:
some application may need this option so i have added it as the keepOpen prop


nealoke:
Also way handier when styling the options xD now I'm opening -> make css change -> repeat which gets annoying. :)

JedWatson:
I'm not keen on having an `isOpen` prop because it complicates things quite a bit... happy to revisit the decision after 1.0 is out but I'm going to close this because it has a large number of conflicts.

Thanks for the PR @athese6

PR #1471, opened 2017-01-09
[Creatable] Fix multiple clicks required to create new option

### LINKED ISSUE #1327: newOptionCreator is called twice for every key pressed?
Created this simple demo to illustrate:
(sorry, I tried to use plunker, but could not import Creatable)

```
import React from 'react'

import {Creatable} from 'react-select'
import '/node_modules/react-select/dist/react-select.css'

const optionList = [
  {value: 1, label: 'one'},
  {value: 2, label: 'two'},
  {value: 3, label: 'three'}
];

export const SelectCreate = React.createClass({

  getInitialState () {
    return {
      options: [
        {value: 2, label: 'two'}
      ]
    };
  },

  onChange (options) {
    console.log('onChange', options);
    this.setState({options: options});
  },

  newOption (obj) {
    console.log('newOption', obj);
    return {
      [obj.labelKey]: obj.label,
      [obj.valueKey]: obj.label
    };
  },

  validateOption (obj) {
    return obj && obj.label && obj.label.length > 1;
  },

  render () {
    return (
      <Creatable
        name="select-create"
        multi={true}
        newOptionCreator={this.newOption}
        isValidNewOption={this.validateOption}
        value={this.state.options}
        options={optionList}
        onChange={this.onChange}
        clearable={true}
      />
    )
  }
});
```

When typing 'qwerty' in the select input field, the console output looks like this:

```
newOption Object {label: "qw", labelKey: "label", valueKey: "value"}
newOption Object {label: "Create option "qw"", labelKey: "label", valueKey: "value"}
newOption Object {label: "qwe", labelKey: "label", valueKey: "value"}
newOption Object {label: "Create option "qwe"", labelKey: "label", valueKey: "value"}
newOption Object {label: "qwer", labelKey: "label", valueKey: "value"}
newOption Object {label: "Create option "qwer"", labelKey: "label", valueKey: "value"}
newOption Object {label: "qwert", labelKey: "label", valueKey: "value"}
newOption Object {label: "Create option "qwert"", labelKey: "label", valueKey: "value"}
newOption Object {label: "qwerty", labelKey: "label", valueKey: "value"}
newOption Object {label: "Create option "qwerty"", labelKey: "label", valueKey: "value"}
```

(It does work BTW; if I press TAB, the new option 'qwerty' is added to the options.)

billyvg:
When you attempt to create a new option, it does not have a `value` property -- the "Create new label" option also does not have a `value` property, so `isOptionUnique` believes they are equal.

Is this the right place to handle the issue?

Fixes #1327


dave-clover:
Can you write a test for this to ensure the behavior doesn't regress when future changes are made?

JedWatson:
Can you provide a full usage example here? I can't reproduce the behaviour you're seeing.

I suspect you're using a custom `newOptionCreator` method that's not setting the `value` property on the new option, the default one sets both the `value` and the `label` property to the input value:

```js
function newOptionCreator ({ label, labelKey, valueKey }) {
	const option = {};
	option[valueKey] = label;
 	option[labelKey] = label;
 	option.className = 'Select-create-option-placeholder';
 	return option;
};
```

billyvg:
That's correct, we have an option whose `value` as an empty string.

PR #1475, opened 2017-01-11
Clearing value bug

michalvankodev:


michalvankodev:
Please ignore :( I'm sorry.. github was more clever then it should be

PR #1480, opened 2017-01-12
added shouldReloadOptions option to Async component

VanCoding:
Hi @JedWatson 

I had the need to add an option to the Async component that lets it automatically reload its options when it receives new props or the user opens the list.

As use case, imagine the following UI:

Client: [Client Select Box]
Invoice: [Invoice Select Box]

The invoice select box lets you select an invoice of the client selected in the client select box.
As soon as you select a different client, and then want to select the invoice, you should get the invoices of the new client.

Currently, the Async component assumes that a call to `loadOptions` with the same search term will always return the same options, and therefore does not reload its options when receiving new props, or opening the list. But in the use case above, that is not the case. `loadOptions` would return different options, if it was called. But it isn't!

That's why I added the option `shouldReloadOptions`, which is a function that, when called, returns a boolean value indicating if a call to `loadOptions` would return different results than before. If it returns true, the Async component then automatically reloads its options.

What do you think?


bobsmits:
Can this be merged, got the same problem?

dave-clover:
If you disable the internal cache, do you get the behavior you want?

blunatic:
+1

VanCoding:
@dave-clover No. This is thought as an addition to disabling the cache.

ajbdev:
Is there anything preventing this from being merged in? I'm using a pretty ugly workaround to mimic this behavior by conditionally rendering two different <Select> components that are nearly identical aside from their loadOptions function. It would be nice to have this as a built in option instead.

obykoo:
Is this at npm ? or not released yet?

alanhr:
I have the same problem on my app. :(
When this Pr will be merged in master?

alanhr:
Could you fix your PR @VanCoding  ?
I need a lot of this feature.

VanCoding:
fixed.

alanhr:
@VanCoding  thanks ;)

tommyalvarez:
Any news on this? I also need to refresh the options due to cascading selects dependency.

eliseumds:
That would be great to have!

rcarausu:
I'm also waiting on this cool little feature :)

asaarnak:
Any news on this ?

flexdinesh:
Hey @VanCoding [v1 of react-select is no longer maintained](https://github.com/JedWatson/react-select/pull/3786). 

Thanks for taking your time out to create this PR but since this PR has changes for `v1`, we cannot merge it right away. If you would like to discuss this feature for the latest versions of `react-select`, please open an issue with feature details and we could move forward with the discussion there. We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest versions.

PR #1482, opened 2017-01-12
V0.9.1 react15

vasyl-purchel:
Updating peer dependency to React 15.3.1 in order to make it possible to use this component in application built with React 15 without throwing errors on unmet peer dependencies.


bassjacob:
Is the component still compatible with 0.14? Should we make the peerdep at least greater than 14 rather than specifically 15?

JedWatson:
We're well past the last release of v0.9 in master, and about to cut a final 1.0.0 release so I don't see how we can easily release a patch that adds v15 peer deps to 0.9, not to mention I haven't tested 0.9 on React 15 and can't confirm it *is* compatible.

v0.6.x is the last release compatible with React 0.13, up to 0.9 was developed and tested with 0.14, 1.0.0 (which was a comprehensive rewrite) is compatible with 0.14 and 15 and tested extensively with 15.

JedWatson:
Going to close this as unmergeable unless someone wants to perform some git magic to create a new branch from the last 0.9.x release and submit this PR against that, sorry @vasyl-purchel 

If you really need this release on npm I'd suggest publishing it as your own [scoped package](https://docs.npmjs.com/getting-started/scoped-packages) from your fork.

PR #1500, opened 2017-01-23
fix arrow flip

tlconnor:
This PR fixes the select control arrow flipping.

The `.Select-arrow` component is not a direct descendant of `.Select-control` so the existing CSS for arrow flipping was not being used.


JedWatson:
Thanks for this @tlconnor

PR #1508, opened 2017-01-27
fixed exception when clearing an Async field that is not set to multi

VanCoding:
In a7782b34abdc9d2beb8c604ee605fb1e81db8bd2, the following bug was introduced:
When clicking the "clear" Button of an Async Component and its `multi` property is not present or is set to `false`, it throws an exception, because the code expects an array as new value.

I've created a temporary fix for this, that you can pull if you want.
It might make sense to ask @DanielHeath if this is the proper way to fix this, though.


JedWatson:
Good fix @VanCoding, thanks. Unless @DanielHeath has feedback, I think leaving the options in place for a single select is reasonable UX, so I'm happy to go with this.

PR #1512, opened 2017-01-30
Fixed bug. No more comparing objects in getFocusableOptionIndex

rndm2:
When using with redux-form focusedOption can be not equal to any element of options array

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -1009,7 +1009,14 @@ const Select = React.createClass({
 
 		let focusedOption = this.state.focusedOption || selectedOption;
 		if (focusedOption && !focusedOption.disabled) {
-			const focusedOptionIndex = options.findIndex(option => option.value === focusedOption.value);
+			let focusedOptionIndex = -1;
+			options.some((option, index) => {
+				const isOptionEqual = option.value === focusedOption.value;
+				if (isOptionEqual) {
+					focusedOptionIndex = index;
+				}
+				return isOptionEqual;
+			});
 			if (focusedOptionIndex !== -1) {
 				return focusedOptionIndex;
 			}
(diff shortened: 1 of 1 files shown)


JedWatson:
This is a good fix for that problem, but `Array.prototype.findIndex` is unfortunately [not supported in Internet Explorer](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) so we need to write a custom compare loop.

Would you mind implementing this in a cross-browser compatible way?

Thanks!

rndm2:
I can write custom compare loop, but not sure where is best place for it. Inside func?

JedWatson:
Actually... can you tell me some more about why the strict equality comparison doesn't work with `redux-form`? the options are private and the focused option in the state should be a reference to the same object.

rndm2:
I am not sure why, I didn't have a time to find reason. But after @@redux-form/CHANGE it is not same object. 
I am using react-select like described here https://github.com/JedWatson/react-select/issues/1513 in first comment.

rndm2:
I created new commit. Now it IE compatible. 
I will now try to find, why objects are not same, but I think there is something in getValueArray or expandValue functions. There some magic with props.

src/Select.js
```diff
@@ -1009,7 +1009,14 @@ const Select = React.createClass({
 
 		let focusedOption = this.state.focusedOption || selectedOption;
 		if (focusedOption && !focusedOption.disabled) {
-			const focusedOptionIndex = options.indexOf(focusedOption);
+			let focusedOptionIndex = -1;
+			options.some((option, index) => {
```
JedWatson: Neat use of `.some` 👍

praveenaj:
+1

tlconnor:
@JedWatson Is it possible to get this merged if you are happy with the fix?

JedWatson:
Sure is, thanks for the bump!

jbschrades:
Should this still work when using a custom `valueKey`?  Doesn't appear to be working for me.

Issue #1516, opened 2017-01-31
Function onMenuScrollToBottom can not fires when the menu is scrolled to the bottom


Viktor-the-great:
resize the window in the browser chrome v55 and scroll to bottom.
https://plnkr.co/edit/HKvicJntbuxEH2PYPZid?p=preview
result is callback onMenuScrollToBottom is not called. A problem with this code block:
```javascript
handleMenuScroll (event) {
	if (!this.props.onMenuScrollToBottom) return;
	let { target } = event;
	if (target.scrollHeight > target.offsetHeight && !(target.scrollHeight - target.offsetHeight - target.scrollTop)) {
		this.props.onMenuScrollToBottom();
	}
}
```
When the browser is resized, scrollTop returns the fractional value, that why (target.scrollHeight - target.offsetHeight - target.scrollTop) is not equal to zero.
I propose to do so
```javascript
handleMenuScroll (event) {
	if (!this.props.onMenuScrollToBottom) return;
	let { target } = event;
	if (target.scrollHeight > target.offsetHeight && target.scrollHeight - target.offsetHeight - target.scrollTop < 1) {
		this.props.onMenuScrollToBottom();
	}
}
```

dvijeshpatel:
hi @Viktor-the-great , can you tell me is issue fixed?

Viktor-the-great:
the issue is no more relevant

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our efforts towards the current major version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue  is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

PR #1519, opened 2017-01-31
onMenuScrollToBottom fires when the menu is scrolled to the bottom on resize window

Viktor-the-great:
Fixed issue #1516.
Due to rounding error in some browsers (Edge) and fractional values in others (Chrome) on window resize I decided to use an additional element to detect end of scrolling. I find the coordinates of the element and comparing the coordinates of the element and container options

### DIFF
--- a/less/menu.less
+++ b/less/menu.less
@@ -31,6 +31,7 @@
 .Select-menu {
 	max-height: (@select-menu-max-height - 2px);
 	overflow-y: auto;
+	border-bottom: 1px solid transparent;
 }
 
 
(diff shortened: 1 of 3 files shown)


Viktor-the-great:
JSDOM does not compute element size and position. So I can't test onMenuScrollToBottom call in this code block:
```javascript
handleMenuScroll (event) {
	if (!this.props.onMenuScrollToBottom) return;
	let { target } = event;
	const { bottom: endPoint } = this.menuListEndPoint.getBoundingClientRect();
	const { bottom } = target.getBoundingClientRect();
	if (target.scrollHeight > target.clientHeight && bottom >= endPoint) {
		this.props.onMenuScrollToBottom();
	}
},
```
Please tell me, what tests should I write?

PR #1522, opened 2017-02-01
Update changelog

Noviny:
Update of changelog files for rc.3

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,41 @@
 # React-Select
 
-## v1.0.0-rc / 2016-09-04
+## v1.0.0-rc.3 / 2017-02-01
+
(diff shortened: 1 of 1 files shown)


jochenberger:
#1226 should be mentioned, since it is a breaking change (not the kind to do from one rc to another, if you ask me).

Noviny:
@jochenberger I wasn't sure how to summarise that one properly, so I'm waiting for @JedWatson to add that one in.

JedWatson:
I'll add that - good point @jochenberger, we'll try not to do that again, tbh I was more focused on the issues it fixed than the fact that it was an API change, sorry!

JedWatson:
Thanks for this @Noviny - much appreciated!

PR #1533, opened 2017-02-06
Ignore punctuation - feature request

willdembinski:
Hi there,

I ran into an issue where my api was returning results with punctuation that was causing the options list to not display anything because of things like commas/single quotes (apostrophes).

I will open an issue and follow the correct procedure - doing this at work at the moment because I had a hot minute during my lunch break.  Would love anyone's input, its very rudimentary atm.


willdembinski:
This is currently failing CI.  Will tinker

willdembinski:
This is pretty weird, all test/linting and coverall runs to completion without error locally.  In no-way would I ever assume that I am not an idiot...but am I being an idiot more so right now?

PR #1541, opened 2017-02-09
Adding a test case for unhandled promise rejections from loadOptions

staab:
I don't see anything in the docs about rejected promises, and I know it's common for promises to get swallowed, but it's definitely bad behavior to silently fail and not trigger a global handler. I've written a test case for it - let me know if you want me to make an issue or anything.

### DIFF
--- a/test/Async-test.js
+++ b/test/Async-test.js
@@ -192,6 +192,25 @@ describe('Async', () => {
 				typeSearchText('bar');
 				expect(asyncNode.querySelectorAll('[role=option]').length, 'to equal', 0);
 			});
+
+			it('should not swallow rejected promises', done => {
+				const handler = sinon.spy();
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @staab - this is a good point, we should definitely not be swallowing rejected promises. I'd prefer to merge a fix for that before this test though, or it'll break master for all other PRs until it's fixed.

Any chance you'd like to have a go at fixing the implemention as well? Otherwise an issue to track it would be helpful 👍

JedWatson:
I've looked further into this and I don't think it's safe for us to throw errors automatically - the preferred solution would be to catch the error before resolving the promise to react-select, so we won't merge this.

If anyone has a better idea on how to implement this, please let me know...

Issue #1521, opened 2017-01-31
React-Select component with properties crash Firefox on Windows 10


devonbalicki:
Affects Firefox 51.0.1 on Windows 10 32 and 64 bit

When loading a React-Select component that has properties set causes the browser window to crash.

```
let props = this.props;
        return (
                <ReactSelect {...props} noResultsText={ '' }/>
        );
```

This loads fine:

```
let props = this.props;
        return (
                <ReactSelect/>
        );
```

Crash report: https://crash-stats.mozilla.com/report/index/20c12273-27b2-4e1c-b49f-874492170131

cc @peixian @tab1293

amolbarewar:
the same issue is happening for our team also. We are facing problem on Firefox version 51.0.1 in Windows Operating system only.
We are using below code snippet.
`<Select
              className = "block"
              clearable = {false}
              name = "hours"
              onChange = {e => onStateChange(e.value, 'scheduletime', 0)}
              options = {hoursSelect}
              searchable = {false}
              value = {(scheduleTime || '00:00').split(':')[0]} />`

domehead100:
I have a user who reported a hang on Firefox 50.0.1 today on a page using react-select 1.0.0-rc.3.  She's on Windows 8.1.  That crash report posted earlier seems to indicate that the hang was caused by Shockwave Flash, I'll check with my user tomorrow.  Will try disabling hardware graphics acceleration, removing Shockwave if it's installed.

The same page works fine for me and some other users on the same version of Firefox.  

Must be either a FF bug or due to some extension that the user has installed, right?  My page works fine in Chrome and IE 11 (not as pleasant of an experience in IE 11, but it works).

sabapk:
Resolved this issue.

So our code base, was using <Select> Component syntax with each prop to be on new line like,
```javascript
<Select
 options = {timezones}
 value = {timezone} />
```
So, we changed our code style to write props inline and surprisingly this one change resolved our issue of browser crashing.
`<Select options = {timezones} value = {timezone} />`

sabapk:
Resolved this issue.

So our code base, was using react-select `Select` Component syntax with each prop to be on new line.
So, we changed our code style to write props inline and surprisingly this one change resolved our issue of browser crashing.

`<Select options = {timezones} value = {timezone} />`

blackChef:
Same issue here. Browser freezing seems to be random on our product.

peixian:
[...]

PR #1556, opened 2017-02-15
Fix aria-owns attribute - Issue #1521

mhubenthal:
(Addresses issue #1521)

It appears that Firefox is getting stuck in some kind of infinite loop due to incorrect use of `aria-owns` attribute. Here's the W3C spec on `aria-owns`:

https://www.w3.org/TR/wai-aria/states_and_properties#aria-owns

In `Select.js`, line 879, when `isOpen` is false, the component assigns `this._instancePrefix + '-value'` to `aria-owns`. That value corresponds to a DOM node which *is* a parent, and thus has a hierarchical relationship, breaking the W3C spec. 

If you visit the demo in Chrome (https://jedwatson.github.io/react-select/) and inspect the HTML, you'll see that in the situation when the component would be breaking Firefox, Chrome has intelligently removed the offending `aria-owns` value.

(This PR uses the same "classNames" approach to generating `aria-owns` already used on line 841.)

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -871,12 +871,17 @@ const Select = React.createClass({
 
 		if (this.props.disabled || !this.props.searchable) {
 			const { inputClassName, ...divProps } = this.props.inputProps;
+
+			const ariaOwns = classNames({
+				[this._instancePrefix + '-list']: isOpen,
+			});
+
 			return (
 				<div
 					{...divProps}
 					role="combobox"
 					aria-expanded={isOpen}
-					aria-owns={isOpen ? this._instancePrefix + '-list' : this._instancePrefix + '-value'}
+					aria-owns={ariaOwns}
 					aria-activedescendant={isOpen ? this._instancePrefix + '-option-' + focusedOptionIndex : this._instancePrefix + '-value'}
 					className={className}
 					tabIndex={this.props.tabIndex || 0}


mhubenthal:
@omgaz I had the same question. In this case the string value assigned to `aria-owns` was invalid because it referred to a parent node. I guess that's an open question to @JedWatson if there should be tests for valid HTML structure...?

mhubenthal:
@omgaz after poking around a little more this morning I have a bit more to add to the issue. It appears that the invalid use of `aria-owns` breaks Firefox when the parent node is a `<span>` but not when it is a `<div>`. I still think this PR is valid for correcting use of `aria-owns`, but changing the parent node to a `<div>` would also fix the issue.

mhubenthal:
@omgaz @JedWatson any more thoughts on this PR?

omgaz:
Seeing as how aria-owns applies to "All elements of the base markup", changing from a span to a div, like you mention, would feel like a workaround and not a fix.

I think this PR looks good.

dalehurwitz:
Hi @JedWatson any headway on this?

abhishekpillai:
Hi @JedWatson, let me know if there's anything I can do to help get this PR merged! Looking forward to this fix. Thanks! 👍

ekwonye-richard:
I had the same issue, my fix was to use readOnly rather than searchable... Use this props

inputProps={{readOnly: bolean}} 

 I was lucky to quickly effect it across my app  because i wrapped react-select inside a component i wrote.

lxerxa:
looking fotward to this fix.

broadhu2:
When will this pull request be approved?

Kadrian:
@JedWatson Could you take another look / merge it?

rnadrag:
👍 on this to merge/release as soon as possible

agirton:
Hi @mhubenthal this looks good to me. I agree this is the right approach.

rnadrag:
any timelines on getting this released?

agirton:
@rnadrag I would like to get the new Async focus bug fixed and then can get a new release out soon.

rnadrag:
sweet!

benelog:
@agirton 
Thanks for merging it.
The next release will be very helpful to our project, too.

PR #1563, opened 2017-02-18
Fix focus on next Element on tab press

BANOnotIT:
When I chose multiple values and I wanted to focus on the next element by pressing `tab` it didn't work so I had to press `esc` first and only afterwards `tab`.
Most clients encounter with this problem so I fix it by adding one more cond in keyDown handler.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -482,7 +482,7 @@ const Select = React.createClass({
 				}
 			return;
 			case 9: // tab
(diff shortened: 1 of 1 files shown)


gwyneplaine:
@BANOnotIT I can not reproduce this in newer versions of react-select. If this is still something you can reproduce, please note down reproduction steps and the react-select version you're using.

Please see below for the latest version of react-select examples: 
http://react-select.netlify.com/

PR #1586, opened 2017-03-06
Added onRemoveValue props

WarisR:
When I remove a value, I also want to remove that value in the option if I created it before by typing in the text field. So I add a onRemoveValue props, it is a callback when a value was removed, It make I can do what I want in my own project. 

- Able to pass onRemoveValue props to any react-select components.

- onRemoveValue would be called when a value is removed (By clearing all values, backspace or click 'x' button)
- In case of remove 1 value, onRemoveValue would be called with a value parameter. (onRemoveValue(value))
- In case of clear all values, onRemoveValue would be called without a parameter. (onRemoveValue())


jossmac:
Version 1 of react-select is no longer supported.

In the best interest of the community we've decided to spend the time we have available on the latest version.

We apologise for any inconvenience.

Please see:

1. [v1 to v2 upgrade guide](https://react-select.com/upgrade-guide)
1. [v1 documentation and examples](v1.react-select.com)

PR #1588, opened 2017-03-06
Changing confusing doc wording

weexpectedTHIS:
Definition of onChange handler parameter says it returns the newValue but it actually returns the newOption (which is a hash)


JedWatson:
Thanks @weexpectedTHIS!

Unfortunately it's conflicted with the updated formatting in the Readme - would like to update your PR? otherwise I'm happy to copy this across myself.

PR #1590, opened 2017-03-08
Fix multi select dropdown list gets cleared issue

katexue:


gwyneplaine:
Thanks for this @katexue, not exactly sure what this PR is fixing so I'm closing this for now. 
Happy for this to be resubmitted with the reference issue (if it's still open). Please also make sure changes are in the /src (and don't commit build files).

PR #1595, opened 2017-03-09
Update tabIndex propType to string or number

### LINKED ISSUE #1132: Possibly wrong propType?
Hello, I have a question. Why is [tabIndex](https://github.com/JedWatson/react-select/blob/master/lib/Select.js#L125) prop considered as a string? IMHO number would be more accurate. Currently I must do conversion like this `tabIndex={String(this.props.tabIndex)}`, otherwise I get "Warning: Failed prop type: Invalid prop `tabIndex` of type `number` supplied to `Select`, expected `string`."

wagoid:
This PR fixes #1132. Integer properties can be set as either a `number` or a `string`, depending on the way you set the property value. When using [IDL](https://github.com/facebook/react/blob/e452e3374135c116ef687a8bb3a5d277e3cde8fb/src/renderers/dom/shared/DOMPropertyOperations.js#L146), `tabIndex` should be an integer, but when using [`element.setAttribute()`](https://github.com/facebook/react/blob/e452e3374135c116ef687a8bb3a5d277e3cde8fb/src/renderers/dom/shared/DOMPropertyOperations.js#L158), the value needs to always be a `string`. React converts values to string when setting the value through `setAttribute`, but I think it's better to keep both of the types for backwards compatibility purposes 😄 
More on content vs IDL attributes [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Content_versus_IDL_attributes).

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -37,6 +37,10 @@ const stringOrNode = React.PropTypes.oneOfType([
 	React.PropTypes.string,
 	React.PropTypes.node
 ]);
+const stringOrNumber = React.PropTypes.oneOfType([
+	React.PropTypes.string,
+	React.PropTypes.number
+]);
 
 let instanceId = 1;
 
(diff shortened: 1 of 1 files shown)


noah-conning:
I need this as well.

agirton:
Hi @wagoid can you please rebase on top of master and add unite tests to handle the new prop type?

gwyneplaine:
Thanks for this @wagoid, this has now been merged into master via #2126

PR #1600, opened 2017-03-12
Fixed Numeric Values Multi-Select

edmondpr:
Selected options weren’t being added to the input in Multi-Select mode
for Numeric Values

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -568,7 +568,13 @@ const Select = React.createClass({
 		/** support optionally passing in the `nextProps` so `componentWillReceiveProps` updates will function as expected */
 		const props = typeof nextProps === 'object' ? nextProps : this.props;
 		if (props.multi) {
-			if (typeof value === 'string') value = value.split(props.delimiter);
+			if (typeof value === 'string') {
+				value = value.split(props.delimiter);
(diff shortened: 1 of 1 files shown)


agirton:
Hi @edmondpr since the test coverage went down by a lot, can you add some unit tests to cover this?

gwyneplaine:
Hi @edmondpr  I'm having trouble reproducing the error case for this PR, would you be able to provide me with steps to reproduce, along with the version of react-select this was occuring in?

edmondpr:
Hi Charles,

It's very easy to reproduce on the demo page of React-Select:
http://jedwatson.github.io/react-select/
Go to the Numeric Values section, make sure Multi-select is checked and try
to add values from the dropdown menu. You'll notice that no value is being
added.
Let me know if you still have issues reproducing.
I've created the fix for the issue here:
https://github.com/JedWatson/react-select/commit/1b06702ef4a15be0ab495995f8ca4a020bbb845f

Cheers,
Edmond

gwyneplaine:
@edmondpr thanks for this, reproduced the error. I've addressed this in a separate PR which accounts for cases where we are interacting with a set of options that have a mixed bag of Strings and Numbers.

Issue #1479, opened 2017-01-12
How to prevent the dropdown list to go off screen?


ghost:
How to prevent the dropdown list to go off screen?

rubencodes:
It depends on the rest of your code, but I'd suggest looking into the `scrollMenuIntoView` option.

robertmaloney:
Also check out the `menuBuffer` property in the [options](https://github.com/JedWatson/react-select#further-options) for a potential solution.

User8765:
isn't there an option to open the outer menu to the opposite direction (i.e. up) , if it reaches to the edge of the window height?

colinhunt:
`menuBuffer` doesn't appear to be working for me. I'm seeing the menu overflow the viewport on the bottom no matter what I set this to. Anyone else?

OmriAharon:
+1
Happens to me as well, `scrollMenuIntoView` doesn't seem to work.

tcosentino:
Unfortunately, @OmriAharon's solution doesn't work for me. My page does not scroll, so a dropdown near the bottom of the page needs to open up rather than down. Seems like this is a common use case that should be solved by default? Seems to be a lot of people with the same issue and no response for a while.

I'd be interested in taking a stab at this if you could recommend what changes should be made @JedWatson? In the meantime I'll use the css hack to manually mark which ones need to open up

hashwin:
FWIW @OmriAharon's solution worked great for me. It opens up the entire dropdown list in the view despite being in a modal. Otherwise, when trying to scroll by clicking with a mouse or swiping in mobile would just close the dropdown list at the end of the screen.

Before:
![before](https://user-images.githubusercontent.com/406115/37142935-8eb99dce-22df-11e8-9153-0dbd8d51f6ff.png)

After:
![after](https://user-images.githubusercontent.com/406115/37142936-8f5384a2-22df-11e8-8ce3-dfa6f165da96.png)

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you feel this issue / pull request is still relevant and you'd like us to review it, please leave a comment and we'll do our best to get back to you.

RebootGG:
`menuPlacement="auto"` now solves this issue. 

> Default placement of the menu in relation to the control. 'auto' will flip when there isn't enough space below the control.

PR #1601, opened 2017-03-13
Fix `scrollMenuIntoView`

OmriAharon:
`srollIntoView` worked incorrectly - only when scrollable container was the `window` object.

Addresses [issue #1479 ](https://github.com/JedWatson/react-select/issues/1479)


chadwilken:
What is the status of this? I am experiencing this issue as well.

gwyneplaine:
@OmriAharon thanks for the PR, unfortunately this introduces a jumping effect that is not what we would like to support, we'll look at a different solution in the future.

PR #1610, opened 2017-03-20
MultiSelection after you add space will not load

mansisce:
In the case of multiselection, it shouldn't accept space after an option is selected. This is being fixed with this change. Check example GithubUsers.


gwyneplaine:
Thanks for this @mansisce, this has been fixed in https://github.com/JedWatson/react-select/pull/1862

Issue #1561, opened 2017-02-16
Select.Async list disappear when two selected


antoine35:
I have found a problem with the select.async component with multi-selection mode (true value for the multi property). Everything works fine when I select one item, but as soon as I select a second item, the drop down list empties itself.But if I start typing one character, the drop down list appears with the suggested items that match that character.

 Thanks

diwakar-s-maurya:
+1. I am also facing the same problem. using _react-select: 1.0.0-rc.3_

antoine35:
it seems to be better with 1.0.0-rc.2

Jimbo80:
+1. 
With searchable=false it is impossible to select more than two items. When searchable is true you can still type to get more results.
Works fine in 1.0.0-rc.1 (couldn't get it to work in rc.2 either).

kornholio83:
Reverted back to rc.1 and works perfectly

traviss-comvex:
I can confirm (after hours debugging) that _react-select@1.0.0-rc.3_ exhibits the same issue as mentioned by @diwakar-s-maurya.

I downgraded to rc2 and it worked as expected

gazpachu:
The PR is not working for me either. All the selected options disappear when starting a new search with a different letter

eddyLazar:
Is there any progress with this issue? Facing the same problem

mehafi:
There is still problem in 1.0.0-rc.5

JedWatson:
I believe I've fixed this, and will release it soon as rc.9

The menu will always close by default for both single and multi-select when an option is selected, but you can change that with the closeOnSelect={false} prop.

To keep suggestions from the Async component after selecting an option, also set onSelectResetsInput={false}

joshuahatcher:
I seem to be experiencing this same behavior on v. 1.2.0. I see that the issue was fixed on rc.9, which was from an earlier release, so I wonder if something happened to break it? I'd prefer not to revert to an rc version if possible.
I can confirm that I'm getting the exact behavior @antoine35 initially described on v. 1.2.0 of both `Select.Async` and `Select.AsyncCreatable`. I tried overriding the `clearOptions()` method to do nothing (see: https://github.com/JedWatson/react-select/issues/1542 ) to no avail.

@JedWatson - any ideas?

PR #1612, opened 2017-03-21
Fix async with multiselect

### LINKED ISSUE #1542: Async Multiselect, options dropdown hides every time an option selected
In Async Multiselect (Select.Async), if multi-select is true then options dropdown hides every time an option selected. To see the dropdown I have to search/type. I want to keep dropdown opened instead of hiding it every time after selecting an option like happens in normal Multiselect. 

A gif is attached showing issue

react-select-async-multi

dekelb:
This PR fixes #1542 and #1561
Original PR that caused this issue: #1508


dekelb:
@JedWatson Did you manage to check this pr? Will appreciate if you can merge and release a new version as this fix a really problematic behavior with the async multiselect.

Thanks!

dekelb:
@JedWatson, @agirton now that this package get a bit more attention, can you please check this pr? Thanks!

agirton:
Hi @dekelb thank you for your contribution. It seems like this would call the API, what's the primary goal for adding the additional method calls?

dekelb:
@agirton the main issue was introduced in https://github.com/JedWatson/react-select/pull/1508/files, in my code I was trying to revert that issue and fix the original problem there.

The original problem was that when the value was changed, the internal list was not cleared, and there was nothing to show after choosing the first value. Clearing the internal state with `this._onInputChange('');` fixed that.

ericnewcomer:
This is the same issue described in #1753 as well, so I'm behind getting a solution merged for this. Haven't looked through this PR yet, but reissuing the empty query is along the lines of what I was proposing.

agirton:
Hi @dekelb haven't pulled this down locally, but couldn't you solve this by just removing the clearOptions method call without having to add two additional inputChange calls?

dekelb:
@agirton I tried using `this.clearOptions` but it didn't work (maybe it's related to some other issues, it looks like it doesn't trigger the relevant callbacks, setting focus for example, getting the relevant list of items, etc). I know this package is getting better attention these days, maybe now with some other fixes the `this.clearOptions` will work, I admit that I didn't try it in v1.0.0-rc.5

gwyneplaine:
thanks for this @dekelb, this has been fixed in a separate body of work by @JedWatson, 
outlined here https://github.com/JedWatson/react-select/issues/1542#issuecomment-328919546
you can find the commit here https://github.com/JedWatson/react-select/commit/c7fe6f44763d680313ef7620e3c5373c87b8e319

PR #1616, opened 2017-03-23
Update README.md

fatman-:
Fix markdown table syntax under 'Further Options'


fatman-:
Not sure of the cause behind TravisCI error! After all, it is just a README edit

fatman-:
#1621 fixes this issue in a better way; ergo, closing this one!

Issue #1597, opened 2017-03-10
onInputChange() result is not used


startswithaj:
This code:
https://github.com/JedWatson/react-select/blob/master/src/Async.js#L151
```
              if (ignoreAccents) {
			inputValue = stripDiacritics(inputValue);
		}

		if (ignoreCase) {
			inputValue = inputValue.toLowerCase();
		}

		if (onInputChange) {
			onInputChange(inputValue);
		}

		return this.loadOptions(inputValue);
```

should be 

```
              if (ignoreAccents) {
			inputValue = stripDiacritics(inputValue);
		}

		if (ignoreCase) {
			inputValue = inputValue.toLowerCase();
		}

		if (onInputChange) {
			**inputValue** = onInputChange(inputValue);
		}

		return this.loadOptions(inputValue);
```

min20:
Creatable.js onInputChange has same issue:
https://github.com/JedWatson/react-select/blob/master/src/Creatable.js#L172

`
onInputChange (input) {
	const { onInputChange } = this.props;

	if (onInputChange) {
		onInputChange(input);
	}

	// This value may be needed in between Select mounts (when this.select is null)
	this.inputValue = input;
},
`

PR #1617, opened 2017-03-23
Fix Creatable.js onInputChange()

min20:
#1597 Fix Creatable.js onInputChange() to use props.onInputChange() and return its result.


min20:
How can I treat Travis 504 bad gateway error? T.T

batmandarkside:
https://github.com/batmandarkside/react-select-2/pull/1

min20:
close with #1669

PR #1624, opened 2017-03-24
Formatting of further options on Readme

tbohnen:

PR #1626, opened 2017-03-25
Fix table formatting on Readme.md

eknkc:


eknkc:
Oh this is a dupe of #1624. Closing.

Issue #1036, opened 2016-06-18
Cannot select menu items on mobile


vitaminwater:
I got on a bug preventing me from selecting any menu items on mobile only.
Turns out, selecting an item would trigger `handleInputFocus` with `Select-input` as target.

Removing the call to `renderInput` for non-searchable select fixed it.

Moeface:
Also having this same issue, though only on iOS devices - both Chrome and Safari. Works fine on Android phones.

adamwan92:
Same bug here. Have to hold the item on mobile for it to work, pretty annoying!

adamwan92:
@vitaminwater Can you explain what you did when you said 

> Removing the call to renderInput for non-searchable select fixed it.

vitaminwater:
@adamwan92 I got it to work by changing something in there code, did not make a pull request because of lack of time, and because I don't really know if I'm not breaking anything else.

Maybe @JedWatson got what I was talking about ? Interested in a PR for this ?

adamwan92:
@vitaminwater A PR or even a copy past of the code that you changed would help very much!

It's weird cause running the demo (jedwatson.github.io/react-select) works very well on mobile.

vitaminwater:
@adamwan92 ok thats' really dirty for now: in `node_modules/react-select/lib/Select.js`, line 938, replace `this.renderInput(valueArray)` with `this.props.searchable ? this.renderInput(valueArray) : null,`.

But honestly I can't really encourage you to do this, the change will be lost as soon as you update the package.

adamwan92:
@vitaminwater thanks so much for the help, but I'm afraid that doesn't solve the issue for me.

peterkuiper:
Same problem here, is there a PR already? If not I can create one. Not sure if the fix by @vitaminwater is good enough

NekR:
I don't know, but I cannot select even on desktop right now (Chrome/macOS)

NGuldager:
Anyone know the status of this? I have a similar issue but when using the solution @vitaminwater supplied, then it works fine.

idanwe:
Fast click has a disable hook, add `needsclick` class to option element https://github.com/ftlabs/fastclick#ignore-certain-elements-with-needsclick

mpigsley:
Any update on this? @peterkuiper did you get a PR ready?

peterkuiper:
[...]

PR #1627, opened 2017-03-27
fixing #1036 (cannot select option on mobile) with vitaminwater's solution

DeBaum:
I have no idea, if this is the way to go, but since this works and nobody made a pull request out of this: here it is!


JedWatson:
Thanks for the PR @DeBaum 

We've made some other changes in this area that I believe have fixed the issue - I can't reproduce any problems using the latest react-select on mobile :)

Issue #1514, opened 2017-01-30
complete:true is not working in loadOptions callback in 1.0.0-rc.2


jariwalabhavesh:
I am returning loadoptions callback with complete: true still it will call loadoptions function again on typing in textbox

```
loadOptions(input, callback) {
        const {AbInstance, filter } = this.props;
        return getDataFromServer({ AbInstance, filter }).then((response) => {
            callback(null, {
                options: response,
                complete: true
            });
        }, (error) => {
            callback(error, null);
        });
}
```

` <Select.Async value={value} clearable={clearable} disabled={disabled} onChange={onChange} valueKey={valueKey} labelKey={labelKey} loadOptions={loadOptions} backspaceRemoves={backspaceRemoves} autoload={true} searchable={true} noResultsText={noResultsText} placeholder={selectPlaceholder} searchPromptText={searchPromptText} autofocus={autofocus} />`

autoload is true so when select load first time it will call loadOptions function, and my loadOptions function taking all data from server and returning in callback with options and complete : true.

Still when ever i search in textbox it will calling loadOptions function again, i thing it should return data from cached result instead of calling loadOptions function again

Any idea, i am using 1.0.0-rc.2

james-gardner:
Not working for me either on 1.0.0-rc.3.

james-gardner:
Line 84 of Async.js and onwards. If anyone can give me a pointer around where the check for 'complete' should go then I'll make a PR. I would imagine that we force the use of the cache.

shlajin:
Yes, bothers me as well. I made a search by "complete" substring through whole repository and there's indeed nothing related. Anybody know if this feature coming anytime soon?

Rauttis:
Same issue here, loadOptions is ran multiple times despite `complete: true`

gazpachu:
@JedWatson any idea of when will this get fixed? I mean the `complete: true` will actually work? I'm suffering this issue in a project in production, having to call the endpoint everytime the user types anything in the input. Thanks!

kinyat:
RC 5 still have this issue :(

juancarlosfarah:
I have the same issue with v1.0.1. Is this still broken?

ryanhomer:
There were major changes in `Async.js` between v1.0.0-beta14 and the next tagged version, v1.0.0-rc.3, in which all references to `complete` were removed. Odd to see such a big change going from beta to RC.

cclarks:
[...]

PR #1629, opened 2017-03-28
Fix multi Async option issues

### LINKED ISSUE #1542: Async Multiselect, options dropdown hides every time an option selected

w3-3w:
The first two commits of this PR fixes #1542 and #1561 respectively.
The 3rd ~ 5th commit resolves an inconsistency problem as described in commit message.

As in #1289 , clearing options after selection for multi Async has been a default behavior. I introduced a new prop `clearOptionsOnSelection` , which defaults to true. For those who still need to keep options on selection, set it to `false` and it will behave as it did in v1.0.0-rc.2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,6 +358,7 @@ function onInputKeyDown(event) {
 	className 	|	string	|	undefined	|	className for the outer element
 	clearable 	|	bool	|	true		|	should it be possible to reset value
 	clearAllText 	|	string	|	'Clear all'	|	title for the "clear" control when `multi` is true
+	clearOptionsOnSelection | bool | true | clears options after selecting an option for Async component when `multi` is true
 	clearRenderer | func | undefined | Renders a custom clear to be shown in the right-hand side of the select when clearable true: `clearRenderer()`
 	clearValueText 	|	string	|	'Clear value'	|	title for the "clear" control
 	resetValue 	|	any	|	null	|	value to use when you clear the control
--- a/examples/src/components/GithubUsers.js
+++ b/examples/src/components/GithubUsers.js
@@ -22,7 +22,7 @@ const GithubUsers = React.createClass({
 	switchToMulti () {
 		this.setState({
 			multi: true,
-			value: [this.state.value],
+			value: this.state.value ? [this.state.value] : null
 		});
 	},
 	switchToSingle () {
--- a/src/Async.js
+++ b/src/Async.js
@@ -31,6 +31,8 @@ const propTypes = {
 	]),
 	onInputChange: React.PropTypes.func,             // optional for keeping track of what is being typed
 	value: React.PropTypes.any,                      // initial field value
+	onBlur: React.PropTypes.func,                    // onBlur handler: function (event) {}
+	onBlurResetsInput: React.PropTypes.bool,         // whether input is cleared on blur
 };
 
 const defaultCache = {};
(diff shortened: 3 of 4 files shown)


gazpachu:
I've tested this PR and even if I set `clearOptionsOnSelection={false}`, the options in the select get removed when I start a new search

w3-3w:
@gazpachu By `start a new search` did you mean clicking somewhere outside the component and focusing on it again? Maybe the third commit is questionable, but I think it is better to clear options on blur, because the keyword typed for search will be cleared on blur, or else it creates inconsistency on a second focus.

gazpachu:
@w3-3w the options on a multi select shouldn't be cleared ever. Only when the user clicks on the 'x' icon. ATM, in rc2 and rc3, the options are gone whenever a new search starts. The only become visible again if the search starts with the same letter as the options that were previously selected. For this reason, I had to disable the search funtionality.

I'm not sure if this is exactly the same issue you guys are talking about in this thread.

gazpachu:
No need to click outside the select, the issue happens just by starting a new search after having added a few options on the async multi select.

gazpachu:
@w3-3w Nevermind my previous messages. I've found the issue. It turns out that I was reloading all the options because of the `input` parameter was changing, so I stopped passing the `input` parameter to the API endpoint and now the search works fine. It doesn't remove any option.

hsingh23:
Does this fix the complete issue here: #1514?

w3-3w:
@hsingh23 No. This PR focuses on option display behaviors.

amit034:
any progress?

picolll:
Can someone merge this PR please. I'm waiting for it for a long time now.

ndbroadbent:
This would be awesome, thanks @w3-3w! It would be great if this could be merged (but there's 
a merge conflict now)

ndbroadbent:
@w3-3w i'm using your branch, and the `clearOptionsOnSelection` option doesn't seem to do anything for me. The options are hidden as soon as I select one.

ndbroadbent:
I've tried downgrading to `1.0.0-rc.2`, but then there's a serious bug when using multiple Selects. They all seem to share the same options instead of loading them individually.

begincalendar:
@w3-3w Thank you for creating this PR.

I have some constructive feedback for you:

- I don't think you should conflate two issues with this (or any PR). As far as I can tell you are trying to fix #1542 and #1561 (which are the same issue, so that's fine), but then you also appear to be adding functionality to clear options when the field is blurred. I think the latter part should be stripped out into another PR. What do you think?
- You appear to be doing some refactoring and bug fixing, as well as adding functionality. Related to the previous point, I think refactoring, bug fixing and feature addition should be split into separate PRs. Yes it takes more work on the part of the contributor, but it makes life easier for reviewers of PRs and also for future maintainers/contributors reviewing (master branch) squashed merge commit history. Would you agree?

JedWatson:
Thanks for the PR @w3-3w 

I believe all of the fixes you've included here have been implemented separately (and slightly differently, see the changelog) but this has been a great reference.

For what it's worth in the future, separate PRs for each feature will make it easier to merge 🙂

PR #1646, opened 2017-04-04
gives possibility to use ref property for Creatable

blacktemplar:
fixes issue #1644


gwyneplaine:
thanks for this @blacktemplar  this has been merged into master.

aaronincincy:
This appears to be causing warnings for me (react-select@1.1.0, react@15.6.2).  I'm using ref={r=>this.selector = r} and getting:

Warning: CreatableSelect: `ref` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop. (https://fb.me/react-special-props)

PR #1652, opened 2017-04-06
Include a className of 'is-clearable' if clearable is true

dan-diaz:
mimicking how we add other classNames like "is-searchable" and "is-disabled"


JedWatson:
Looks good to me

PR #1670, opened 2017-04-14
Duplicates/Index Fields for Select

cglane:
Added two props to the Select element
1)allowDuplicates
This prevents filtering out selected values from the _visibleOptions array
2)trackByIndex
Adds index prop to the Value element
Removes element based on its index and not the 'value' of the Value element


agirton:
Hi @cglane this PR includes our build files. It also has drag and drop so it's doing more than your original description. Please clean up PR so its easier to review.

JedWatson:
Thanks for the PR @cglane 

I'm going to close it for now because the use-case for the new props isn't clear, and the commits include a lot of conflicts and build artefacts, as well as irrelevant history.

But I'm open to discussing both props individually if you still think they are important to support - in the future please make sure to keep limit PRs to a single feature at a time and only include changes to `src`.

PR #1673, opened 2017-04-19
Drag drop2

cglane:


agirton:
Hi @cglane thank you for the PR. We are going a different direction for drag and drop.

Issue #948, opened 2016-05-09
Make Searchable Act More Like Native HTML Input


gojohnnygo:
I was wondering if there were design decisions on why searchable select does _not_ act more like native HTML inputs.

For example:
1.) `ValueComponent` appears to be HTML input value, but it is not. 
2.) When you click the `ValueComponent`, you expect the cursor to appear where you clicked in the text.

I have [this PR](https://github.com/JedWatson/react-select/pull/946), which I need to fix (dropped .2 in coverage) and some commits ready, but before I continue: 
1.) Is this something worthwhile for the community? 
2.) Was this not explored for some reason?

JedWatson:
I think these were just oversights, @gojohnnygo. The event cancelling thing is clever, while the input value clear seems like a bit of a hack? Either way, I've merged your changes - would be very happy with any other improvements you can make!

gojohnnygo:
Can you take a look [this comparison](https://github.com/JedWatson/react-select/compare/master...gojohnnygo:master) and give me your thoughts at my first stab?

**Summary**
1. Changed behavior for searchable single select only.
2. `select` has a value >  `ValueComponent` clicked (the empty space) > the `input` field will have the text of the `ValueComponent`'s label > cursor will be at the end of the text. 
3. `select` has a value > `ValueComponent`'s clicked (the value label) > the `input` field will have the text of the `ValueComponent`'s label > text will be highlighted (image below). This is similar to the Facebook's search bar behavior. 

![screen shot 2016-05-11 at 5 59 38 pm](https://cloud.githubusercontent.com/assets/197015/15201435/cf5bdc4e-17a5-11e6-8856-dad38322fa46.png)

The whole input value is selected (vs putting the cursor where you clicked) bc I'm not sure if it's worth the trouble. It seems like it can be done but won't elegant.

With the changes above, there is one side effect to the options. When you click on the the component, the drop down activates as usual but with only one option: the value the `select` already has. Feature or bug?  ¯\_(ツ)_/¯

TODO: When component is focused, the cursor should be at the end of the value.

ericchernuka:
Just wondering if this is still being actively pursued?  I'm oddly enough looking to build something that fits exactly this implementation.

travisbloom:
@gojohnnygo @JedWatson  did this ever get implemented? If not, any code you can point me to so I can work on this?

gojohnnygo:
[...]

PR #1674, opened 2017-04-20
Make Single Select Act More Natively

gojohnnygo:
## Overview
This PR makes single select act more natively.

## Issue
https://github.com/JedWatson/react-select/issues/948#issuecomment-289309520

## Considerations
1. Because core functionality has changed, there are some breaking tests. I didn't want to take time to try to fix them if this new functionality was rejected.
2. There was a CSS fix to correct a pixel jump when switching from input to value label and vice versa. I tested on the latest Mac Safari, Chrome, and FF. How far back do we need to test? What about IE?

#### New
---
new-focus
#### Old
---
old-focus

#### New
---
new-click-inside

#### Old
---
old-click-inside

#### New
---
new-click-outside

#### Old
---
old-click-outside


gojohnnygo:
Any feedback on this? :)

Like I said, there are breaking tests bc things have are fundamentally changed. I don't want to try to fix them if people don't agree with the changes.

src/Select.js
```diff
@@ -339,31 +340,38 @@ const Select = React.createClass({
 			});
 		}
 
+		const inputLabel = this.getInputLabel();
+
+		// determines if the user mouse downed on the actual ValueComponent excluding placeholder text
+		const isOnValue = this.valueComponent && (this.valueComponent.valueNode === event.target) ? true : false;
```
agirton: This probably could be simplified to just be `this.valueComponent && (this.valueComponent.valueNode === event.target)`

src/Select.js
```diff
@@ -339,31 +340,38 @@ const Select = React.createClass({
...
+
+		// clears value so that the cursor will be a the end of input then the component re-renders
+		this.input.getInput().value = '';
```
agirton: Just like the check below we should probably check `getInput` is a function before we access it.

src/Select.js
```diff
@@ -339,31 +340,38 @@ const Select = React.createClass({
...
 				isOpen: true,
 				isPseudoFocused: false,
+				inputValue: inputLabel
+			}, () => {
+				isOnValue && this.input.select();
 			});
 		} else {
 			// otherwise, focus the input and open the menu
-			this._openAfterFocus = this.props.openOnFocus;
+			this._openAfterFocus = true;
```
agirton: Why was this removed?

src/Select.js
```diff
@@ -165,6 +165,7 @@ const Select = React.createClass({
 
 	getInitialState () {
 		return {
+			hasInputChange: false,
```
agirton: Nit: to follow other naming schemes `hasInputChanged` would probably be better.

src/Select.js
```diff
@@ -415,13 +425,20 @@ const Select = React.createClass({
...
-		var isOpen = this.state.isOpen || this._openAfterFocus || this.props.openOnFocus;
+		const isOpen = this.state.isOpen || this._openAfterFocus || this.props.openOnFocus;
 		if (this.props.onFocus) {
 			this.props.onFocus(event);
 		}
+
+		// To make the react-select act more natively, render the input value with the corresponding label on focus. However, we 
+		// cannot do this for custom valueRenderer's bc the label can be manipulated during render (e.g. option.label.toUpperCase())
+		// so the option labels could be inconsistent.
+		let inputLabel = this.getInputLabel();
```
agirton: It doesn't look like this is reassigned or used elsewhere. Could you just call this when update the state on line 441?

agirton: Also it seems like users with valueRenderer should be able to benefit from this. Maybe we can warn users against manipulating the value?

scss/control.scss
```diff
@@ -147,14 +147,22 @@
...
+		padding: -1 (($select-input-internal-height - 14) / 2 - 1) 0 (($select-input-internal-height - 14) / 2);  /* For IE 8 compatibility */
 		-webkit-appearance: none;
 
 		.is-focused & {
 			cursor: text;
 		}
 	}
+}
 
+// in FF, the input field jumps a pixel when enabled, this adjust for that so the label and input are perfectly aligned
```
agirton: Using vendor specific hacks should not be used. If this selector ever goes away then the original issue may reappear. To ensure they are perfectly aligned I think we should spend a bit more time digging into why they aren't instead of adding a band aid :smile:

gojohnnygo: @agirton I don't disagree with you, but: https://giphy.com/gifs/13XW2MJE0XCoM0

I tried to find a fix that's not vendor specific before your comment but didn't get anywhere. I would welcome any assist from the community. :D

src/Select.js
gojohnnygo: Good point.

src/Select.js
gojohnnygo: @agirton Yeah, I think a warn makes sense. I removed `valueRender` from the `if`, but depending on what the users' func returns, it could blow up.

JedWatson:
Thanks for this @gojohnnygo 

I love the effort you've gone to with the PR, but am not going to merge as-is because it's a major behaviour change for react-select. The ideas are good, so my next goal with this project is to work out how to make react-select more flexible (and have less props) and it would be great to revisit how this could be made an option once that's in place.

bluespore:
@JedWatson Any update on whether the above can be achieved through configuration now? These changes are exactly what I'm looking for and I'm keen to understand how to integrate.

Issue #1602, opened 2017-03-14
Better positioning system


FezVrasta:
Right now react-select positions the dropdowns right below the originating input element.
This is problematic in situations when there's not enough space on the bottom of the page or of its scrollable container because the page will be enlarged to make the dropdown fit.

If you used a positioning engine like [`react-popper`](https://github.com/souporserious/react-popper) the dropdown could be positioned where there's more available space. The library is very lightweight and supports all your use cases.

blakedietz:
As I'm currently gathering requirements for a project this was one of the features that we were looking for. This seems like a nice feature add.

Grsmto:
FYI in v2 you can now do `menuPlacement="auto"`. Not sure why this isn't default tho.

jossmac:
@gwyneplaine might be time to make the move, perhaps in our next major.

CC @JedWatson

ebonow:
The prop `menuPlacement` has been created since the creation of this issue. While the merits of @jossmac's suggestion doesn't go unnoticed, the move to react-hooks is a much larger undertaking and there are other existing issues to address menu positioning concerns (ie: available space defined from "bottom of window" instead of "bottom of ancestor node with no overflow"). As such, I will be marking this as closed.

PR #1677, opened 2017-04-22
feat: Implement menu auto flipping

FezVrasta:
This PR aims to implement #1602 and #284.
Supersedes #1232, and is a foundation for #810, #697, #637.

Basically, it adds `react-popper` to the equation, making it possible for the dropdowns to flip automatically in case there's not enough space on the bottom.

`react-popper` is based on Popper.js, which is the same library [being integrated in Bootstrap v4](https://github.com/twbs/bootstrap/pull/22444) to replace Tether.

Right now there's a bug with single options dropdown, that, when filled with an option, are misplaced.
I'm investigating the issue but I wanted to gather feedbacks to see if it is worth it or you don't plan to merge this.

Thank you!

### DIFF
--- a/less/control.less
+++ b/less/control.less
@@ -76,7 +76,6 @@
 }
 
 .is-open > .Select-control {
-	.border-bottom-radius( 0 );
 	background: @select-input-bg;
 	border-color: darken(@select-input-border-color, 10%) @select-input-border-color lighten(@select-input-border-color, 5%);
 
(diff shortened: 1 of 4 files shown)


agirton:
Hi @FezVrasta thank you for the PR and great work on Popper! I'm a bit concerned that it basically takes over the package. However, I know this feature is important to many of our users. This would be good to have after a 1.0 release.

alexmattson:
What is the status on this?

juanpujol:
Is there any plans to merge this?

JedWatson:
Hey @FezVrasta thanks for the PR - I've used popper before and it's a great solution for this.

However, I don't want to build it directly into react-select, as it's not always necessary and I'm trying to slim the core package down... it should be possible to do this using a custom `menuRenderer` component, or if not let's focus on making our API more extensible so you could.

Would be great to get an example added (if it is possible), in the same way @bvaughn has done with `react-select-virtialised` to demonstrate how to use it.

Issue #1615, opened 2017-03-23
option for manually closing dropdown menu when onBlur


bryan-foong:
Hi, thanks so much for this awesome component!

Is it possible to have a function to allow user manually closing the dropdown instead of closing it automatically when blurring? Because what I would like to do is to fire an action before the dropdown closes.

something like additional prop :

```onBlurCloseMenu={false}```

Then we can do:

```
onBlur={() => {
    setTimeout(() => {
        this.setState({
            dropdownVisible: false
        });
    }, 1000);
}}
```

green-arrow:
Bumping this. Being able to manually control when the dropdown menu closes would definitely be useful for many things, such as custom animations. Right now, animating on open is easy with the `onOpen` event, but animating out is not possible.

Fwiw, my specific use case is showing the dropdown menu fixed to the bottom of the screen on mobile devices with a semi-transparent overlay on the rest of the page. Animating in is not an issue, but animating out with a slide down would need something like @bryan-foong suggests.

If you're interested in having this functionality, I can take a look at the code and see if I can get a PR going.

vsc-github:
BUMP! There could be a function that can be called to manually close the dropdown whenever/wherever required.

sag1v:
would be nice for a "always-opened" feature

sergiubologa:
Any news regarding this option?

JohnReagan:
Have another use case for this. I'm using react-select to pick time frames, with a custom Option component to allow custom date selection, see image below: 
![image](https://user-images.githubusercontent.com/6519232/71635324-819ec480-2bf1-11ea-971c-fb26b9cac39a.png)
I'd like to focus the datepicker (using react-datepicker) to allow keyboard input; however, this triggers a blur on the input and auto closes the menu. An optional prop to prevent closing on blur would work great here. Could check that prop to optionally call onMenuClose in the onInputBlur function here: https://github.com/JedWatson/react-select/blob/292bad3298f2cafad6767f2134bd79a9c27e4073/src/Select.js#L1108

dluk:
Is there any workaround for this issue?

ebonow:
Greetings,

This functionality and the PR associated with this was resolved in version 2 with the addition of the prop `isMenuOpen` which can be used to programmatically decide whether the menu should be open or closed.

To address the initial question as to how to accomplish such a thing in v2+, [here is a codesandbox](https://codesandbox.io/s/react-select-delayed-close-onblur-3yp76)


[...]

PR #1691, opened 2017-04-27
Add autoClose and requestClose props

green-arrow:
When autoClose is set to false, the requestClose prop function will be invoked when the dropdown needs to close. The consumer will be responsible for calling 'close()' on the component instance to close the dropdown.

Addresses #1615


green-arrow:
@JedWatson - I know you're super busy, just wanted to bump this now that the merge conflict has been resolved and possibly get your thoughts on it.

green-arrow:
@JedWatson - I'm going to leave this open, but not resolve the merge conflicts until you let me know if this is something you'd like to put in the library. I know there are a lot of changes going on right now, so just let me know if / when you want this feature.

Thanks for all your work on this library 😄

PR #1658, opened 2017-04-10
Migrate to prop-types and create-react-class packages

agirton:
In React 15.5 they are starting the process to migrate away from having `PropTypes` and `createClass` under the main `React` object. This PR updates all of the related files needed to use the new packages.

I ran all of the unit tests and also tested the examples app after the changes. Both had no errors/issues.

PR #1702, opened 2017-05-04
Refactor: migrate React components to ES6 classes.

rgrocha:
Also use new React (as per v15.5.0) recommendation to migrate from React.PropTypes to 'prop-types' package.

No new feature introduced.

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -1,51 +1,11 @@
-import React, { Component, PropTypes } from 'react';
+import React from 'react';
+import PropTypes from 'prop-types';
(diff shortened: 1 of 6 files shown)


laytong:
Duplicate of https://github.com/JedWatson/react-select/pull/1658?

jochenberger:
IMHO, this PR should have been merged instead of #1717 as it doesn't add a dependency on `create-react-class`.

agirton:
Thank you for the PR. Merged in another PR with ES2015 changes.

PR #1703, opened 2017-05-05
:arrow_up: fix proptypes / react version

guillaumecrespel:


cheeaun:
Duplicate of #1658 and for some reason, all files renamed to `kriya-select` 😅

PR #1705, opened 2017-05-05
Add prop to control whether to filter options in multi mode

ignatif:
Hi. I've met a case where i didn't want options to disappear after i select them in Multiselector. So I made a new prop called `filterOptionsMulti`. Default value is true, README.md is updated


ignatif:
My checks have failed because of `Error: Cannot find module 'prop-types'`. I guess code in master is not valid

gwyneplaine:
@ignatif  thanks for this, closing this PR in favor of https://github.com/JedWatson/react-select/pull/1891, which is close to making it into master.

PR #1711, opened 2017-05-08
Add placeholderOnOpen prop to view placeholder onOpen

athomann:
Add placeholderOnOpen prop to view placeholder onOpen.  Some users didn't realize the dropdown was a typeahead because the value was still showing.  This PR adds an option to be able to show the placeholder when `isOpen` to allow instructions to be shown.


JedWatson:
Thanks for the PR @athomann!

Could you resolve the conflicts please? you don't need `React.` now, we've upgraded to the `prop-types` package.

Also, let's call this prop `displayPlaceholderWhenOpen` for clarity

athomann:
Done!

Drecula:
I monkey patched this to handle the negative case

`	if (!valueArray.length || this.props.displayPlaceholderWhenOpen && isOpen) {
			if(!this.state.inputValue && !this.props.displayPlaceholderWhenOpen && isOpen) return null
			return !this.state.inputValue ? _react2['default'].createElement(
				'div',
				{ className: 'Select-placeholder' },
				this.props.placeholder
			) : null;
		}`

TallOrderDev:
I don't think this is documented on the ReadMe

TallOrderDev:
@athomann @JedWatson ^^^^

carlosbensant:
**Floating Label:**
I'd like to create a Floating Label theme of this package (just using CSS), but it expects the placeholder to persist along with the current value.

The **displayPlaceholderWhenOpen** prop is an outstanding option for what I'm expecting to see.

When are you expecting to merge this?

JedWatson:
@carlosbensant this change still won't actually do what you want, because the placeholder will still disappear when the user types in the input

@athomann sorry for dropping this one, got overwhelmed by issues and PRs for a while. I'm going to release 1.0 final and then revisit how configurable this is to cut down on the crazy number of props we support.

carlosbensant:
@JedWatson Thanks for taking your time to reply. 

I've tried to explain my concern in a much cleaner way. Could I ask you kindly to check this issue #2152 I've opened recently.

tiffwu:
@JedWatson This would be huge for my team as well. I'm fine with the placeholder disappearing when the user starts typing.

jayaranigarg:
@JedWatson : Can we take this PR in? It will be of huge help!!

Rall3n:
For a solution to keep the placeholder around (with current version) check out [this comment](https://github.com/JedWatson/react-select/issues/2152#issuecomment-470889515).

PR #1716, opened 2017-05-10
fixed drop down list position of display.

Lenic:
1. fixed drop down list position of display.
2. add prop-types package.

Signed-off-by: leixuewei <leixuewei@neutrey.com>


adnasa:
8K PR 👀

JedWatson:
@Lenic thanks for the PR, would you mind resubmitting with just the relevant parts in `src` changed please? the build artefacts shouldn't be included, and please don't reformat code you're not actually changing. Also if you rebase, you'll get the update to use `prop-types` that has been merged.

Lenic:
Thank you for your warning. This is my first pull request, and I'll take care of it later.

agirton:
Hi @Lenic it looks like you are setting the style to fixed on the outer menu. What was the primary goal for this? I'm not sure if this is the behavior all of our users want.

gwyneplaine:
Thanks for the effort @Lenic, we've updated to prop-types in a separate PR, and as @agirton has mentioned, I'm not entirely sure that a fixed position navigation is something we want to enforce for all of our users. 

Happy to revisit the idea in a separate PR (without bundled dist), am closing this in the meantime.

PR #1728, opened 2017-05-17
[fix] Make getFocusableOptionIndex use valueKey

Craga89:
Currently the `getFocusableOptionIndex` method assumes that the value for a particular option is stored under the `value` key, rather than using the `valueKey` prop like elsewhere. This causes the up/down arrows and option hovering to not work


agirton:
Hi @Craga89 thank you for your contribution, we merged this fix in #1733.

PR #1731, opened 2017-05-18
Always pass valueArray to menuRenderer

chronick:
My application's designers wish to display a visual indicator for selected items instead of hiding them from the dropdown list when using a multi selector. Always passing the `valueArray` to the menu renderer seemed to be the best and least-impact way to do so.

The default behavior of the select box appears to be unaffected by this change, since the actual hiding of options happens in the default `filterOptions` function.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -1121,7 +1121,7 @@ const Select = createClass({
 					{this.renderClear()}
 					{this.renderArrow()}
 				</div>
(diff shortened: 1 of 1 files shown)


MrOrz:
Just encountered this and the fork worked like a charm.

Conditionally pass in valueArray does not make any sense to me either.

gwyneplaine:
@chronick  thanks for this, this particular fix has been pulled into master as part of https://github.com/JedWatson/react-select/pull/1850

PR #1732, opened 2017-05-18
Add inputValue to menuRenderer

chronick:
We wish to override the `optionRenderer` to highlight each search result with the main inputs' query. This means that the option at some level will need access to the input's value. I originally attempted to append the input value to the data by overriding `filterOptions`, but that broke filtering and selection. This seemed to be the most reasonable way to do it, and does not affect default behavior.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -968,6 +968,7 @@ const Select = createClass({
 			return this.props.menuRenderer({
 				focusedOption,
 				focusOption: this.focusOption,
+				inputValue: this.state.inputValue,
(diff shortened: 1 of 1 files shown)


agirton:
Hi @chronick thank you for your contribution. While this would be helpful and is a very small change, I want to try and get only critical features for the 1.0 release. We can add merge this after release.

gwyneplaine:
@chronick, thanks for this, merging this in now as its value added and has zero impact on default behaviour.

Issue #1513, opened 2017-01-30
Using with redux-form not focus active selected option


rndm2:
When using with redux-form 

```
  renderSelect = (props) => {
    const { input, options, disabled, onFormChange } = props;

    return (
      <Select
        className={`custom-select custom-select-${input.name}`}
        clearable={false}
        name={input.name}
        value={input.value}
        options={options}
        disabled={disabled}
        onChange={(value) => {
          input.onChange(value);
          onFormChange({ target: { name: input.name, value: value.value } });
        }}
        onBlur={() => input.onBlur(input.value)}
      />
    );
  }
```

After selecting any value and then opening select again focused option will be always first in list because in getFocusableOptionIndex focusedOption searched across options array which contains objects, and they are not always equal with focusedOption (after @@redux-form/CHANGE for example). 

I have created PR on this, https://github.com/JedWatson/react-select/pull/1512
But maybe there is another solution?

Betree:
PR #1512 introduced a new bug : it uses `option.value` to compare objects, thus breaking focus for objects with a different value key (in my case the value key is `id`).

I've made a new PR that uses the `valueKey` property to compare objects.

agirton:
Closing this as it has been fixed in #1733

PR #1733, opened 2017-05-18
Fix selected option focus when valueKey is not "value"

Betree:
Related to #1513


agirton:
Hi @Betree thank you for the contribution! However it looks like you updated the lib file instead of the src file. Do you mind updating?

Betree:
My bad, I'll fix this right now

agirton:
Thank you!

PR #1739, opened 2017-05-20
Update README.md

hegelstad:
Added note about bower installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,13 @@ The easiest way to use React-Select is to install it from NPM and include it in 
 npm install react-select --save
 ```
 
+React-Select is also available on Bower, but you have to specify version number to get the latest version:
+
+```javascript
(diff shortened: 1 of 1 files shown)


agirton:
Thanks @hegelstad. We hope to release a new version soon, so that version will be out of date. Also I just noticed bower is now advising users to use yarn/NPM and weback on their homepage. Wondering if we should still support it 🤔.

hegelstad:
Rather just remove bower support.

JedWatson:
I'm definitely removing bower support after 1.0.0 (don't want to remove it before then) but in the meantime also don't want to promote it at all...

If anybody's wondering why, there are better tools now, and great cdns like [unpkg(https://unpkg.com/) and supporting bower means checking in our builds which is a nightmare.

Thanks for the PR anyway @hegelstad

PR #1741, opened 2017-05-22
break dependency cycle between Select.js and Async.js

### LINKED ISSUE #1517: Circular dependency issues with Webpack 2 and react-select 1.0.0-rc.2
Maybe this is due to specific configuration issues, but in case anyone else has the same problem -- after updating to Webpack 2 (2.1.1) and react-select (1.0.0-rc.2), our usages of Select.Async began breaking all over the place with the following error:

> Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: object. You likely forgot to export your component from the file it's defined in. Check the render method of `Async`.

After banging my head against the wall for a couple of hours, I've figured out this is related to circular dependencies in react-select -- Select requires Async and Async requires Select. In the following code (from src/Async.js):

```
import React, { Component, PropTypes } from 'react';
import Select from './Select';
import stripDiacritics from './utils/stripDiacritics;

console.log(Select);
```

The log will return an empty object. If, instead, I require Select inside the `defaultChildren` function (for debugging purposes), it will log the Select component correctly.

I'm a bit torn on where to report this issue -- this is related to the changes on Webpack 2, but I reckon circular dependencies will always be somewhat error-prone.

The quick fix I've found was to always pass the following as a prop to Select.Async:

```
children={function (props) {
    return <Select {...props} />;
}}
```

But this obviously is a hack-ish solution, and I would prefer to find a better one. I'd appreciate any input you could provide on the issue!

jochenberger:
Add a separate entry module to avoid a circular dependency between `Select.js` and `Async.js`.
fixes #1517

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "react-select",
   "version": "1.0.0-rc.4",
   "description": "A Select control built with and for ReactJS",
-  "main": "lib/Select.js",
+  "main": "lib/index.js",
   "style": "dist/react-select.min.css",
   "author": "Jed Watson",
   "license": "MIT",
--- a/src/AsyncCreatable.js
+++ b/src/AsyncCreatable.js
@@ -1,6 +1,8 @@
 import React from 'react';
 import createClass from 'create-react-class';
 import Select from './Select';
+import Async from './Async';
+import Creatable from './Creatable';
 
 function reduce(obj, props = {}){
   return Object.keys(obj)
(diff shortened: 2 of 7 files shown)


jochenberger:
See also https://github.com/rollup/rollup-plugin-commonjs/issues/105

agirton:
Hi @jochenberger thank you for this. This seems like the right way to go, just not sure if we want to bring this in before 1.0. @JedWatson open to your thoughts on this one.

jochenberger:
Closing in favor of #1855

Issue #1316, opened 2016-10-20
autosize=false blocks the opening of the menu after being focused


geraldodev:
Hi,

When autosize=false the code on handleMouseDown blocks the menu of being opened for the second time after getting the focus. You will not be able to open it by clicking. When you focus anywhere else and click back the menu opens again. This is does not affect  AutosizeInput.

```
        if (event.target.tagName === 'INPUT') {
            return;
        }
```

regards,
geraldodev

dave-clover:
I'm also having trouble with this line of code.  It's breaking our change that allows you to right click and paste into the select component.

@gojohnnygo was the original author of that code.  Would you mind elaborating on what issue it is fixing?

gojohnnygo:
@dave-clover can you link me the line in question?

dave-clover:
Hey @gojohnnygo, thanks for the quick response.  This line:

https://github.com/JedWatson/react-select/blob/master/src/Select.js#L322

We're probably going to have to make a change to fix the original issue in this thread, but we don't want to break the thing you were fixing.

gojohnnygo:
Hey @dave-clover,

That fixed a minor UI quirk. Here's the PR describing what it does. https://github.com/JedWatson/react-select/pull/946

cbergmiller:
I have submitted a PR that should solve this issue: #1746

JedWatson:
I've been able to reproduce this and resolved it differently to #1887, see the referenced commit for more information

PR #1746, opened 2017-05-23
[FIX] #1316 enable opening if autosize=false and click on input

cbergmiller:
This is a proposed fix for issue #1316 
The menu should open when a user clicks on the input after a selection has been made and the input is still in focus.
Aborting handleMouseDown when clicked on the input element only when `autosize=false` will fix the issue.
Downside is that it's not possible to change the location of the cursor by clicking when entering a search string.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -318,7 +318,7 @@ const Select = createClass({
 			return;
 		}
 
-		if (event.target.tagName === 'INPUT' && this.props.autosize) {
+		if (event.target.tagName === 'INPUT') {
 			return;
 		}
 
--- a/src/Select.js
+++ b/src/Select.js
@@ -1120,8 +1120,8 @@ const Select = createClass({
 					onTouchMove={this.handleTouchMove}
 				>
 					<span className="Select-multi-value-wrapper" id={this._instancePrefix + '-value'}>
-						{this.renderValue(valueArray, isOpen)}
 						{this.renderInput(valueArray, focusedOptionIndex)}
+						{this.renderValue(valueArray, isOpen)}
 					</span>
 					{removeMessage}
 					{this.renderLoading()}
--- a/src/Select.js
+++ b/src/Select.js
@@ -318,7 +318,7 @@ const Select = createClass({
 			return;
 		}
 
-		if (event.target.tagName === 'INPUT') {
+		if (event.target.tagName === 'INPUT' && this.props.autosize) {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


agirton:
Hi @cbergmiller  thank you for the contribution! However, this feels like a mask to the problem and not the right solution. Mind taking another go at it? Tests would be helpful 😄.

dave-clover:
This comment doesn't add much, but we've had a lot of trouble with this if check: ```

if (event.target.tagName === 'INPUT') {
	return;
}
```

https://github.com/JedWatson/react-select/issues/1316

It messes with the right click and paste hack we have.

cbergmiller:
I found another fix for #1316
Changing the order of `Select-value` and `Select-input` changes the event target of the mouseDown event to `Select-value` which passes the `event.target.tagName === 'INPUT'` condition.
Works fine in the browser (chrome/FF/edge) but may be a riskier change.

I also added tests, but i'm having trouble to get the options menu to open in the test when `autosize=false` (even when the input is not focues yet). Any help?

cbergmiller:
I spent several hours trying to write tests that cover #1316.
In my opinion this is not possible with the current testing framework because:
- jsdom does not implement the window.focus() method
- findAndFocusInputControl() from Select-test does not work for `autosize=false`
- #1316 fails because `Select-value` and `Select-input` are layered in case of `autosize=false`. Simulating events on specific dom nodes with `TestUtils.Simulate` cannot reproduce the problem and prevent regression.
Since migrating to another testing framework will likely not be an option (phantomjs?), i would propose to either accept one of the two fixes i suggested, or to remove the `event.target.tagName === 'INPUT'` from PR #946. Functionality is (slightly) broken for `autosize=false` with `searchable=true` as is.

agirton:
Hi @cbergmiller thank you for taking the time to dive into this a bit further. This extra check seems to have caused a lot more issues, wondering if we should just remove it. #946 was just to fix a minor UI quirk, but if it causes a lot more issues with using the component, I'm not sure if it's worth it. Let me think about this a bit more.

ls-guillaume-lambert:
So this won't get merged to fix this issue? Any workarounds in the mean time?

JedWatson:
I can't reproduce this issue in the examples, so it may have been resolved by other changes in 1.0.0

Will comment on the issue, I'm not confident about merging this reordering though without also being able to reproduce this problem so I'm going to close this.

Thanks for the PR @cbergmiller

JedWatson:
Actually I dig further into trying to reproduce it, and worked it out. I've pushed some changes that fix it - see #1316 for details

PR #1756, opened 2017-05-25
fix not clearing when given invalid values

ticklemynausea:
when given an invalid value (not in the range of specified possible values), the select component shows the clear select icon despite being empty. this was determined to be caused by renderClear() being incorrectly called.

this PR simplifies the logic that determines when the clear icon is to be rendered.


agirton:
Hi @ticklemynausea thank you for your contribution. It seems there is a lot of churn in this area lately. Do you mind adding a test to show that the x should not show with an invalid value? 

Also per guidelines do not include the build files in your PR. We generate these during deployment.

gwyneplaine:
thanks for this @ticklemynausea, I've pulled out the crux of your PR into https://github.com/JedWatson/react-select/pull/2100/ to be merged into master

Issue #1352, opened 2016-11-03
ignoreCase = true (default behavior) lowercases all user input


caguthrie:
This code in ```Async.js``` causes all user input to be ```toLowerCase```'d:

```
if (ignoreCase) {
  inputValue = inputValue.toLowerCase();
}
```

Why does this exist?  If A user types an uppercase character and a lowercase character is printed, that would be confusing for the user.  I'm happy to submit a PR removing this conditional unless there is a workaround I'm not aware of or if there is a valid reason for this code.

erfangc:
+1

ee08b397:
why

scags9876:
Is this a duplicate of https://github.com/JedWatson/react-select/issues/1106 ?

saviomuc:
@JedWatson Any chance we can get this fixed? There are already some PRs like this one https://github.com/JedWatson/react-select/pull/1329. Honestly I'd prefer to just have ignoreCase not lowercasing the input at all. Happy to provide a PR if you agree.

alexanderteno:
Bump: I'm suffering from this on an Async call, seems if I backup and retype the input it doesn't swap capitalization on a per letter basis. It's very strange. Anyone know a workaround or some sort of soft fix?

BossGrand:
Is there any work or thought being done on this? Is there a workaround?

BorodinDK:
+1

ericnewcomer:
FWIW I'm still seeing this behavior when ignoreCase={true}. I think the fixes proposed previously to not manipulate the user input during async evaluation seems like the right one. For some reason that PR was never merged. Is a new PR welcomed to address this?

agirton:
Hey @ericnewcomer depending on the outcome of #1329, we may need a new PR to just remove all mutation of the `inputValue`

ericnewcomer:
Also, I did confirm that commenting out that single line does seem to do the trick. I'm really unsure why we would want to maintain that behavior in any case. We can muck with a copy of the input all we want (removing diacritics, etc), but we ought not overwrite what the user is typing.

sponrad:
> remove all mutation of the inputValue

That gets my vote for this. Simple fix.

not-an-aardvark:
I created a new PR at https://github.com/JedWatson/react-select/pull/1858.

Issue #1214, opened 2016-09-13
Doesn't recognize some special characters like ö,ä


heaks:
I have Select.Async component with some options (e.g. loadOptions={this._loadOptions}). But it doesn't work well with special characters (ö,ä). It doesn't find an option which contains "ä", no matter what you type "ä" or "a". Any solutions ?

knegusen:
I believe there is a prop for this that is not documented. I searched in the source code and found `ignoreAccents` and when setting that one to false it did not replace special characters with common ones (like å -> a). You can find the map in the file stripDiacritics.js. Don't know if this is the correct solution though.

markoandersson:
I just encountered the same issue. Yes, setting `ignoreAccents` to `false` works. 

It seems that this concerns only `Async` component as I'm using basic `Select` component and special characters are working with it just fine. I'm also using custom `filterOptions`-function with `Select.Async` so maybe it's combination of those things.

thevangelist:
Thanks @markoandersson! You saved my day!

JedWatson:
This is better documented now

Issue #664, opened 2015-12-12
Select.Async replaces german letters


kuuup:
When I type german letters like ä, ö, ü and ß they are replaced by a, o, u and s. This behavoir leads to wrong search results.

StevenLangbroek:
There's a prop called `ignoreAccents` you can set to `false` to negate this behavior. See https://github.com/JedWatson/react-select/blob/master/src/Async.js#L43. @JedWatson want me to add some docs around this?

kuuup:
Works :thumbsup:
Thank you.

I'll leave this one open for Jed.

JedWatson:
That would be great, thanks @StevenLangbroek!

jochenberger:
I wonder if stripping the diacritics is a good default behavior when querying remote data sources.

jochenberger:
IMO, `Async` should have `filterOption={()=>true}`. The filtering should be done by the remote endpoint.

JedWatson:
In hindsight I don't think these props should have defaulted to `{true}`, but I don't want to push a breaking change at the moment to update that. The props are now more clearly called out in the documentation, which should make it easier for users to understand what's going on and how to manage the behaviour.

PR #1762, opened 2017-05-29
don't apply additional filters to the result that is returned from the remote endpoint

jochenberger:
See #1352, #1214, #664


jochenberger:
Not sure what to do with those test. I'd say they should be removed, because they don't make sense anymore.

agirton:
Hi @jochenberger thank you for your contribution. Instead of removing them the default could be false. This way if users still want this behavior they can opt in. This feels like a breaking change though. Let me give this a bit more thought.

agirton:
Hi @jochenberger removing this behavior doesn't seem the right thing to do so can you update the PR to just default to false? We will need to communicate this in the docs as well.

jochenberger:
I updated the PR. Any suggestions on how the docs should be updated? I assume it should be mentioned in the "Async options" section.

JedWatson:
Hey @jochenberger sorry for the runaround on this one - I'm not inclined to change the default behaviour before the next major version, as it's a big breaking change for anyone using the component and expecting the current behaviour.

Going to close this (aiming to get all PRs merged or closed) but will definitely consider less opinionated defaults in the next major version.

Thanks!

Issue #1651, opened 2017-04-06
[Call setState in lifecycle function] Async.js 


zhangzhinan:
When i'm used width inferno, it throw me an error.

According to React Docs said:

https://facebook.github.io/react/docs/react-component.html#componentwillupdate

Note that you cannot call this.setState() here. If you need to update state in response to a prop change, use componentWillReceiveProps() instead.
![image](https://cloud.githubusercontent.com/assets/16774464/24755100/7fa9dd30-1b0b-11e7-9ed8-5fa484e64dbe.png)

agirton:
Hi @zhangzhinan `componentWillReceiveProps` seems like the better approach here. Since it's just syncing the props. Open to PR's to fix this :smile:

jochenberger:
This was fixed in #1651 and should be closed.

PR #1765, opened 2017-05-30
[FIX] #1651 moved option prop sync to componentWillReceiveProps

cbergmiller:
This is a fix for #1651


agirton:
Thank you for the tests. This looks good to me.

Issue #1760, opened 2017-05-26
Why `onInputChange` expects a value to be returned?


roman01la:
Hello!

I didn't expect that `onInputChange` handler would make use of return value to set a new `inputValue` in `Select` component. It happens [here](https://github.com/JedWatson/react-select/blob/9943711e256d61aab937445c1f69502de43f780b/src/Select.js#L449). Essentially it means that `onInputChange` is an *interceptor* rather than *handler*.

What's the motivation behind this behavior? Is it something that can be changed?

Thanks!

cbergmiller:
This allows you to manipulate input that is entered into the input field. See [here](https://github.com/JedWatson/react-select#updating-input-values-with-oninputchange).

You can always return the unchanged new input value if you just want to get the updated value.
Changing this behavior would break downwards compatibility of the API.

roman01la:
Thanks for response! I'd at least add a note somewhere to avoid confusion. `onInputChange` name is usually means a handler function.

agirton:
Hi @roman01la would you like to submit a PR to update the documentation? :smile:

agirton:
Closing as the README has been updated

PR #1773, opened 2017-06-03
[ADD] Clarified the onInputChange prop signature in the docs

cbergmiller:
Updated docs on `onInputChange` as requested in #1760.


agirton:
Looks good! Thanks! Just a few nits then we can merge this in.

PR #1777, opened 2017-06-06
Migrate ReactTestUtils to react-dom package

mcamis:
react-addons-test-utils was deprecated as of version 15.5.0:

https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html

### DIFF
--- a/test/Async-test.js
+++ b/test/Async-test.js
@@ -17,7 +17,7 @@ var expect = unexpected
 
 var React = require('react');
 var ReactDOM = require('react-dom');
-var TestUtils = require('react-addons-test-utils');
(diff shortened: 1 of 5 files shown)


JedWatson:
@agirton it's fine to update this since we're in control of the version used for our tests, the public interface isn't affected at all.

@mcamis I'd prefer to not update `TestUtils => ReactTestUtils` everywhere though, for the sake of avoiding potential conflicts with forks needlessly.

If you could update without that change, it would be great, otherwise I'm happy to port this fix in.

gwyneplaine:
This has been fixed in a separate PR, thanks for this @mcamis

PR #1783, opened 2017-06-09
- Android webview touchstart and touchend triggers and opens and imme…

yocokirk:
…diately closes select and thus we ignore the the touchend event

- If Android and not Chrome, ignore touchend

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -311,10 +311,22 @@ const Select = createClass({
 		this.clearValue(event);
 	},
 
+	disableEventList () {
+		// Android webview touchstart and touchend triggers and opens and
+		// immediately closes select and thus we ignore the the touchend event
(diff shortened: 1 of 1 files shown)


agirton:
Hi @yocokirk thank you for your contribution! I'm a bit concerned with the UA sniffing. Could we just use touchend? I don't have a WebView to test this out on so can't confirm.

JedWatson:
@yocokirk thanks for the PR :)

I'm also pretty concerned that I don't know which versions this is relevant to. Do you have any information about that you can share? And to echo @agirton's comment, is there a simpler (possibly safer) way we could solve this?

agirton:
Closing as we haven't heard back. Feel free to reopen if you would like to discuss further 😊

PR #1789, opened 2017-06-14
Always allow accents and case in the input

sponrad:
It would be nice to always allow accents and case in the Input regardless of if `ignoreAccents` or `ignoreCase` are true, especially when using create mode (since you may want to create a value that has accents or is case-sensitive).

I think they can safely be removed from here for filtering purposes since they are still applied to the input value in `defaultFilterOptions.js`

As discussed in https://github.com/JedWatson/react-select/pull/1329 and https://github.com/JedWatson/react-select/issues/1352.

---
This change is [Reviewable](https://reviewable.io/reviews/jedwatson/react-select/1789)

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -136,15 +136,7 @@ export default class Async extends Component {
 	}
 
 	_onInputChange (inputValue) {
-		const { ignoreAccents, ignoreCase, onInputChange } = this.props;
-
-		if (ignoreAccents) {
-			inputValue = stripDiacritics(inputValue);
-		}
-
-		if (ignoreCase) {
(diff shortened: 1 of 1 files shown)


agirton:
Hi @sponrad I apologize if my comment's were too vague. What I meant was that we wouldn't mutate the original input but create a new instance and manipulate that. This way we can keep the functionality but not impact the users input.

sponrad:
@agirton Ah, I was unaware of lost functionality by removing this, since I saw the login in `defaultFilterOptions.js` also. Re-reading the comments it looks more involved than just deleting these statements. Closing for now.

Issue #575, opened 2015-11-06
Support for CSS namespacing


cosmin-nicula:
Are there any plans to allow namespacing the Sass classes?

For example, define a prefix variable inside Sass and then use it in all the Sass class definitions. So, basically instead of this:

```
.Select-control {
    ...
}
```

...you'll have this:

```
.#{$css-prefix}Select-control {
    ...
}
```

This has further implications inside the .js code, as the classNames are hardcoded (https://github.com/JedWatson/react-select/blob/master/src/Select.js#L459):

```
var className = classNames('Select-input', this.props.inputProps.className);
```

This could be easily mitigated by propagating through the `context` the CSS prefix.

JedWatson:
I get where this is coming from, but I'm really not keen to introduce this level of complexity into the codebase at this stage, and am looking at moving to a css-in-js solution because maintaining two independent css codebases is making the maintenance of react-select more difficult than it should be.

Hopefully if the `Select-` namespace for the control clashes with other components in anyone's apps, there are other ways to work around it.

PR #1798, opened 2017-06-20
create css namespace

agutoli:
Way to specify css prefix `<Select {...props} cssPrefix="my-css-prefix" />`


agutoli:
related issue #575

agirton:
Hi @agutoli thank you for the PR. I'm not sure this is the route we want to go, but it's not something we want before 1.0 release. :smile:

agutoli:
I perfect understand, thanks for your feedback. :)

JedWatson:
Hey @agutoli sorry to decline this PR but I'm uneasy about introducing this complexity into the codebase at this point. To be honest, I really need to simplify things a bit so I and the other contributors have an easier time maintaining this package.

I'm also looking to move to a css-in-js solution in the next major version so that we don't have to maintain two separate stylesheet sources, which has been a real maintenance burden.

Thanks for the PR though - I really appreciate it 👍

agutoli:
Hi @JedWatson, no problem at all. Actually it was like a suggestion and maybe is not a good moment like you said. 

I really like this package/component, but recently I was working for a project that I had some specifics needs, so I have created a package based on `react-select` but with `css-in-js` approach, shadowDOM, css-variables. I think it can be useful for you about this possible refactoring.

Node: This package has not the most of features of `react-select`, it`s like a concept (still alpha).

JedWatson:
@agutoli sounds cool, feel free to drop a link here and I'll take a look!

agutoli:
Ohh... haha sure... 
https://github.com/agutoli/react-styled-select

agutoli:
complementiong: @JedWatsonI did a article about my "motivations" to write this pkg.
Original article: https://medium.com/trainingcenter/react-css-variables-shadowdom-styled-components-576f6c562d50 
Translated one (sorry about my bad english): https://medium.com/@agutoli/react-css-variables-shadowdom-styled-components-en-4a5f69b4b2f3

PR #1799, opened 2017-06-20
Fix Usage Docs Example

rohmanhm:
Before this change, console log in browser will throw
> Selected: [object Object]

After this change console log in browser will throw some nice information.
> Selected: {"value":"three","label":"Three"}


agirton:
Thanks @rohmanhm!

PR #1803, opened 2017-06-20
Update README.md to include 'valueComponent' prop

lstuartfry:
Thanks to some very helpful feedback I received in this repo (and reading through some issues other users were having) I found there was a valueComponent prop that I could utilize that made my life so much easier.  I think it needs to be included in the list of available props in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,7 +400,7 @@ function onInputKeyDown(event) {
 | loadingPlaceholder | string\|node | 'Loading...' | label to prompt for loading search result |
 | tabSelectsValue | bool | true | whether to select the currently focused value when the `[tab]` key is pressed |
 | value | any | undefined | initial field value |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -400,6 +400,7 @@ function onInputKeyDown(event) {
 | loadingPlaceholder | string\|node | 'Loading...' | label to prompt for loading search result |
 | tabSelectsValue | bool | true | whether to select the currently focused value when the `[tab]` key is pressed |
 | value | any | undefined | initial field value |
+| valueComponent | func | undefined | function that returns a user-created component to render in the value field `() => <CustomComponent ... />` |
```
agirton: The default value for a `valueComponent` is a Value component and not undefined. :smile:

This is a nit but to match the wording/format of the other documentation can you update it to say
function which returns a custom way to render/manage the value selected. `<CustomValue />`

lstuartfry:
No problem, changes have been committed.

PR #1830, opened 2017-06-27
Support event default prevention for shouldKeyDownEventCreateNewOption

tonybaroneee:
The `Create` component supports a `shouldKeyDownEventCreateNewOption` prop, which is a useful place for us to determine if we should create an option or not based on certain criteria. More specifically, we have a use case to prevent the creation of an option that doesn't pass certain criteria. Therefore, we were hoping that we could simply call `event.preventDefault()` within this callback when determining that the ENTER key was pressed and the criteria hasn't been met.

Unfortunately, `shouldKeyDownEventCreateNewOption` does not get passed the entire event object for some reason. Rather, it receives a stripped-down version of the event, containing only the `keyCode` attribute. This PR restores the parameter to be the full event object so that we can call `preventDefault` in our override function and thwart the option creation if need be.

Perhaps this event stripping is done for a reason, but it wasn't apparent to us. Thank you!

### DIFF
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -186,7 +186,7 @@ const Creatable = createClass({
 		if (
 			focusedOption &&
 			focusedOption === this._createPlaceholderOption &&
-			shouldKeyDownEventCreateNewOption({ keyCode: event.keyCode })
+			shouldKeyDownEventCreateNewOption(event)
 		) {
 			this.createNewOption();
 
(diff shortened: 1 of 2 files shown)


tonybaroneee:
Hrm, not sure why the coverage percentage would change from this PR.

tonybaroneee:
My apologies, it looks like `isValidNewOption` will do exactly what we want. This PR might still have some merit for another use case, but I'm closing this for now.

PR #1833, opened 2017-06-28
prevent reset of inputValue when hit 'down' key, after close menu

ericfong:
When I switch to another browser tab and come back. Or just closed the option menu.

And hit keyboard 'down' key. It will clear inputValue and I need to type again to search


jossmac:
Cannot reproduce

PR #1837, opened 2017-06-28
Prevent default event from firing on Tab and Enter press

karaggeorge:
When navigating through options, pressing **Tab** should select the option but keep focus on the element. Next **Tab** press, would focus on the next element. 

This behavior works on Creatable only when selecting the "Create new..." option because of this extra line: https://github.com/JedWatson/react-select/blob/master/src/Creatable.js#L194

However, when selecting an existing option, the **Tab** event escapes and focuses the next element.

Fixed in this PR by adding `event.preventDefault()` on Select.js 's `handleKeyDown()`
Same was done for the **Enter** key

### Actual:
- Pressing **Tab** in one of the existing options:
gifs1
You can see that after pressing **Tab**, the checkbox is focused (not desired)

- Pressing **Tab** in the "Create new ..." option:
gifs2
This works properly

### Expected:
- This is what should happen when pressing **Tab** in one of the existing options:
gifs3
The item is selected and the focus remains on the select component


agirton:
Hi @karaggeorge,

Thank you for such a detailed PR and issue! I’m a bit concerned that this may cause an unexpected change for users who are used to jumping to the next focusable item after selecting an item. It seems like the user will need to hit tab twice to do that with this PR is that correct?

karaggeorge:
@agirton,

Yes, they would need to press tab once to select the highlighted option and once again to tab away. 

It seems to me that this is the generally expected behavior. For example, when using an autocomplete feature, and press tab to select an option you expect the tab press to select the autocomplete option but not to enter an extra tab after it:
gifs5

Another example is that if I press enter to select an option in a select component, I expect it to select the option but not submit a form (on enter) that contains the select. 

Of course, if the `tabSelectsValue` prop is set to false, tab should just focus the next item. But I think if it's meant to select an option it should do that without actually losing focus of the element. In the case that this is not the desired behavior, the behavior of Creatable should change since it's inconsistent, where the tab focuses away if you select an existing option but retains focus on the element if you select the "Create new ..." option.

I can see the argument that people might be used to the select behaving this way, so maybe an appropriate prop to trigger the effect or not could work?

lamoglia:
Hi there, @agirton 
As @karaggeorge proposed, a new prop to conditionally enable this new behavior is the only thing left for this PR being merged? If it is so, I could submit a new PR to comply.

karaggeorge:
@lamoglia I can add the prop if needed. My question is, what should the default behavior of the prop be. To me, it seems like the default behavior should be the expected behavior like discussed above, but if the problem is that people that have been using this will now get altered behavior, then maybe the default behavior be the existing one? 

In either case, when I have an answer I can add the prop to this so it can be finalized

lamoglia:
Ok @karaggeorge, it seems reasonable to keep the current behavior as the default.

karaggeorge:
@lamoglia @agirton any proposed name for the prop?

agirton:
Hi @karaggeorge my biggest hesitation is that we would have to add yet _another_ prop to the package. With the growing number of props we have, this package could become very difficult to maintain.

karaggeorge:
@agirton I completely agree. I honestly don't think adding a prop is the _right_ thing to do here. I think the default behavior should be the expected behavior we see everywhere else. I think the problem becomes when we try to stay away from changing a behavior that people are used to (even if that's the wrong one).

I think that if this doesn't get implemented, the Creatable one should at least change to match this since this is inconsistent between the two types of select.

JedWatson:
Thanks @karaggeorge 

I think the right call here is to change the default behaviour; it brings the behaviour of react-select closer inline with native selects, and can still be disabled with the `tabSelectsValue` prop anyway.

Issue #1710, opened 2017-05-08
Changing required programatically doesn't work


NGuldager:
If the required attribute is bound to a variable it sets the required attribute on the input field correctly when changing from `false` to `true` but doesn't remove it again when it is changed from `true` to `false`.

agirton:
HI @NGuldager looks like it's due to this right here: https://github.com/JedWatson/react-select/blob/9943711e256d61aab937445c1f69502de43f780b/src/Select.js#L197. It's just checking for `falsy` values. I don't see any open PRs to fix this. Would you like to fix this?

karaggeorge:
@agirton Here is a PR to fix this issue:
https://github.com/JedWatson/react-select/pull/1839

JedWatson:
Fixed by #1839

PR #1839, opened 2017-06-28
Changing required prop doesn't work

karaggeorge:
Fix for this issue - https://github.com/JedWatson/react-select/issues/1710

On `componentWillMount()`, the `required` prop is set to the right value or it defaults to `false`.
When receiving new props, if the new `required` prop is `true`, it sets the state value correctly, if it is `false`, it  didn't updated the state correctly to `false`

Also adds `required` as an available prop in the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -394,6 +394,7 @@ function onInputKeyDown(event) {
 | optionRenderer | func | undefined | function which returns a custom way to render the options in the menu |
 | options | array | undefined | array of options |
 | placeholder | string\|node | 'Select ...' | field placeholder, displayed when there's no value |
+| required | bool | false | applies HTML5 required attribute when needed |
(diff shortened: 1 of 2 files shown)


agirton:
Hi @karaggeorge thank you for the PR. I'm a bit concerned that this has the potential of being called multiple times because of this being in the componentWillReceiveProps lifecycle. The default behavior is false so this would be called every time props changed. 

It might be better to just check in the first block if the new required prop is different from the old. Thoughts?

karaggeorge:
@agirton Agreed.

It seemed like the `setState` has to be called every time `required` is true even if the prop was true before as well. I added the extra check to only set the `required` on the state if the prop changed from being `true` to `false`

JedWatson:
Looks good, thanks @karaggeorge!

PR #1848, opened 2017-07-02
ES-2015

lagun4ik:


agirton:
Hi @lagun4ik thank you for the PR, however there are already PR's that convert most of the code base to ES6 classes.

PR #1823, opened 2017-06-26
convert Select to ES6 class

jochenberger:

PR #1734, opened 2017-05-18
remove create-react-class dependency from main sources

jochenberger:
Basically a rebase of #1702 onto master
see also #1658, #1727

PR #1849, opened 2017-07-04
remove create-react-class dependency from main sources

jochenberger:
Follow-up for #1734, #1820, #1821, #1822, #1823, #1824

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
     "url": "https://github.com/JedWatson/react-select.git"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


agirton:
Hi @jochenberger do you want to rebase on top of master to fix the issues? I hope also to be merging in a fix for the issue we have where two tests will timeout periodically.

jochenberger:
This will probably not help much without https://github.com/JedWatson/react-input-autosize/pull/85

Issue #1847, opened 2017-07-01
if I don't want to filter options while in multi mode 


kcliu:
Instead of filter options in 'multi' mode, if we have to rewrite filterOptions, we may also want to do something to the selected options ((e.g apply a style, remove option if click again). Therefore, pass valueArray and removeValue()  to menuRenderer would be needed.

agirton:
Hi @kcliu can you give us a bit more information? Are you proposing a feature?

kcliu:
yes, I think it's more like a feature, just like what pr #1731 did, to add more flexibility for the options. when we use multiselect mode, we may not always want to filter the selected options. Therefore, making `menuRenderer` more flexible would be a good direction.

Below is the example that I want to leverage `menuRenderer` to acheive. But actually I can't do it because the lack of `valueArray` and `removeValue()` in `menuRenderer`. 
<img width="304" alt="screen shot 2017-07-01 at 10 34 52 pm" src="https://user-images.githubusercontent.com/161968/27762953-90e00170-5ead-11e7-960e-637b39eb9cf9.png">

newswim:
This seems like a pretty obvious enhancement. I have an identical mock-up that I'm trying to recreate.

agirton:
Hi @kcliu i see. But this doesn’t look like a select component you are looking for a menu component, right?

kcliu:
@agirton Nope, the mock-up is just for demonstration.

rrgrs:
+1

I too am interested in being able to render the menu in the way described by kcliu as it's much more mobile friendly.

newswim:
Is seems like [this PR](https://github.com/JedWatson/react-select/pull/1731) might fix this, and it's only a tiny change to pass along the valueArray.

agirton:
Hi @newswim my concern is some people may be expecting this old behavior, so it's best to wait till after 1.0 is released.

JedWatson:
I've stepped through the code and this should be safe (not make a difference) because the function signature remains the same, and by default selected options aren't being passed to the menuRenderer anyway. So I've merged that PR, which (in combination with the new `removeOnSelect` prop in PR #1891) will let you do what you're after.

PR #1850, opened 2017-07-04
pass removeValue() and always pass valueArray

kcliu:
this is a pr for #1847


JedWatson:
Thanks @kcliu, will leave some notes in that issue too.

PR #1851, opened 2017-07-04
Use Node 6.x

jochenberger:


jochenberger:
I opened this because I thought it might fix the test failures, but they were fixed elsewhere, so I'm closing this.

PR #1853, opened 2017-07-04
fix tests

jochenberger:
Inspired by #1848


agirton:
Thank you! @jochenberger

PR #1855, opened 2017-07-05
Bugfix/remove circular refs

agirton:
This PR is fairly simple in that we just remove the circular dependencies. It seems most bundlers handle this ok, but may cause issues with Scope hoisting found in Webpack 3 (with their `ModuleConcatenationPlugin`) or Rollup.

example: Async Component relies on Select Component and Select relies on Async.

Have gone through all the tests and verified manually that the examples still work as expected.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -4,7 +4,7 @@ var initGulpTasks = require('react-component-gulp-tasks');
 var taskConfig = {
 
 	component: {
-		name: 'Select',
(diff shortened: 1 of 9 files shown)


jochenberger:
See #1741

agirton:
@jochenberger haha, I even commented on it! I searched before I started and github showed no results. Ah well 😄.

PR #1856, opened 2017-07-05
Add pkg.module

jochenberger:
https://github.com/rollup/rollup/wiki/pkg.module


jochenberger:
This is probably not correct, we probably need to transform the JSX at least.

PR #1858, opened 2017-07-05
Avoid mutating user input when ignoring case/accents (fixes #1352)

### LINKED ISSUE #1352: ignoreCase = true (default behavior) lowercases all user input

not-an-aardvark:
Previously, when the ignoreCase or ignoreAccents options were set to true in an Async select, the visible user input would be transformed (e.g. it would appear all lowercase even when the user was typing capital letters). This commit updates the component to transform the input internally without changing the input value itself.

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -131,26 +131,28 @@ export default class Async extends Component {
 				isLoading: true
 			});
 		}
-
-		return inputValue;
 	}
 
 	_onInputChange (inputValue) {
 		const { ignoreAccents, ignoreCase, onInputChange } = this.props;
+		let transformedInputValue = inputValue;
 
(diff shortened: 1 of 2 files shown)


agirton:
Thank you for the PR and the additional unit test!

zackify:
Any idea on when this will be released?

PR #1862, opened 2017-07-06
Trim search text from beginning and the end. (fixes #1861)

### LINKED ISSUE #1861: Filter text is not trimmed
This is a problem when you type a filter starting or ending with space characters

Type " new south wales" or "new south wales " in the first example
**Actual:**
No options shown
**Expected**
New South Wales option should be shown

Whitespaces wrapping the filter should be trimmed so that intended value could be matched. This is also the default behaviour 
in https://harvesthq.github.io/chosen/ and http://selectize.github.io/selectize.js/

serkanozer:


agirton:
Hi @serkanozer thank you for the PR. Since users can add this in by overriding the `filterOptions` going to hold off on this one for now unless other users are needing it. The bigger the surface area, the harder it is to maintain.

serkanozer:
Thanks @agirton . You are right actually we are using `filterOptions` in our app but the drawback with is that its not a method for extension but replacement. When I want to add trimming I also need to define other basic defaults  (`ignoreCase` `ignoreAccents`). 

Also, this is the default behavior in http://selectize.github.io/selectize.js/ and https://harvesthq.github.io/chosen/ .Would be good to add it to defaults here as well

Issue #1554, opened 2017-02-15
Clear button overflows `value` when placed in narrow container


jochenberger:
If a `Select` is placed inside a narrow container element, the clear button overflows the value text.
![react-select](https://cloud.githubusercontent.com/assets/370733/22967739/fb55d0f4-f367-11e6-9a57-cc62ade9e377.png)

https://jsfiddle.net/5e0g6qjt/1/

dekelb:
Setting `padding-right` on the relevant selector will fix this:

    .Select-placeholder, .Select--single > .Select-control .Select-value {
        padding-right: 40px;
    }

JedWatson:
Fixed by #1866

PR #1866, opened 2017-07-11
Fix clearable padding style

mmiyauchi:
Adjust padding-right because prevent text from overlapping `×` button.

####  (English / Japanese)
before

### After

after


jochenberger:
fixes #1554 (and duplicate #1916)

JedWatson:
Thanks @mmiyauchi!

PR #1876, opened 2017-07-17
Export select prop types

ecbrodie:
The purpose of this PR is to allow the prop types from the `Select` component to be exported. The use case that this satisfies is to allow an application using the `react-select` to build a custom wrapper around the `Select` component (ie, to allow for customizations that serve business needs of the application, such as specific default prop values), but not need to copy-paste prop type code from `react-select` to satisfy the prop-type checks. This duplicated prop-type code can simply be replaced with an import of `SelectPropTypes` from `react-select`.

I also made a couple other minor improvements, to hopefully improve the development experience for `react-select` contributors:

* Adds an `.nvmrc` file. The README calls out that Node v4 should be used for development environments. Using this file enables better UX for contributors using `nvm`
* Removes an outdated line from `CONTRIBUTING.md`

Feedback welcome. I hope this PR adds value to this project. Thank you.

### UPDATE

One more thing I should note. I originally tried to export `SelectPropTypes` from the same file as the `Select` component, `select.js`. I did this as a named export, alongside the default export of `Select`. Unfortunately, this ended up breaking all tests of the `Select` component. I narrowed down the reason to behaviour of ES6 transpilation in this project. Before my changes (master branch), `lib/Select.js` would contain this line:

```javascript
module.exports = exports['default'];
```

However, with the named export included, the `export` object would have its `default` and `SelectPropTypes` properties set accordingly, but that latter line for `module.exports` would not appear. I decided to work around this by keeping a single export (the default) in `Select.js` and extracting `SelectPropTypes` into its own file with an `export default`. I am open to hearing other suggestions to better export `SelectPropTypes`. This could also be a pattern for the prop types of other components for future PRs, as needed.


ecbrodie:
@JedWatson do you have any feedback on this PR? Thank you.

jochenberger:
@ecbrodie, can't you just use `Select.propTypes`?

ecbrodie:
The feeling when you look back at your own PR from 3 months ago and say "Shit, what was I thinking?"

You are right @jochenberger . `propTypes` are indeed exported. Derp.

PR #1880, opened 2017-07-18
Trigger onInputChange on clear

phamthaibaoduy:
I'm using Async component, when I click on clear button, list of options is not reloaded with empty string. Which is IMO a little bit strange.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -568,6 +568,9 @@ class Select extends React.Component {
 		}
 		event.stopPropagation();
 		event.preventDefault();
+		if (this.props.onInputChange) {
+			this.props.onInputChange('');
+		}
 		this.setValue(this.getResetValue());
 		this.setState({
 			isOpen: false,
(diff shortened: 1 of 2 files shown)


plantvsbirds:
an empty string is probably not good, since there are people who also use object as options. what about just `undefined`?

phamthaibaoduy:
@plantvsbirds sorry but it is onInputChange, not onChange

plantvsbirds:
my bad :)

gwyneplaine:
@phamthaibaoduy, can you confirm that the bug you're fixing is still, reproducible with the current rc of react-select? 

At the moment clearValue calls `this.handleInputValueChange` with an empty string, which in turn invokes this.props.onInputChange with the passed in value. So the additional onInputChange here is seemingly redundant.

Unfortunately, I have been unable to reproduce this on my end, PR closed until further notice

phamthaibaoduy:
Sorry for late response, did you check this issue yourself?

phamthaibaoduy:
I checked code in rc10, look like the issue is fixed. Thank you @gwyneplaine

PR #1881, opened 2017-07-19
add ref existential check fixes #899

### LINKED ISSUE #899: Uncaught TypeError: Cannot read property 'offsetTop' of null
Hi i'm trying to use this module in a meteor 1.2 project, but when i click the component for select an option,i get the following error message. 

```
Select.js:193 Uncaught TypeError: Cannot read property 'offsetTop' of null
```

I used the example code for test in my project.

### LINKED ISSUE #1704: Suggestions on fixing - "Only a ReactOwner can have refs"
Hello, I've just stumbled across this plugin and would really like to use it however I have this issue (I'm sure we have all seen this before): "Only a ReactOwner can have refs. You might be adding a ref to a component that was not created inside a component's `render` method, or you have multiple copies of React loaded". Apologises for creating another bug about this issue, there are a few others however I didn't want to reopen old issues.

I've checked the additional details here: https://facebook.github.io/react/warnings/refs-must-have-owner.html

I'm not using any refs, I'm simply using the select as basically as I can.
I don't have multiple instances of react, I've checked the other node modules folders and ran "npm ls react" which only returns one instance. ( I'm using npm and gulp if that makes a difference )
I've updated node and npm
I've deleted all node modules and reinstalled
I've followed this link: https://gist.github.com/jimfb/4faa6cbfb1ef476bd105

Is there anything else I can try before giving up?

Regards

plantvsbirds:
Check my changes at https://github.com/JedWatson/react-select/pull/1881/files#diff-04191649796841200b33375bdff7c78fR118 It's just one line. other files are built.

Frankly I don't know how it occurred. I am just importing module B in project A, and exported a wrapped `react-select` component. Both A, B are built using webpack and I got a ton of warning as descrbied in #899. Using `react-select` directly in A cause no problem.

Anyway, it's good defensive programming, and I strongly encourage you guys to do so when referring to any result from `findDOMNode`. 

And there's a lifecycle warning (updating state when unmounted). Will probbly fix that too


plantvsbirds:
I should've merged master before any change, but I am not doing that. Closing

PR #1894, opened 2017-07-27
fix bug, if wrote "hasOwnProperty"

SuhushinAS:


agirton:
Hi @SuhushinAS thank you for the PR. What’s the use case?

SuhushinAS:
Hello!
Sorry for my English.
See the attached picture to get bug:
screenshot_1

JedWatson:
Nice, good fix @SuhushinAS

JedWatson:
@jochenberger what's going on here:

The cache uses the `Object` prototype, the bug here is that adding a property called `hasOwnProperty` to the cache object overrides it. So calling the method on `Object.prototype` is a neat workaround.

PR #1895, opened 2017-07-27
Update README.md

seelts:
Now the component throws an exception if `onChange` is not a function.
So, it forces developer to specify it explicitly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ You can also use the standalone build by including `react-select.js` and `react-
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


agirton:
Hi @seelts if this throws an error then this is a regression. What is the error?

jochenberger:
Might be related to https://github.com/JedWatson/react-select/commit/a7782b34abdc9d2beb8c604ee605fb1e81db8bd2#diff-149b30fde0306e461dc519d4853f1e50R186

JedWatson:
I fixed the regression, see #1988

Issue #1904, opened 2017-08-01
weird error in Options handleTouchEnd


blacktemplar:
I am tracking javascript errors of all users of one of my tools where I use react-select and yesterday I got the following error:

Uncaught TypeError: Cannot read property 'dragging' of undefined

The error occurs in the handleTouchEnd event of the Option component. Unfortunately I was not able to reproduce the error, but I checked the Option component and saw there that methods like handleMouseDown or handleTouchStart get bound in the constructor but not the method handleTouchEnd. So maybe adding a bind for this method would fix this error, but as I cannot reproduce it its hard to tell. Is it intended that handleTouchStart is bound in the constructor and handleTouchEnd not?

Anyway do you have any clues why this error might happen? I never saw that this is undefined...

gwyneplaine:
@blacktemplar  thanks for logging this issue. Your PR has been merged into master and is in the latest release. 

Issue closed

PR #1905, opened 2017-08-01
binds handleTouchEnd in Option component

blacktemplar:
fixes possible wrong this object in handleTouchEnd method,
possibly fixes issue #1904


agirton:
Hi @blacktemplar thank you for this. It looks like I need to add a few more since those were missed.

PR #1908, opened 2017-08-02
Pass isOpen to renderValue

ariannedee:
It can be useful for renderValue to know if the menu is opened. This can be used to add a 'Type to search' text.

context-chooser2

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,7 +402,7 @@ function onInputKeyDown(event) {
 | value | any | undefined | initial field value |
 | valueComponent | func | <Value /> | function which returns a custom way to render/manage the value selected `<CustomValue />` |
 | valueKey | string | 'value' | the option property to use for the value |
-| valueRenderer | func | undefined | function which returns a custom way to render the value selected `function (option) {}` |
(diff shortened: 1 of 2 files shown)


himynameisdave:
This would be super useful 👍

JedWatson:
Thanks for the PR @ariannedee 

This looks like a useful behaviour, thanks for the gif illustrating the use-case. I'm happy to look at ways to achieve this, but I don't think this is the right implementation.

I'm concerned that this creates an inconsistency between the signature of the `valueRenderer` prop between single and multi-select inputs (you'll notice that when `props.multi` is true, the `renderLabel` method is passed `renderLabel(value, i)`)

Fundamentally when you're trying to do is hide the value and show the placeholder when the menu is open, even when a value is selected. I'd be happy to look at supporting a specific prop that renders the placeholder instead of the value when the menu is open (and works for both single and multi-selects). I'm also planning some changes post-1.0 and will take this into account if you don't open another PR before then.

Thanks again, and sorry for not accepting the PR!

PR #1923, opened 2017-08-11
Fix onMenuScrollToBottom does not work in chrome 58.0

deevus:
Determining whether the user had scrolled to the bottom of the menu used an integer to boolean coercian. Replacing the coercian with an integer comparison instead fixes the problem.


deevus:
Hello?

JedWatson:
Thanks @deevus!

PR #1924, opened 2017-08-11
Customizable value remove icon

R0nd:


JedWatson:
Thanks for the PR @R0nd but this seems like unnecessary complexity.

You can already provide a custom `Value` component, which is pretty simple, and you can extend the provided one if you want to do what you want without adding more props to the Select:

```js
class MyValueWithCustomRemoveIcon extends Value {
  renderRemoveIcon() { /* your custom component goes here */ }
}
<Select valueComponent={MyValueWithCustomRemoveIcon} {...props} />
```

PR #1931, opened 2017-08-17
add inputDisabled option

untitledboy:
Add **inputDisabled** option for Select.
The most common case: you can choose only 10 users. When you have 10 users, you cannot add 11th. But you still can remove one of those 10.
So **disabled** won't work. Because **disabled** works for options as well. Yet **inputDisabled** disables just input without interacting with options.

### DIFF
--- a/examples/src/components/Multiselect.js
+++ b/examples/src/components/Multiselect.js
@@ -24,6 +24,7 @@ var MultiSelectField = createClass({
 	getInitialState () {
 		return {
 			disabled: false,
+			inputDisabled: false,
 			crazy: false,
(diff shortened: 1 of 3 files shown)


JedWatson:
Thanks for the contribution @untitledboy! much appreciated.

Unfortunately I'm not inclined to merge this, because disabling the input has other unintended effects to do with managing focus and accessibility (e.g. you couldn't use the keyboard to remove items once the maximum has been reached)

We could potentially support a `maxItems` option for multi-select which stops the menu from being opened or adding values when it is reached, but we still shouldn't disable the input in that case.

PR #1932, opened 2017-08-18
simplify stringifyValue

dysfunc:
JSON.stringify will handle type casting (to string) properly for num, bool, objects, and arrays. This shortens the logic for handling the various cases. You could wrap this in a try/catch to prevent cyclical issues blowing things up but I don't think that's necessary.


JedWatson:
Thanks @dysfunc 👍

PR #1933, opened 2017-08-18
simplify constructor method binding

dysfunc:
Saves some bytes and makes it a little easier, at a glance, to see which methods you're binding to


JedWatson:
Thanks @dysfunc, agreed this is a bit clearer 👍

PR #1942, opened 2017-08-24
New prop/api: onValueRemove

cjtafoya:
Passes removed values to prop function onValueRemove.

Minimal overhead for implementing through a callback which utilizes existing remove method here, as opposed to manually tracking/matching in a HOC.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -557,6 +557,7 @@ class Select extends React.Component {
 	removeValue (value) {
 		var valueArray = this.getValueArray(this.props.value);
 		this.setValue(valueArray.filter(i => i !== value));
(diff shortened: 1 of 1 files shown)


JedWatson:
I'm actually trying to _reduce_ the API surface area... what's the argument for not just listening to `onChange` here? that'll be more consistent (e.g. if the control is cleared, all values would be removed in a way that skips this)

JedWatson:
@cjtafoya thanks for the PR. I'm going to close it for now because I'm trying to reduce the number of open PRs, feel free to reopen with an answer to my question above though and I'm happy to reconsider.

Issue #1539, opened 2017-02-07
Weird focus behaviour wiht react-datepicker


tume:
There is already issue in react-datepicker: https://github.com/Hacker0x01/react-datepicker/issues/730 

Posting here also because not certain which component should fix this. 

I have repository for reproducing it here: https://github.com/tume/select-datepicker-conflict

You can see the issue when first opening the calendar and then switching to react-select.
I have some more speculation about the cause in react-datepicker issue but it looks like to be problem with onBlur/mousedown event ordering if I'm correct. Would appreciate if you have any comments on this or how to fix it....

buz-zard:
Same here:
`"react-datepicker": "^0.40.0"`
`"react-select": "^1.0.0-rc.3"`

Also focusing on ordinary input doesn't flicker, just react-select input produces this.

ibcWillJust42:
same problem

gilbertKaradja:
I am starting to think its a problem with react-slect

swbullis:
@gilbertKaradja I think your might be right here.  From what I can gather so far it looks like it might be react-select not allowing https://github.com/Pomax/react-onclickoutside to work or vice versa.

JedWatson:
I'm pretty sure this is an issue with the way react-datepicker is asynchronously refocusing the input when it is blurred. I don't totally understand the use case for that though, so it might be important to support. Will comment on that project to learn more.

theGABS:
+1

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

If you feel this issue / pull request is still relevant and you'd like us to review it, please leave a comment and we'll do our best to get back to you.

PR #1946, opened 2017-08-29
WIP: Fix the focus behaviour with react-select and react-datepicker

nothingrandom:
In reference to #1539 and Hacker0x01/react-datepicker#730

This is an imperfect fix to the bug. It needs further work and suggestions, open to hearing what that might be.

### DIFF
--- a/lib/Select.js
+++ b/lib/Select.js
@@ -367,51 +367,56 @@ var Select = (0, _createReactClass2['default'])({
 	handleMouseDown: function handleMouseDown(event) {
 		// if the event was triggered by a mousedown and not the primary
 		// button, or if the component is disabled, ignore it.
-		if (this.props.disabled || event.type === 'mousedown' && event.button !== 0) {
-			return;
-		}
+    var self = this;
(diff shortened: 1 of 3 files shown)


JedWatson:
Hey @nothingrandom, thanks for the PR 👍 

Some quick feedback on it - if you click the "files changed" tab you'll see that all the indentation has changed, which makes it really hard to see what the actual intended change is here. You could also avoid creating a reference to `self` by using an arrow function in the timeout: `() => { ... }` and I'm not sure what the value of creating a second reference to `evt` is.

Also, please only submit changes to the `src` folder 🙂 

With that in mind, I'm trying to sort out as many PRs as possible so I'm going to close this one. I've had a look at react-datepicker and I think the solution may be a change in that component (it's re-focusing in a timeout after any blur from what I can tell) and we would need to understand the impact of removing the `.preventDefault()` here before this could be merged.

If we can't find a solution in `react-datepicker` I'm open to looking at ways to change `react-select` but let's see if we can solve it there first.

PR #1949, opened 2017-08-29
add isOpen prop that allows owners to override the isOpen state

EnriqueVidal:
I've rebased current `master` against @bobbyrenwick's original PR, hopefully it gains some traction.

Rel #817


EnriqueVidal:
My bad I had not noticed this was not clean, closing.

PR #1950, opened 2017-08-29
Add alwaysOpen prop support

EnriqueVidal:
Rather than juggle a prop tied to the state provide support for `alwaysOpen`.

```es6
const options = [
  { label: "Vanilla", value: "vanilla" },
  { label: "Chocolate", value: "chocolate" },
  { label: "Strawberry", value: "strawberry" },
];

export default () => (
  <Select options={options} alwaysOpen />
);
```

Rel #817


JedWatson:
Thanks for the PR @EnriqueVidal, I'm going to close it until we have more clarity on #1989

PR #1952, opened 2017-08-30
Resolve #709

### LINKED ISSUE #709: Option to preserve inputValue after select item in multiselect

davidroeca:
This is essentially a verbatim copy of #750 with merge conflicts resolved and handling the updated api. I've tested it with my application and it seems to be working as expected. Thanks to @nosovsh for most of the thought/code involved here.

Let me know if you'd like me to push the builds to this branch as well--I'm not clear on how this library gets published

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -520,7 +520,7 @@ class Select extends React.Component {
 			this.setState({
 				inputValue: this.handleInputValueChange(updatedValue),
 				focusedIndex: null
-			} , () => {
+			}, () => {
 				this.addValue(value);
 			});
 		} else {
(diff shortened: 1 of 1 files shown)


JedWatson:
Looks good, thanks for picking this up @davidroeca and to @nosovsh for the original implementation 👍

PR #1953, opened 2017-09-01
New build pipeline for react-select

gwyneplaine:
### Features
- The gulp build process has been replaced by the following: 
  - UMD build to dist via rollup
  - lib is now transpiled straight from src via babel-cli 
  - examples and dev workflow now built through webpack.


.babelrc
```diff
@@ -0,0 +1,9 @@
+{
+  "ignore": ["node-modules/**", "src/index.umd.js"],
+  "presets": ["es2015", "stage-0", "react"],
```
agirton: Instead of `stage-0` and `es2015` could we use `env`? If we don't provide any configuration, it will use the latest browsers. It's also the recommend preset.

rollup.config.js
```diff
@@ -0,0 +1,44 @@
...
+      'prop-types',
+      'react-input-autosize',
+      'react',
+      'react-dom',
+    ],
+    plugins: [
+      babel({
+        babelrc: false,
+        exclude: 'node_modules/**',
+        presets: [['es2015', { 'modules': false }], 'stage-0', 'react'],
```
agirton: Webpack 3 would benefit by this default as well. Maybe we can remove most of this babel config and keep it in the `babelrc` file?

webpack.config.js
```diff
@@ -0,0 +1,65 @@
...
+    port: 8080,
+  },
+  module: {
+    rules: [
+      {
+        test: /\.js$/,
+        exclude: [/node_modules/],
+        use: [{
+          loader: 'babel-loader',
+          options: { presets: ['es2015', 'stage-0', 'react'] },
```
agirton: Same as comment from rollup config. Let's just use the `babelrc` config since there is no real difference.

.babelrc
gwyneplaine: @agirton yup, we've made a note of using the env preset, please be on look out for a PR to implement this once this one has been closed out 👍

rollup.config.js
gwyneplaine: good point, I've fixed up the babel config to share common presets and plugins across the two configurations

webpack.config.js
gwyneplaine: yup, done 👍

examples/src/standalone.html
```diff
@@ -15,10 +15,10 @@ <h1 style="margin-top: 40px;">React Select</h1>
 	</div>
 	<script src="//unpkg.com/react@0.14.2/dist/react.min.js"></script>
 	<script src="//unpkg.com/react-dom@0.14.2/dist/react-dom.min.js"></script>
-	<script src="//unpkg.com/three.js@0.73.0/build/three.min.js"></script>
 	<script src="//unpkg.com/classnames@2.0.0/index.js"></script>
+	<script src="//unpkg.com/prop-types@15.5.10/prop-types"></script>
 	<script src="//unpkg.com/react-input-autosize@0.6.3/dist/react-input-autosize.min.js"></script>
-	<script src="standalone.js"></script>
+	<script src="../../dist/react-select.js"></script>
```
JedWatson: Just a note for later: when you clean up the contributing / dev guide, make sure to explain why this file exists and that it depends on the build

src/index.umd.js
```diff
@@ -0,0 +1,10 @@
+import Select from './Select';
```
JedWatson: This file should also have a comment explaining why it exists

johnnypez:
The added dependency here for stage-0 seems unnecessary. Our project now won't build unless we add the stage-0 which we don't want to do. Can that be addressed?

agirton:
Hi @johnnypez can you try the latest build (rc7)? We recently fixed an issue with where lib wasn't precompiled.

johnnypez:
Hey @agirton, we hit this problem on rc7. We were good up to rc5 I think.

I'll have to check later but maybe it's because we're not requiring the dist version. We've had the following for a long time:

```import Select from 'react-select'```

but maybe we should be pulling in the dist version?

```import Select from 'react-select/dist/react-select'```

agirton:
Odd. What's failing in your build?

johnnypez:
It bitches about `stage-0` being a missing dependency. We have `stage-2`.

agirton:
@johnnypez I haven't come across this type of issue before. What is complaining that `stage-0` is missing?

johnnypez:
It's webpack that complains about the missing dependency but you know what, I think this is down to the non-standard way that we're building our app. Sorry to have wasted your time.

KoAi:
@agirton, same issue for me
`ERROR in ./node_modules/react-select/lib/index.js
Module build failed: Error: Couldn't find preset "stage-0" relative to directory "D:\\project\\node_modules\\react-select"`

agirton:
Hi @KoAi please submit an issue with your build setup and we will take a look.

KoAi:
After installing babel-preset-stage-0, it solves the problem. Sorry and thanks

gwyneplaine:
@neptunian No worries, glad this could be of use elsewhere 👍

PR #1955, opened 2017-09-01
don't keep generated files in the repository

jochenberger:


agirton:
Hi @jochenberger, @JedWatson wants to keep the generated files in the repo so users don't have to try and do the building on their side.

JedWatson:
Adding context here, we're still supporting bower with our und builds and I don't think that's safe to change without a major (breaking) version bump. Builds will be removed in the future with 2.0.0 but for now should remain.

jochenberger:
I'm not entirely familiar with your build/release process, but I don't think that it would be changed by that commit. Or don't you run the build task as part of the release?

PR #1962, opened 2017-09-04
Export the Value component

Grahack:
Or is there a way to pre-fill the Creatable component with known values?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,14 +2,17 @@ import Select from './Select';
 import Async from './Async';
 import AsyncCreatable from './AsyncCreatable';
 import Creatable from './Creatable';
+import Value from './Value';
 
 Select.Async = Async;
 Select.AsyncCreatable = AsyncCreatable;
 Select.Creatable = Creatable;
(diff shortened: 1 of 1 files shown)


Grahack:
Sorry, I managed to pre-fill the component with an array of strings. I only missed `multi={true}`.

sakulstra:
@Grahack why did you close this pull request? I actually wanted to commit exactly the same one right now :dancer: 

Exporting Value would allow us to write valueComponent without rewriting everything from scratch:
```
import {Value} from 'react-select'

class ValueComponent extends Value{
render () {
const title = generateTitle(this.props.value)
		return (
			<div className={classNames('Select-value', this.props.value.className)}
					 title={title}
			>
				{this.renderRemoveIcon()}
				{this.renderLabel()}
			</div>
		);
	}
}
```

Grahack:
Because I don't need it anymore. I can reopen it for you.

JedWatson:
👍 thanks @Grahack

PR #1967, opened 2017-09-07
Fix wrapping on Firefox in SCSS files

mwilliamson:
This is the same fix as #1300, but applied to the SCSS files.


JedWatson:
Thanks @mwilliamson!

PR #1975, opened 2017-09-10
Prevent UglifyJs from choking on defaultArrowRenderer.js

circlingthesun:
UglifyJs chokes on defaultArrowRenderer.js as destructuring is not es5. This fixes that. The alternative is to change the main entry point of this module from "lib/index.js" to "dist/react-select.js".


agirton:
Hi @circlingthesun thank you for the contribution. However this will be fixed when #1976 is merged.

PR #1979, opened 2017-09-11
Remove `jsnext:main` entry

jochenberger:
The file does not exist, this sneaked in in https://github.com/JedWatson/react-select/pull/1953/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R6

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,6 @@
   "version": "1.0.0-rc.7",
   "description": "A Select control built with and for ReactJS",
   "main": "lib/index.js",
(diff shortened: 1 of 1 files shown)


jochenberger:
cc @gwyneplaine

gwyneplaine:
yes that definitely shouldn't be there, thanks for that @jochenberger
lgtm

jochenberger:
See also #1981

JedWatson:
I've fixed it the other way, by adding the es build.

Issue #1980, opened 2017-09-11
Need tests for: menu does not open on click if `searchable` is `true`


jochenberger:
Follow-up for https://github.com/JedWatson/react-select/commit/27462611929254d91a82b6b7b8cbdec6b9ca80a1#diff-04191649796841200b33375bdff7c78fR231

JedWatson:
Thanks @jochenberger

jochenberger:
`git bisect` blames f7cd6c2fa5b9b157a39ccaa22ebad421a5c224e2

jochenberger:
Or rather #1580.

jochenberger:
Maybe the default value for `openAfterFocus` should be `true`.
I wonder why the `Select -> with simple options -> should show the options on mouse click` test does not fail.

JedWatson:
`openAfterFocus` is intended to control whether the menu opens when the control is focused (which it doesn't by default, this is by design)

However the menu *should* open when the control is clicked on (regardless of whether it was previously focused or not), this is the regression in the fix you've linked to.

Not sure why the test isn't failing 😞

jochenberger:
The test succeeds because `handleInputFocus` is called before `handleMouseDown`. In the browser (Firefox at least), it's the other way around. That's why, in the test, the menu is opened in https://github.com/JedWatson/react-select/blob/v1.0.0-rc.7/src/Select.js#L257

jochenberger:
Oh, I get it, the tests focus the select from `beforeEach`:
```
...
    at Select.handleInputFocus (/home/jochen/git/upstream/react-select/src/Select.js:320:14)
...
    at Object.focus (/home/jochen/git/upstream/react-select/node_modules/react-dom/lib/ReactTestUtils.js:349:18)
    at findAndFocusInputControl (/home/jochen/git/upstream/react-select/test/Select-test.js:130:24)
    at createControl (/home/jochen/git/upstream/react-select/test/Select-test.js:150:4)
    at Context.<anonymous> (/home/jochen/git/upstream/react-select/test/Select-test.js:261:15)
```
So there is no test for "no focus -> click -> should show menu". Should we create one or modify the existing test case?

jochenberger:
See also #1527

JedWatson:
I've updated this to reflect fixing the tests for this, let's track also fixing the issue over in #1994

JedWatson:
I found the issues with the tests for this and comprehensively fixed them. Required quite a bit of digging, see the commit for more details.

PR #1986, opened 2017-09-12
WIP: Select should open when clicked

jochenberger:
fix for #1980

### DIFF
--- a/test/Select-test.js
+++ b/test/Select-test.js
@@ -272,6 +272,7 @@ describe('Select', () => {
 		});
 
 		it('should show the options on mouse click', function () {
+			// the input already has the focus here (see beforeEach -> createControl)
 			TestUtils.Simulate.mouseDown(ReactDOM.findDOMNode(instance).querySelector('.Select-control'), { button: 0 });
 			var node = ReactDOM.findDOMNode(instance);
 			expect(node, 'queried for', '.Select-option', 'to have length', 3);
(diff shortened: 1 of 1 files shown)


jochenberger:
actually, I can only get the test to run if I revert 3f102e51079d59af862923bb01d10181b50b2b34

jochenberger:
I can't figure out why the test fails. It gets to https://github.com/JedWatson/react-select/blob/v1.0.0-rc.8/src/Select.js#L264 and to https://github.com/JedWatson/react-select/blob/v1.0.0-rc.8/src/Select.js#L181, but not to https://github.com/JedWatson/react-select/blob/v1.0.0-rc.8/src/Select.js#L315.

ls-guillaume-lambert:
This is a regression in the latest RC releases?

jochenberger:
I suspect that jsdom does not fire the focus events that should trigger `handleInputFocus`. Upgrading it does not seem to help.

jochenberger:
cc @JedWatson

JedWatson:
@bruderstein any chance you can help shed some light on this? see #1980 for context

JedWatson:
OK I worked through this separately and committed a fix.

It was important to manage the `initialFocus` option for `createControl`, and explicitly call `findAndFocusInputControl` at the right time to correctly simulate user interaction.

Thanks for pointing me in the right direction @jochenberger

PR #1990, opened 2017-09-12
Fix peerDependencies to be compatible with 'prop-types' package

MichaelDeBoey:
`prop-types` [requires](https://github.com/reactjs/prop-types#compatibility) `React ^0.14.9` or `React ^15.3.0`


JedWatson:
Thanks @MichaelDeBoey

JedWatson:
I've fixed this on react-input-autosize as well

PR #1993, opened 2017-09-13
Update build

### LINKED ISSUE #1992: Cannot read property 'bool' of undefined
## Accessing PropTypes from "react" instead of "prop-types"
https://github.com/JedWatson/react-select/blob/master/lib/Select.js#L1212

PropTypes should be used from "prop-types" package and not from "react". All other proptypes are from the prop-types package, except for this one. Not sure how this error came to pass. I can submit a PR if you need.

imbhargav5:
I think a build was pending. 
I just ran 'npm run build' and created this PR.

New build auto fixes #1992


imbhargav5:
This commit fixed the issue in the source.
https://github.com/JedWatson/react-select/commit/a693fc476cca92b674d09c7bd193393ae620bd23

The build step resolves the same issue in the lib folder.

Thanks @MrLeebo @JedWatson

gwyneplaine:
thanks for this @imbhargav5, this has been fixed in a separate PR that has been merged into master and will be part of the next rc

PR #1998, opened 2017-09-13
Add variable for focused input background

aronstrandberg:
Add a variable to make it possible to configure the input's background color when focused. 

I was trying to theme the focused state in a project, but found it bit cumbersome, hence the desire for these changes. I'd be happy to make any requested changes from your side.

### DIFF
--- a/less/select.less
+++ b/less/select.less
@@ -15,6 +15,7 @@
 // control options
 @select-input-bg:                   #fff;
 @select-input-bg-disabled:          #f9f9f9;
+@select-input-bg-focus:             #fff;
 @select-input-border-color:         #ccc;
 @select-input-border-radius:        4px;
 @select-input-border-focus:         @select-primary-color;
--- a/less/select.less
+++ b/less/select.less
@@ -15,7 +15,7 @@
 // control options
 @select-input-bg:                   #fff;
 @select-input-bg-disabled:          #f9f9f9;
-@select-input-bg-focus:             #fff;
+@select-input-bg-focus:             @select-input-bg;
 @select-input-border-color:         #ccc;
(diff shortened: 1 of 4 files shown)


JedWatson:
Thanks @aronstrandberg, happy to accept this addition.

Would you please add it to the `scss` files as well though? we need to keep them in sync with the `less`, unfortunately this is a manual process.

less/select.less
```diff
@@ -15,6 +15,7 @@
 // control options
 @select-input-bg:                   #fff;
 @select-input-bg-disabled:          #f9f9f9;
+@select-input-bg-focus:             #fff;
```
JedWatson: Please make this default to `@select-input-bg` in case that has been customised, otherwise this would be a breaking change.

aronstrandberg:
@JedWatson Pushed some commits to adress these things. I couldn't find the equivalent of `.Select-focus-state` in the scss files, so I'm not 100%, but I think they should be similar now – is there a way to test the examples with the scss files to make sure?

aronstrandberg:
@JedWatson I just rebased this on master to adjust for upstream changes. Do you think this is ready to merge?

aronstrandberg:
@JedWatson?

JedWatson:
Looks great @aronstrandberg, thanks!

abdennour:
Do not release feature with big bug : 

```
Failed to compile.

Module build failed: 
		background: $select-input-bg-focus;
             ^
      Undefined variable: "$select-input-bg-focus".
      in /Users/abdennour/workspace/idax/node_modules/react-select/scss/control.scss (line 70, column 15)

```

PR #2000, opened 2017-09-13
Fixing css states to be scoped with Select selector, closes #1951.

### LINKED ISSUE #1951: Stateful CSS classes should always be chained
Hey!

I don’t actually use react-select myself, but I’m currently doing an audit for a client who does. I noticed that you use [stateful CSS classes that aren’t associated with anything](https://github.com/JedWatson/react-select/blob/af0aa5ae14dd741c3dd1c9b06733569f413c88ba/scss/control.scss#L65).

This selector:

```css
.is-open > .Select-control {
  […]
}
```

…means that, if I have react-select inside of my own modal component, like so:

```html
<div class="modal  is-open">
  […]
</div>
```

…we’d end up with a false positive, and your CSS would kick in when it shouldn’t.

- - -

As a general rule, all stateful classes (e.g. `.is-open`, `.has-dropdown`, etc.) should always be chained with something:

```css
.Select.is-open > .Select-control {
  […]
}
```

The new snippet leaves us immune to false positives and subtree collisions.

Thanks, 
H

agirton:
Ran this locally and compared each of the demos against the live version
and saw no differences.

### DIFF
--- a/less/control.less
+++ b/less/control.less
@@ -29,16 +29,78 @@
 	}
 
 	// handle disabled state
-	&.is-disabled > .Select-control {
-		background-color: @select-input-bg-disabled;
-		&:hover {
-			box-shadow: none;
+	&.is-disabled {
+		.Select-arrow-zone {
+			cursor: default;
+			pointer-events: none;
+			opacity: 0.35;
+		}
+
+		> .Select-control {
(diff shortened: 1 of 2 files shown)


JedWatson:
Thanks @agirton, and @csswizardry for the detailed issue!

PR #2002, opened 2017-09-13
autofocus --> autoFocus

gwyneplaine:
`autoFocus` prop added, deprecation warning added for use of `autofocus` prop

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -79,10 +79,10 @@ class Select extends React.Component {
 
 	componentDidMount () {
 		if (typeof this.props.autofocus !== 'undefined' && typeof console !== 'undefined') {
-      console.warn('Warning: The autofocus prop will be deprecated in react-select1.0.0 in favor of autoFocus to match React\'s autoFocus prop');
+			console.warn('Warning: The autofocus prop will be deprecated in react-select1.0.0 in favor of autoFocus to match React\'s autoFocus prop');
 		}
 		if (this.props.autoFocus || this.props.autofocus) {
-      this.focus();
+			this.focus();
 		}
 	}
 
--- a/src/Select.js
+++ b/src/Select.js
@@ -1051,8 +1051,8 @@ Select.propTypes = {
 	addLabelText: PropTypes.string,       // placeholder displayed when you want to add a label on a multi-value input
 	arrowRenderer: PropTypes.func,        // Create drop-down caret element
 	autoBlur: PropTypes.bool,             // automatically blur the component when an option is selected
-	autofocus: PropTypes.bool,            // [DEPRECATED] autofocus the component on mount
-	autoFocus: PropTypes.bool,						// autofocus the component on mount
+	autofocus: PropTypes.bool,            // deprecated; use autoFocus instead
+	autoFocus: PropTypes.bool,            // autofocus the component on mount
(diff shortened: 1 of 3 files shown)


src/Select.js
```diff
@@ -78,8 +78,11 @@ class Select extends React.Component {
 	}
 
 	componentDidMount () {
-		if (this.props.autofocus) {
-			this.focus();
+		if (typeof this.props.autofocus !== 'undefined' && typeof console !== 'undefined') {
+      console.warn('Warning: The autofocus prop will be deprecated in react-select1.0.0 in favor of autoFocus to match React\'s autoFocus prop');
```
JedWatson: Please fix the indentation here

gwyneplaine:
@JedWatson indentation fixed and tests added

jochenberger:
Wouldn't it be better to use a custom function for deprecation warning that returns an `Error` if the old `autofocus` property is set?
And I think that `examples/src/components/States.js` should be updated as well.

gwyneplaine:
Thanks for the feedback @jochenberger, what's the purpose of returning an Error object in this case?

jochenberger:
IMHO, returning an `Error` is more idiomatic React. Also, it will only warn in development mode and doesn't need the check for `console`. Also, it will not fail if `console` is available but `console.warn` is not. ;-)

src/Select.js
```diff
@@ -1048,7 +1051,8 @@ Select.propTypes = {
 	addLabelText: PropTypes.string,       // placeholder displayed when you want to add a label on a multi-value input
 	arrowRenderer: PropTypes.func,        // Create drop-down caret element
 	autoBlur: PropTypes.bool,             // automatically blur the component when an option is selected
-	autofocus: PropTypes.bool,            // autofocus the component on mount
+	autofocus: PropTypes.bool,            // [DEPRECATED] autofocus the component on mount
+	autoFocus: PropTypes.bool,						// autofocus the component on mount
```
JedWatson: @gwyneplaine please fix the in-line indentation here, should be using spaces for consistency

JedWatson:
@jochenberger what makes you say that returning an error is more idiomatic react?

we're explicitly continuing to support the `autofocus` prop for 1.0.0, just putting a deprecation warning in for dev mode - this is modelled specifically after the way that react deprecates features between versions

JedWatson:
@jochenberger I'm going to merge this as-is for now, I may have misunderstood what you meant with the error so please feel free to open another PR with how you meant for this to be implemented.

I checked up on whether we should check for `console.warn` but it's [available](https://developer.mozilla.org/en-US/docs/Web/API/Console/warn#Browser_compatibility) in more browsers than react-select supports and all node versions so I think we're safe to just use it.

JedWatson:
p.s. thanks @gwyneplaine 😁

jochenberger:
I'm fine with that. I thought that returning an `Error` might be more idiomatic since it's the way that custom validators should be implemented (https://github.com/facebook/prop-types). But other modules seem to warn directly too, however they usually seem to use [`warning`](https://github.com/BerkeleyTrue/warning).
> just putting a deprecation warning in for dev mode
I think that the warning will also appear in production. Maybe that should be changed. `warning` would take care of that.

PR #2003, opened 2017-09-14
strip proptypes in production build (fixes #1882)

### LINKED ISSUE #1882: Production build should not include proptypes definitions
The proptypes specifications should be stripped from the production build.
The definitions could be wrapped in `if (process.env.NODE_ENV !== 'production')` clauses, but that won't get replaced in the browserify bundling step.
That probably makes it a feature that could go into https://github.com/JedWatson/react-component-gulp-tasks, but it might be a better idea to switch to Webpack or even Rollup on the long run. @JedWatson, any advice how this should be solved?

jochenberger:
File size changes:
```
react-select.es.js: 82822 -> 82970
react-select.js: 84232 -> 84380
react-select.min.js: 43966 -> 41700
```
See also #1876.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "babel-loader": "^7.1.1",
     "babel-plugin-external-helpers": "^6.22.0",
     "babel-plugin-istanbul": "^4.1.4",
+    "babel-plugin-transform-react-remove-prop-types": "^0.4.8",
     "babel-preset-es2015": "^6.24.1",
     "babel-preset-react": "^6.24.1",
     "babel-preset-stage-0": "^6.24.1",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -13,10 +13,16 @@ const globals = {
 	react: 'React',
 };
 const external = Object.keys(globals);
-const babelOptions = {
-	babelrc: false,
-	presets: [['es2015', { modules: false }], 'stage-0', 'react'],
-	plugins: ['external-helpers'],
+const babelOptions = (production) => {
+	let result = {
+		babelrc: false,
+		presets: [['es2015', { modules: false }], 'stage-0', 'react'],
+		plugins: ['external-helpers'],
+	};
+	if (production) {
+		result.plugins.push('transform-react-remove-prop-types');
+	};
+	return result;
 };
 
 export default [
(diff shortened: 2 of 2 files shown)


jochenberger:
I guess we need to replace `'process.env.NODE_ENV'` in the other two bundles too. But I'm not sure what we should replace it with in the es6 build.

jochenberger:
And what should happen to `lib/*.js`? With this PR merged, they would contain the `if (process.env.NODE_ENV !== 'production')`wrappers. Maybe we should only add the https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types plugin to the `.min.js` build.

agirton:
Hi @jochenberger instead of putting the `propTypes` behind a flag, we could probably just use `babel-plugin-transform-react-remove-prop-types`. This would make it much cleaner. Thoughts?

JedWatson:
> we could probably just use `babel-plugin-transform-react-remove-prop-types`

yes please 🙂

jochenberger:
I'm not so sure. Don't we want to have the checks in the `src` and `lib` files? I don't do node apps myself, but I imagine that people who do want the propTypes checks removed in production. If we only remove them in the bundles via `babel-plugin-transform-react-remove-prop-types`, they will always end up with the propTypes checks. However, if the `if (process.env.NODE_ENV !== 'production')` are present in the `src` and `lib` files, they will also effect the users' apps/bundles. Unless *they* use `babel-plugin-transform-react-remove-prop-types` too. I don't really know what's the best practice here, since I don't have much experience with node.
What do others do? `react-router` uses `babel-plugin-transform-react-remove-prop-types` for instance.

JedWatson:
I think that it's not safe to have `if (process.env.NODE_ENV !== 'production')` anywhere other than in `src`. We can't guarantee anything about the environment the code may be run in now or in the future, which doing so breaks.

I don't think there's any concern having proptypes in place in a node environment, if the consumer requires the lib entry. React ignores proptypes in production anyway so this is only an optimisation for client-side package size.

Given the consistency of other major packages using babel-plugin-transform-react-remove-prop-types I think that's the way to go, and let's not introduce environment specific code in our source that we don't need to (I'm happy to follow react-router's configuration for this in terms of which builds to apply the plugin to)

jochenberger:
Done. Now the only file size change is `react-select.min.js: 44246 -> 41927`.

JedWatson:
Neat, thanks @jochenberger!

Issue #2009, opened 2017-09-15
1.0.0.rc10 async "caching" error


timhwang21:
Bug: When user types faster than responses can finish, "intermediary" search queries do not ever get attempted.

Example: 
![sep-15-2017 18-31-47](https://user-images.githubusercontent.com/5831434/30505685-42eb4f6a-9a44-11e7-843a-682cb8198547.gif)

I type "asdf" while throttling the network. Queries are fired for "a", "as", "asd", and "asdf". However, when I start deleting letters, the queries for the first three search strings are never attempted. Note the placeholder text says 'Type to search' rather than 'No results.'

timhwang21:
Some more detail: it seems like part of the issue relates to the "intelligent caching" of input strings that have already been fetched. When the user types faster than the fetches can complete, it seems like "no options" are being cached for the intermediate search strings.

timhwang21:
Ah, it seems like the original issue is due to the component hitting the Github API rate limit 😄 oops.

Regardless, I think there is room for optimization here: 

Let's say the user types `'asdf'`. Queries will be sent for `'a'`, `'as'`, `'asd'`, and `'asdf'`. However, because [only the most recent callback is ran](https://github.com/JedWatson/react-select/blob/master/src/Async.js#L99), only the last result is cached, even though the first three requests may complete successfully.

@JedWatson what do you think about moving lines 104-106 outside the `if (callback === this._callback)` block? This means that completed network requests will still be cached, even if they're not used. The result would be that when the user hits delete from `'asdf'`, there won't be a redundant request sent for `'asd'`, as we've already cached that response. If it sounds good to you, I can file a PR. Thanks!

timhwang21:
A related issue occurs when handling paginated endpoints:

Let's say I do a search for `'Bob'`. Three requests are fired for `'B'`, `'Bo'`, and `'Bob'`, but only the third is cached.

Then, I clear the input and search for `'Bob'` again. Upon entering `'Bo'`, the search is re-fired, and upon entering `'Bob'`, I cache hit and a new `this._callback` is not set, so the callback for `'Bo'` is fired, resetting the options array with a new set of options.

This normally would be fine, but with paginated endpoints, the options for `'Bob'` may not be a subset for those of `'Bo'` (there may be a hundred `'Boa'`s for example). Thus, we end up in a state where the search string is `'Bob'` but the options are for `'Bo'`, which can result in `'No results found.'`. 

[...]

PR #2012, opened 2017-09-18
Async> cache async response regardless of req order

timhwang21:
Addresses #2009.

There is a WIP test -- I can't figure out how to simulate the exact
scenario that this patch fixes in the test. The current test returns
positive even without my change. Suggestions would be appreciated --
thanks!

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -96,14 +96,14 @@ export default class Async extends Component {
 		}
 
 		const callback = (error, data) => {
+			const options = data && data.options || [];
+
+			if (cache) {
+				cache[inputValue] = options;
+			}
+
 			if (callback === this._callback) {
 				this._callback = null;
-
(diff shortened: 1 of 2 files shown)


JedWatson:
Thanks @timhwang21 

This looks good to merge, I'll see what I can do to get the test working

JedWatson:
@timhwang21 I fixed the test so it works correctly now. There are a couple of changes I made (including slightly different math for the delay and different cache sets for each input) but the key here was making the test asynchronous by adding the `cb` argument and then calling it after the assertions.

timhwang21:
Hey @JedWatson ! Was wondering if you plan to cut a new release any time soon? This feature would be valuable for my use case 🙂

(Additionally, I filed this follow-up based on @youtogod 's suggestion: #2042. Thanks!)

PR #2013, opened 2017-09-19
Make this work in preact.

2betop:
Without key prop the input element unmounted and remounted after input changed, and the "onblur" handler make the panel dismissed soon.


JedWatson:
Thanks @2betop!

Let me know if you're interested in adding an example entrypoint that uses preact so we can continue to test it and support preact going forward 👍

PR #2020, opened 2017-09-22
Support Webpack 2, Webpack 3, rollup.

mschnee:
Current LTS versions of Webpack and Node honor the standardized `module` field in the `package.json`.  As I am using webpack 3 and was confused when it wasn't resolving your awesome optimized dist builds, I decided to PR the fix for y'all (and leave the old `jsnext:main` as there are plenty of systems that still use it)


JedWatson:
Thanks for picking that up @mschnee!

PR #1891, opened 2017-07-26
removeSelected prop (round 2), for optionally keeping selected values in dropdown

banderson:
I forked what @dherran did with #882 to resolve conflicts with master and to implement PR feedback from @agirton, so defer to #882 and the original issue for more details: #803 

This is in a holding state until we decide what to do here. But this should work as expected, and all tests are passing.

PR #2024, opened 2017-09-27
removeSelected: bool

firedev:
When `removeSelected` is set to `false` you can add same item multiple
times

Changed removal logic to remove values based on position


gwyneplaine:
Thanks for this @firedev, we've opted with a removeSelected PR that's more generic. Please see the following PR. https://github.com/JedWatson/react-select/pull/1891

PR #2027, opened 2017-10-02
Added id in the input select

thecreazy:
For testing with Jest it's necessary can use and custom id to search the component,
with this edit we can use the:

`wrapper.find('#our-custom-id') `

command.


JedWatson:
Good idea 👍 thanks!

PR #2034, opened 2017-10-04
Don't use `ReactDOM.findDOMNode`

ojab:
Fixes https://github.com/JedWatson/react-select/issues/899


ojab:
Apparently Option's `onFocus` is not tested, so coverage has dropped although unchanged testsuite is passed :/

ojab:
...but now we can better see actually untested parts!

src/Select.js
```diff
@@ -875,11 +871,9 @@ class Select extends React.Component {
 		}
 	}
 
-	onOptionRef(ref, isFocused) {
-		if (isFocused) {
-			this.focused = ref;
-		}
-	}
+       onOptionRef(ref) {
```
JedWatson: I think `onFocusedOptionRef` would be clearer, since the function is now only called when the option is focused, and it could really be merged into the `onFocus` handler for options.

src/Option.js
```diff
@@ -64,6 +64,7 @@ class Option extends React.Component {
 
 	onFocus (event) {
 		if (!this.props.isFocused) {
+			this.props.onOptionRef(this.ref);
```
JedWatson: Given the rest of the changes, this would be clearer if we just passed the ref as a third argument to the `onFocus` handler (below) and made the corresponding update in Select.js

yuri-sakharov:
Hi, 
Will be better when conflicts will be resolved and added unit test for this feature.
Thanks.

jnachtigall:
@ojab I hit the same issue (Interestingly only when using `react-select` locally via `npm link`). Now with version 1.1 being out would you mind cleaning this PR up to get it merged? 😄

PR #2035, opened 2017-10-04
ARIA fixes

paulmillr:


lib/Value.js
```diff
@@ -94,7 +104,9 @@ var Value = function (_React$Component) {
 				'span',
 				{ className: 'Select-value-icon',
 					'aria-hidden': 'true',
+					'tabIndex': 0,
```
StommePoes: random question: why is aria-hidden=true on an element who can get focus with tabindex=0? This can result in a sort of nameless thingie when you focus on it.

JedWatson:
Thanks for this - greatly appreciate the accessibility improvements.

Two of the inclusions here have been raised separately (see #2037 and #2036)

The remaining issue (being able to focus on values in the select) is actually something I'm exploring separately - not sure what the best design pattern is here, but I'm tempted to use something similar to the "pseudo-focus" that options get when the menu is open...

Unfortunately, this PR makes changes to the `lib` files which are overridden by the build system so it's not possible to actually preview the changes on the netlify deployment. With that in mind I'm going to close this PR since it can't be merged.

If you're keen to contribute ARIA fixes it would be really helpful to have a separate PR for each change, with changes only made to the `src` files so they can be discussed, tested and merged independently.

PR #2040, opened 2017-10-06
Remove option `addLabelText` from README and propTypes

jjeising:
The option `addLabelText` is now replace by `promptTextCreator`in
`Creatable` and only existed in documentation and propTypes anymore.


JedWatson:
Thanks for catching that @jjeising

PR #2042, opened 2017-10-08
Set isLoading to false if cache hit

timhwang21:
From @youtogod in the comments of #2012:

> Add this would be better : `this.setState({ isLoading: false, options: cache[inputValue] });`. Otherwise when hitting the cache during a promise, the loading icon will remain and the menu won't appear.


JedWatson:
Thanks for following this up @timhwang21, I'll cut a new release with this and the other fix in it shortly.

Issue #2041, opened 2017-10-07
Update examples page with code/options used...


joetidee:
Please could you update the examples page with the exact code/options that were used to generate each example:
http://jedwatson.github.io/react-select/

JedWatson:
Fixed by #2051

joetidee:
Still can't see anything on the examples page that shows the code/options used (?)

JedWatson:
It's in the source and will be published with the next version when the examples are published

Preview here: https://deploy-preview-2074--react-select.netlify.com

PR #2051, opened 2017-10-11
Adds source links to each example

damonbauer:
This addresses https://github.com/JedWatson/react-select/issues/2041


JedWatson:
Thanks @damonbauer

PR #2055, opened 2017-10-12
Skip rendering arrow wrapper when custom arrow renderer returns falsy value

### LINKED ISSUE #1700: Remove arrow and functionality
Passing a function that returns `undefined` to the `arrowRenderer` prop will remove the arrow, but not the functionality of the arrow. How can I completely remove the arrow and functionality?

mtlewis:
Closes #1700.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -837,6 +837,10 @@ class Select extends React.Component {
 		const isOpen = this.state.isOpen;
 		const arrow = this.props.arrowRenderer({ onMouseDown, isOpen });
 
+		if (!arrow) {
+			return null;
+		}
+
 		return (
 			<span
 				className="Select-arrow-zone"
--- a/test/Select-test.js
+++ b/test/Select-test.js
@@ -3782,6 +3782,24 @@ describe('Select', () => {
 		});
 	});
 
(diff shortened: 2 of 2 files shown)


lifejuggler:
Wouldn't it be better if there is a separate option to disable the arrow functionality like disableArrow: true/false instead of an override like this?

mtlewis:
I don’t have strong feelings either way, but my reasoning for doing it this way is that it avoids having two props which have to have an order of precedence. I do agree that a disableArrow prop would make the ‘no arrow’ use case more readable though!

lifejuggler:
Same for me just that it would be easier to use the default arrowRenderer when someone wants them, but I am fine with this change... just wish it would get merged ASAP

mtlewis:
Just a note that we've worked around this for our use case by adding a rule like the below, which is not as neat as having react-select support skipping arrow rendering completely, but does get the functionality to where we want it today. 

```
.myReactSelect .Select-arrow-zone {
  display: none;
}
```

JedWatson:
I think this makes sense as a fix for the problem, and is better than adding another prop.

Thanks @mtlewis!

PR #2063, opened 2017-10-15
Destructured props and state in Select,js and Value.js

gjvpaet:
Destructured props and state for more readability.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -30,7 +30,7 @@ const stringOrNode = PropTypes.oneOfType([
 let instanceId = 1;
 
 class Select extends React.Component {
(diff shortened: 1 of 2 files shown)


JedWatson:
Thanks for the PR @gjvpaet - this is something I'd love to do.

Unfortunately I'm not going to merge it at this time, because I'm working through a lot of the open PRs trying to merge as many as possible, and large refactors like this cause a lot of conflicts and make that much harder to do.

(believe me, I've been dying to do this, and also tighten the linter rules, but don't want to throw away the enormous number of fixes that have been contributed that I am triaging at the moment)

PR #2064, opened 2017-10-15
refactor componentWillUnmount()

riophae:


src/Select.js
```diff
@@ -143,11 +143,7 @@ class Select extends React.Component {
 	}
 
 	componentWillUnmount () {
-		if (!document.removeEventListener && document.detachEvent) {
-			document.detachEvent('ontouchstart', this.handleTouchOutside);
```
riophae: `detachEvent` is for supporting IE8 here. However, IE8 doesn't support touch events, right?

JedWatson: Yeah, you're right about this.

JedWatson:
Thanks for the cleanup, @riophae

PR #2066, opened 2017-10-17
add className '.Select-input--value' for easier access and styling

ivanzusko:
For easier access and styling process.

### DIFF
--- a/scss/control.scss
+++ b/scss/control.scss
@@ -150,7 +150,7 @@
 	padding-right: $select-padding-horizontal;
 	vertical-align: middle;
 
-	> input {
+	&--value {
 		background: none transparent;
 		border: 0 none;
 		box-shadow: none;
--- a/src/Select.js
+++ b/src/Select.js
@@ -810,7 +810,7 @@ class Select extends React.Component {
 		}
 		return (
 			<div className={ className } key="input-wrap">
-				<input {...inputProps} />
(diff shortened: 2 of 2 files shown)


JedWatson:
@ivanzusko this would be a breaking change for anyone with a custom stylesheet. What's the benefit here? I'm not sure what's hard about this, in terms of customising the styling...

ivanzusko:
@JedWatson First of all, thank you for your library, second of all, thank you for your reply!

There is nothing hard about styling with `> input`, but as far as I know, the browser reads CSS from right to left. So, in the case with `input` selector, it will take "more" time for a browser, because he will go through all of the inputs. Of course, there is no practical difference, but still, I think it's better to avoid of using this.

> this would be a breaking change for anyone with a custom stylesheet

Probably then do not add changes for `scss/control.scss`, but only add the className for `input`? In this case, it will only give possibility to style it, if necessary, without touching the custom stylesheet.

JedWatson:
@ivanzusko no worries!

I've reviewed this a bit more closely and `className` is actually already defined in `inputProps`, as it can be controlled by the `Select` component's `inputProps` prop - which I believe would override this.

So I'd prefer to leave this as-is for now, but if we introduce breaking changes to the stylesheet in the future I'll make sure we add an explicit className for this element.

Thanks for the PR!

PR #2067, opened 2017-10-17
Fix triggering onChange method when value is null

ozcanovunc:
Deleting value inside Select causing onChange to be triggered when content becomes empty. Which is unwanted behavior in my opinion.

### DIFF
--- a/lib/Select.js
+++ b/lib/Select.js
@@ -601,7 +601,7 @@ var Select = function (_React$Component) {
 						return i[_this3.props.valueKey];
 					}).join(this.props.delimiter) : value[this.props.valueKey];
 				}
-				this.props.onChange(value);
+				value && this.props.onChange(value);
 			}
 		}
 	}, {


JedWatson:
This doesn't make sense to me - there are several code paths that lead to the value being set to `null`, in which case react-select should absolutely call the `onChange` prop.

Why wouldn't you want to be notified when the value is cleared?

ozcanovunc:
I get notified even though I did nothing indicating I'm done editing the value.

onChange doesn't get called when the value becomes null first time but after that keep pressing backspace causes onChange to be called multiple times even though value is not changed.

Some other solution might be introduced too to solve this.

JedWatson:
I can't reproduce this. Could you please link to a demo in https://codesandbox.io that reproduces the behaviour you're talking about?

In my tests, looking at console.log output linked to `onChange`, it is only called when the value changes (including being changed from a `null` to a value, or vice-versa)

The select behaves very similarly to react inputs, calling `onChange` for every modification to the value.

If you can demonstrate the issue I'd be happy to help ensure it's fixed correctly, but in the meantime I'm going to close this PR as the changes have been made to `lib` but that's part of the build; changes need to be made to `src` so we couldn't merge this as is.

Thanks for the PR though, and please don't be discouraged :)

Issue #2073, opened 2017-10-19
Clear button displays even when value is an empty string


mariegodon:
As a result of https://github.com/JedWatson/react-select/issues/1394, the clear button now appears even when value is and empty string. This is also present in 1.0.0 release candidate. 

This is confusing since there is the placeholder displayed with the clear button, which doesn't make sense, as shown in image below.

![image](https://user-images.githubusercontent.com/18152777/31748952-df9d1b80-b444-11e7-8d0f-2529a16c7030.png)

mariegodon:
Implemented https://github.com/JedWatson/react-select/pull/2074 as potential fix

PR #2074, opened 2017-10-19
do not show clear button if value is an empty string

mariegodon:
The clear button should not display if value is an empty string

Fixes https://github.com/JedWatson/react-select/issues/2073


JedWatson:
Thanks for this @mariegodon 

Not sure why the test timed out, but it doesn't seem to be related to this so I'll merge and sort that out separately.

Issue #1826, opened 2017-06-26
How to keep value in textbox search after choosing it


cnits:
I saw option onBlurResetsInput to set whether to clear input on blur or not.

And I also want to do that after choosing a value. Please show me how to do?

Thanks in advance!

DigitalFlow:
Hey there,

just had this issue myself.
You'll have to save the selected value somewhere (preferably in your state) and then pass it as value element to the Select on the next render.

Kind Regards,
Florian

mshwery:
This would be useful – sometimes with a searchable Select the user may want to place the cursor in the currently selected value to alter it, or select part of the text to change/remove it.

The display value is misleading because it _appears_ as though you can do this ^, but in fact you cannot. It might be worthwhile to add a `keepInputValue`.

Another way to potentially go about this would be to provide a custom renderer for the _input_ itself.

Either way you'd want to use the input as the display for the selected value instead of the ValueRenderer. Although, there might be a clever way to approach this that supports both the `ValueRenderer` prop as well as keeping the selected value in the input.

jeremymarc:
any update on this?

jeremymarc:
#2075

mcpolandc:
@jeremymarc See my comment on #2075

mcpolandc:
@mshwery 

`This would be useful – sometimes with a searchable Select the user may want to place the cursor in the currently selected value to alter it, or select part of the text to change/remove it.`

I think the ability to re-edit the input after selecting a value would be an excellent enhancement to the project. I will put up a PR to this effect if you don't mind.

jeremymarc:
@mcpolandc  @gwyneplaine I don't think "backspaceRemoves" is what we are talking about ; I want, when clicking  in the search field, to select the current value so you can delete it by typing any key, something like that:

Input field without focus
https://www.dropbox.com/s/6u5nl8cfu02ay41/Screenshot%202017-12-07%2013.12.44.png?dl=0

Input field with focus:
https://www.dropbox.com/s/k3ic6h4slxe41cl/Screenshot%202017-12-07%2013.13.05.png?dl=0

That was the purpose of my PR #2075

magdalko:
anything changed to this issue? It would be awesome to have options to edit the selected text to make new search (doesnt have to write from scratch the names to search)

hollandmatt:
I've been having a similar issue to @jeremymarc - here's an example app: https://codesandbox.io/s/q27lnon96

And here's the behaviour:

[...]

PR #2075, opened 2017-10-19
Add an option (keepInputValue) to prevent clearing the text input value

jeremymarc:
I need an option to prevent clearing the text input field value - discussion #1826

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -526,11 +526,12 @@ class Select extends React.Component {
 			});
 		} else {
 			this.setState({
-				inputValue: this.handleInputValueChange(''),
+				inputValue: this.handleInputValueChange(this.props.keepInputValue ? value[this.props.labelKey] : ''),
 				isOpen: !this.props.closeOnSelect,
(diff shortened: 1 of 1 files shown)


mcpolandc:
We already have `backspaceRemoves` that when explicitly set to false should satisfy your use case, right?

gwyneplaine:
Hi @jeremymarc, really appreciate the work you've done here, however I'm closing this PR for the following reasons: 

- as mentioned by @mcpolandc, backspaceRemoves seems to fulfill your use case here. 
- adding an additional prop is not necessary, as we have onSelectResetsInput, which serves the same purpose. This needs to be extended to support single select as well, however doing so is more appealing than adding an additional prop. 
- I'm still unclear as to the use case behind this, for multi it makes sense since there's an intent and a possibility of selecting additional values with the same filter (not clearing the filter/input value so as to preserve a particular set of options makes sense here), I don't see this being the case in single. Happy to discuss this further in  #1826  👍

jeremymarc:
I will comment in #1826 so

PR #2077, opened 2017-10-19
Adds lint and test pre-commit hooks

mcpolandc:
It occurred to me while reading the contribution guide that the project might benefit from automatic push hooks to lint and test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,6 +102,6 @@
     "ui"
   ],
   "lint-staged": {
-    "*.js": ["eslint", "git add"]
+    "*.js": "eslint"
   }
 }
--- a/package.json
+++ b/package.json
@@ -38,17 +38,18 @@
     "extract-text-webpack-plugin": "^3.0.0",
     "html-loader": "^0.5.1",
     "html-webpack-plugin": "^2.30.1",
+    "husky": "^0.14.3",
     "isomorphic-fetch": "^2.2.1",
     "istanbul": "^0.4.5",
     "jsdom": "^9.12.0",
     "less": "^2.7.2",
     "less-loader": "^4.0.5",
     "less-plugin-clean-css": "^1.5.1",
+    "lint-staged": "^4.3.0",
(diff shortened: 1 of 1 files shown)


JedWatson:
@mcpolandc thanks! good idea.

Thoughts on using `husky` and `lint-staged` with a `pre-commit` hook though? (instead of pre-push, errors would be caught before commit)

mcpolandc:
Hi @JedWatson! Yeah, running the linter on only the staged files when committing makes a lot of sense. I will make changes to this effect 😄 

Just extending on this... I wonder if we should maybe add the test cover check at the commit stage as well. Maybe `husky`, `lint-staged` or some other package can provide this functionality (exit with 1 if test coverage has dropped, thus failing commit). What are your thoughts on this?

mcpolandc:
Changed this to run lint (only staged js files) and test scripts on commit rather than push.

package.json
```diff
@@ -97,5 +100,8 @@
     "react-component",
     "select",
     "ui"
-  ]
+  ],
+  "lint-staged": {
+    "*.js": ["eslint", "git add"]
```
JedWatson: What is the point of `git add` here? unless `eslint` is being run with the `--fix` flag (which it's not and I'm happy to leave it that way) this wouldn't be necessary; and it has the side-effect of preventing you from committing only staged chunks of a file without making it an obvious thing...

Surely just `eslint` on its own would be sufficient (and safer!) here?

JedWatson:
Thanks @mcpolandc! Left a note there for feedback but otherwise this looks great!

package.json
mcpolandc: I knew that I didn't want to have it fix linter errors automatically but I didn't really understand what git add was doing. Reading further down the readme of lint-staged helped with that 😛. I totally agree with what you're saying here so making the change.

JedWatson:
Great, thanks!

PR #2078, opened 2017-10-20
Updated the comment for the deleteRemoves option.

dider7:
Updated the comment for the deleteRemoves option.


gwyneplaine:
thanks for this @dider7

PR #2080, opened 2017-10-20
Muti Select: Prevent selected option with long text from line breaking

therahl:
Options with long names cause a line break in multiselect between the icon and the text.  Tried searching for a related issue, but didn't see one.

Before:
image

After: 
image


JedWatson:
Thanks @therahl - I'd *love* to merge this but unfortunately [css tables are only supported from IE11](http://caniuse.com/#feat=css-table) and at the moment, react-select supports IE10.

Definitely plan to bump this in a future major version but in the meantime, I'd encourage anyone who experiences this issue and can live with IE11+ to add their own stylesheet with the fix applied overriding the built-in react-select classes.

PR #2084, opened 2017-10-23
expose children in AsyncCreatable.js

gwyneplaine:
This PR ports over the fantastic work that @DanielHeath has done to refactor AsyncCreatable to expose the wrapped Select component as a child (as we do in Async and Creatable respectively).


JedWatson:
re: coverage, this change removes unnecessary code that was previously covered, so the change is due to having less code overall. feel free to ignore.

Issue #1476, opened 2017-01-11
Numeric Values with Multi-Select option not working


pelhage:
Unable to get multi select working with options that have integers as values. This would be particularly useful if trying to collect a list of Ids.

1. Visit http://jedwatson.github.io/react-select/ down to `Numeric Values`

2. Check `Multi-Select` from the checkboxes

3. Select an option from the drop down 

You will then see that the select component does not populate

IndifferentDisdain:
+1; this one is hurting me pretty badly at present.

paulirwin:
The example @pelhage references seems to not be working because of the "simpleValue" attribute being added. In that example, it is not simple values that are being passed in, but rather an array of objects with numeric value properties.

I confirmed that with 1.0.0-rc2 multi-select _does_ work with numeric property values, as long as you do _not_ use the simpleValue option. 

It does not seem like passing an array of numbers (rather than an array of objects with number values) works with the simpleValue option, so I think that option is broken. You can work around this by mapping the array of numbers to an array of objects with `myArray.map(i => { label: i, value: i })`

Example that works with objects with numeric values (this was scaffolded from create-react-app):

```JS
import React, { Component } from 'react';
import logo from './logo.svg';
import './App.css';
import "react-select/dist/react-select.css";
import Select from "react-select";

class App extends Component {
  constructor(props) {
    super(props);
    this.state = {
      products: [
        { label: "Foo", value: 1},
        { label: "Bar", value: 2},
        { label: "Baz", value: 3},
        { label: "Fuz", value: 4},
      ],
      selectedValues: []
    }
  }

  onChange(values) {
    this.setState({
      selectedValues: values
    });
  }

  render() {
    return (
      <div className="App">
        <div className="App-header">
          <img src={logo} className="App-logo" alt="logo" />
          <h2>Welcome to React</h2>
        </div>
        <form>
          <div style={{ width: 300, margin: "1em"}}>
            <Select multi options={this.state.products} onChange={this.onChange.bind(this)} value={this.state.selectedValues} />
          </div>
          <ul style={{ width: 300 }}>
            {this.state.selectedValues.map(i => <li>{i.value}</li>)}
          </ul>
        </form>
      </div>
    );
  }
}

export default App;
```

amitripshtos:
I'm having same issue and it's quite critical since most of the time i need to send to the form list of IDs - which are numbers

newswim:
Would this PR resolve the issue?
[...]

PR #2086, opened 2017-10-23
Fix/numeric multi select

gwyneplaine:
Added string coercion to condition in expandvalue()
Fixes the bug referenced in #1600 and in #1476


tauren:
Any chance we can get a release cut with this in it?

gwyneplaine:
@tauren yup, I've talked to @JedWatson , and the intention is to get a new release cut within the next couple of days. Stay tuned

PR #2089, opened 2017-10-24
onOpen: run after dom has rendered

macobo:
This PR changes the onOpen event handler to be called after the dom has rendered and refs have been initialized.

This is useful since one of the uses of `onOpen` is to change the menu css dynamically as it opens. Currently this requires an `setInterval` hack which can result in flickering.

Any and all feedback is welcome! :)

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -99,14 +99,6 @@ class Select extends React.Component {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


JedWatson:
Good call @macobo, thanks!

Issue #2096, opened 2017-10-25
Expose Option for import


bebepeng:
Hi! 
Would it be possible to expose the base Option in the index.js? I would like to be able to use the internal Option since i dont have a usecase for customizing the handleMouseDown, handleMouseEnter, and handleMouseMove methods. I see that the base Option already take children.

I would like to be able to do something like this...

```
<Select
      ...
      optionComponent={MyExtendedOption}
 />
```

```
const MyExtendedOption = (props) => {
    return (
        <Option {...props}>
            <div>{props.value.label.extracustomstuff}</div>
            <OTHERCOMPONENT/>
        </Option>
    );
};
```

This is similar to how one might use the exported "Value" component currently.

If you are open to pull request, I could also create one. Thanks!

jperkelens:
I happened to be working on opening this issue with a PR when you submitted it, I think this is what you were asking for.

tsteele2013:
I'm currently on the roadmap to V2 migration, but any way to enable this in V1?

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue  is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

PR #2097, opened 2017-10-25
Exports Option Component

jperkelens:
Exports the Option Component to assist in minor customizations. (Addresses: #2096)


gwyneplaine:
thanks for this @jperkelens, this has been merged into master

PR #2098, opened 2017-10-25
Add keyboard accessibility to multiselect

jasonchangxo:
I came across this issue while working on accessibility items throughout our app. The "x" icon on each of the multiselect pill values are currently not accessibly via keyboard. Pressing backspace to delete the value is a viable workaround, but is not obvious for a user who cannot see the screen and relies on a screen reader.

### DIFF
--- a/src/Value.js
+++ b/src/Value.js
@@ -8,6 +8,7 @@ class Value extends React.Component {
 		super(props);
 
 		this.handleMouseDown = this.handleMouseDown.bind(this);
+		this.onKeyDown = this.onKeyDown.bind(this);
 		this.onRemove = this.onRemove.bind(this);
 		this.handleTouchEndRemove = this.handleTouchEndRemove.bind(this);
 		this.handleTouchMove = this.handleTouchMove.bind(this);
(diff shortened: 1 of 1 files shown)


JedWatson:
@seancurtis can you take a look at this please?

JedWatson:
Thanks for the PR @jasonchangxo 

I've had a look at the implementation and thought quite a bit about it; what I'd like to do long term is provide a way to "pseudo-focus" selected values using the keyboard instead of making them tab focusable (similar to how menu items are handled)

With that in mind I'm not going to merge this, but you and anyone else could absolutely implement it using the `valueComponent` prop if this is the behaviour you want in your app.

JedWatson:
Just a note; I've linked to this PR from the Readme with a note about the current accessibility limitations.

src/Value.js
```diff
@@ -28,6 +29,14 @@ class Value extends React.Component {
 		}
 	}
 
+	onKeyDown (event) {
+		if (event.keyCode === 13 || event.keyCode === 32) {
```
Suhas010, line 33: We should define key codes in constants with proper names

PR #2100, opened 2017-10-26
fix not clearing when given invalid values, from #1756

gwyneplaine:
Cleaning up and adding tests to @ticklemynausea's initial PR https://github.com/JedWatson/react-select/pull/1756


JedWatson:
Thanks

@gwyneplaine it looks like this can be optimised so we don't need to call `getValueArray` in every render - can you look into a subsequent pr that fixes this?

PR #2108, opened 2017-10-28
removing getInputValue function

uxtx:
Was doing some cleanup and testing work to try to bump up the coverage and ran across this unused `getInputValue` function - don't see it called anywhere, react-select seems to compile and tests run perfectly without it - looks like it was part of the initial work for `allowCreate` (https://github.com/JedWatson/react-select/commit/7fe723f1195d4dd5012ffb7f5c6b4b383daf5172) , but it seems unused now.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -701,10 +701,6 @@ class Select extends React.Component {
 		return this._focusedOption;
 	}
 
-	getInputValue () {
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks for this @uxtx, much appreciated!

PR #2121, opened 2017-10-31
Extracting and using directly the findDOMNode.

leonascimento:


JedWatson:
Thanks @leonascimento

Issue #2128, opened 2017-11-03
`onSelectResetsInput=false` is not respected


kevinhos:
`react-select` works really well in our project so far. Really great component.

I noticed `onSelectResetsInput` doesn't seem to be respected. On selection, or on creation and selection, the `onInputChange` handler is called with an empty value instead of the current input value. Example:

https://codesandbox.io/s/j7zkqq5l69

Steps to reproduce:
* type into input
* select an option OR create an option
* notice the console output after selection `Input:''`

Please let me know if there's some part of the API I misunderstand.

kevinhos:
I see now this is only an option for `multi`.

PR #2129, opened 2017-11-03
check onSelectResetsInput prop for non-multi

kevinhos:
Address issue #2128 in which the `onSelectResetsInput` prop is not respected.


kevinhos:
This is apparently an option only for `multi`.

PR #2140, opened 2017-11-10
Creatable: when `promptTextCreator` returns only the label, changes don't fire on first Enter key press

karaggeorge:
#### Reason
It was caused because when creating a new choice, there is a check to see if the option is unique (even though there is a check for that to even display the new option creator). This check was checking the value against a list of filtered options. This filter was the passed down filter prop and was basically checking the option against itself with the addition of the prompt text. Therefore, by accident if there was a prompt text it was passing, but when using `(label) => label`, it was failing.

#### Solution
It has been fixed to check against the list of options available.

### DIFF
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -89,7 +89,7 @@ class CreatableSelect extends React.Component {
 	}) {
 		const { isOptionUnique } = this.props;
 
-		options = options || this.select.filterOptions();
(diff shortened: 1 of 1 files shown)


JedWatson:
Great, thanks @karaggeorge 

👍 props for both the fix and explanation!

PR #2144, opened 2017-11-14
Added new prop 'preferStartMatch'

forum-is:
New property `preferStartMatch` to prioritize options that match a given filter input at the **beginning** of the value/label. Options that start with the filter input will be listed first, other matches follow. This offers more flexibility than just setting `matchPos = "start"`.

Also added to tests with `preferStartMatch=true` and both, `matchProp` set to either `"value"` or `"label"`.


forum-is:
After discussion we will try amending the existing prop `matchPos` instead of introducing a new property.

PR #2147, opened 2017-11-15
New option for property matchPos: preferStart

forum-is:
New option for property `matchPos` to prioritize options that match a given filter input at the **beginning** of the value/label. Options that start with the filter input will be listed first, other matches follow. This offers more flexibility than just setting` matchPos = "start"`.

Also added to tests with `matchPos="preferStart"` and both, `matchProp` set to either `"value"` or `"label"`.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -11,7 +11,7 @@ import classNames from 'classnames';
 
 import defaultArrowRenderer from './utils/defaultArrowRenderer';
 import defaultFilterOptions from './utils/defaultFilterOptions';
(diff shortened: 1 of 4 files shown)


JedWatson:
Thanks for the PR @forum-is, but since this is already possible to do in user-land I'm inclined to leave the core library simple. This adds a fair bit of code duplication too, which I'm not keen on.

It might be nice to add an example that implements this though, so other users can easily see and reproduce the behaviour?

PR #2154, opened 2017-11-20
Export default util functions

Deiru2k:
This is useful when trying to alter default behavior of react-select without necessarily re-inventing things already in use in react select.
In fact, we're currently using copy-pasted defaultOptionRenderer from react selects code.
Things would be way nicer if you could also import such functions from react-select package.


JedWatson:
Good idea @Deiru2k

Issue #2127, opened 2017-11-02
openOnClick prop doesn't work


elghali:
Hi, I don't want the menu to open on click but only when the user writes some characters that do exist in the `options`, so I tried to use the `openOnClick={false}` prop but it didn't do the job, so I wrapped the `Select` comp within a `span` which doesn't allow it to open and it works, but the problem is that now even when the arrow is clicked it won't open either,
```  
     dontOpenOnClick = () => {
    this.conditionSelector.select.closeMenu();
  }
   <span onFocus={this.dontOpenOnClick}>
                <Select.Creatable
                  ref={(node) => this.conditionSelector = node}
                />
   </span>

**Comment by CodeMeNatalie** [comment]
+1

**Comment by karaggeorge** [comment]
It seems like the implemented functionality for this is that `openOnClick` is talking about opening the menu when clicking on a not-focused select. This behavior works. If you click a not-focused Select, the menu will not open, but the Select will get focused. If, however, you click a focused Select, the menu will open normally.

Willing to write a PR if that was not the desired behavior.

**Comment by elghali** [comment]
I really can't seem to find the use case, this is my test code
```             <Select 
              options={optionsSelect}
              name="form-field-name"
              value="one"
              openOnClick={false}
            />
```
and I tried to click on the select while it's not-focused, but it immediately opens, so I can't find how does the behavior really works

karaggeorge:
@elghali 
Here's a very simple test case: https://plnkr.co/edit/coSxiEvUgxjdwedhMtyK?p=preview
Try clicking on the select. It will focus it, but won't open it

elghali:
ok, but now even when you click on the arrow it doesn't open if it doesn't already have focus! and that's the main problem

karaggeorge:
That's what I was saying in my first comment. I don't know what the purpose of the prop was in the first place. If the prop is there to make sure that the first click on the component focused it but doesn't open it, then not opening on the arrow click would also be acceptable.

elghali:
Exactly

gwyneplaine:
hi @elghali the intention of the **openOnClick** prop is to enable the behaviour where clicking on an unfocused select will open the menu, as @karaggeorge has correctly surmised. Therefore setting **openOnClick** to false won't do what you're expecting, the example case that was provided above is very much the expected behaviour.

JedWatson:
Closed by #2165 

[...]

PR #2165, opened 2017-11-23
handleMouseDownOnArrow fix when openOnClick prop is set to false

elghali:
this pull request is intended to fix the issue #2127, now if I set the `openOnClick` prop to `false`, the selectable opens when the arrow is clicked but not if the field is clicked


JedWatson:
The original intention of `openOnClick` was to block opening unless the field is focused, but you've convinced me - we should toggle the open state when the dropdown icon is clicked regardless of the prop.

Thanks for the PR!

PR #2170, opened 2017-11-25
Add conditional scrolling into view of focused option

mtlewis:
Previously the focused option would be scrolled to the top of the scroll window when the menu is opened, regardless of whether it was already in view.  An example of a situation where this is problematic is when a list starts with one or more disabled items. In this situation, the disabled items are scrolled out of view when the menu opens, and because they're at the top of the list it's hard to notice they're there.

This PR modifies the behavior such that when the menu is opened, it will only scroll the focused option into view if it's not already fully in view.

The current behavior is reproducible in the live demo:
1. Open the live demo at http://jedwatson.github.io/react-select/
2. Change the states demo to the US dataset
3. Without typing, open the menu by clicking the input.
4. Notice that it's hard to tell that there's a disabled option at the top of the list, especially in environments where the scrollbar is not always visible.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -108,7 +108,19 @@ class Select extends React.Component {
 		if (this.menu && this.focused && this.state.isOpen && !this.hasScrolledToOption) {
 			let focusedOptionNode = findDOMNode(this.focused);
 			let menuNode = findDOMNode(this.menu);
-			menuNode.scrollTop = focusedOptionNode.offsetTop;
+
+			const scrollTop = menuNode.scrollTop;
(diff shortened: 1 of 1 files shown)


mtlewis:
If this behavior is not something we can impose on everyone, I'd be happy to look at adding a prop controlling whether we do this check or not.

JedWatson:
This looks like a good change, I'm happy to make it across the board 👍 

Thanks for the PR and especially the clear instructions to reproduce! Made this really easy to test and merge.

mtlewis:
No worries! Thanks @JedWatson for the quick turnaround, will look out for other ways to contribute more in the future.

PR #2171, opened 2017-11-25
Update code styles in Creatable.js

Chopinsky:
There're a couple of code style issues in `Creatable.js` that should probably be addressed for an easy win (warnings on those code have been thrown when running lint on my test project): 1. removing an unnecessary space when binding `onOptionSelect` to the control instance; 2. removing `options` variable in function `filterOptions` since it's not used; 3. adding default condition to the switch statement in function `shouldKeyDownEventCreateNewOption`.

No functional changes are expected from this PR.


JedWatson:
Thanks @Chopinsky, much appreciated!

Issue #1397, opened 2016-11-29
Feature Request: Support for Chrome autofill


casschin:
I have some address fields that I want to be able to complete with [Chrome's autocomplete feature](https://developers.google.com/web/updates/2015/06/checkout-faster-with-autofill), but I'm having trouble getting it to select a value in my state input that's using react-select. I've tried swapping it out with a native select input and the autofill works properly there.

I also tried using autocomplete on the [demo](http://jedwatson.github.io/react-select/) inputs and wasn't able to get it working there.

Is this a known issue?

Ryanthegiantlion:
Had the same issue. I see that the component uses a hidden type input field which chrome will not autofill.

A workaround I used is to wrap the select component and place an absolutely positioned text input above with zero width, height and opacity. Then set the select input value in the on change of the hidden input.

shreyansqt:
Hey guys, I intend to work on a fix for this. Any inputs from the contributors would be appreciated.

Ryanthegiantlion:
So one potential avenue is suggested by the workaround I mentioned above. There is a hidden input which is used to store the value and make sure the data is posted back with a form submission. Look into hiding the input via css rather than creating an input with hidden type. It does seem like a cleaner solution to keep the hidden input but I just saw no way with my brief investigation to get chrome autofill to play well with that.

jzaefferer:
For "hiding via CSS", this snippet might be useful: https://github.com/jquery/jquery-ui/blob/74f8a0ac952f6f45f773312292baef1c26d81300/themes/base/core.css#L17-L26

Might be overkill in this case, since this isn't hiding-for-accessibility.

> A third option would be to try and get chrome to play well with the existing hidden input but afaik you can't do that.

Chrome not filling out hidden inputs is probably a security feature, to prevent harmless looking forms (like a search) to capture important data like credit card details.

mmintel:
I am currently working on a fix for this issue and created a pull-request over here: https://github.com/JedWatson/react-select/pull/2174. Any help is very appreciated :)

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you feel this issue / pull request is still relevant and you'd like us to review it, please leave a comment and we'll do our best to get back to you.

jzaefferer:
[...]

PR #2174, opened 2017-11-27
Adds support for browser autofill

mmintel:
Hi there,

I am preparing a fix for https://github.com/JedWatson/react-select/issues/1397, if you'd like to help me you can find my progress over here: https://github.com/sevenval/react-select.

What I basically changed was just updating the inputValue when the hidden input changes. To help with styling I am also adding a class "is-autofilled" in that case.

The autofilling itself is already working but I still have some styling issues. The problem is that Chrome ignores many css rules when autofill was applied. It's really hard to recreate the "natural autoComplete behaviour" because Chrome already applies the yellow background to inputs when it's just "suggesting" values, but I didn't find a way to react to those suggestions via javascript.

nov-27-2017 11-40-48

Also it currently only works `autosize={false}` because this adds a `width="5px"` on the input field, still have to figure out why this is needed and how I can optimize this.

Cheers!


jzaefferer:
If I understand this correctly, finishing this without the autocomplete styling would be a lot easier (since that is already done)?

Recreating the native autocomplete styling seems unnecessarily complicated. It doesn't seem necessary, and getting the styling right across platforms (browser, OS) might be really hard. 

Maybe we can start with a PR without the styling and get that merged (or we publish a fork for a while and test that in production).  Meanwhile, do a second PR that deals with the styling. If the project owners find that unnecessary or too risky, it can be left out.

mmintel:
If I remove all style changes except for the required `.Select-hidden` class it looks like this:
nov-28-2017 11-01-09

JedWatson:
Thanks for working ont his @mmintel!

Just a heads up, you seem to be working in the `dist` and `lib` files; you want to be changing only the `src`, `less` and `scss` files (both stylesheet languages need to be kept in sync). Everything in `lib` and `dist` are build artefacts and should be excluded from PRs.

mmintel:
@JedWatson I am only working on the src files, the changes from the dist files come from `yarn build`

jzaefferer:
> If I remove all style changes except for the required .Select-hidden class it looks like this:

Gotcha, that's definitely worse. Hopefully you can get the proper styles working then.

mmintel:
Turns out it's more than complicated to get the styling right, but finally I found this "hacky" way to detect autofill suggestions: https://medium.com/@brunn/detecting-autofilled-fields-in-javascript-aed598d25da7.  Still working on it but it looks much better now.

mmintel:
I'm already pretty close to native browser behaviour. I will now have a closer look to the differences between `multiselect` and `autosize`.
dec-06-2017 15-28-32

mmintel:
@JedWatson @jzaefferer I pushed my recent changes including some unit tests that "should normally" work but I need to test them without a `value` on the select. 

So I was trying to set the wrapper like this:
```JS
wrapper = createControlWithWrapper({
  value: false,
  autoComplete: 'test',
  multi: false,
  name: 'field',
  options: options,
  simpleValue: true,
});
```

And I am always getting this error message:
```
     TypeError: Cannot read property 'tagName' of null
      at Object.change (node_modules/react-dom/lib/ReactTestUtils.js:323:35)
      at simulateAutoFill (test/Select-test.js:123:22)
      at Context.<anonymous> (test/Select-test.js:660:4)
```

when I set it like this it works but is not what I am trying to test:
```JS
wrapper = createControlWithWrapper({
  value: 2,
  ...
});
```

same result when I try to change the `value` via `wrapper.setPropsForChild`.

It's crashing here:
```JS
var simulateAutoFill = (text) => {
  TestUtils.Simulate.change(getHiddenInput(instance), { target: { value: text } });
};
```

because `getHiddenInput(instance)` returns `null` when setting the `value` to `false`, which I dont understand because it seems like the hidden input is not even rendered in this case.. can't see any condition in the code that could lead to this result.

Help really appreciated! :)

src/Select.js
```diff
@@ -596,12 +651,39 @@ class Select extends React.Component {
...
 		this.setValue(this.getResetValue());
+		this.clearAutofilled();
 		this.setState({
 			isOpen: false,
 			inputValue: this.handleInputValueChange(''),
 		}, this.focus);
 	}
 
+	setAutofill () {
+		const hasValue = this.getValueArray(this.props.value).length;
```
jzaefferer: Why is this line here? `hasValue` isn't used, `getValueArray` doesn't seems to have any side effects.

examples/src/components/AutoComplete.js
```diff
@@ -0,0 +1,79 @@
...
+	toggleCheckbox (e) {
+		let newState = {};
+		newState[e.target.name] = e.target.checked;
+		this.setState(newState);
+	},
+	render () {
+		var options = [
+			{ value: 'UK', label: 'United Kingdom' },
+			{ value: 'ES', label: 'Spain' },
+			{ value: 'DE', label: 'Germany' },
```
jzaefferer: We should probably add a few more countries here, so that users and other contributors can also test this with their specific autofill country.

examples/src/components/AutoComplete.js
```diff
@@ -0,0 +1,79 @@
...
+import PropTypes from 'prop-types';
+import Select from 'react-select';
+
+const STATES = require('../data/states');
+
+var AutoCompleteField = createClass({
+	displayName: 'AutoCompleteField',
+	propTypes: {
+		label: PropTypes.string,
+		searchable: PropTypes.bool,
```
jzaefferer: This isn't used. `autoComplete` is used, but missing

examples/src/components/AutoComplete.js
```diff
@@ -0,0 +1,79 @@
...
+			country: newCountry,
+			selectValue: null,
+		});
+	},
+	updateValue (newValue) {
+		this.setState({
+			selectValue: newValue,
+		});
+	},
+	toggleCheckbox (e) {
```
jzaefferer: This looks like it was copied from another demo, but isn't needed here.

less/control.less
```diff
@@ -128,6 +133,15 @@
 		direction: rtl;
 		text-align: right;
 	}
+
+	&.is-autofilled > .Select-control,
+	&.is-autofill > .Select-control,
+	&.is-autofill.is-focused .Select-input,
+	&.is-autofill.is-focused:not(.is-open) > .Select-control {
+		@media (-webkit-min-device-pixel-ratio:0) {
+			background-color: #faffbd;
```
jzaefferer: Let's add a comment here explaining this particular color value, like
```
// simulate Chrome's native autofill styling
```

scss/control.scss
```diff
@@ -285,8 +285,18 @@
 	position: relative;
 }
 
+// visually hidden fields
 
-
+.Select-hidden {
```
jzaefferer: Is this file supposed to have the same content as the `less` file? If yes, its missing a bunch of changes.

src/Select.js
```diff
@@ -59,10 +64,17 @@ class Select extends React.Component {
 			'onOptionRef',
 			'removeValue',
 			'selectValue',
+			'setAutofill',
```
jzaefferer: Looks like these four don't have to be here, since they're all called directly, and not used as event handlers. So no need for binding them.

src/Select.js
```diff
@@ -59,10 +64,17 @@ class Select extends React.Component {
...
+			'clearAutofill',
+			'setAutofilled',
+			'clearAutofilled',
 		].forEach((fn) => this[fn] = this[fn].bind(this));
 
+		this.autoCompleteEnabled = !this.props.multi && this.props.autoComplete;
```
jzaefferer: This looks like it should be a method that get's called where needed, since these props could change and the constructor runs only once.

src/Select.js
```diff
@@ -147,6 +162,22 @@ class Select extends React.Component {
 		this.toggleTouchOutsideEvent(false);
 	}
 
+	handleAutoFillAnimation ({ target, animationName }) {
```
jzaefferer: `target` isn't used, can it be removed?

src/Select.js
```diff
@@ -147,6 +162,22 @@ class Select extends React.Component {
...
+		} else {
+			this.clearAutofill();
+			this.clearAutofilled();
+		}
+	}
+
+	addAutoFillListener () {
+		// hack to receive -webkit-autofill event
+		// @see https://medium.com/@brunn/detecting-autofilled-fields-in-javascript-aed598d25da7
+		this.inputField && this.inputField.addEventListener('animationstart', this.handleAutoFillAnimation, false);
```
jzaefferer: I prefer using `if (this.inputField) ...` for this kind of expression. If that's common in this repo, I guess its fine to leave as-is.

src/Select.js
```diff
@@ -1084,6 +1185,7 @@ Select.propTypes = {
 	'aria-labelledby': PropTypes.string,  // HTML ID of an element that should be used as the label (for assistive tech)
 	arrowRenderer: PropTypes.func,        // Create drop-down caret element
 	autoBlur: PropTypes.bool,             // automatically blur the component when an option is selected
+	autoComplete: PropTypes.string,				// support for form auto completion
```
jzaefferer: comment indent doesn't line up anymore

scss/control.scss
jzaefferer: PS: this is resolved. I miss GitLab's "Mark resolved" feature here.

examples/src/components/AutoComplete.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import createClass from 'create-react-class';
+import PropTypes from 'prop-types';
+import Select from 'react-select';
+import { COUNTRIES } from '../data/countries';
```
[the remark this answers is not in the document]
mmintel: I implemented it this way because other files in /data had the same structure

examples/src/components/AutoComplete.js
```diff
@@ -0,0 +1,56 @@
...
+
+var AutoCompleteField = createClass({
+	displayName: 'AutoCompleteField',
+	propTypes: {
+		autoComplete: PropTypes.string,
```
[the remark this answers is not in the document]
mmintel: Ah right

src/Select.js
```diff
@@ -147,6 +157,26 @@ class Select extends React.Component {
 		this.toggleTouchOutsideEvent(false);
 	}
 
+	isAutoCompleteEnabled() {
+		return !this.props.multi && this.props.autoComplete;
```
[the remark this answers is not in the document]
mmintel: Good idea

jzaefferer:
> This branch has conflicts that must be resolved

Could you look into those?

Also for new props a readme update would be great.

Since this is currently mostly-Chrome, do you see any way to make this work in other browsers as well?

examples/src/components/AutoComplete.js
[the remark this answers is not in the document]
jzaefferer: ok *mark as resolved*

src/Select.js
jzaefferer: nevermind

mmintel:
@jzaefferer I updated the README and resolved the conflicts. However I don't know how to test this with Safari and Firefox, because for me autofill doesn't work at all in those browsers. In Safari it just shows me an icon for autofill suggestions but that doesn't do anything (although the information in my contacts are filled in correctly). Firefox doesn't give me any suggestions, also there seem to be no way to edit saved autofill values in Firefox. Also tried this with other forms to validate it's not only a problem with react-select.

README.md
```diff
@@ -1,30 +1,24 @@
...
-[![Build Status](https://travis-ci.org/JedWatson/react-select.svg?branch=master)](https://travis-ci.org/JedWatson/react-select)
-[![Coverage Status](https://coveralls.io/repos/JedWatson/react-select/badge.svg?branch=master&service=github)](https://coveralls.io/github/JedWatson/react-select?branch=master)
-[![Supported by Thinkmill](https://thinkmill.github.io/badge/heart.svg)](http://thinkmill.com.au/?utm_source=github&utm_medium=badge&utm_campaign=react-select)
+[![NPM](https://img.shields.io/npm/v/react-select.svg)](https://www.npmjs.com/package/react-select) [![CDNJS](https://img.shields.io/cdnjs/v/react-select.svg)](https://cdnjs.com/libraries/react-select) [![Build Status](https://travis-ci.org/JedWatson/react-select.svg?branch=master)](https://travis-ci.org/JedWatson/react-select) [![Coverage Status](https://coveralls.io/repos/JedWatson/react-select/badge.svg?branch=master&service=github)](https://coveralls.io/github/JedWatson/react-select?branch=master) [![Supported by Thinkmill](https://thinkmill.github.io/badge/heart.svg)](http://thinkmill.com.au/?utm_source=github&utm_medium=badge&utm_campaign=react-select)
```
jzaefferer: Why did these change?

README.md
```diff
@@ -1,30 +1,24 @@
...
 
-React-Select
-============
+# React-Select
```
jzaefferer: Or this? Or the various other minor changes that are unrelated? I don't think we should mix a more general cleanup into this PR.

README.md
```diff
@@ -336,6 +336,29 @@ function onInputKeyDown(event) {
 />
 ```
 
+### Supporting browser autofill
+
+Add the `autoComplete` prop to your select. Please notice that this feature is ignored for a Multiselect. The `autosize` prop should be set to `false` because autosize will shrink the input so the clickable area to trigger autofill suggestions does not expand over the whole select field.
```
jzaefferer: Let's link to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autocomplete for a list of values.

```
The value should be from [this list of possible values](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autocomplete).
```

README.md
```diff
@@ -336,6 +336,29 @@ function onInputKeyDown(event) {
...
+	simpleValue
+	name="country"
+	value={this.state.selectValue}
+	onChange={this.updateValue}
+	autoComplete="country"
+	onAutoFill={this.handleAutoFill}
+/>
+```
+
+You can use the `onAutoFill` callback to react to autofill state changes.
```
jzaefferer: What's the usecase for that?

README.md
jzaefferer: Could also give some examples where a select is commonly used, like 'cc-exp-month', 'cc-exp-year'

README.md
mmintel: if you want to react to autofill state change on another level.. for example if you want to transport an `is-autofilled` to a parent node.

jzaefferer: `to autofill state changes, for example to adjust the styling of a wrapper component.`?

mmintel:
@JedWatson would you mind to review this again? We've put a lot of work into this in the last weeks and also implemented this in one of our projects successfully. The "coveralls" check is complaining about a decreased coverage but I've implemented some tests to make sure autoComplete is working so that should not be the problem. You will find a quick explanation in the README and you can try it out yourself in the examples. Please let us know :)

jzaefferer:
Looks like this branch has conflicts again. Would be nice to know if this could get merged at some point soon - the longer this stays open, the more conflicts will pile up.

davidcoleman007:
please merge this.  How can I help?  we need it.

davidcoleman007:
@JedWatson @jzaefferer @mmintel I have resolved the conflicts and tested it, my work is here:
https://github.com/davidcoleman007/react-select

Can we get this merged in please?

I'll raise my own PR if that is what you wish.

davidcoleman007:
@CarlosOlave I have raised a clean PR with no conflicts which is also still pending:
https://github.com/JedWatson/react-select/pull/2395

bladey:
Hi all,

I'm closing this issue in favor of #2395 which is more up to date with a newer PR.

In an effort to sustain the `react-select` project going forward, we're closing issues that appear to now be redundant based on their age.

However, if you feel this issue is still relevant, please leave a comment and we'll do our best to get back to you, and we'll re-open it if necessary.

PR #2175, opened 2017-11-27
Make options DRAGGABLE!

memorylorry:
In some cases, we wanna make our table columns orderable, when I dynamically choose our databases metris. So I try to use jquery-ui plugins to make options could be draggable. Here is my code, I have add an draggable example in index page. Check it.


JedWatson:
Thanks for the PR @memorylorry but I'm not keen on including jQuery and jQuery UI in the package.

You should be able to implement this without changing react-select itself by using a custom `valueRenderer` method, if not we can look at making that more flexible so you can.

Also just a note for future PR's, please only include changes to the `src` folder, not the build (lib / dist)

memorylorry:
OK. I have a try.

PR #2179, opened 2017-11-28
Fixed search for invalid label and/or value

yuri-sakharov:
We have a results in dropdown if user search "undefined" word when value and/or label are undefined or if search "null" word when value and/or label are null.

### DIFF
--- a/src/utils/defaultFilterOptions.js
+++ b/src/utils/defaultFilterOptions.js
@@ -1,6 +1,10 @@
 import stripDiacritics from './stripDiacritics';
 import trim from './trim';
 
+function isValid(value) {
+	return typeof (value) !== 'undefined' && value !== null && value !== '';
+}
+
 function filterOptions (options, filterValue, excludeOptions, props) {
 	if (props.ignoreAccents) {
 		filterValue = stripDiacritics(filterValue);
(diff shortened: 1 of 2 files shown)


src/utils/defaultFilterOptions.js
```diff
@@ -20,24 +20,35 @@ function filterOptions (options, filterValue, excludeOptions, props) {
 		if (excludeOptions && excludeOptions.indexOf(option[props.valueKey]) > -1) return false;
 		if (props.filterOption) return props.filterOption.call(this, option, filterValue);
 		if (!filterValue) return true;
-		var valueTest = String(option[props.valueKey]);
-		var labelTest = String(option[props.labelKey]);
+
+		var value = option[props.valueKey];
+		var label = option[props.labelKey];
+		var hasValue = typeof (value) !== 'undefined' && value !== null && value !== '';
+		var hasLabel = typeof (label) !== 'undefined' && label !== null && label !== '';
```
liorbentov: Maybe extract this logic into a function?

src/utils/defaultFilterOptions.js
```diff
@@ -20,24 +20,35 @@ function filterOptions (options, filterValue, excludeOptions, props) {
...
+
+		if (!hasValue && !hasLabel) {
+			return false;
+		}
+
+		var valueTest = hasValue ? String(value) : null;
+		var labelTest = hasLabel ? String(label) : null;
```
liorbentov: Why not include this in the hasX checks?

yuri-sakharov: It's because there is situation when only one of label or value is invalid and still need to search in valid field.

JedWatson:
Thanks for this @yuri-sakharov! Great PR 👍

yuri-sakharov:
@JedWatson thank you too.

Issue #2155, opened 2017-11-20
Cannot clear value manually (not truly controlled) when there is no match


zivester:
It seems that the component is not controlled when the search result is not a match of the options.

Example: https://plnkr.co/edit/b85HMDbt0jUZjkwCE7jG?p=preview

Using `1.0.0-rc.10`.

I use the component to do a complex search from a number of filters (`react-select` being one of them).  My external function is unable to clear the component by simply settings `value={}` to the empty string/undefined/null.

agirton:
Hi @zivester value should be a `falsey` value. In your example you are just setting the label and value to an empty string. You should actually be setting the `value` of React `state` to `null`. Does that help?

zivester:
No that doesn't work. I've tried numerous combinations of setting value to `null`, `undefined`, and empty string but none of them work for me.  Feel free to play with the plnkr, I'm pretty sure it's a bug.

yuri-sakharov:
Hi, 
I fixed it and it's waiting to marge https://github.com/JedWatson/react-select/pull/2179

JedWatson:
@zivester can you confirm whether this is fixed by #2179? preview here: https://deploy-preview-2179--react-select.netlify.com/

zivester:
Doesn't seem to be working for this bug.  It doesn't ever seem to clear the display "label" value, but it does clear the `x` clear button on certain values:

`value` equal to `null`, `undefined` or the empty string will hide the `x` button.  Display string remains unchanged.

`value` as an object with `value.value` as `null`, `undefined` or the empty string does nothing.

I'm not sure what is desired with the new `1.0.0` release, but I thought it was suppose to support the full object syntax only now? `value = { label: 'Label', value: 'value' };`? Which doesn't seem to be affected by that commit at all.

I also wasn't sure how to use the preview you linked, so I just tested locally with `master` branch w/ that most recent commit.

agirton:
Hi @zivester the example you gave us includes some incorrect code. 

The `onChange` handler should not be
```
onChange(val) {
      console.log("onChange: " + val);
      const value = { label: val || '', value: val || '' };
      this.setState({ value });
  }
```
but 
```
onChange(val) {
      console.log("onChange: " + val);
      this.setState({ value: val });
  }
```
The reason is because you are still passing in an object and we don't do a diff to check if all of the values on the object are empty.

zivester:
[...]

PR #2183, opened 2017-11-30
Clear input value on receiving props with another value.

yuri-sakharov:
Hi,
Should be fixed this bug https://github.com/JedWatson/react-select/issues/2155.
Thanks.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -101,6 +101,10 @@ class Select extends React.Component {
 			// Used to be required but it's not any more
 			this.setState({ required: false });
 		}
+
+		if (this.state.inputValue && this.props.value !== nextProps.value) {
+			this.setState({ inputValue: this.handleInputValueChange('') });
+		}
 	}
 
 	componentDidUpdate (prevProps, prevState) {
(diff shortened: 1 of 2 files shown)


JedWatson:
Solid fix @yuri-sakharov, thanks!

yuri-sakharov:
@JedWatson thank you too.

jharris4:
FYI this was a breaking change for me (and took me a little while to track down after upgrading!)

My value prop (actually a list of values since I'm in multi mode) are coming from an immutable collection which I was converting to an array of values in a render function.

I had to refactor things to instead do the conversion in componentWillReceiveProps (only if the immutable collection changed) and assign to state for use in the render function.

I can see the value of this PR, but it might be helpful to draw attention to it in the docs or something in case other people get tripped up by this as well.

yuri-sakharov:
Hi, @jharris4 
Thank you for comment I fixed it in https://github.com/JedWatson/react-select/pull/2299

hiredgun:
@yuri-sakharov was this fix merged?

yuri-sakharov:
@hiredgun I don't know and don't have rights for merge. The question should be addressed to collaborators.

Issue #2184, opened 2017-11-30
Default isOptionUnique could crash on null options


Chopinsky:
*   _What's your use case? (Tell us about your application and what problem you're trying to solve.)_
When using creatable select control with custom `loadOptions` and default `isOptionUnique`, the provided options property could end up with null or undefined. This will cause problems if user would then try to create a new option and add it to the list -- because in this workflow, we would check if the new option is unique by calling `isOptionUnique`, where we will invoke `filter` function on `options`, and in this edge case it could be null or undefined and then crash. 

Here's an example exception trace:
![image](https://user-images.githubusercontent.com/5376280/33410060-15c52dfe-d544-11e7-9ecc-b9c6ab26ac75.png)

JedWatson:
Fixed by #2185

PR #2185, opened 2017-11-30
Fix/is option unique crash

Chopinsky:
This PR is trying to fix issue #2184, where for some edge cases, if we try to create a new option while the `options` property is null or undefined, the control will crash because function `isOptionUnique` will call `filter` on `options` before checking if it's valid to be performed on.


JedWatson:
Yep, looks great. Thanks @Chopinsky!

PR #2192, opened 2017-12-03
Fix state value in README example

srishanbhattarai:
Change `this.state.value` to `this.state.selectedOption.value`. The example runs now.


JedWatson:
Wow, my bad - thanks!

PR #2195, opened 2017-12-04
document breaking change of removing getInputValue

turadg:
Our app on rc10 broke when we went to the release.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -15,6 +15,7 @@
 
 ## v1.0.0 / 2017-11-23
 
+* breaking; removed `getInputValue` function - [see PR](https://github.com/JedWatson/react-select/pull/2108)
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks for that @turadg - sorry I missed it in the release notes.

Out of curiosity - it was removed in #2108 because it appears to be unused. What was your use-case for it?

turadg:
We have a `Select` component that wraps react-select. Looking at its history, we built it on a fork of react-select. In Oct 2016 we switched back to npm distro of react-select on 1.0.0-rc.2. Our `handleInputKeyUp` went from getting the input value from the DOM to getting it by `getInputValue`.

JedWatson:
Thanks for the reply @turadg. Is your use case now covered by the `handleInputKeyUp` event?

turadg:
That's our handler for `onKeyUp` event. To get it working again I just replaced `.getInputValue()` with `.state.inputValue`. So all good.

Issue #1252, opened 2016-09-26
Clearing value in multi select causes option list to expand


jamaica474:
Given multiselect with several options selected:
![0390b46807ac2ef6d68958bfa393018a](https://cloud.githubusercontent.com/assets/7143553/18844059/a726a2d0-83e9-11e6-918f-0fbc3eb36e7a.png)

When I click on either 'X' (clear single or clear all):
![yniu5k 1](https://cloud.githubusercontent.com/assets/7143553/18844162/13714cec-83ea-11e6-93f4-dc32e46a4077.jpg)

Option list expands: 
![5lf9ix](https://cloud.githubusercontent.com/assets/7143553/18844198/4a0eb7ee-83ea-11e6-916a-ab7aa6e7b558.jpg)

I would say this behavior is unexpected, when user clicks clear select should only clear itself, nothing else should happen.,

dylanfpaul:
+1

puckybreg:
+1

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue  is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

PR #2198, opened 2017-12-07
Don't open drop down menu when clear values

Chopinsky:
This PR is to address issue #1252, where when "clear" button is clicked upon, the drop down menu will prompt to open. This issue is caused by involuntarily firing the focus event after clearing values, where we will prompt the menu regardless of the context.

To address this issue, we will set a flag in the `clearValue` function, and then reset the flag after the focus handler finishes.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -237,6 +237,7 @@ class Select extends React.Component {
 					isPseudoFocused: false,
 				});
 			}
+
 			return;
 		}
 
@@ -259,6 +260,8 @@ class Select extends React.Component {
 			this.focus();
 
 			let input = this.input;
+			let toOpen = true;
+
 			if (typeof input.getInput === 'function') {
 				// Get the actual DOM input if the ref is an <AutosizeInput /> component
 				input = input.getInput();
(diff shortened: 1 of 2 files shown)


garrettdehnketurpin:
Just curious, on when this branch will be merged?

Regards,
Garrett

JedWatson:
Opening when the clear button is clicked was originally intended behaviour, but having tested the two this does make more sense. Seems like if the user intention is to clear the field, it doesn't necessarily also convey an intention to open the menu... right?

This also makes the "clear" behaviour consistent with the behaviour of the "remove" icon on values in a multi-select, so I'm happy to merge it.

Would you mind resolving the conflict @Chopinsky? I've taken a look and am not 100% confident how to preserve your fix without digging deeper into it.

JedWatson:
Nevermind, I'm about to release a version and want to include this so I've dug in deeper and resolved the conflicts. Thanks for the PR @Chopinsky!

samuelcolvin:
Any chance of getting a release with this included?

PR #2203, opened 2017-12-08
Add re-display of text input in Select component on edit

gwailoTr0n5000:
Useful for the case that you want to restore partially entered state or enable editing of a previously entered/selected option.


JedWatson:
Hey @gwailoTr0n5000, thanks for the PR!

I think that this would introduce some lack of clarity around the state of the select for users; `inputValue` looks indistinguishable from the selected value with the default styling, so using the prop would mean the select _looks_ like it has a value, but it would not.

Also, because the prop doesn't sync with state after initial mount, it would be confusing for people used to `value` or `xValue` props being controlled; it behaves more like the `defaultValue` prop on react input components.

Even that is slightly different; there are other events (e.g. focusing then blurring the select) that would reset the value unexpectedly.

I'm not totally adverse to your use-case, but I'm not comfortable introducing the new prop while it could cause that kind of confusion for users.

Let me know what you think.

Also, just for the future - please make sure PRs only include changes to the `src` files!

jossmac:
@gwailoTr0n5000 please ping the thread if this is still a concern, and needs to be re-opened.

PR #2205, opened 2017-12-11
Use onSelectResetsInput for single select

### LINKED ISSUE #1092: Keeping inputValue
Hi, is there a way to be able to set the inputValue based on a prop?

lachiet:
Allow onSelectResetsInput to persist the inputValue on single selects. This feature is already working for multiselect.

Fixes #1092


JedWatson:
Thanks @lachiet

Issue #2182, opened 2017-11-29
Two 'selects' can be open by clicking on an arrow


an4ger:
I have 2 selects on one page. Click on first select arrow and than click on second select arrow. Both selects stay opened.

![react-select-2opened](https://user-images.githubusercontent.com/5645926/33377756-3fd93a00-d51b-11e7-82a1-406092dc4372.png)

Problem is located in: [https://github.com/JedWatson/react-select/commit/f6920deefd935d366cddff14cb45b7f6f950c08b](https://github.com/JedWatson/react-select/commit/f6920deefd935d366cddff14cb45b7f6f950c08b)

Return statement was removed and event not bubble to main handleMouseDown

```
 		if (!this.state.isOpen) {
-			return;
+			this.setState({
+				isOpen: true,
+			});
 		}
 		// prevent default event handlers
 		event.stopPropagation();
 		event.preventDefault();
```

kurtinatlanta:
I submitted PR #2213 to fix this.

JedWatson:
Fixed by #2213 with thanks to @kurtinatlanta

vladlucapitech:
Thank's, when to expect a new release that contains this fix?

PR #2213, opened 2017-12-12
Fix issue #2182

kurtinatlanta:
There was a change made in commit f6920de that results in multiple dropdowns being open at the same time because stopping propagation of the arrow mouse down event in select 2 when opening its dropdown does not allow the blur event to happen for select 1, so the dropdown for select 1 remains open.


JedWatson:
Great catch @kurtinatlanta, thanks!

sloyless:
@JedWatson pls

PR #2215, opened 2017-12-13
onSelectResetsInput regression fixed

dehamilton:
Recent changes to componentWillReceiveProps prevented the input from retaining values even if onSelectResetsInput was false. This PR resolves that regression and adds tests around onSelectResetsInput, which I didn't notice anywhere. Also updated package.json to use cross-env for all scripts.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -102,7 +102,7 @@ class Select extends React.Component {
 			this.setState({ required: false });
 		}
 
-		if (this.state.inputValue && this.props.value !== nextProps.value) {
+		if (this.state.inputValue && this.props.value !== nextProps.value && this.props.onSelectResetsInput) {
 			this.setState({ inputValue: this.handleInputValueChange('') });
 		}
 	}
@@ -565,6 +565,7 @@ class Select extends React.Component {
 			this.hasScrolledToOption = false;
 		}
 		const updatedValue = this.props.onSelectResetsInput ? '' : this.state.inputValue;
+		console.log(this.props.onSelectResetsInput, this.state.inputValue, updatedValue);
 		if (this.props.multi) {
 			this.setState({
 				focusedIndex: null,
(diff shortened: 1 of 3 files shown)


test/Select-test.js
```diff
@@ -2129,6 +2129,66 @@ describe('Select', () => {
 
 	});
 
+	describe('with multi=true and onSelectResetsInput=false', () => {
```
yuri-sakharov: more accurately `describe('with multi=true and different onSelectResetsInput', () => {`

test/Select-test.js
```diff
@@ -2129,6 +2129,66 @@ describe('Select', () => {
...
+		it('should have retained inputValue after accepting selection', () => {
```
yuri-sakharov: it('should have retained inputValue after accepting selection if onSelectResetsInput=false ', () => {

test/Select-test.js
```diff
@@ -2129,6 +2129,66 @@ describe('Select', () => {
...
+
+			clickArrowToOpen();
+			typeSearchText('two');
+			pressEnterToAccept();
+			setValueProp('two'); // trigger componentWillReceiveProps
+
+			expect(instance.state.inputValue, 'to equal', 'two');
+		});
+
+		it('should have reset the inputValue after accepting selection', () => {
```
yuri-sakharov: it('should have reset the inputValue after accepting selection if onSelectResetsInput= true  or not set', () => {

JedWatson:
Thanks for the fix @dehamilton and for the review, @yuri-sakharov. I'm about to publish a new version so I've made the changes you suggested directly, so I can get this merged 👍

src/Select.js
```diff
@@ -102,7 +102,7 @@ class Select extends React.Component {
 			this.setState({ required: false });
 		}
 
-		if (this.state.inputValue && this.props.value !== nextProps.value) {
+		if (this.state.inputValue && this.props.value !== nextProps.value && this.props.onSelectResetsInput) {
```
yuri-sakharov: I think better `if (this.state.inputValue && this.props.value !== nextProps.value && nextProps.onSelectResetsInput) {`

yuri-sakharov: Hi, @JedWatson 
Sorry I didn't see the merge. Can you please also add this change?
Thank you.

PR #2219, opened 2017-12-14
Update README.md

sakshij:
In the documents it says, `import 'react-select/dist/react-select.css';` will load the css. However, the component appears un-styled. 

However it works fine using `import '!style-loader!css-loader!react-select/dist/react-select.css';`
This is when I am using it with webpack

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ You can then import react-select and its styles in your application as follows:
 
 ```js
 import Select from 'react-select';
-import 'react-select/dist/react-select.css';
+import '!style-loader!css-loader!react-select/dist/react-select.css';
 ```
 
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks for the PR @sakshij 

The syntax here actually depends on how you've got webpack configured. With `create-react-app`, which I think it probably a reasonable baseline, you don't need to specify the loaders inline. You only need to do that if you don't have the loaders configured in your `webpack.config.js`

I'd be happy to add a note that, in the example, the assumption is you're using webpack with the `css-loader` and `style-loader` configured, but I don't want to change the default example. There are other things that can go wrong too (e.g not having style-loader dependency installed) and I don't want to start going down the rabbit hole of helping people set up webpack.

PR #2221, opened 2017-12-14
Fixed bug onSelectResetsInput=false for single select

yuri-sakharov:
The bug: Select shows selected value as empty when multi=undefined or false and onSelectResetsInput=false but selected value is not empty.


yuri-sakharov:
I see that this feature people used and need to find better way to solve the bug instead  return to previous behavior.

PR #2222, opened 2017-12-15
Removed autofocus prop and fixed all warnings on running tests.

yuri-sakharov:


JedWatson:
Thanks @yuri-sakharov, but I'm not ready to merge a breaking change because it would require a major version bump and (imo) removing backwards compatibility isn't sufficient on its own to justify. Let's do this in the next major.

Good catch with the tests, though - I'll port that in!

PR #2226, opened 2017-12-16
Regression fix for single select with onSelectResetsInput=false

yuri-sakharov:
Hi, 
We got bug on single select with `onSelectResetsInput=false` after https://github.com/JedWatson/react-select/pull/2205
The bug doesn't show selected value when inputValue not empty. You can reproduce it after add `onSelectResetsInput=false` to single (`multi=false` or not set) `Select`.
The bug example:  
bug
And fix:
fix


yuri-sakharov:
Last commit was removed.

JedWatson:
Thanks for catching this regression, @yuri-sakharov!

yuri-sakharov:
You are welcome

aryeguy:
👍

PR #2228, opened 2017-12-17
Refactored handleKeyDown switch

yuri-sakharov:


JedWatson:
Thanks @yuri-sakharov 

Just out of curiosity, what's your thinking behind this change? Do you find it clearer to repeat the `event.preventDefault` and not return from the switch?

yuri-sakharov:
You are welcome

Issue #1663, opened 2017-04-12
Input field of mutiselect creatable field is not cleared after new tag cration


Falenos:
For reasons unknown the input field of a mutiselect creatable field is not cleared after new tag creation.
I click on create newTag, newTag is displayed on the field and next to it the input stays as is

Is there a method I can call when a new tag is stored to clear the input?

I know its supposed to work by default, but in my case it is not happening.

arackaf:
@Falenos Same here.  Here's a gnarly workaround for you - this is MobX code, so yours may be a bit different.  

Grab a ref to the creatable component like so

```javascript
<Creatable disabled={saving || frozen} placeholder="Tag this section" ref={el => this.creatableEl = el} onNewOptionClick={obj => store.addNewTag(obj, this.creatableEl)} onChange={store.setTags} value={store.rawTags} multi={true} options={sectionTagStore.allTags} />
```

note the ref, and note that I'm passing that ref into the onNewOptionClick method.  Here's the addNewTag method

```javascript
@action addNewTag = (obj, creatableEl) => {
    this.tags.push(sectionTagStore.createTag(obj));
    creatableEl.select.setState({inputValue: ''});
}
```

So the ref to the Creatable component has a select property on it which (I assume) is the underlying multi component.  Set *its* state to clear the inputValue.  And done.

Falenos:
@arackaf thanks a lot for the tip. It got me really close but still not there, it really is a bit gnarly though. I am new in react and from your suggestion I checked out refs for the first time...

What happened is that my component that renders the "Select" is stateless and is not declared as a Class maybe that's why it's not working, but I need to research further. The creatableEl.select.setState(); gives me nothing. There is a State object in Select so creatableEl.select.State exists but the .setState() method is not available. The code is below, if you have any suggestions...

import React, { PropTypes } from "react";// Artlimes: For our Style/Type/Content tags
import Select from "react-select";

function TagSelect(props) {
  const handleNewCategoryTagSave = (tag)=> {
    // Blank tags cannot be saved
    if ((!tag || !tag.value) ||
      (typeof tag.value === "string" && tag.value.trim().length === 0)) {
      return false;
    }

    props.onNewCategoryTagSave(props.productId, tag.value, props.name);
  };
  const selectValue = ()=> {
    const value = props.value;
    const multiSelect = props.multi;
    if (!multiSelect && Array.isArray(value)) {
      return props.value[0];
    }

    return value;
  }

  return (
[...]

PR #2241, opened 2017-12-20
Sets context of onNewOptionClick

MrLeebo:
Literally just changing
```js
onNewOptionClick(option);
```
to
```js
onNewOptionClick.call(this, option);
```

This allows you to access `this.select.selectValue(option)` within your `onNewOptionClick` callback without having to generate a bunch of extra refs. 

Should not break anything, because `this` would have been undefined previously.

### DIFF
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -31,7 +31,7 @@ class CreatableSelect extends React.Component {
 			// Don't add the same option twice.
 			if (isOptionUnique) {
 				if (onNewOptionClick) {
-					onNewOptionClick(option);
+					onNewOptionClick.call(this, option);
 				} else {
 					options.unshift(option);
 


JedWatson:
Cheers @MrLeebo 

I'm not sure about this though, shouldn't you just be updating the `value` prop with the new option that's been created? I don't understand why you'd want to call `this.select.selectValue(option)` directly.

MrLeebo:
@JedWatson I took inspiration from issues like https://stackoverflow.com/questions/44549186/how-to-close-menu-onnewoptionclick and https://github.com/JedWatson/react-select/issues/1663. In short, they end up using a ref to do something inside their `onNewOptionClick` callbacks. I do the same thing, but in my case the Select component is nested underneath a couple layers of higher-order components, which would all need to pass their refs up the tree. That's a lot of work. My option creator is async because it saves to the server and I want to call `selectValue` to select a temporary option object while the AJAX request is in flight. The `value` prop gets updated when the server responds with the new option ID.

This PR doesn't let people do anything they don't already do, its just a quality of life change to make it easier.

```js
if (onNewOptionClick) {
  onNewOptionClick(option);
} else {
  options.unshift(option);

  this.select.selectValue(option); // Sometimes you still want this to happen even though you're overriding onNewOptionClick for other reasons
}
```

JedWatson:
Thanks for the context - I understand it better now.

Wouldn't it be better to address the misbehaviour by causing Creatable to clear the inputValue when a new option is selected, rather than leaving it up to the consumer? I still don't like the idea of making it easier for people to use internal API or force state changes. Among other things, it makes it much harder to track how changes to react-select might break how people are using it.

MrLeebo:
I thought about proposing something like that, such as 

```
if (onNewOptionClick) {
  const canceled = onNewOptionClick(option);
  if (!canceled) {
    this.select.selectValue(option);
  }
} else {
  options.unshift(option);

  this.select.selectValue(option);
}
```

But that would be a breaking change if anyone has a handler that already returns something and it doesn't address async.

MrLeebo:
@JedWatson, I just tested it and at least for multi select controls, changing `value` after the new option got created does NOT currently reset the input value. I'm assuming that's a bug, however even when that is fixed I think we'd still want a way to reset the component without changing the current value. For example, if my async new option creator comes back with an error and I want to reset the select in the `.catch()`.

I sympathize with you wanting to avoid exposing the internal APIs, but maybe at least one of these internals needs to be made public because it seems like it would get awkward to try to use props to trigger the reset. What if we "publicize" selectValue and do 
```js
if (onNewOptionClick) {
  onNewOptionClick(option, this.select.selectValue);
}
```

Then, in a later release if you/me/someone else comes up with a better approach, we can change it to 
```js
if (onNewOptionClick) {
  onNewOptionClick(option, () => (deprecate("Using selectValue is deprecated, try X instead."), this.select.selectValue(arguments)))
}
```

bladey:
Hi @MrLeebo,

Thanks for this pull request.

PR #2244, opened 2017-12-22
Fixed console error in GitHub users example

yuri-sakharov:
Appears error after clicking on "Creatable?" checkbox in GitHub users (Async with fetch.js) example.

The error:
`app.js:323 Warning: GithubUsers is changing an uncontrolled input of type checkbox to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components`


JedWatson:
Thanks for catching that @yuri-sakharov

Issue #2237, opened 2017-12-20
Auto Scroll to select option not work in searchable case


shikelong:
How to scroll to select option? Seem no props / method provided now.

The gif: 
![gif](https://user-images.githubusercontent.com/4499997/34189922-f09fbc70-e578-11e7-9332-3200737a381e.gif)

Regards.

yuri-sakharov:
Hi @shikelong.
This is the bug and looks fixed in https://github.com/JedWatson/react-select/pull/2245

Please check it on demo: https://deploy-preview-2245--react-select.netlify.com/

shikelong:
Thank you!

PR #2245, opened 2017-12-22
Fixed focused option. Issue #2237

yuri-sakharov:
Fixed https://github.com/JedWatson/react-select/issues/2237


JedWatson:
This looks good, thanks for the fix @yuri-sakharov!

Issue #2263, opened 2018-01-06
Add a SelectAll function to the menuRenderer props


coder-guy22296:
The project I'm working on requires the ability to select all of the filtered results e.g. searching a list of people and want to select all the results with a particular last name.

I have a working patch and I am working on a pull request for this feature.

I tried adding a button to the options list in my custom renderer that when you click it, it would call selectValue on all the options passed in. needless to say, it was hacky and did not work, it only added the last item.

Inspired by: https://github.com/Aidurber/react-picky/

Chopinsky:
I think it's reasonable to add a configurable option to the `react-select`, such that if multi-selection is enabled, we can let users to select all the available options in the dropdown list. I will work on the PR to this feature.

ghoullier:
@Chopinsky Good news!

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our efforts towards the current major version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue  is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

PR #2265, opened 2018-01-07
Feature: adding select all option to multi-select enabled control 

Chopinsky:
This PR will address issue #2263, where we lack the ability to select all options in the dropdown menu. This is will make multi-values selections easier if users would want to select a range of the filtered options.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -661,17 +661,11 @@ class Select extends React.Component {
 	}
 
 	selectAllValues (values) {
-		console.log('select all', values);
-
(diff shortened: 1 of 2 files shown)


LillaEli:
@jossmac  I was just wondering why hasn't this been merged?! Sounds like a usable feature!

jossmac:
@LillaEli if we were to introduce unlimited features the API would become too large, and basically unusable. The approach we took in v2 was to expose low-level API that allow consumers to implement the features they desire.

PR #2266, opened 2018-01-08
Add explicit docs on how to disable client side filtering

jnachtigall:
I found the docs in the Readme not explicit enough to answer the question on how one would really disable client side filtering (eventually I found it to be answered in an issue). I think a small code snippet to answer this would be good.


JedWatson:
Thanks @jnachtigall

PR #2268, opened 2018-01-09
Bump react-input-autosize to fix SSR issues

rkkautsar:
JedWatson/react-input-autosize#121


rkkautsar:
Need to pull JedWatson/react-input-autosize#124 too. This might need some tests in the future

Issue #2208, opened 2017-12-12
React Select v2 and the future


JedWatson:
I've spent the last few months cleaning up react-select v1 and thinking about the future. Now I'm ready to share some exciting news, and begin discussing plans openly.

The project has grown way beyond my original use-case to become one of the most popular react components: with nearly a million downloads per month, statistically over one in seven people installing `react-dom` also install `react-select` 😆

Over the years the API has also grown organically to accommodate a huge amount of flexibility and customisation, and I've been reluctant to do much that would break it or invalidate open PRs before they can be reviewed.

On to the news: for the last half-year, with several team members from Thinkmill I've been working with Atlassian to develop their new UI library [Atlaskit](https://atlaskit.atlassian.com) which is being used in Jira, Stride, Confluence, Bitbucket and many other products. We've had the opportunity to do a lot of deep thinking on component architecture, and are now turning our attention to Select components... which I am hoping to use react-select as the base for.

With that in mind, I'm embarking on a rewrite of the component using these new ideas and much more modern techniques. I've got a working prototype of the new architecture, and from now I'll be working publicly in a `v2` branch. Here's what you need to know:

* v1.x is stable and will continue to be maintained, at least until v2 is released with feature parity
* The `v2` branch is experimental; I'll update here as it evolves
* v2 will use css-in-js (my current prototype uses glam, this may change) and will continue to be themeable
* v2 won't include less/scss stylesheets. I've got some ideas on how to continue shipping a classNames-based API for everyone who's already invested in stylesheet customisations... we'll see
* I'm working on a new concept for customising the component which will consolidate a lot of the current props and make the component more modular, more maintainable while increasing flexibility
* Despite re-writing the architecture, much of the existing code will be brought forward; there is a huge amount of maturity in this project that I'm keen to protect

Some specifics:

* I'm using flow
* Internal state management has been cleaned up significantly - how options are handled and data is passed are both more powerful and more consistent
* You'll be able to override _all_ the components to completely customise the UI
* Some functions will also be configurable when you want custom behaviour
* Accessibility, performance and flexibility are all first-class concerns

My goal is that many of the requirements people open issues and PRs for will become easily customisable so they don't require changes to the react-select core. If (_if_) I get this right, it'll both make the component more useful and also help me stay on top of things as a maintainer. It also opens the potential for an ecosystem of packages and customisations to start developing on top of react-select, instead of the many forks that currently exist.

webmobiles:
[...]

PR #2269, opened 2018-01-09
Add new props controlClassName and menuContainerClassName

marccerrato:
Hello there,

We want to use `react-select` in a [backpack](https://github.com/Skyscanner/backpack) component and we need to customize some styles to meet the library UI standards.

In the library we use SASS (for defining and reusing styling rules), and CSS Modules (which appends a hash to every CSS class name we define), not allowing us to override the `react-select` classes from our style definitions.

We found that best option we have to customize the styles would be to provide some extra classes to some of the `react-select` elements. In some cases, this is already possible (props `className` and `optionClassName`) but for others it is not: the ones we added in this PR.

I'm aware that a classNames-based API approach is planned for the v2 as explained in #2208, but we would like to use it as soon as possible, and other options as inline styling or styled components are not possible in our case (use of SASS variables/mixins/breakpoints/pseudo-selectors).

Would you consider approving this PR until you move to the v2? 😄


barbu110:
This is not reflected in the docs.

flexdinesh:
Hey @marccerrato [v1 of react-select is no longer maintained](https://github.com/JedWatson/react-select/pull/3786). 

We highly recommend [upgrading from v1 to v2](https://react-select.com/upgrade-guide). We have change the style customization architecture in v2. You might find it simpler and more convenient.

Issue #2202, opened 2017-12-08
Selector has wrong height if autosize=false and arrowRenderer=null


kozakdenys:
The issue appears if `autosize={false}` and `arrowRenderer={null}` on the same time.
Can be related to this issue https://github.com/JedWatson/react-select/issues/2082
An example of code https://plnkr.co/edit/i5Cdg3e8f8JPBZowFLMj?p=preview

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you feel this issue / pull request is still relevant and you'd like us to review it, please leave a comment and we'll do our best to get back to you.

PR #2276, opened 2018-01-10
Fix rendering with arrowRenderer and no autosize.

### LINKED ISSUE #2082: React Multi has broken with prop autosize=false
Are you reporting a bug or runtime error?
### Bug

I have a problem when I am using the prop autosize={false} and multi={true} the height of the component is extremely long, See:
dropdown broken

*  Test Case:
https://plnkr.co/edit/xLJjLV5YQuTZTWA5XIWs?p=preview

harrykao:
The container for the input element lacked `display: inline-block`,
which is present on both the `AutosizeInput` component and the disabled
version of the input component.

This should fix #2082 and #2202.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -925,7 +925,7 @@ class Select extends React.Component {
 			);
 		}
 		return (
-			<div className={ className } key="input-wrap">
+			<div className={ className } key="input-wrap" style={{display: 'inline-block'}}>
 				<input id={this.props.id} {...inputProps} />
 			</div>
 		);


JedWatson:
Thanks for the fix @harrykao

natefoundry:
This commit may have unintentionally broken the width and vertical alignment of the hidden input field. I have `autosize=false` and if I type a lot of characters in the text box it cuts off before reaching the end of the total width of the text box (tested on Chrome).

Adding the following scss fixed the issue for me:

```
  /* Fixes issue with input width and position introduced by commit: d235dee2 */
  .Select-input {
    display: block !important;

    & > input {
      display: inline-block !important;
    }
  }
```

ls-guillaume-lambert:
@natefoundry I also got this issue, I will implement the fix you have there. Should that be opened as a PR?

image

Issue #2279, opened 2018-01-11
Sass missing property from LESS


mantissa7:
react-select > less > control.less has a property on line 191 in `.Select-input > input { width: 100% }`, however, in the equivalent selector in the scss folder that property is missing, which causes some issues with responsive grid systems when the input should be less than ~160px.

PR #2280, opened 2018-01-11
Add property to control.scss from control.less

mantissa7:
Fixes https://github.com/JedWatson/react-select/issues/2279


JedWatson:
Thanks for that @mantissa7

PR #2291, opened 2018-01-12
Del key should work normal if not deleteRemoves (fixes #2282)

### LINKED ISSUE #2282: Accepting PR? <Del> should work like normal when `Creatable` and deleteRemoves is `true`
**Steps to reproduce:**

1. Go to the last example *Custom tag creation* at http://jedwatson.github.io/react-select/
2. Type in some stuff, create a typo and use the `<Del>` (delete) to fix it. 

**Expected results:**

For Creatable and when deleteRemoves is disabled, then like in any other text input field, as a user I want to be able to use the `<Del>` key

**Actual results:**
Just nothing happens when pressing `Delete` key.

**Solution:**

Would you accept a PR on this? It seems to be a simple matter of just moving the event.preventDefault() inside the `if` clause at:

https://github.com/JedWatson/react-select/blob/7bc9363ad139546e296778774b7d16e7cdf72d55/src/Select.js#L548-L553

At the moment the `<del>` is just always captured.

jnachtigall:
Fixes #2282 as described in the ticket. Same behaviour as for `backspace`

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -546,8 +546,8 @@ class Select extends React.Component {
 				this.focusStartOption();
 				break;
 			case 46: // delete
-				event.preventDefault();
 				if (!this.state.inputValue && this.props.deleteRemoves) {
+					event.preventDefault();
 					this.popValue();
 				}
 				break;


JedWatson:
Thanks @jnachtigall, happy to have this fixed 👍

PR #2292, opened 2018-01-12
Keep dropdown open during load options (fixes #2281)

### LINKED ISSUE #2281: Accepting PR? Keep dropdown menu open when waiting for 'loadOptions' to return
At the moment the `AsyncCreatable` always collapses (closes) the dropdown when executing the `loadOptions` promise/callback. Then when the results are fetched then the dropdown is expanded (opened) again. For quick ajax responses this results in some kind of "flashing" of the menu. See this screencast:
react-select-flashes

I would rather like to have it open all the time like e.g. www.google.com does it when presenting its search suggestions in the dropdown.

It's rather simple to implement this, just change (comment out) this src:
https://github.com/JedWatson/react-select/blob/99acc31346acacaee23cefecd2ab9b13e3d672bb/src/Async.js#L133-L140

I just did this and it looks like I want now:

react-select-keep-open

Would you accept a PR for this, maybe using a new `Async` prop called `keepMenuOpenDuringLoading`?

jnachtigall:
This allows for a dropdown to stay open as one is used to from DDG or google. Fixes #2281

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,7 +304,7 @@ You can manipulate the input by providing a `onInputChange` callback that return
 function cleanInput(inputValue) {
     // Strip all non-number characters from the input
     return inputValue.replace(/[^0-9]/g, "");
-}   
+}
 
 <Select
     name="form-field-name"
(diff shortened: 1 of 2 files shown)


JedWatson:
Thanks @jnachtigall, would you mind updating this to exclude the `dist` and `lib` changes? Just need the `src` and readme updates.

jnachtigall:
> would you mind updating this to exclude the dist and lib changes? Just need the src and readme updates.

@JedWatson  I tried to rebase and push again with cleaner commits. Better?

BTW, why aren't `dist/` and `lib/` in `.gitignore` in the first place? This confused me a lot at first when not knowing if changes need to be made to `src/` or/and `lib/` (did not know that lib is autogenerated).

jnachtigall:
@JedWatson Any chance of getting this merged? I am using it in production but would like to get in merged here too :)

kidmillions:
+1 to this, would love to see this merged.

flexdinesh:
Hey @jnachtigall [v1 of react-select is no longer maintained](https://github.com/JedWatson/react-select/pull/3786). 

We highly recommend [upgrading from v1 to v2](https://react-select.com/upgrade-guide). I think this issue might already be fixed or non-existent in v2.

PR #2294, opened 2018-01-15
Use getScrollParent to determine the container to scroll 

nicocrm:
For issue #1267 

I can't test the actual functionality with jsdom, but I could add a test to make sure the code is exercised, if you are good with the approach


mathieumg:
@jossmac @JedWatson If I rebase this and add unit tests, is there a chance of it getting merged?

jossmac:
@mathieumg You'll need to update this PR or open a new one against the [v1 branch](https://github.com/JedWatson/react-select/tree/v1.x)

mathieumg:
@jossmac I ended up realizing this was already shipping with v2 when going through the code, even though it's not really documented. My scrollable parent wasn't `relative`, which caused the feature to not work; I changed that and it now works properly. 👍

PR #2295, opened 2018-01-16
V2 custom styles framework

JedWatson:
This is a basic implementation of what I think we'll ship for customising styles in v2.

@jossmac @treshugart keen for your review & feedback, if you're on board I'll apply the technique to the rest of the components.


src/components/SingleValue.js
```diff
@@ -1,25 +1,33 @@
...
-  />
-);
+export const css = ({ isDisabled }: { isDisabled: boolean }) => ({
+  ...marginHorizontal(spacing.baseUnit / 2),
+  color: isDisabled ? colors.neutral40 : colors.text,
+  position: 'absolute',
+});
+
+const SingleValue = ({ getStyles, ...props }: ValueProps) => {
+  const { isDisabled, data, ...cleanProps } = props;
```
ebonow, line 23: Would this be potentially breaking for any users who rely on `data`?

Issue #2277, opened 2018-01-11
componentWillReceiveProps resets input even if values are the equal


choonkending:
# Bug report

Currently in Select https://github.com/JedWatson/react-select/blob/master/src/Select.js#L144 , we set input value to ''" even if the value is equal. 

```js
if (this.state.inputValue && this.props.value !== nextProps.value && nextProps.onSelectResetsInput {
      this.setState({ inputValue: this.handleInputValueChange('') });
}
```

## Why is this a problem?
If `<Select>` is rendered within a stateful React Component which sets state, and is triggered while the Select box is focused, this will cause the input to be reset. In my use case, we have an `onScroll` event which triggers when the user scrolls. 

As soon as the `componentWillReceiveProps` is triggered, the shallow equality check will reset the input _even though nothing changed_.

**Anything that causes a rerender of Select will potentially reset the input when the user is typing**

```js
var naiveIsObjectEqual = function(o1, o2) {
    return Object.keys(o1).reduce(function(acc, cur) {
        if (o1[cur] !== o2[cur]) {
            acc = false;
        }
        return acc;
    }, true);
};

var naiveArrayIsEqual = function(v1, v2) {
    if (v1.length !== v2.length) return false;
    return v1.reduce(function(acc, cur) {
        var isElementEqual = v2.reduce(function(innerAcc, innerCur) {
            if (!naiveIsObjectEqual(innerCur, cur)) {
                innerAcc = false;
            }
            return innerAcc;
        }, true);

        if(!isElementEqual) {
            acc = false;
        }

        return acc;

    }, true);
};


// Ignore the above equality implementations, they are just a slightly deeper equality check
if (this.state.inputValue && naiveArrayIsEqual(this.props.value, nextProps.value) && nextProps.onSelectResetsInput) {
    this.setState({ inputValue: this.handleInputValueChange('') });
}
```

# Next Steps
I will attempt to reproduce this issue on codepen. I am unsure whether this issue specifically affects AsyncCreatable or affects all Select components. 

However, if the maintainers are already aware of this issue, I am more than happy to submit a PR to help with this. Please let me know!

asaarnak:
Using `onSelectResetsInput={false}` workaround for now.
Propably this commit introduced the bug where input is reset too often: https://github.com/JedWatson/react-select/commit/ae26bf80d10112b1407e1d34e7fa00caa1fe8699

diegoarcega:
FYI: having the same issue here.

[...]

PR #2299, opened 2018-01-17
Fixed regression inputValue reset on componentWillReceiveProps ( issue #2277 )

yuri-sakharov:
Hi,
Following code:
`if (this.state.inputValue && this.props.value !== nextProps.value && nextProps.onSelectResetsInput) {
this.setState({ inputValue: this.handleInputValueChange('') });
}`
was inserted for https://github.com/JedWatson/react-select/issues/2155 and it provides issues like https://github.com/JedWatson/react-select/issues/2277. Fortunately @gwyneplaine provided better solution via ref and reset inputValue in componentWillReceiveProps is redundant.


yuri-sakharov:
Irrelevant for react-select 2

PR #2306, opened 2018-01-18
fix flow errors in v2 branch

gwyneplaine:
Fix flow issues in v2 branch. 
These include errors in: 
- animated/MultiValue.js
- components/MultiValue.js
- animated/transitions.js


JedWatson:
Thanks @gwyneplaine

PR #2307, opened 2018-01-18
Move prop declarations out of child components

jossmac:
Waiting on jedwatson/react-input-autosize#130 to land, before cleaning up the Input component.


JedWatson:
Heaps cleaner, thanks.

Not sure about the snapshot changes tho? see https://travis-ci.org/JedWatson/react-select/builds/330210028

Could you check those properties are supposed to be missing and update the snapshot as part of the PR please?

PR #2310, opened 2018-01-19
Loading indicator component

jossmac:
Created a default component for `LoadingIndicator`, which replaces the `ClearIndicator` when async data is being loaded:

- **HACK:** keyframes added by using a [style tag inside component](https://github.com/JedWatson/react-select/pull/2310/files#diff-c637ef640dbff2a6e7fbe1ff68462b28R132)
- Introduced `isLoading` boolean to Select's props object


JedWatson:
`isLoading` should be a prop, not state

PR #2317, opened 2018-01-22
[WIP] V2 component props experiment

JedWatson:
**DO NOT MERGE**

I'm opening this PR for discussion and review. It demonstrates different approaches to how we could encapsulate styles, aria, and other props to make customising components in v2 simpler.

@jossmac as discussed, I think the main areas people will want to be able to independently customise are:

- styles + theme
- component structure (specific ref: Group, MultiValue)
- component wrappers (e.g Transition implementation)
- components and props themselves (advanced mode)
- content (various labels, formatters, messages & aria)
- functional logic (filtering, selected state, etc)

We should also look at how, given the setup, variations on the UI (e.g Checkboxes / Radio menus in Atlaskit) would be implemented.

The two use-cases I've been thinking about are:

- how simple could it be to re-organise the `MultiValue` component, to put the "cross" on the left?
- how simple could it be to set the colour of options and values as per #2295

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -3,6 +3,9 @@
 import React, { Component, type ElementRef } from 'react';
 import glam from 'glam';
 
+import { className } from './utils';
+import { borderRadius, colors, spacing } from './theme';
+import { marginVertical, paddingVertical, paddingHorizontal } from './mixins';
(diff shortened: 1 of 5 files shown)


JedWatson:
This has been superseded by other work, I'm going to close it now.

PR #2319, opened 2018-01-23
Adding basic Async implementation

JedWatson:
WIP, need to add more features to the base Select component to improve the UX for this


JedWatson:
I'm not totally happy with the state of the UX yet, but want to progress a few other parts of the architecture now so it's good to merge. Will open an issue with outstanding tasks.

PR #2329, opened 2018-01-25
Update code example on README

dirkdunn:
Example code does not work out of the box, the correct logic is "const value = selectedOption || selectedOption.value;", not &&.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ class App extends React.Component {
   }
   render() {
   	const { selectedOption } = this.state;
-  	const value = selectedOption && selectedOption.value;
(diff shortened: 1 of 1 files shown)


JedWatson:
I'm not sure why the code in the readme isn't working for you but `const value = selectedOption || selectedOption.value;` would *absolutely* break with an `undefined is not an object` error.

JedWatson:
I copied the example in the Readme into a codesandbox, it works as expected: https://codesandbox.io/s/y2p1808251

If there's anything else that's unclear in the documentation that would help avoid the problem you're experiencing please let me know. Thanks!

PR #2340, opened 2018-01-31
Isolate inner props

jossmac:
Isolate properties necessary to functionality, making it simpler for consumers to pass the correct props to components if they are replaced.


JedWatson:
Nice work @jossmac, this is much cleaner 👍

Issue #2335, opened 2018-01-29
Categorized options feature


yanshiyason:
Hi, thank you for this library. It has been a pleasure to work with so far!

I would like to propose a feature. 

I am working on an app where the options are categorized (nested) like so:

![image](https://user-images.githubusercontent.com/11358371/35498429-92e61b52-0511-11e8-8a4e-e086cd6d524e.png)

When toggling the parent, all children toggle along with it.

I am accomplishing this by using the [setValue](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L597) method defined on the `Select` element. I pass a `ref` from the `Select` component to a custom `Menu` component rendered through the `menuRenderer` prop which in turn passes it down to the "category's parent option" which finally uses it to toggle it's children. I thought it would be nice to offer a user friendly way of accomplishing this functionality.

There is a working demo of a jQuery library which accomplishes this [here](https://www.jqueryscript.net/demo/Handy-Customizable-Dropdown-Select-Plugin-With-jQuery-JQSelect/). Look at the `multi / group` one.

Please let me know if you would be interested in this feature and I can open an initial PR.

JedWatson:
Hi @yanshiyason, thanks for creating this issue - it's great to see what you've been able to do with react-select.

Not sure if you've seen but I've been working on a new architecture for v2 which would make this much easier to implement, so I don't think it's likely we'll extend v1 to support this more cleanly but I would be very interested in your feedback on v2 and whether it makes this case better for you.

The API is still undergoing a lot of change so whatever you do with it in the next week or so should be considered experimental, but I'm hoping to finalise it and make the first production release by the end of February.

See #2289, and especially check out the "Experimental" examples.

yanshiyason:
Hi @JedWatson , thank you for taking the time to look into this!

I took a look at v2 and I tried implementing a first rough version.
I am not sure it fits in with the overall architecture but if you can checkout it out and let me know a better way to go about it, it would be greatly appreciated!

Here is the PR #2341.

Also, I couldn't find the API for keeping the menu opened when selecting options.. is that yet to be implemented in v2?

Thanks!

yanshiyason:
I forgot to upload a screenshot, will do so tomorrow!

JedWatson:
@yanshiyason no need, a preview build was automatically generated for your PR so I can see what you've done: https://deploy-preview-2341--react-select.netlify.com 😄 

[...]

PR #2341, opened 2018-01-31
[v2] Allow option groups to be toggled together

yanshiyason:
ref #2335

### DIFF
--- a/examples/pages/Categorized.js
+++ b/examples/pages/Categorized.js
@@ -1,0 +1,148 @@
+// @flow
+import React, { Component } from 'react';
+import { withValue } from 'react-value';
+
+import { colors } from '../../src/theme';
+
+import { groupedOptions } from '../data';
+
+import { Li, Ul, Span } from '../../src/primitives';
+import { Link, H1 } from '../components';
+
+import Select from '../../src';
+const SelectWithValue = withValue(Select);
+
+class Group extends Component<*, *> {
+  render() {
+    const { children, getStyles, Heading, innerProps, label } = this.props;
+    return (
+      <Li
+        aria-label={label}
+        css={getStyles('group', this.props)}
+        {...innerProps}
+      >
+        <Heading {...this.props} />
+        <Ul>{children}</Ul>
+      </Li>
+    );
+  }
+}
+
+const getGroupHeadingStyles = ({
+  isSelected,
+  isDisabled,
+}) => defaultStyles => ({
+  ...defaultStyles,
+  backgroundColor: isSelected ? colors.primary : 'transparent',
+  color: isDisabled
+    ? colors.neutral20
+    : isSelected ? colors.neutral0 : 'inherit',
+  ':hover': isDisabled
+    ? null
+    : {
+        background: '#999',
+        color: '#fff',
+      },
+});
+
+class GroupHeading extends Component<*, *> {
+  children() {
+    const { group: { key }, menuOptions: { render } } = this.props;
+    const selectedGroup = render.find(i => i.key == key);
+    return selectedGroup ? selectedGroup.children : [];
+  }
+  allOptionsSelected = () => {
+    const { isOptionSelected } = this.props;
+    const selectValue = this.props.getValue();
+    return this.children().map(i => isOptionSelected(i, selectValue)).every(isTrue => isTrue);
+  };
+  valuesIncludingChildren = () => {
+    const { isOptionSelected } = this.props;
+    const selectValue = this.props.getValue();
+    return [
+      ...this.children().filter(i => !isOptionSelected(i, selectValue)).map(i => i.data),
+      ...selectValue,
+    ];
+  };
+  valuesExcludingChildren = () => {
+    const { isOptionSelected } = this.props;
+    const selectValue = this.props.getValue();
+    return selectValue.filter(i => !isOptionSelected(i, this.children())).map(i => i.data);
+  };
+  toggleGroup = () => {
+    const { setValue } = this.props;
+
+    if (this.allOptionsSelected()) {
+      setValue(this.valuesExcludingChildren());
+    } else {
+      setValue(this.valuesIncludingChildren());
+    }
+  };
+  render() {
+    const { getStyles, label, isDisabled } = this.props;
+
+    const isSelected = this.allOptionsSelected();
+
+    const defaultStyles = getStyles('groupHeading', this.props);
+    const styles = getGroupHeadingStyles({ isSelected, isDisabled })(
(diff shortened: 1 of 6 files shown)


JedWatson:
Thanks for the PR, and sorry about the conflicts!

So I would be really interested to see if you can implement this by providing custom components - see the Animated and Calendar (under `Experimental`) examples for what I mean.

In theory this should be possible to implement as an extension and require very little change to the core Select component, which would be ideal.

We'd probably have to pass some additional props down to the Group component (like `setValue`, the current `selectValue` state, etc) but otherwise I think it would work, and it would be a really good test of the new architecture.

yanshiyason:
Thank you for your quick response! I am going to take your advice and make
a 2nd PR tomorrow morning!

Thanks again for reviewing!

examples/pages/Categorized.js
```diff
@@ -0,0 +1,120 @@
...
+      menuOptions,
+      selectValue,
+      selectValues,
+      ...others
+    } = this.props;
+    const { Heading } = components;
+    return (
+      <Li
+        aria-label={label}
+        css={getStyles('group', this.props)}
```
yanshiyason: I am not sure how how I can pass a styles props in a way that makes it extensible when reusing this component?

examples/pages/Categorized.js
```diff
@@ -0,0 +1,120 @@
...
+          {label}
+        </Heading>
+        <Ul>{children}</Ul>
+      </Li>
+    );
+  }
+}
+
+const Option = props => (
+  <Div style={{ paddingLeft: '2em' }}>
```
yanshiyason: I wrapped the `<Option>` in  a `<Div>` and because I am not sure how I can merge new styles into the `<Option>`'s props..

yanshiyason: actually, I can't do that.. The following compiles but it wipes out the initial styles of the component instead of adding to them:

```
<SelectComponents.Option
      {...props}
      innerProps={{
        ...props.innerProps,
        css: { ...props.innerProps.css, paddingLeft: '2em' },
      }}
    >
```

yanshiyason: Ahh, I found the answer in the `Experimental.js` file..

```
const getOptionStyles = defaultStyles => ({
  ...defaultStyles,
  display: 'inline-block',
  width: '12%',
  margin: '0 1%',
  textAlign: 'center',
  borderRadius: '4px',
});
```

and then:
```
...
const defaultStyles = getStyles('option', props);
const styles = getOptionStyles(defaultStyles);
...
```

then

```
<Span {...cleanProps} css={styles} ref={innerRef}>
```

yanshiyason:
Actually, I might just need to change `Option` and `GroupHeader`, let me change that..

src/Select.js
```diff
@@ -886,8 +900,13 @@ export default class Select extends Component<Props, State> {
                 'aria-label': label,
                 role: 'group',
               }}
-              Heading={GroupHeading}
```
yanshiyason: I thought this prop should be passed through the `components` props, rather than directly.. is that correct?

src/components/Group.js
```diff
@@ -22,14 +25,15 @@ type Props = PropsWithStyles & GroupProps;
 export const groupCSS = () => paddingVertical(spacing.baseUnit * 2);
 
 const Group = (props: Props) => {
-  const { children, getStyles, Heading, label, innerProps } = props;
+  const { components, getStyles, children, label, ...innerProps } = props;
+  const { GroupHeading } = components;
```
yanshiyason: and here it should be retrieved through the `components` prop. Is this correct?

src/Select.js
yanshiyason: nvm, I figured it out.. this was correct.

src/Select.js
```diff
@@ -886,8 +900,13 @@ export default class Select extends Component<Props, State> {
...
+              components={{ Heading: GroupHeading }}
               getStyles={this.getStyles}
+              menuOptions={menuOptions}
+              isOptionSelected={this.isOptionSelected}
+              selectValue={selectValue}
+              selectValues={this.selectValues}
+              group={group}
               {...group}
```
yanshiyason: I need to pass these 4 props to make the `Categorized` component work...
Not sure if this fits well in the overall architecture of the code.

I don't need to pass `group` though since I am already spreading it's attributes, I will remove that right now.

yanshiyason: actually, I need the `key` prop inside the `group` object but if I remove `group={group}` and rely on the attributes assigned by `{...group}` but even though I change [this line](https://github.com/JedWatson/react-select/pull/2341/files#diff-cdaff1a639e18d02379ab62c90551f29R56) to 
```
const { key, menuOptions: { render } } = this.props;
```

it breaks the code... I will need to investigate a bit..

JedWatson:
@yanshiyason nice, that one was for you 😁 

If you get a chance I'd love your feedback on #2351, it should also be helpful for what you're doing because it standardises the state and API that is provided to all the internal components.

yanshiyason:
@JedWatson I think it's great!
It was hard to keep track of which props were getting passed down to all children components!
But by sticking them in the `commonProps` attribute, you can just refer to the `getCommonProps` method and know right away which props they all share.

```
return {
    clearValue,
    getStyles,
    getValue,
    hasValue,
    isMulti,
    options,
    selectOption,
    setValue,
    selectProps: props,
  };
```

I am going to rebase this PR as soon as you merge #2351!

JedWatson:
@yanshiyason that PR has been merged now, see how you go :)

Hopefully things are getting easier... let me know if there’s anything else that you think we could do to improve the customisation.

yanshiyason:
@JedWatson I noticed that the options being passed as `commonProps` are the options unaltered. The options I am using are inside the `menuOptions.render` object and they come from the [`buildMenuOptions`](https://github.com/JedWatson/react-select/blob/v2/src/Select.js#L295) method.

```
options.forEach((item, itemIndex) => {
          ...
          
          render.push({
            type: 'group',
            key: `${itemIndex}-${toKey(itemLabel)}`,
            label: itemLabel,
            children,
          });
          
          ...
          
    return { render, focusable };
  }
```

I still need to pass down the `menuOptions` prop.. but it doesn't seem optimal to pass down 2 props which are a variation of the same data.. any ideas?

JedWatson:
Hey @yanshiyason thanks for the collaboration on this PR.

I'm going to close it because it's well behind `v2` now and I don't want to add this in as core functionality but if you have anything else you'd like to propose adding in the future, let me know!

PR #2343, opened 2018-02-01
Make examples one scrollable page

jossmac:
Notes:
- The default route returns all pages accessible via their hash e.g. "/#animated"
- Navigation scrolls to the applicable section
- Individual pages are still accessible i.e. navigating to the "/animated" route will only return the `<Animated />` component

Later:
- Implement a scrollspy solution to highlight the selected nav-item


jossmac:
Keeping pages separate for now

jossmac:
See #2402

PR #2351, opened 2018-02-05
[v2] Refactor internal component props

JedWatson:
I've refactored the props that are passed to the internal components, including:

- More consistently passing `innerProps` to all components
- Creating and passing `commonProps` to all components on render, so we can ensure consistency around the API

This is a pretty big change and has some rough edges, so will need review and testing before merge.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -162,6 +162,7 @@ export default class Select extends Component<Props, State> {
   static defaultProps = defaultProps;
   blockOptionHover: boolean = false;
   components: SelectComponents;
+  commonProps: any; // TODO
   controlRef: ElRef;
   focusedOptionRef: ?HTMLElement;
   hasGroups: boolean = false;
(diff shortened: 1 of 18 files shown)


src/Select.js
```diff
@@ -162,6 +162,7 @@ export default class Select extends Component<Props, State> {
   static defaultProps = defaultProps;
   blockOptionHover: boolean = false;
   components: SelectComponents;
+  commonProps: any; // TODO
```
treshugart: Could at least be `{ [string]: any }`.

src/Select.js
```diff
@@ -998,6 +1012,24 @@ export default class Select extends Component<Props, State> {
       return <input name={name} type="hidden" value={value} />;
     }
   }
+  getCommonProps() {
```
treshugart: I think we should experiment with context here, especially after depending on React 16.3.

src/Select.js
```diff
@@ -437,7 +440,7 @@ export default class Select extends Component<Props, State> {
       onChange ? () => onChange(newValue, { action }) : undefined
     );
   };
-  selectValue = (newValue: OptionType) => {
+  selectOption = (newValue: OptionType) => {
```
treshugart: You mentioned something about a `setOptions` or `selectOptions` (plural) here based on grouping functionality. While that'd be a good idea to support, most probably, I think it should have a single entry point like this one and the value could be `Array<OptionType>`. If you're selecting one, you just pass an array of one item etc.

treshugart:
I made some comments, but overall this looks good to go in now and we can address other things as needed.

JedWatson:
Looks good to go! I'll publish this a `alpha.4`

PR #2354, opened 2018-02-07
V2 a11y and primitive cleanup

jossmac:
- options and groups are now `Div`s for a less "noisy" assistive experience
- removed primitives `Li`, `Ul`, `Em`, and `Strong`
- remove `mixins` and replaced references
- rename `SROnly` --> `A11yText`, hopefully more descriptive
- use indexes for group/option IDs
- implement `formatGroupLabel` property and added an example to the home page
- groups employ `aria-labelledby` [heading] rather than duplicate label declarations
- groups now receive "options" not "children"
- simplify and future-proof prop type definitions for builtins by leveraging `typeof`


JedWatson:
👍 nice work @jossmac!

Issue #2099, opened 2017-10-25
Remove focused item on menu close


bunkscene:
When a dropdown items is hovered, it gets highlighted.  If you close the menu and reopen it, the last hovered item is still in it's hover state.  

![focusisretained](https://user-images.githubusercontent.com/2693678/32017611-3e42704a-b995-11e7-9d3c-50b091c7d9b6.gif)

I would like to suggest that this does not happen.  Is there a use case where retaining the last hovered item is desired?  If so, then could there be an option to clear the hovered state?

This creates an issue for me because when I focus on the input field, I automatically open the dropdown.  If you then tab out, that hovered item becomes selected.  If there is a selection from a previous opening of the dropdown, I don't want this selected without user action.

cmsheffield:
I would also very much like to see this issue resolved, or at least make this behavior optional.

The behavior is even stranger if you've previously selected an item. If an item has been selected, then you expand the menu again and hover over another item, then close the menu by clicking outside of it, if you expand the menu one more time, it scrolls to the highlighted item, instead of to the selected item. I would expect the latter behavior.

joeHillman:
... also upvoting this issue to be resolved.

mirovelk:
I've just spend 2 hours trying to figure out how to disable this behaviour. It's super weird. Upvoting.

mirovelk:
After trying basically every other similar library out there I conclude that this one in still the best :) ...so I hopefully fixed the issue and created a pull request.

tsi:
+1

jossmac:
Resolved by #2736

PR #2356, opened 2018-02-07
Reset focused option when menu is closed

mirovelk:
This should fix the issue #2099 ...the current behaviour is confusing from the UX point of view, see the issue for more information.

Changes on 2 places: 
- when closing the menu manually (using the keyboard/mouse)
- when focus is lost


jossmac:
Duplicate of #2736

PR #2359, opened 2018-02-08
Reduce to get render/focusable menu options

albertogasparin:
This way it is easier to understand where `render` and `focusable` get populated.
`acc` stands for accumulator, which is a generic variable name I use in `reduce` in order to quickly identify the subject/output.


JedWatson:
Neat. looks good to me, thanks @albertogasparin!

PR #2360, opened 2018-02-09
V2 menuIsOpen, inputValue as controllabe props

JedWatson:
I've done some significant refactoring to move `menuIsOpen` and `inputValue` from internal state to props that can be controlled.

This PR also includes a new component, `StateManager` which is the component actually exported and cleanly handles the default behaviour of the new props by managing them as internal state that can be overridden by props.

Based on my initial testing, it looks good (and is arguably simpler from an implementation perspective) and these has been a highly requested feature.

@jossmac would appreciate your review + testing before we merge this in!

### DIFF
--- a/src/StateManager.js
+++ b/src/StateManager.js
@@ -3,16 +3,25 @@
 import React, { Component } from 'react';
 
 import Select from './Select';
+import { type OptionType } from './types';
 
-export default class StateManager extends Component<*, *> {
-  constructor(props: any) {
-    super(props);
-    this.state = {
-      inputValue: this.props.inputValue || '',
-      menuIsOpen: this.props.menuIsOpen || false,
-      value: this.props.value || null,
-    };
-  }
+type Props = {
+  inputValue?: string,
+  menuIsOpen?: boolean,
+  value?: OptionType,
+};
+type State = {
+  inputValue: string,
+  menuIsOpen: boolean,
+  value: OptionType | null,
+};
+
+export default class StateManager extends Component<Props, State> {
+  state = {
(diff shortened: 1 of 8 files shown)


src/index.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import Select from './Select';
```
jossmac: We should export the "stateless" component also, perhaps `SelectStateless`?

src/StateManager.js
```diff
@@ -0,0 +1,61 @@
+// @flow
+
+import React, { Component } from 'react';
+
+import Select from './Select';
+
+export default class StateManager extends Component<*, *> {
+  constructor(props: any) {
```
jossmac: - [ ]  unnecessary constructor

src/StateManager.js
```diff
@@ -0,0 +1,61 @@
+// @flow
+
+import React, { Component } from 'react';
+
+import Select from './Select';
+
+export default class StateManager extends Component<*, *> {
```
jossmac: - [ ] needs `Props` and `State` type declarations

src/StateManager.js
```diff
@@ -0,0 +1,61 @@
...
+    this.setState({ menuIsOpen: false });
+  };
+  render() {
+    const inputValue = this.getSelectProp('inputValue');
+    const menuIsOpen = this.getSelectProp('menuIsOpen');
+    const value = this.getSelectProp('value');
+    return (
+      <Select
+        {...this.props}
+        inputValue={inputValue}
```
jossmac: - [ ] alphabetize prop declarations

PR #2361, opened 2018-02-09
V2 prop cleanup

jossmac:
- fix for bug where multi-selects couldn't be "not clearable"
- new `isSearchable` property (defaults to `true`)
- more robust edge detection for the menu
  - new `menuPlacement` property (defaults to `'bottom'`)
  - new `menuShouldFlip ` property (defaults to `true`)


jossmac:
@JedWatson resolved conflicts

Issue #1370, opened 2016-11-16
Fixed header and footer on the dropdown menu


rdiazv:
I need to do two things on the menu:
1. Use a custom search input fixed on top of the menu.
2. Add a fixed button at the bottom of the menu.

Is there any way to do them? (I'm currently using 1.0.0-rc.2). If not, I can create a PR adding props `menuFooter` and `menuHeader`, or something like that.

adamcai76:
Is `menuFooter` and `menuHeader` implemented?

agirton:
Hi @rdiazv it sounds like you need to customize the `renderOuter`.  We currently don't support this, but I've heard more uses cases where people need to swap out the entire menu. Will look at this again after 1.0 final.

kiranbabu189:
is the feature added?

kyleholzinger:
I figured this issue may have been left behind, but this feature is now in v2 via the replaceable components API. I think this is good to close

PR #2373, opened 2018-02-15
#1370 Feature: Custom outer component

growlycode:
Allows the end-user to supply a custom outerRenderer or outerComponent, to modify the menu container.
e.g. to add a header or footer, search field etc.

We're currently using this in a project, so any comments gratefully received even if you choose not to accept the PR.

I haven't added any documentation, but can do if there's any interest in this solution.

### DIFF
--- a/examples/src/app.js
+++ b/examples/src/app.js
@@ -25,8 +25,8 @@ ReactDOM.render(
 		<GithubUsers label="GitHub users (Async with fetch.js)" />
 		<NumericSelect label="Numeric Values" />
 		<BooleanSelect label="Boolean Values" />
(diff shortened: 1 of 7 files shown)


ssinha-mdsol:
@jossmac how can we do this in v2 if I need to add a static footer ?

jossmac:
Use the [components API](https://react-select.com/components) to replace whatever you want. For a static footer you'd have something like:

```jsx

const Menu = ({ children, ...props }) => (
  <components.Menu {...props}>
    {children}
    <div>
      Your Custom Footer
    </div>
  </components.Menu>
);

<Select
  components={{ Menu }}
/>
```

PR #2376, opened 2018-02-16
Selected disabled items are now flagged with CSS class

wijnbladh:
Thanks for this nice component. I was doing some demos with this some time ago, and in our UI flow it was possible that other selections made by the user invalidated an option previously selected in the react-select. Therefore I added a minimal change that allowed me to highlight selected-but-currently-invalid-options.


jossmac:
Thank you @wijnbladh 🙂

PR #2382, opened 2018-02-19
adds build step to make flow types available to consumers

petegleeson:
Hello! 👋  

I added a step to the build script that uses `flow-copy-source` to copy over each file in src and add the `.flow` postfix.

I published the lib with these changes to a local npm registry and tested it out in another project. Below is a gif the flow types in action.

consumer-of-react-select

I did get some grief from flow when I tried to use the `components` prop on Select. I plan to investigate these further when I get a chance.


JedWatson:
Thanks @petegleeson!

PR #2391, opened 2018-02-21
Add initial search input value

pawlowskim:
I just needed to pre-populate search input value with custom text so I have added optional props for this one.  I'm only not sure about tests. Could someone take a look a test file?


jiangweiliu-aa:
hi @pawlowskim How can you get permission to create a PR directly? I have a PR as well, However, I can't commit it because of permission issue.

ajaymathur:
@pawlowskim From my understanding:

- when you move out of select the search text resets and select closes. Therefore in this case it will be wrong since it will show a search text but select will be closed?

Can you please create a example also?

pawlowskim:
@ajaymathur It depends how you configure the select component. PR doesn't change anything beside that you can set it up to show query text when you first time render it. By default it's always empty text.

JedWatson:
Thanks for PR @pawlowskim 

I'm concerned that this could interact weirdly with the async and creatable wrappers - the initial loaded results and input value would be out of sync.

This is actually already implemented in v2, would you be OK if we closed this and left it out of v1? It would be simpler than checking all the edge case behaviours and writing tests for it here.

pawlowskim:
If it's implemented in v2, and it might impact other parts of the component then you should definitelly close the PR.

PR #2407, opened 2018-02-28
passing labelKey and valueKey into optionRenderer and valueRenderer

jiangweiliu-aa:
**Changes made**:

Make `valueRenderer` and `optionRenderer` access `labelKey` and `valueKey`.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -568,7 +568,7 @@ class Select extends React.Component {
 	}
 
 	getOptionLabel (op) {
-		return op[this.props.labelKey];
+		return op.option[this.props.labelKey];
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


src/Select.js
```diff
@@ -568,7 +568,7 @@ class Select extends React.Component {
 	}
 
 	getOptionLabel (op) {
-		return op[this.props.labelKey];
+		return op.option[this.props.labelKey];
```
jossmac: Is this (`op.option`) intended?

jossmac: Is this (`op.option`) intended?

flexdinesh:
Hey @jiangweiliu-aa [v1 of react-select is no longer maintained](https://github.com/JedWatson/react-select/pull/3786). 

PR #2408, opened 2018-02-28
Fix #1667: Do not output invalid empty 'aria-owns' attr

### LINKED ISSUE #1667: Aria-owns is empty when the searchable option is true
When the select is used as autocomplete, the aria-owns of the input is empty:

```
<input id="personal_state" aria-errormessage="personal.address.state-error" aria-invalid="false" autocomplete="off" role="combobox" aria-expanded="false" aria-owns="" aria-haspopup="false" aria-activedescendant="react-select-5--value" value="" style="width: 5px; box-sizing: content-box;">
```

Can we have a configuration where we can add the parent id? This way we can use the input props to add the attribute manually.

Complete markup:
```
<div class="Select select Select--single is-searchable has-value">
    <div class="Select-control"><span class="Select-multi-value-wrapper" id="react-select-5--value"><div class="Select-value"><span class="Select-value-label" role="option" aria-selected="true" id="react-select-5--value-item">CA</span>
    </div>
    <div class="Select-input" style="display: inline-block;">
        <input id="personal_state" aria-errormessage="personal.address.state-error" aria-invalid="false" autocomplete="off" role="combobox" aria-expanded="false" aria-owns="" aria-haspopup="false" aria-activedescendant="react-select-5--value" value="" style="width: 5px; box-sizing: content-box;">
        <div style="position: absolute; top: 0px; left: 0px; visibility: hidden; height: 0px; overflow: scroll; white-space: pre; font-size: 14px; font-family: Verdana, Geneva, sans-serif; font-weight: normal; font-style: normal; letter-spacing: normal;"></div>
    </div>
    </span><span class="Select-arrow-zone"><span class="Select-arrow"></span></span>
</div>
</div>
```

jnachtigall:
Fixes #1667 

`aria-owns=""` is not a valid attribute. I noticed this when using the [aXe accessibility checker](https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/). 

image

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -876,7 +876,7 @@ class Select extends React.Component {
 			'aria-haspopup': '' + isOpen,
 			'aria-label': this.props['aria-label'],
 			'aria-labelledby': this.props['aria-labelledby'],
-			'aria-owns': ariaOwns,
+			...(ariaOwns ? { 'aria-owns': ariaOwns } : {}),
 			className: className,
 			onBlur: this.handleInputBlur,
 			onChange: this.handleInputChange,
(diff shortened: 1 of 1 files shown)


flexdinesh:
Hey @jnachtigall [v1 of react-select is no longer maintained](https://github.com/JedWatson/react-select/pull/3786). 

PR #2410, opened 2018-03-01
show at least one unselected option while open select

jiangweiliu-aa:


jossmac:
Thanks for the PR @jiangweiliu-aa

This isn't the desired default behaviour for react-select

PR #2411, opened 2018-03-01
update to v0.3 of codesandboxer

Noviny:
Released v0.3 of `react-codesandboxer`, and updated exampleWrapper to use the newer version.

This was a fairly major API rewrite, but should be a much more stable base to build on. Easiest upgrade after that is to get a 'loading' svg, so we can show this to people while they are waiting to fetch files.

Biggest wins:
codesandboxer handles all file fetching itself, pushing nothing back onto you.
Now can fetch any number of files from github, and send them to codesandbox.
Has security around only fetching a file once.
Resolves all relative imports.

### DIFF
--- a/docs/ExampleWrapper.js
+++ b/docs/ExampleWrapper.js
@@ -2,50 +2,18 @@
 
 import glam from 'glam';
 import React, { Component } from 'react';
-import CodeSandboxer, { replaceImports } from 'react-codesandboxer';
-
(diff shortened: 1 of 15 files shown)


Noviny:
@jossmac How did you generate the svg icons? We can now get an `isLoading` update from the render prop, but would need a new icon for it.

jossmac:
This is awesome! Thank you @Noviny

Issue #1128, opened 2016-07-31
<div> can't be nested in <span> according to HTML spec


MilovanovM:
The component renders:
<img width="354" alt="div-span" src="https://cloud.githubusercontent.com/assets/1742065/17276507/5ebdabce-5734-11e6-8d9b-7750bb5f80c8.png">

alainds:
I use your component and I do the same comment : the rendered HTML code is invalid. You can't put a <div> into a <span>. 
If <div> was replaced <span>for the 2 elements "Select-input"  and "Select-placeholder", it would resolve the problem.  (In the code of the plugin, it's just a parameter).
Thanks.

JedWatson:
Closing as this has been resolved in v2, ref #2289

PR #2413, opened 2018-03-02
make html valid for multi select wrapper

apdrsn:
This fixes this issue #1128 by changing span to div to make it valid html as ```<div>``` is not allowed as child of a ```<span>```

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -1173,10 +1173,10 @@ class Select extends React.Component {
 					onTouchStart={this.handleTouchStart}
 					style={this.props.style}
 				>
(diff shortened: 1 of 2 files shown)


JedWatson:
Thanks for the PR @apdrsn

This is a nice improvement but I'm not inclined to change the DOM structure of v1 at the moment. There are potential knock-on effects (e.g custom CSS selectors for styling, test suites, etc) that could be impacted in userland and that would require a major version bump.

This is resolved in `v2` which will be landing in master soon so I'm going to close this for the current version.

Issue #2181, opened 2017-11-29
Warning: `ref` is not a prop


sabina4:
Hello,

I am using ref in <Select.Creatable ref={(node) => this.selector = node} /> and getting this warning:
CreatableSelect: `ref` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop.

aaronincincy:
This appears to be causing warnings for me (react-select@1.1.0, react@15.6.2). I'm using ref={r=>this.selector = r} and getting:

Warning: CreatableSelect: ref is not a prop. Trying to access it will result in undefined being returned. If you need to access the same value within the child component, you should pass it as a different prop. (https://fb.me/react-special-props)

I believe this is because ref is reserved, so you can't really pass it through like you are. If you renamed it to selectRef for Creatable (and updated the documentation for this exception to the Select API) I believe that would solve the issue.

My code seems to still work ok (it was working before upgrading from rc10), but I'd like to get rid of the warning if possible.

jonfreedman:
I am also seeing this warning in the console, it would be great to be able get rid of it.

This is with react 16.2.0 & react-select 1.2.0

x2es:
Also got this warning.
React docs suggest to use `kindOfCustomRef` prop instead `ref`
https://reactjs.org/docs/refs-and-the-dom.html#exposing-dom-refs-to-parent-components

lynxtaa:
As temporary solution, you can wrap Creatable in class-based component like this: 

```jsx
import { Creatable as ReactSelectCreatable } from 'react-select'
class Creatable extends React.Component {
  render() {
    return <ReactSelectCreatable {...this.props} />
  }
}
```

jfengel:
I am also seeing this error, but oddly, only via react-virtualized-select. When I use the component directly, it presents no problems.

The error, as far as I can tell, refers to the `ref: PropTypes.func` on line 293. When I remove it, the problem goes away.

Moejoe90:
Any updates regarding this?

diogocorrea:
+1

donperi:
+1

Stardog:
I get this warning. I am using ref on a Creatable to clear the inputValue after creating a new entry.
```
newOptionClick(value) {
  this.props.newOptionClick(value)
  this.tagSelect.current.select.setState({inputValue: ''})
}

render() {
return <Creatable
          ref={this.tagSelect}
          multi={true}
[...]

PR #2416, opened 2018-03-05
Use `innerRef` prop instead of `ref` to avoid warning

meriadec:
2018-03-05_1841x150

relatable: #2181

### DIFF
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -154,7 +154,7 @@ class CreatableSelect extends React.Component {
 
 	render () {
 		const {
-			ref: refProp,
+			innerRef: refProp,
 			...restProps
 		} = this.props;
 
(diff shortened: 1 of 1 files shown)


sungwoncho:
This change will break apps that are relying on `ref` prop. Should we consider deprecating `ref` while adding `innerRef` rather than removing `ref` altogether?

JedWatson:
Thanks for the PR @meriadec 

I agree with @sungwoncho - we can't make this change without a major bump, and I'm inclined to implement the new `ref` forwarding from React 16.3 as the solution to this. In the meantime, we could add a `selectRef` prop to get rid of the warning if somebody wants to PR that.

Dem0n3D:
@JedWatson, please, do something with this warning already.

PR #2430, opened 2018-03-13
Add use cases for clear with tab, click and  touched and dragged

RajaBellebon:
@gwyneplaine , adding cases for : 

- should display the options menu when tapped
- should not display the options menu when touched and dragged
- remove any selection - add smoketest for removal on keyboard action.
- on tapping `clear` use iphone viewport to validate tap event
- on tapping and dragging `clear`
- click on Clear button, menu shall remain closed


gwyneplaine:
lgtm

PR #2432, opened 2018-03-13
ExampleWrapper can now show code examples inline

Noviny:
Added in displaying example code to example wrapper.


gwyneplaine:
amazing!

PR #2436, opened 2018-03-14
[V2] test cases

ajaymathur:


JedWatson:
👍 looks good @ajaymathur

src/Select.js
```diff
@@ -335,10 +335,14 @@ export default class Select extends Component<Props, State> {
   }
   onMenuClose() {
     this.onInputChange('', { action: 'menu-close' });
-    this.props.onMenuClose();
+    if (this.props.onMenuClose) {
```
ajaymathur: added these guards to not beak the component in case these are not provided

src/Select.js
```diff
@@ -335,10 +335,14 @@ export default class Select extends Component<Props, State> {
...
+      this.props.onMenuClose();
+    }
   }
   onInputChange(newValue: string, actionMeta: InputActionMeta) {
-    this.props.onInputChange(newValue, actionMeta);
+    if (this.props.onInputChange) {
```
ajaymathur: added these guards to not beak the component in case these are not provided

src/__tests__/AsyncCreatable.test.js
```diff
@@ -0,0 +1,31 @@
...
+  const asyncCreatableWrapper = mount(<AsyncCreatable />);
+  expect(asyncCreatableWrapper.find(Select).exists()).toBeTruthy();
+});
+
+test('render decorated select with props passed', () => {
+  const asyncCreatableWrapper = mount(<AsyncCreatable className="foo"/>);
+  expect(asyncCreatableWrapper.find(Select).props().className).toBe('foo');
+});
+
+test('to show the create option in menu', () => {
```
gwyneplaine: Please adjust this case to account for the fact that we only want to show the create option element, after the options have been loaded asynchronously.

src/__tests__/StateManaged.test.js
```diff
@@ -0,0 +1,612 @@
...
+import cases from 'jest-in-case';
+
+import { OPTIONS } from './constants';
+import Select from '../';
+import { components } from '../components';
+
+const { Control, Menu, MultiValue, Placeholder, Option } = components;
+
+const BASIC_PROPS = { options: OPTIONS, name: 'test-input-name' };
+
```
gwyneplaine: Please add test cases for validating the expected behaviour of exposed controlled props: 
1. props.inputValue
2. props.menuIsOpen 
3. props.value

src/__tests__/AsyncCreatable.test.js
ajaymathur: done

src/__tests__/StateManaged.test.js
ajaymathur: done

ajaymathur:
@gwyneplaine PR is ready for review 🙂

PR #2437, opened 2018-03-14
[V2] Bcon/v2 documentation 2

Noviny:
NB: This branch will seriously mess up until the `@atlaskit/docs` PR is merged, and that package has been released, updated version is being used here.


Noviny:
Okay, I branched this off v2, and I think that is why there are conflicts.

PR #2447, opened 2018-03-19
[v2] Clee/v2 surface select ref

gwyneplaine:
Passing through a prop to expose the ref of the base Select component  isn’t as simple as I thought it would be. 
Given that stateManager will be wrapping around the outer most component., which means it potentially wraps around Selects that are not the base select. 

For instance Creatable is currently
```
<stateManagerWrapper>
	<Creatable>
			<Select> 
```

while Select is currently 
```
<stateManagerWrapper>
     <Select> 
```

This solution attempts to reconcile this in the best way i can. Very open to feedback


JedWatson:
So, while not ideal, this isn't a bad approach to things.

@bvaughn have you seen any better ideas on how to handle getting a ref to a component that's wrapped in HOCs or parents?

... if only we had some sort of `forwardRef` thing... 😉 

@gwyneplaine the other idea I had (which would introduce less API and could be supported indefinitely) would be to implement `focus` and `blur` forwarding from the `Async`, `Creatable` and `StateManager` components.

gwyneplaine:
@JedWatson done, we're now forwarding the focus and blur methods instead of making the entire ref available to the consumer

bvaughn:
@JedWatson I assume that using `forwardRef` isn't an option because you don't want to raise the minimum version of React in your dependencies?

This PR approach seems reasonable.

JedWatson:
👍 Thanks @gwyneplaine and for the feedback @bvaughn

PR #2449, opened 2018-03-19
[v2] make the children of multi-value-remove customisable

gwyneplaine:
Previously this was an internally declared child <CrossIcon/>
This change makes <CrossIcon/> a default value of the children prop of MultiValueRemove, allowing users to customise the child of the MultiValueRemove container.


JedWatson:
I see @jossmac's influence here 😄 

LGTM

PR #2457, opened 2018-03-21
Character item focus

johnnytomcat:
When in a non searchable field, tap a key and focus to that item. Ex: US States -Tapping 'T' would focus to Texas, tapping it again for focus Tennessee.


gwyneplaine:
@johnnytomcat, thanks for the PR, however I think we should examine this as a future feature of react-select v2, as this is difficult to do in v1 without compromising existing functionality.  @JedWatson  would like to get your thoughts on this.

flexdinesh:
Hey @johnnytomcat [v1 of react-select is no longer maintained](https://github.com/JedWatson/react-select/pull/3786). 

Thanks for taking your time out to create this PR but since this PR has changes for `v1`, we cannot merge it right away. If you would like to discuss this feature for the latest versions of `react-select`, please open an issue with feature details and we could move forward with the discussion there. We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest versions.

PR #2460, opened 2018-03-22
[V2] Tests

ajaymathur:
Appologies to raise another PR. But as discussed with @gwyneplaine ,  I have moved all the select props related test cases into select.

I have kept some basic tests of clicking option in statemanged and keyboard interaction to make sure it works as expected and comes in handy if there any any future changes.


gwyneplaine:
thnx for this @ajaymathur! Never apologise for contributing,  👍 lgtm!

PR #2462, opened 2018-03-23
Bumps flow-bin to version 0.68

petegleeson:
I have been looking at upgrading Atlaskit to Flow version 0.68. To get there, react-select needs to be compatible with 0.68. Upgrade was fairly straightforward. I'll highlight some of the most important changes.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -583,11 +583,12 @@ export default class Select extends Component<Props, State> {
     if (!this.state.isFocused) {
       this.openAfterFocus = true;
       this.focusInput();
-    } else if (!this.state.menuIsOpen) {
+    } else if (!this.props.menuIsOpen) {
       this.openMenu('first');
     } else {
       this.onMenuClose();
     }
+    // $FlowFixMe HTMLElement type does not have tagName property
     if (event.target.tagName !== 'INPUT') {
       event.preventDefault();
     }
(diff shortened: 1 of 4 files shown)


src/Select.js
```diff
@@ -599,9 +600,7 @@ export default class Select extends Component<Props, State> {
     }
     if (this.props.isDisabled) return;
     const { isMulti, menuIsOpen } = this.props;
-    if (!this.focused) {
```
petegleeson: `this.focused` didn't exist in the class so now we just call `this.focusInput()` everytime

src/Select.js
```diff
@@ -583,11 +583,12 @@ export default class Select extends Component<Props, State> {
     if (!this.state.isFocused) {
       this.openAfterFocus = true;
       this.focusInput();
-    } else if (!this.state.menuIsOpen) {
```
petegleeson: This looked like a legit bug that the upgrade uncovered.

src/Select.js
```diff
@@ -599,9 +600,7 @@ export default class Select extends Component<Props, State> {
     }
     if (this.props.isDisabled) return;
     const { isMulti, menuIsOpen } = this.props;
-    if (!this.focused) {
-      this.focusInput();
-    }
+    this.focusInput();
```
gwyneplaine: good catch! This afaik is a bug, but I'll let @JedWatson  confirm this. but ideally we should still be doing a check for whether or not the input is focused, the conditional should be `if (!this.state.isFocused)` 

Once again, I'll let either @JedWatson  or @jossmac  confirm this

src/Select.js
gwyneplaine: it is indeed, thanks for this!

src/Select.js
JedWatson: I'll have to look at this tomorrow - we _were_ tracking the instance property because `setState` is async, and unreliable for this. So the solution may be to ensure that is re-implemented.

JedWatson: Actually, got a chance to look through and I believe this is safe to do without the check. So going to merge 👍

PR #2465, opened 2018-03-25
Remain focused on option after selecting it

### LINKED ISSUE #2169: focusedOption jumps when selecting options after disabling closeOnSelect and removeSelected
I configured the Select component to accept multiple values (`multi={true}`),
to stay open when a value is selected (`closeOnSelect={false}`), and to keep
selected values listed (`removeSelected={false}`). When I now select an element
in the dropdown, the component tells its `menuRenderer=` function to highlight
an adjacent element as the `{ focusedOption }`, until the mouse moves by any
further amount.

The reason I want to keep selected options listed in an open menu is that I'm
using a custom `optionComponent=` that visually indicates whether it's presently
selected. The focus jumping away from the mouse in this scenario made a buggy
impression on all users I showed it to so far.

## Demo ##
This Plunker should illustrate the suprising behaviour:
https://plnkr.co/edit/qMnBup1tUnTYeqrEdIFQ?p=preview

tpbowden:
Currently, selecting an option in multi select with `removeSelected=false` and `closeOnSelect=false` causes the focused option to jump which looks odd, especially when using a mouse. This ensures it stays correctly selected.

Fixes #2169 (see issue for more details + demo)

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -646,6 +646,7 @@ class Select extends React.Component {
 		const visibleOptions = this._visibleOptions.filter(val => !val.disabled);
 		const lastValueIndex = visibleOptions.indexOf(value);
 		this.setValue(valueArray.concat(value));
+		if(!this.props.closeOnSelect) { return; }
 		if (visibleOptions.length - 1 === lastValueIndex) {
 			// the last option was selected; focus the second-last one
 			this.focusOption(visibleOptions[lastValueIndex - 1]);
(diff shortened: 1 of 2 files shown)


fedde-s:
I tried the result of merging this branch into `master`, and I can confirm that I could no longer reproduce the issue after that. And the test included does start failing if I break the new line of code.

fedde-s:
And the failing automated test seems to have failed in the same way (‘Deploy directory 'docs/dist' does not exist’) for #2482, the last pull request that was accepted into `master`.

jossmac:
Thanks @fedde-s 🙂

PR #2468, opened 2018-03-26
[v2] fix: resolve active descendent

jossmac:


ajaymathur:
@jossmac We can remove skip from these test cases ? https://github.com/JedWatson/react-select/blob/v2/src/__tests__/Select.test.js#L1098

NOTE: in multi select test -> we need to add `isMulti` to prop. it got skipped here.

jossmac:
@ajaymathur i'll leave that with you. This fix is supposed to be discrete.

PR #2471, opened 2018-03-26
[V2] update async docs

ajaymathur:
Working on docs for async component.

I am inclined towards the given kind of format, ( though there could be better heading for TL;DR ).

- Having a separate markdown than rest of head data will enable us to add an edit button which takes us directly to the markdown content. ( and keeps it simple, though it maybe because I have habit of looking at this doc format. ).

- TL;DR - gives minimum code( not working ) to understand loadOptions. Devs can read code faster than text  🤔)

cc. @JedWatson , @gwyneplaine , @Noviny , @jossmac , @treshugart

### DIFF
--- a/docs/pages/async/Docs.js
+++ b/docs/pages/async/Docs.js
@@ -1,0 +1,99 @@
+// @flow
+
+import React from 'react';
+import { Props } from '@atlaskit/docs';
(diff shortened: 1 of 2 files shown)


ajaymathur:
I have merged this in docs, PR. I will be working on that.

PR #2472, opened 2018-03-27
[V2] remove wildcard from lib ignore path in flowconfig

gwyneplaine:
We want to be ignoring only our lib files, not lib files in other directories (i.e. node_modules).
Ignoring lib files of our dependencies can cause potential dependency type resolution issues.


JedWatson:
Good catch, thanks @gwyneplaine

PR #2479, opened 2018-03-28
Add maxItems option for multi select

kiruh:
Instead of disabling input for multiselect (#1931) I added **maxItems** option which stops the menu from being opened when max items is reached.

### DIFF
--- a/examples/src/components/Multiselect.js
+++ b/examples/src/components/Multiselect.js
@@ -27,6 +27,7 @@ var MultiSelectField = createClass({
 			disabled: false,
 			crazy: false,
 			stayOpen: false,
+			maxItems: false,
 			value: [],
 			rtl: false,
 		};
@@ -46,7 +47,7 @@ var MultiSelectField = createClass({
 	},
 
 	render () {
-		const { crazy, disabled, stayOpen, value } = this.state;
+		const { crazy, disabled, stayOpen, value, maxItems } = this.state;
 		const options = crazy ? WHY_WOULD_YOU : FLAVOURS;
 		return (
 			<div className="section">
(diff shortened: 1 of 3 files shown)


jossmac:
Thanks for the PR @kiruh,

You will be able to achieve the same result in v2 with the `menuIsOpen` property. More info and docs http://bit.ly/react-select-v2

chrisehlee:
Hello @jossmac , is there an example for this? The docs [here](https://deploy-preview-2289--react-select.netlify.com/upgrade-guide) wasn't all too explicit about how to achieve this !?!?

jossmac:
@chrisehlee I had a play around in CodeSandbox -- here's one way you could achieve "max options" with the v2 API:

https://codesandbox.io/s/pwq3xv3nx7

chrisehlee:
@jossmac Thank you so much for this 

> I chose to show an alternative "no results message", rather than just not opening the menu, which I think would feel unresponsive and be a bad experience for the user.

super intuitive and is fine for me .. quick question; How do i get only labels from the selected options in the handleChange function, i only get expected results when using a single select .. a multi select brings back the entire array (sorry i'm a bit late on js arrays)

jossmac:
As of beta 7 it's part of the `onChange` handler, which takes the shape:

```js
onChange = (
  newOptionsArray,
  {
    action, // which action took place to cause `onChange` to be called
    option, // the option that was operated on, see `action` above
  }
) => void;
```

Actions can be one of:

```
'clear' | 'create-option' | 'deselect-option' | 'pop-value' | 'remove-value' | 'select-option' | 'set-value'
```

You can [view the change log](https://github.com/JedWatson/react-select/blob/v2/HISTORY.md) for more updates 🙂

chrisehlee:
You've been super helpful. Will be checking it out over the weekend. Thanks a ton.
________________________________

PR #2480, opened 2018-03-29
[V2] Fix animation transition group prop warning

Blasz:
This fixes a react warning when rendering a disabled select that uses animated components.
The cause was animated component transition group props being spread on a div element.
Moving the disabled input render into Input component itself allows the props to be stripped by the respective Animated Input component.

I also refactored the cypress test file to show each example using animated components as well.

Finally I've updated the animated component names to distinguish them with the default components when debugging using react dev tools.


Blasz:
cc @jossmac

JedWatson:
Merged via #2496

PR #2489, opened 2018-04-04
[V2] add pretty-proptypes

Noviny:
@gwyneplaine


gwyneplaine:
@JedWatson this is the body of work for subsuming pretty-proptypes into the rs repo, as well as the addition of the 'overrides' feature for manually overriding particular proptype definitions via an `overrides` object (see example below). Would be good to get your eyes on this

```javascript
<Props
     props={require('!!extract-react-types-loader!../../PropTypes/Select')}
     overrides={{
        components: (props) => <Prop {...props} shapeComponent={() => null} type="All Components 
        Object" /> 
     }}
 />
```

JedWatson:
Looks good, thanks @Noviny

PR #2492, opened 2018-04-04
Exposed promptText as a prop in Select.Creatable, default value set to 'Create option'

RamYadlapalli:
Exposed promptText as a prop in Select.Creatable, default value set to 'Create option'


madmus:
I fully support this PR highly needed option because of context in which creatable/select is used, and of course i18n support instead of hardcoded strings.

RamYadlapalli:
@JedWatson Can you look into this PR?

jossmac:
This is available as `promptTextCreator` property.

PR #2508, opened 2018-04-10
small updates for s4 specific functionality

keithm-thompson:


keithm-thompson:
please ignore --- one of those days 😅

PR #2517, opened 2018-04-13
use external pretty-proptypes

Noviny:
Additional things:

- Removed a couple of now unused dependencies
- We are now using a custom button (which we likely want to customise more) for expand/collapse
- The expand/collapse animation is smoother
- Displays of function calls in defaultProps now display the name of the function, rather than their function definition


JedWatson:
🎉 looks great @Noviny but there's an issue with the build (reference to a missing file)

Mind fixing that up?

PR #2519, opened 2018-04-17
V2 relative value container

gwyneplaine:
Fix for long single values in the value container not being properly truncated. 
This was a regression introduced in resolving the FF flex bug. 

As the FF flex bug can no longer be replicated under the same conditions, it is safe to revert this change. 

Example added to Tests.js in docs to track this fix (will added cypress automation in a subsequent PR)

### DIFF
--- a/cypress/integration/select_spec.js
+++ b/cypress/integration/select_spec.js
@@ -36,7 +36,7 @@ describe('New Select', function() {
       it('Should expand the menu when click on the arrow ' + view, function() {
         cy
           .get(selector.toggleMenus)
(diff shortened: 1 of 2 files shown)


JedWatson:
@gwyneplaine snapshots and cypress tests are failing, can you sort those out?

Should this be included before beta.1?

gwyneplaine:
@JedWatson  yes this should be included in beta.1, snapshots updated. Cypress tests have been adequately adjusted for now, (this will be my next priority following beta.1)

PR #2524, opened 2018-04-17
Add references between `backspaceRemoves` and `deleteRemoves`

IanVS:
I recently had a bug in a field which I did not want to allow users to clear out.  I set `clearable` and `backspaceRemoves` to `false`, but didn't realize there was also a separate prop for `deleteRemoves`.  This PR just adds a little note to each prop to reference the other, so that users are aware they both exist.


JedWatson:
👍 thanks @IanVS

PR #2526, opened 2018-04-18
change color used to highlight changed sections on the components page

Noviny:
Small change so that the color used in `/components` to indicate the changed section is not the same as the focus outline


JedWatson:
Nice change 👍

PR #2528, opened 2018-04-18
V2 fix ie11 (bugs)

gwyneplaine:
ie11 bug fix for placeholder allignment in the select control


gwyneplaine:
@JedWatson  @jossmac  this should be a satisfactory fix to ie11 bugs with absolutely positioned elements within a flex parent, and also a satisfactory fix to our website not working in ie11.

gwyneplaine:
@JedWatson  jest snapshots updated and tests resolved, this should be good to pull in

PR #2535, opened 2018-04-20
Export indicator icons for use in customizations

jesstelford:
Enables things such as changing the size of the indicator icon, like so:

```javascript
// @flow

import React from 'react';

const DropdownIndicator = (props) => {
  return (
    <components.DropdownIndicator {...props}>
      <components.DownChevron size="2rem" />
    </components.DropdownIndicator>
  );
};

export default () => (
  <Select
    components={{ DropdownIndicator }}
    options={[...]}
  />
);
```

**eg:** 
screen shot 2018-04-20 at 10 39 57 pm

### DIFF
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -15,6 +15,8 @@ import {
   DropdownIndicator,
   LoadingIndicator,
   IndicatorSeparator,
+  DownChevron,
+  CrossIcon,
 } from './indicators';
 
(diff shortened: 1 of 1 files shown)


jesstelford:
Welp, I'm pretty sure I didn't `npm uninstall eslint`, so not sure what's up with Travis :/

Issue #2217, opened 2017-12-14
Enter event


ayozebarrera:
Hi, good work!

I think the problem comes with #212...

Every input item on a form, if its focused, it will submit the form **(if not empty!)** if you press enter key... but react-select will open the list... why? I don't see any good reason for that...

Anyway, if this will be the behaviour, how can I solve it? 

thanks

alexxmde:
Same issue here

sstern6:
Having the same problem

thiagolelis-luizalabs:
You can override the onInputKeyDown function, like this:

```
    onInputKeyDown(event) {
        switch (event.keyCode) {
            case 13: // ENTER
                event.preventDefault();
                break;
        }
    }
```

And on Select

```
<Select
    {...otherProps}
    onInputKeyDown={onInputKeyDown}
/>
```

sstern6:
@thiagolelis-luizalabs thanks but wasnt this issue fixed and is supposed to be intended functionality, not supposed to be overwritten?

Issue #518, opened 2015-10-12
Enter is swallowed when searchable is false


ptim:
I'm using v0.7.0, and find that:
- when `searchable` is false, and
- I've focussed a field
- hit the down arrow to highlight an option
- hit enter to select the option
- hit enter on the closed Select, 

...that the field doesn't budge: the form is not submitted, and 

The select is inside a form element, and the form is submitted as expected if `searchable` is true.

Here's my Select:

```
<Select className="form-control"
  name="gender"
  placeholder="Gender"
  options={options}
  value={this.props.patient.gender}
  onChange={this.handleGenderChange}
  searchable={false}
/>
```

I found this related issue which was previously merged: https://github.com/JedWatson/react-select/issues/212

Couldn't immediately fathom why, looking at the source...

Thanks for the awesome component :)

jossmac:
Version 1 of react-select is no longer supported.

In the best interest of the community we've decided to spend the time we have available on the latest version.

We apologise for any inconvenience.

Please see:

1. [v1 to v2 upgrade guide](https://react-select.com/upgrade-guide)
1. [v1 documentation and examples](v1.react-select.com)

PR #2539, opened 2018-04-20
Pressing ENTER should propagate when a closed Select has focus

MrLeebo:
I'm opening this PR to start a conversation rather than to get what's here merged necessarily. There appears to be competing motivations for better accessibility support vs native `<select>` compatibility. I think the pendulum has swung too far away from native compatibility with a recent change. In v1.2.1 if I have this:

```jsx
<form onSubmit={props.handleSubmit}>
  <Select />
</form>
```

I should be able to press ENTER when the Select menu is closed to submit the form. This is how it worked in v1.0.0, but #1428 changed the behavior to open the menu instead. I don't know what an appropriate compromise should be for accessibility, but not allowing ENTER to propagate runs counter to a lot of people's expectations, I think. It seems that this should be fine, because they can press virtually any other key on the keyboard to open the menu, but I'm not an expert so I was hoping to discuss.

Related issues: #2217 and #518

I created a build branch with these changes applied so if you want to try out these changes in an app before this PR gets merged, you can change your react-select dependency to:
```json
"react-select": "git://github.com/MrLeebo/react-select.git#build",
```

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -485,13 +485,13 @@ class Select extends React.Component {
 				this.selectFocusedOption();
 				break;
 			case 13: // enter
+				if (!this.state.isOpen) {
+					break;
+				}
+
 				event.preventDefault();
 				event.stopPropagation();
-				if (this.state.isOpen) {
-					this.selectFocusedOption();
-				} else {
-					this.focusNextOption();
-				}
+				this.selectFocusedOption();
 				break;
 			case 27: // escape
 				event.preventDefault();
--- a/test/Select-test.js
+++ b/test/Select-test.js
@@ -4408,16 +4408,6 @@ describe('Select', () => {
 
 			});
 
-			it('expands the drop down when the enter key is pressed', () => {
-
-				return expect(wrapper,
-						'with event', 'keyDown', KEY_ENTER, 'on', <div className="Select-control" />,
-						'queried for', <input role="combobox" />)
-						.then(input => {
-							expect(instance.state.focusedOption, 'to equal', { value: 'one', label: 'label one' });
-						});
-
-			});
 		});
 	});
 


degr:
@MrLeebo, Thanks for decision. Unfortunately I need this future yesterday, but PR still in open status.

So, currently I found this way how to override default behavior:

```javascript

export default class InternalSelect extends Select {
    handleKeyDown(event) {
        if (this.props.disabled) return;
        if (typeof this.props.onInputKeyDown === 'function') {
            this.props.onInputKeyDown(event);
            if (event.defaultPrevented) {
                return;
            }
        }
        if(event.keyCode === 13) {
            if (!this.state.isOpen) {
                return;
            }
            event.preventDefault();
            event.stopPropagation();
            this.selectFocusedOption();
        } else {
            super.handleKeyDown(event);
        }
    }
}
```

gwyneplaine:
@MrLeebo, thanks for this. This is behaviour that has been removed from v2, specifically for cases like this. In react-select v2, enter no longer opens the menu or stop event propagation. 

We won't be making this change in v1, as it is a major breaking change to functionality

degr:
Issue not resolved in V2, it event become worse.
If in V1 I can inherit Select and override onKeyDown function, in V2 it's impossible, because library exports `manageState(Select)`, which use Select as composition element, so, inheritance is impossible.
Now, when I click enter button, this lines of code will be executed:
```javascript
switch (event.key) {
   ....code
    case 'Enter':
        if (menuIsOpen) {
          if (!focusedOption) return;
          if (isComposing) return;
          _this7.selectOption(focusedOption);
        } else {
          _this7.focusOption('first');
        }
        break;
   ...code
}
event.preventDefault();
```

degr:
I solve issue. To do it, I pass in all my selects "formSubmit" handler, I start listen `onMenuOpen` and `onMenuClose` events, and inside of `onKeyDown` handler I was able to determinate that select have closed drop-down list, that `enter` key was pressed and call `formSubmit` instead of html submit event. Realy guys, in V1 it was more simple.

degr:
Actually, I found if you will throw and exception inside of your onKeyDown function, event will not be prevented, and, most probably, it will bubble to your form. Yes, throw exception.

jitenderchand1:
Any update on this? This supposed to be default behaviour

TrySound:
@gwyneplaine This shit is broken. We have react-select in 50 forms and they are not accessible.

gwyneplaine:
@TrySound noted, PR for it incoming;

PR #2545, opened 2018-04-23
[v2] Force min-width of select to widest option

jesstelford:
Avoids the issue of widths jumping around when selecting a wide value

**DO NOT MERGE**. This is a 100% WIP, and I haven't even tried to run the code.

I've left comments inline about specifics, but you can see from the gifs below the difference:

**Before / "fluid" width**
react-select-fluid-width

**After / fixed width**
react-select-forced-width

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -1038,48 +1038,54 @@ export default class Select extends Component<Props, State> {
       MultiValueRemove,
       SingleValue,
       Placeholder,
+      ValueSpacer,
     } = this.components;
     const { commonProps } = this;
-    const { isDisabled, isMulti, inputValue, placeholder } = this.props;
+    const { isDisabled, isMulti, inputValue, placeholder, options } = this.props;
     const { selectValue } = this.state;
 
     if (!this.hasValue()) {
       return inputValue ? null : (
-        <Placeholder {...commonProps} key="placeholder" isDisabled={isDisabled}>
-          {placeholder}
-        </Placeholder>
+        <ValueSpacer key="placeholder" values={{ placeholder, options }}>
+          <Placeholder {...commonProps} isDisabled={isDisabled}>
+            {placeholder}
+          </Placeholder>
+        </ValueSpacer>
       );
     }
     if (isMulti) {
       return selectValue.map(opt => (
-        <MultiValue
-          {...commonProps}
-          components={{
-            Container: MultiValueContainer,
-            Label: MultiValueLabel,
-            Remove: MultiValueRemove,
-          }}
-          isDisabled={isDisabled}
-          key={this.getOptionValue(opt)}
-          removeProps={{
-            onClick: () => this.removeValue(opt),
-            onMouseDown: e => {
-              e.preventDefault();
-              e.stopPropagation();
-            },
-          }}
-          data={opt}
-        >
-          {this.formatOptionLabel(opt, 'value')}
-        </MultiValue>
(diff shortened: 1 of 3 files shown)


src/Select.js
```diff
@@ -1038,48 +1038,54 @@ export default class Select extends Component<Props, State> {
...
-    const { isDisabled, isMulti, inputValue, placeholder } = this.props;
+    const { isDisabled, isMulti, inputValue, placeholder, options } = this.props;
     const { selectValue } = this.state;
 
     if (!this.hasValue()) {
       return inputValue ? null : (
-        <Placeholder {...commonProps} key="placeholder" isDisabled={isDisabled}>
-          {placeholder}
-        </Placeholder>
+        <ValueSpacer key="placeholder" values={{ placeholder, options }}>
```
jesstelford: This gives a wrapper over any values selected (note: I only really thought about the single value case)

src/components/containers.js
```diff
@@ -135,3 +137,34 @@ export const IndicatorsContainer = (props: IndicatorContainerProps) => {
...
+// Value Spacer
+// ==============================
+
+export type ValueSpacerProps = CommonProps & {
+  /** Set when the value container should hold multiple values. This is important for styling. */
+  values: any,
+};
+export const valueSpacerCSS = () => ({
+  display: 'flex ',
+  flexDirection: 'column',
```
jesstelford: We're stacking a new element on top of the "Value" element.

src/components/containers.js
```diff
@@ -135,3 +137,34 @@ export const IndicatorsContainer = (props: IndicatorContainerProps) => {
...
+  alignItems: 'flex-start',
```
jesstelford: Keep them left-aligned

src/components/containers.js
```diff
@@ -135,3 +137,34 @@ export const IndicatorsContainer = (props: IndicatorContainerProps) => {
...
+});
+export const ValueSpacer = (props: ValueSpacerProps) => {
+  const { children, values } = props;
+  return (
+    <Div
+      css={getStyles('value-spacer', props)}
+      {...innerProps}
+    >
+      <div style={{ margin: 0, padding: 0, height: 0, visibility: 'hidden' }}>
```
jesstelford: This "new element" is a zero-height div that can't be seen, but which still effects layout / flow of the DOM.

src/components/containers.js
```diff
@@ -135,3 +137,34 @@ export const IndicatorsContainer = (props: IndicatorContainerProps) => {
...
+        {values.placeholder && <Placeholder>{values.placeholder}</Placeholder>}
+        {values.options && values.options.length && values.options.map(
+          option => <SingleValue>{option}</SingleValue>
+        )}
```
jesstelford: We then fill the hidden element with all the possible values (including the placeholder), which will force width to the widest element.

Since they use the `<Placeholder>`/`<SingleValue>` components, the style should be identical to that of the actual visibly selected element, meaning no width jumping!

src/Select.js
jesstelford: Ideally it's an option for users to opt-into this behaviour.

jesstelford:
The problem I found with modifying the `ValueContainer` component was it also contains the `<input>`, which threw things out when I was trying to do my flexbox trickery.

I'm still finding my way around the v2 API, so I may have overlooked something!

PR #2548, opened 2018-04-24
added universal/ssr rendering support

falconmick:
I did a quick find all and there are a bunch of other window/navigator calls in the code base, so there might be more SSR issues, however at this current point this was the only change required to get the basic Select Hello World to work.

As I move forward using this v2 branch I'll make PR's as I find issues and solve them.

Cheers.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -256,7 +256,7 @@ export function isTouchCapable() {
 // ==============================
 
 export function isMobileDevice() {
-  return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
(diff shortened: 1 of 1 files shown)


falconmick:
oh and If I get time perhapse I can upgrade the test bench to use SSR, that is provided the test suite is doing headless testing and not virtual mounting

baer:
Thanks for this PR! using `navigator` prevents v2 from working with Gatsby and Next.js. It's probably worth attaching this PR to a broader issue to scour the codebase.

JedWatson:
Thanks @falconmick! Will ship this with `beta.2` now

Issue #1934, opened 2017-08-18
shouldKeyDownEventCreateNewOption should hook on KeyPress instead of KeyDown 


rpominov:
`event.keyCode` has different meaning on `KeyDown` and `KeyPress`. On `KeyDown` it refers to code of actual physical key on keyboard, while on `KeyPress` it refers to a character that will be typed. See https://stackoverflow.com/a/11030750

Because currently it's hooked to `KeyDown` event, it's impossible to enter other characters that share physical key with the comma.

For example on the following keyboard with the default `shouldKeyDownEventCreateNewOption` I'm unable to enter "<" and "Б". Also the default callback doesn't handle comma key in the number row. And I can't add it either because I would need to sacrifice "6" and "^".

![image](https://user-images.githubusercontent.com/825702/29476221-995c4050-846b-11e7-969d-31391ebf9dcc.png)

See also #345

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you feel this issue / pull request is still relevant and you'd like us to review it, please leave a comment and we'll do our best to get back to you.

Issue #345, opened 2015-07-29
Unexpected behaviour when type query in Russian


sergeylaptev:
Hello, when I type some queries in Russian and use "б" letter, input always interrupt. The root of evil at this line https://github.com/JedWatson/react-select/blob/master/src/Select.js#L427. On a russian keyboard we have "б" letter, whereas in english it's ",". We can fix this, by checking charCode instead of key number.

bruderstein:
Unfortunately charCode isn't set for keyDown events, only keyPress events. 
Please excuse my lack of knowledge about Russian, but I presume you do have a comma key on your keyboard (or at least, can enter one somehow), can you enter it in http://www.w3.org/2002/09/tests/keys.html and post the results here?

Keyboard events in browsers are broken, and i18n just makes it worse, so I'm not sure there's a good solution to this, but we can take a look :)

webvit:
**Character     "б"**
Keyboard Event Properties:
<pre>
Data     keydown     keypress     keyup
keyCode     188     1073     188
charCode     —     1073     —
</pre>

bruderstein:
Thanks for posting.  I meant can you post what happens when you press your "comma" key?

webvit:
"comma [б]" key works as "enter" key, adds tags in multi mode
and delimiter option does not override "comma" to another symbol

bruderstein:
Yes, but I need to know what http://www.w3.org/2002/09/tests/keys.html says, when you press your comma key, not your б key, so that we can work out how we can decide what the user pressed.

б is simply being treated as comma, because it has the same keycode.  On keyPress, we see the character, and I fear we may have to wait for keyPress before acting on ','.  However, I'd like to see what the w3 link above says when you press your version of your comma.  

Many thanks for assisting!

webvit:
I understood,

**Firefox, Mac OS:**
<pre>
Data	keydown	keypress	keyup
keyCode   188 	0 	188
charCode   —   44   —
Character 	,
</pre>

**Chrome, Mac OS:**
<pre>
Data	keydown keypress keyup
keyCode   188	44	188
charCode	—  44	  —
Character	,
</pre>

**Safari:**
<pre>
Data	keydown	keypress	keyup
keyCode	  188	44	188
charCode	—	 44	—
Character	,
</pre>

bruderstein:
ok, great, thanks. That was kinda what I was afraid of. I'll have a bit of a play and see if I can come up with a solution.

[...]

PR #2554, opened 2018-04-24
Fix creating  new option on key down on unexpected keys

benjamin-pirotte:
related to #345 & #1934

This PR:

- Pass full event to *shouldKeyDownEventCreateNewOption* for more customization.
- Use event.key instead of event.keyCode in default *shouldKeyDownEventCreateNewOption* props, to avoid unexpected behaviour  on key down.
- Update README.md to reflect new *shouldKeyDownEventCreateNewOption* interface.
- Update Creatable tests


flexdinesh:
Hey @benjamin-pirotte  [v1 of react-select is no longer maintained](https://github.com/JedWatson/react-select/pull/3786). 

Thanks for taking your time out to create this PR but since this is relevant to v1, we recommend maintaining your own v1 fork if you want to address `v1` issues. We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest versions.

I'm closing this issue now. Feel free add your thoughts in #3981 and reference this issue/PR if needed.

PR #2564, opened 2018-04-26
[v2] fix group component in experimental

gwyneplaine:
Fixed bug where custom group component in Experimental.js (calendar example) wasn't passing cx down to the heading component.


JedWatson:
Thanks @gwyneplaine!

PR #2565, opened 2018-04-26
More accessors, set inputValue via props

skogsbaer:
While developing a quite advanced auto complete component for medicaments, I needed access to some of the select component's state. Also, I needed to ability to set the inputValue via props. So here is the PR!


skogsbaer:
I would appreciate feedback on this PR. If it can go in, I'll take care of the failing checks.

jossmac:
Hi @skogsbaer,

Thanks for the PR -- unfortunately, in v1, the `inputValue` is tightly coupled with async and creatable variants.

You'll be pleased to know that this is not an issue in v2. For more info and docs please checkout http://bit.ly/react-select-v2

PR #2568, opened 2018-04-27
[V2| WIP] Jesstelford forced width

gwyneplaine:
An expansion on @jesstelford #2545 notably:
- We no longer map over and render each hidden component
- We now call 'getLongestOption' which iterates over the passed in options array to retrieve the 'longest' option, according to an evaluation of what's currently passed in as 'getOptionLabel'. This does mean that our longest option is now an approximation, but I think this is an adequate trade off compared to rendering the entire options list. 
- getLongestOption flattens the options array so that we're also accounting for option groups. 
- We also pass in the Placeholder and SingleValue components from the components object, as opposed to directly importing them into the file. This allows us to use the placeholder and singlevalue components passed in by the user.

Theoretically this should be an performance optimisation for longer (long) lists. Will add a performance benchmark soon

### DIFF
--- a/src/components/Placeholder.js
+++ b/src/components/Placeholder.js
@@ -10,13 +10,17 @@ export type PlaceholderProps = CommonProps & {
   children: Node,
   /** props passed to the wrapping element for the group. */
   innerProps: { [string]: any },
-};
+} & State;
 
-export const css = () => ({
+export type State = {
+  cssPosition: 'relative' | 'absolute',
+}
+
+export const css = ({ cssPosition }: State) => ({
   color: colors.neutral50,
   marginLeft: spacing.baseUnit / 2,
   marginRight: spacing.baseUnit / 2,
-  position: 'absolute',
+  position: cssPosition || 'absolute',
   top: '50%',
   transform: 'translateY(-50%)',
 });
--- a/src/components/containers.js
+++ b/src/components/containers.js
@@ -3,9 +3,8 @@ import React, { Component, type Node, type ElementRef } from 'react';
 
 import { Div } from '../primitives';
 import { spacing } from '../theme';
-import Placeholder from './placeholder';
-import SingleValue from './SingleValue';
-import type { CommonProps, KeyboardEventHandler } from '../types';
+import type { CommonProps, KeyboardEventHandler, OptionsType, OptionType } from '../types';
+import type { SelectComponents } from '../components';
 
 // ==============================
 // Root Container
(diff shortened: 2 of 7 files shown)


src/components/containers.js
```diff
@@ -135,3 +136,120 @@ export const IndicatorsContainer = (props: IndicatorContainerProps) => {
...
+            </SingleValue>
+          }
+        </Div>
+        {children}
+      </div>
+    );
+  };
+}
+
+// export const ValueSpacer = (props: ValueSpacerProps) => {
```
gwyneplaine: Left this in for performance comparison metrics later.

src/components/containers.js
```diff
@@ -135,3 +136,120 @@ export const IndicatorsContainer = (props: IndicatorContainerProps) => {
...
+    this.longestOption = null;
+  }
+
+  componentDidUpdate (prevProps: ValueSpacerProps) {
+    if (prevProps.options !== this.props.options) {
+      this.longestOption = this.getLongestOption(this.props.options);
+    }
+  }
+
+  flattenOptions = (options: OptionsType) => {
```
gwyneplaine: reduce over the passed in options list, if its a nested array, flatten it out.

src/components/containers.js
```diff
@@ -135,3 +136,120 @@ export const IndicatorsContainer = (props: IndicatorContainerProps) => {
...
+    return options.reduce((acc, current) => {
+      if (Array.isArray(current.options)) return [...acc, ...current.options];
+      return [...acc, current];
+    }, []);
+  }
+
+  getLongestOption = (options: OptionsType) => {
+    const { getOptionLabel } = this.props;
```
gwyneplaine: here we use the passed in 'getOptionLabel' fn to do a best guess of which option will appear the longest. There are problems with this approach, that mostly stem from customisation, however we sacrifice a little bit of accuracy here for not having to render every option as a value to the dom'

src/components/containers.js
```diff
@@ -135,3 +136,120 @@ export const IndicatorsContainer = (props: IndicatorContainerProps) => {
...
+
+    // $FlowFixMe
+    const { getStyles } = commonProps;
+    const { Placeholder, SingleValue } = components;
+    return (
+      <div>
+        <Div
+          css={getStyles('valueSpacer', this.props)}
+        >
+          {placeholder &&
```
gwyneplaine: here we render the placeholder still, but no longer render every option as a singleValue.

src/components/containers.js
```diff
@@ -135,3 +136,120 @@ export const IndicatorsContainer = (props: IndicatorContainerProps) => {
...
+      isDisabled,
+      formatOptionLabel,
+      placeholder,
+      components,
+      ...commonProps
+    } = this.props;
+
+    // $FlowFixMe
+    const { getStyles } = commonProps;
+    const { Placeholder, SingleValue } = components;
```
gwyneplaine: Notice here that we pass in the Placeholder and SingleValue component from the internally held components object. This allows us to use the custom placeholder and singlevalue components supplied by the consumer (should one be supplied).

src/components/Placeholder.js
```diff
@@ -10,13 +10,17 @@ export type PlaceholderProps = CommonProps & {
...
+  cssPosition: 'relative' | 'absolute',
+}
 
-export const css = () => ({
+export const css = ({ cssPosition }: State) => ({
   color: colors.neutral50,
   marginLeft: spacing.baseUnit / 2,
   marginRight: spacing.baseUnit / 2,
-  position: 'absolute',
+  position: cssPosition || 'absolute',
```
gwyneplaine: the base style for placeholder and singlevalue set the position value to 'absolute' which neutralises the core premise of this implementation. I don't if this is the best way to go about fixing this for *just the instance of singleValue and placeholder we use in ValueSpacer, and I'd like to discuss this and its alternatives in the context of not complicating user customisations.

src/components/containers.js
```diff
@@ -135,3 +136,120 @@ export const IndicatorsContainer = (props: IndicatorContainerProps) => {
...
+          {placeholder &&
+            <Placeholder
+              {...commonProps}
+              cssPosition="relative"
+              isDisabled={isDisabled}
+            >
+              {placeholder}
+            </Placeholder>}
+          {this.longestOption &&
+            <SingleValue
```
gwyneplaine: I'm also a little unhappy about this at the moment, ideally this should be a lightweight dummy component that realistically replicates the width of the passed in SingleValue or Placeholder components, it's a little icky that we're not only rendering these twice, but also configuring them with a full featureset.

PR #2572, opened 2018-04-27
[v2] Allow Creatable option first

iojichervo:
There are cases, like if you want to submit `" "` as a new option, where you need to scroll down to the `Create " "` option to select this

Allowing to set the `Create new...` option first (like it was before v2) makes it easier to handle it

### DIFF
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -29,6 +29,9 @@ export type CreatableProps = {
      created, and `onChange` will **not** be called. Use this when you need more
      control over what happens when new options are created. */
   onCreateOption?: string => void,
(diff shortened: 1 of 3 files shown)


gwyneplaine:
thanks @iojichervo. I've merged this body of work in here #2638. 
That PR contains a merge conflict resolution and a minor change to the prop, instead of a boolean, we'd prefer it to be a prop that expects either the string value of 'first' or 'last' (defaulting to last). 
We also renamed the prop `createOptionPosition` as opposed to `createOptionFirst` just for parity with our current patterns around positioning props like this.

PR #2581, opened 2018-05-02
[V2] live region implementation

gwyneplaine:
This PR includes work to substitute the existing minimal aria-label based accessibility implementation with an aria-live-region. 

We remove the following labels: 
### Group
- aria-labelledby
- aria-label 
- aria-expanded 
- role 

### Menu
- aria-multiselectable 
- role 
- id 
- innerRef (innerRef is now a prop exposed from Menu, I'm happy to revert this change as while logical it is not consistent with the the api we expose for the rest of our components) 

### Option 
- aria-selected 
- role 

## aria live region Implementation 
- Added an A11yText element with two children 
  - a string representing value events (select-option, deselect-option etc) 
  - a string representing context 
    - focused-value
    - focused-option
    - results 
    - instructions for what to do depending on where the user is focused in the select. 
- See inline code comments for details


src/Select.js
```diff
@@ -4,6 +4,15 @@ import React, { Component, type ElementRef, type Node } from 'react';
 
 import { createFilter } from './filters';
 import { DummyInput, ScrollBlock, ScrollCaptor } from './internal/index';
+import {
```
gwyneplaine: We import in helpers for constructing the relevant aria live messages. These can eventually be exposed as a command api like so 
```javascript 
export const a11yCommands = {
 valueFocus,
 optionFocus, 
 resultContext 
 instructions 
 valueEvent
}
```

src/Select.js
```diff
@@ -224,7 +232,7 @@ export const defaultProps = {
   pageSize: 5,
   placeholder: 'Select...',
   screenReaderStatus: ({ count }: { count: number }) =>
```
gwyneplaine: Keeping screenReaderStatus for now to not break the existing accessibility API, but i think we should do away with this in favor of the more verbose pattern specified above.

src/Select.js
```diff
@@ -236,12 +244,22 @@ type MenuOptions = {
 };
 
 type State = {
```
gwyneplaine: Internally we now have two new pieces of state information. 
* ariaLiveContext: (a string value holding instructions for assistive tech based on the context of the user (in menu, focused on value, in input) 
* ariaLiveSelection: a string value holding instructions for assistive tech based on value events that occur (select-option, deselect-option, pop-value etc.)

src/Select.js
```diff
@@ -635,6 +672,16 @@ export default class Select extends Component<Props, State> {
   // ==============================
   // Helpers
   // ==============================
+  announceAriaLiveSelection = ({ event, context }: { event: string, context: ValueEventContext }) => {
+    this.setState({
+      ariaLiveSelection: valueEventAriaMessage(event, context),
```
gwyneplaine: this calls the valueEventAriaMessage helper with the passed in event and context to resolve to the next value of ariaLiveSelection in state.

src/Select.js
```diff
@@ -635,6 +672,16 @@ export default class Select extends Component<Props, State> {
...
+    });
+  }
+  announceAriaLiveContext = ({ event, context }: { event: string, context?: InstructionsContext }) => {
+    this.setState({
+      ariaLiveContext: instructionsAriaMessage(event, { ...context, label: this.props['aria-label'] }),
```
gwyneplaine: this calls the instructionsAriaMessage helper with the passed in event and context to resolve to the next value of ariaLiveContext in state.

src/Select.js
```diff
@@ -1063,23 +1111,23 @@ export default class Select extends Component<Props, State> {
...
   // ==============================
-
-  renderScreenReaderStatus() {
-    const { screenReaderStatus } = this.props;
-    return (
-      <A11yText aria-atomic="true" aria-live="polite" role="status">
-        {screenReaderStatus({ count: this.countOptions() })}
-      </A11yText>
-    );
+  constructAriaLiveMessage () {
```
gwyneplaine: each invocation of constructAriaLiveMessage creates an aria live message to represent the following: 
- focusedValue 
- focusedOption 
- results (amount of options available) and inputvalue 
- ariaLiveContext

We construct this every time on render, in order for us to not have to imperatively infer value focus or option focus change in event calls or within componentWillReceiveProps. valueFocus and optionFocus can change due to any number of user interactions and non user interaction based side-effects from async loading to controlled prop changes outside of our intended scope. Likewise with results.

src/Select.js
```diff
@@ -1481,6 +1506,16 @@ export default class Select extends Component<Props, State> {
     }
   }
 
+  renderLiveRegion () {
```
gwyneplaine: if the input (and hence the select) is not focused, don't render the A11yText elements.

src/accessibility/index.js
```diff
@@ -0,0 +1,30 @@
+export type InstructionsContext = { isSearchable?: boolean, isMulti?: boolean, label?: string };
+export type ValueEventContext = { value: string };
```
gwyneplaine: I think it's actually better here to pass in value as an object and the function getOptionLabel as a function argument here. 

This solves for Creatable select, where we need to be able to either fire an announcementEvent from the wrapping Creatable HoC, or infer whether or not a value isNew to change the returned event string. However this also requires that the user apply getOptionLabel manually, which doesn't seem great, given that it's a prop that they've passed in with the implicit understanding that we'd be handling application for them.

src/accessibility/index.js
```diff
@@ -0,0 +1,30 @@
...
+    case 'deselect-option':
+    case 'pop-value':
+    case 'remove-value':
+      return `option ${value}, deselected.`;
+    case 'select-option':
+      return `option ${value}, selected.`;
+  }
+};
+
+export const valueFocusAriaMessage = ({ focusedValue, getOptionLabel, selectValue }) => `value ${getOptionLabel(focusedValue)} focused, ${selectValue.indexOf(focusedValue) + 1} of ${selectValue.length}.`;
```
gwyneplaine: see above for argument against this way of doing things. Of course, passing in the focusedValue as a string here, means that instead of getOptionLabel and selectValue, we'll probably have to pass through selectValueSize (number) and focusedValueIndex (number).

src/accessibility/index.js
```diff
@@ -0,0 +1,30 @@
...
+export const optionFocusAriaMessage = ({ focusedOption, getOptionLabel, options }) => `option ${getOptionLabel(focusedOption)} focused, ${options.indexOf(focusedOption) + 1} of ${options.length}.`;
```
gwyneplaine: same as above

src/accessibility/index.js
```diff
@@ -0,0 +1,30 @@
...
+export const resultsAriaMessage = ({ inputValue, screenReaderMessage }) => `${screenReaderMessage} for search term ${inputValue}.`;
```
gwyneplaine: we ingest a screenReaderMessage which is the returned value of calling the screenReaderStatus prop.

PR #2584, opened 2018-05-04
V2 multi value keyboard

gwyneplaine:


docs/examples/BasicMulti.js
```diff
@@ -8,6 +8,14 @@ export default () => (
     defaultValue={[colourOptions[2], colourOptions[3]]}
     isMulti
     name="colors"
+    onValueClick={onValueClick}
```
JedWatson: I thought we decided to leave this out of the basic implementation?

gwyneplaine: fixed!

PR #2589, opened 2018-05-07
[V2] Add isDisabled class to Option

eemeli:
Previously, `Option` would be rendered with just the `isFocused` and `isSelected` classname variants, i.e. without the `isDisabled` variant. This makes styling disabled options practically speaking impossible using CSS and `<Selected className>`.

This adds the `isDisabled` classname variant to `Option`.


gwyneplaine:
@JedWatson  this looks good to me.

JedWatson:
Thanks @eemeli! will release this shortly.

Issue #2583, opened 2018-05-03
Content Overflowing in IE 11 (2.0.0-beta.2)


kidroca:
# Content Overflowing in IE 11 (2.0.0-beta.2)

- Text Input (Select Value) appears with scrollbar next to the indicator icons 
- Dropdown menu content is overflowing outside the menu

![image](https://user-images.githubusercontent.com/12156624/39568313-7450fefc-4eca-11e8-8e3f-4a4277a3def0.png)

## Plunk example 

https://codesandbox.io/s/4j2y5k6y09

I'm not sure whether IE 11 is targeted for support, Edge seems to work fine 

*Edit: Not IE 10 -> IE 11*

jprosevear:
IE10 has been unsupported by MS for > 2 years:
https://www.microsoft.com/en-us/windowsforbusiness/end-of-ie-support

kidroca:
Oops, sorry my bad this is IE 11

gwyneplaine:
@kidroca  thanks for this, we'll look into patching this

gwyneplaine:
Closed by #2593 #2596

PR #2593, opened 2018-05-09
V2 ie11 fixes

gwyneplaine:
IE11 fixes for https://github.com/JedWatson/react-select/issues/2583

### DIFF
--- a/src/components/Menu.js
+++ b/src/components/Menu.js
@@ -238,6 +238,7 @@ export const menuCSS = ({ maxHeight, placement }: MenuState) => ({
   maxHeight: maxHeight,
   position: 'absolute',
   width: '100%',
+  overflowY: 'scroll',
   zIndex: 1,
 });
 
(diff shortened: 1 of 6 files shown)


src/components/Menu.js
```diff
@@ -327,8 +327,9 @@ export type MenuListProps = {
 export type MenuListComponentProps = CommonProps &
   MenuListProps &
   MenuListState;
-export const menuListCSS = () => ({
+export const menuListCSS = ({ maxHeight }: MenuListComponentProps) => ({
   flexGrow: 1,
+  flexBasis: (maxHeight - spacing.baseUnit*2),
```
gwyneplaine: In IE11 flex-basis auto is ignored. We need to specify an initial size for our flex element, otherwise our content will overflow. 

Luckily we have access to the maxHeight of our menu.

gwyneplaine:
e2e tests pass locally, looks to be a CI dependency issue with cypress @JedWatson, if you could rerun the e2e runner in CI that would be excellent. I don't have write access to do so unfortunately.

JedWatson:
Thanks @gwyneplaine 

I've pushed a change to the docker image used in CircleCI to the `v2` branch, which installs the official cypress container and will hopefully resolves the issue. Going to merge this and try to get CI fixed in the main branch.

PR #2596, opened 2018-05-10
remove flex from menu to address ie11 concerns

gwyneplaine:
see #2583


JedWatson:
@jossmac we're going to remove this for now because it severely breaks IE11 support, and isn't actually giving us a lot.

Let's revisit when we get back to header / footer implementations in the Menu.

PR #2601, opened 2018-05-14
Fixed issue with console warnings about onExited & in attributes when…

MikeLimeRocket:
Fixed issue of console warnings when isSearchable was false due to spreading all props.
DummyInput now passes an explicit list of props to Input

### DIFF
--- a/src/internal/DummyInput.js
+++ b/src/internal/DummyInput.js
@@ -3,28 +3,48 @@
 import React from 'react';
 import { Input } from '../primitives';
 
-export default (props: any) => (
-  <Input
-    {...props}
-    css={{
-      // get rid of any default styles
-      background: 0,
-      border: 0,
-      fontSize: 'inherit',
-      outline: 0,
-      padding: 0,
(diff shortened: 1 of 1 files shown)


src/internal/DummyInput.js
```diff
@@ -3,28 +3,48 @@
 import React from 'react';
 import { Input } from '../primitives';
 
-export default (props: any) => (
```
JedWatson: A safer way to do this would be to pull out the props that we know could be included (for the `Animated` components) and exclude them. That way we don't need to maintain a white-list of all props.

e.g

```js
export default ({ in, out, ...props }: any) => {
```

MikeLimeRocket: Had considered that option but is it only these props for Animated components that we would ever need to exclude?

MikeLimeRocket: Disregard my last question. I have discussed with @gwyneplaine and have better context now on this one.

gwyneplaine:
Merged!

PR #2611, opened 2018-05-15
New filterMaxResults option to easily limit the default filtering

### LINKED ISSUE #126: Limiting number of options displayed in dropdown
For example, how would one only show the 5 closest matches to input text in the dropdown, despite having 1000 options which are searchable.

peterbartos:
New filterMaxResults option to easily limit the default filtering function results count. 
It shows the first `filterMaxResults` items. FIX #126

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,7 @@ function onInputKeyDown(event) {
 | `delimiter` | string | ',' | delimiter to use to join multiple values |
 | `disabled` | boolean | false | whether the Select is disabled or not |
 | `escapeClearsValue` | boolean | true | whether escape clears the value when the menu is closed |
+| `filterMaxResults` | number | undefined | limit the number of results being shown when using default filtering options. Shows first `filterMaxResults` items. |
 | `filterOption` | function | undefined | method to filter a single option `(option, filterString) => boolean` |
 | `filterOptions` | boolean or function | undefined | boolean to enable default filtering or function to filter the options array `([options], filterString, [values]) => [options]` |
 | `id` | string | undefined | html id to set on the input element for accessibility or tests
--- a/src/Select.js
+++ b/src/Select.js
@@ -990,6 +990,7 @@ class Select extends React.Component {
 				filterValue,
 				excludeOptions,
 				{
+					filterMaxResults: this.props.filterMaxResults,
 					filterOption: this.props.filterOption,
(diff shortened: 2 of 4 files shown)


gwyneplaine:
@peterbartos  thanks for this PR, just a few queries: 
1. What is your reason for reducing results after filtration in this way? If it's to avoid rendering a long list of results for performance reasons, I would imagine react-select virtualized would provide an adequate  resolution here.
2. I'm a little worried that we're arbitrarily reducing the number of results from filterOptions, without a good way of viewing all of the results. This seems like functionality that would be best paired with pagination, but seems incomplete on its own. 
3. Aside from the two concerns above, the slice on filterMaxResults would probably be better placed on the filterOptions method within the Select, as opposed to the default prop value, so that filterMaxResults logic is always applied on filtration, irrespective of the value of this.props.filterOptions. 

Would love your thoughts, thanks again

peterbartos:
Hi @gwyneplaine thank you for your attention

1. The reason I find this useful is a simple solution for #126. In our case we have few hundreds of options and there's a lag on filtering in the cases it displays many results. You are right, the virtualized package seems like a solution too, but this update would help without having to include extra packages.

1. I see your worries of lacking pagination. I think for the use cases reported in #126 the pagination is not needed – as the benefit is to show only the closest matches. I'm concerned that the pagination would be hard for this update.

1. You are right. If you think this feature is still possible to include without the pagination, I would quickly move the `slice` into Select as you suggest.

whiterook6:
Any update on this? Would love to see maxFilterResults included in react-select.

vitexikora:
Quick temp solution:
```jsx
const resultLimit = 10
let i = 0
return <Select filterOption={({label}, query) => label.indexOf(query) >= 0 && i++ < resultLimit} />
```

flexdinesh:
Hey @peterbartos  [v1 of react-select is no longer maintained](https://github.com/JedWatson/react-select/pull/3786). 

Thanks for taking your time out to create this PR but since this is relevant to v1, we recommend maintaining your own v1 fork if you want to address `v1` issues. We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest versions.

I'm closing this issue now. Feel free add your thoughts in #3981 and reference this issue/PR if needed.

PR #2615, opened 2018-05-16
[v2] Add brief class name explanation to Styles docs

fozcodes:
IMO this is an important addition, especially for anyone upgrading. Moving styles from your project's
CSS into JS can be pretty painful in an already existing codebase. Also, if the rest of your project
is using CSS/Sass/etc then having the styles for a single component like this in JS can be hard to maintain
or abstract.

I think it's important that users know they have either option and if they wanna jump down either rabbit hole
they can.

Just my two cents...


gwyneplaine:
thanks for this @fozcodes

PR #2622, opened 2018-05-17
[v2] Use getOptionValue to compare values when filtering selectValue

### LINKED ISSUE #2403: selectOption comparison not working for multi Select
In Select.js 

```
selectOption = (newValue: OptionType) => {
    const { isMulti } = this.props;
    if (isMulti) {
      if (this.isOptionSelected(newValue, selectValue)) {
        this.setValue(
          selectValue.filter(i => i !== newValue),
          'deselect-option'
        );
      } else {
        this.setValue([...selectValue, newValue], 'select-option');
      }
    } else {
      this.setValue(newValue, 'select-option');
    }
  };
```

`i !== newValue` this comparison compares if two JSON objects are the same, not if their contents is the same, which causes the select not to recognize the unselected value as it might be a different JSON in memory, but with the same {label, value} contents.

inv8der:
This should fix the issue described in #2403 by allowing you to customize the comparison of `newValue` to the `selectValue`. This is the same logic used inside of `isOptionSelected` (in the default case), so unless there is some use case I'm not considering, this change should be safe.

Closes #2403


JedWatson:
This looks like a good fix, thanks @inv8der 

I've just updated this PR with latest from master, which will hopefully fix the CI fail - happy to merge this one that passes 👍

PR #2623, opened 2018-05-18
Add "popout" example to advanced docs

jossmac:
- new prop `controlShouldDisplayValue`


JedWatson:
I'm inclined to spend a bit more time on the naming of that prop, but otherwise looks great 👍 

We can easily rename it later (before the next beta release though!)

PR #2628, opened 2018-05-18
[v2] Fix Warning: Unknown prop `css` on <div> tag in SingleValue

guigs:
I believe this should have been removed on cf2557dd53a817f8cd55f26e96db5bc239d61980.


gwyneplaine:
thanks for this @guigs, merged

Issue #2614, opened 2018-05-16
Why do you collect code in a single bundle?


retyui:
**Problem:**
I try reduce my bundels size. In my project I use one import:
```js
import Select from 'react-select'; // 84 KB
```
I would like to have the opportunity to do so:
```js
import Select from 'react-select/dist/Select';
```

**Describe the solution you'd like**
Why do you collect code in a single js bundle?
Use flatten struct!
Like as: `lodash-es`, `date-fns`, `react-router`, `ramda`, `recompose`

TrySound:
@retyui recompose will remove path imports in the next major release. We rely on treeshakability which works well as you can see [here](https://github.com/acdlite/recompose/blob/master/.size-snapshot.json#L16-L19).

retyui:
@TrySound 
![image](https://user-images.githubusercontent.com/4661784/40385275-b7f69498-5e0e-11e8-85a9-c55eb0abf6c3.png)

TrySound:
@retyui I show you example. That size number is generated with webpack itself. Treeshaking works well in webpack 4. Just use it. Use tooling to be sure that treeshaking works.

Path imports is import hell. It's illness when file has more imports than code. Treeshaking solve this.

PR #2634, opened 2018-05-22
Add src code to npm package

retyui:
I'm working on reducing the bundle in my project.
We are very serious about optimizing the size of the final gang for some reason the client.

At the moment I only use one import
```js
```
But in the end all the code gets to the gang, half of which remains there dead part.

But I would like to import only the necessary parts as for example made in `lodash-es, date-fns, react-router, ramda, recompose`
Example:
```js
import { Select } from  'react-select';
// or
import Select from 'react-select/es/Select';
```


retyui:
https://github.com/JedWatson/react-select/issues/2614

PR #2646, opened 2018-05-24
[V2] Flowtype upgrade to v0.72.0

lunij:
When the consumer of react-select v2 already uses Flowtype v0.72.0 then react-select v2 beta-6 is reporting Flow errors. This is solved by upgrading Flowtype and fixing the necessary type issues.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -23,6 +23,7 @@ import {
 
 import {
   defaultComponents,
+  type PlaceholderOrValue,
   type SelectComponents,
   type SelectComponentsConfig,
 } from './components/index';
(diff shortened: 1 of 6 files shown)


src/Select.js
```diff
@@ -1141,7 +1142,7 @@ export default class Select extends Component<Props, State> {
       />
     );
   }
-  renderPlaceholderOrValue() {
+  renderPlaceholderOrValue(): ?PlaceholderOrValue {
```
mike1808: I think `Node | Node[]` is enough. `renderPlaceholderOrValue` is not used outside of `render`

PR #2650, opened 2018-05-25
Update README.md

spankie:
add check for to handle where `selectedOption` is null in the example `handleChange` function


JedWatson:
Thanks @spankie

PR #2652, opened 2018-05-28
[v2] Re-render when defaultOptions changes

jesstelford:
WIP.

Discussed with @JedWatson - needs tests probably.

Current workaround is to set a key on the select item which forces it to rerender (`key={this.props.defaultOptions}`).


few-bits:
Hi @JedWatson, what about this MR for v2?

PR #2658, opened 2018-05-30
[V2] add newValue to set-value actionMeta in onInputChange

gwyneplaine:
Solution to support a use case where in the use of Async Select you may want to call loadOptions after the select value has been changed via user interaction. Please see an example of this case here 
https://codesandbox.io/s/18wmw5mk1j

added a `newValue` property to our InputActionMeta, when setValue is called, `onInputChange('', { action: 'set-value', newValue });`. I like this approach, but I’m also weary of overloading our actionMeta with conditional attributes like this.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -488,7 +488,7 @@ export default class Select extends Component<Props, State> {
   }
   setValue = (newValue: ValueType, action: ActionTypes = 'set-value') => {
     const { closeMenuOnSelect, isMulti, onChange } = this.props;
-    this.onInputChange('', { action: 'set-value' });
+    this.onInputChange('', { action: 'set-value', newValue });
     if (closeMenuOnSelect) {
       this.inputIsHiddenAfterUpdate = !isMulti;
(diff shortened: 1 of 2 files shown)


lukebatchelor:
(I may be responsible for the attrocity above, apologies).

The above solution should work well for what I'm trying to do, but just to put some more information here.

Because of the way we call Asyncs (and hence the consumers) `handleInputChange` handler on every `[input-change, menu-close, item-select, input-blur]`, all of those events (excluding input-change) will reset the input field (except in multi-select). 
I'm not sure if that's by design or not?

I guess at a high level one part what I would want to achieve is to be able to do something similar to the `onSelectResetsInput = false; closeMenuOnSelect = false` example here https://deploy-preview-2289--react-select.netlify.com/advanced but using a non-multi select. (Specifically, the onSelectResetsInput = false; part). That example only works because its a multi-select without `closeMenuOnSelect={false}` flag.

Not sure if all that makes sense...

PR #2659, opened 2018-05-30
[v2] making group options work with CreatableSelect

dpnolte:
Dear Jed and other Maintainers,

Hereby a pull request to support group options also with CreatableSelect.
Without this PR, the isValidNewOption builtin will fail with group options, because it does not compare the underlying options array in the group. Secondly, the compareOption method didn't use the GetOptionLabel and GetOptionValue builtins for any custom option types. 

The main changes are in Creatable.js.
The isValidNewOption is now called with option label/value builtins as parameters in the componentWillReceiveProps event. In the PR, the isValidNewOption will now check if the option is a group before comparing it for similarity to any existing option. In the compareOption function, the builtins GetOptionLabel and GetOptionValue builtins will be applied to extract label and value.

For illustration purposes, I've also added an example 'CreatableGrouped', and added it to the Creatable document page.

In any case, thanks for the wonderfull component. I've switched from a different multiselect component to this one, and I really like it.

Cheers,
  Derk

### DIFF
--- a/docs/examples/CreatableGrouped.js
+++ b/docs/examples/CreatableGrouped.js
@@ -1,0 +1,37 @@
+import React from 'react';
+
+import CreatableSelect from '../../src/Creatable';
+import { colourOptions, groupedOptions } from '../data';
+
+const groupStyles = {
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'space-between',
+};
+const groupBadgeStyles = {
+  backgroundColor: '#EBECF0',
+  borderRadius: '2em',
+  color: '#172B4D',
+  display: 'inline-block',
+  fontSize: 12,
+  fontWeight: 'normal',
+  lineHeight: '1',
+  minWidth: 1,
+  padding: '0.16666666666667em 0.5em',
+  textAlign: 'center',
+};
+
+const formatGroupLabel = data => (
+  <div style={groupStyles}>
+    <span>{data.label}</span>
+    <span style={groupBadgeStyles}>{data.options.length}</span>
+  </div>
+);
+
+export default () => (
+  <CreatableSelect
+    defaultValue={colourOptions[1]}
+    options={groupedOptions}
+    formatGroupLabel={formatGroupLabel}
+  />
+);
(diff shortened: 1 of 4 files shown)


marc0n3:
Hi @dpnolte, using your modification with a complex **getOptionLabel** i had to add
 `if (opt.__isNew__) return opt.label;` 
to mi code.

also using **filterOption** i had to add 
`function (opt,filter) {
				const data = opt.data;
				if (data.__isNew__) return opt.label;`

and in **grouped component** without **filterOption** defined i get the "Create" entry only if the label in the entry match the filter unless i add:

filterOption = function (opt,filter) {
				const data = opt.data;
				if (data.__isNew__) return opt.label;
				return createFilter()(opt,filter);
			}

Just for you to know, 
thanks for your work!

dpnolte:
Hi @marc0n3, thanks for sharing  👍  what happened if you didn't add the check for __isNew__? Did you use getNewOptionData built-in to define data for the 'Create' entry so that this entry also contains a value? cheers

lusa:
@JedWatson can this be merged into v2? I'm having this issue and this fixes it for me.

sarink:
@dpnolte is your fork up to date? I need this and can't wait any longer, so I was hoping to point directly at your fork in the meantime.

johannesdold:
In which version will this be fixed? It is not in 2.1, I guess.

Rudgey84:
Can anybody help me implement the workaround for this please - https://codesandbox.io/s/k5pyr75xor

landerton:
A workaround for me is to not use the creatable component at all and just append a create option to my list in `loadOptions`.

```
loadOptions = (value) => {
    return this.getSuggestions(value).then(
      rsp => {
        const suggestions = this.filterGroupedSuggestions(rsp.data)
        const creatable = {
          label: `Search "${value}"`,
          value: value,
          group: 'search'
        }
        return [creatable, ...suggestions]
      }
    )
  }
```

I already perform specific actions based on the option's `group` name. So in my case, I just added a new group called `search`.

ebonow:
Greetings everyone, 

I followed this back to verify the root of the issue. It would seem that this was resolved at some point in v2 as it appears to be working in react-select v2.4.4

demo: [codesandbox](https://codesandbox.io/s/react-select-pr-3338-creatable-support-with-groups-je1bn)

PR #2670, opened 2018-06-04
Update Select.js to meet accessibility requiremenets

ewagulik:
Currently there is an accessibility violation. Adding aria-label to buttons will make it possible for screen readers to convey the meaning of visual elements.


gwyneplaine:
@ewagulik  thanks for this as of https://github.com/JedWatson/react-select/pull/2581 (released in beta.7) we've intentionally added aria-hidden to all of the elements specified in this PR, as removing them from the knowledge of screen reading devices is better for accessibility than making them visible. 

All behaviour exposed by these elements, are accessible to non-sighted users through other means, closing this issue for now, happy to discuss this further :)

Issue #2675, opened 2018-06-05
[v2] Allow for unclearable (preset) options


daggmano:
Implement unclearable options in v2 (as provided by `clearableValue: false`) in previous version.

Ishneetkaur:
Implement unclearable options in v3 (as provided by clearableValue: false) in the previous version.

ebonow:
Greetings,

Based on the reply to the PR https://github.com/JedWatson/react-select/pull/2683, I will be closing this issue.

It would seem that this inspired the [Fixed Options](https://react-select.com/home#fixed-options) example on the react-select documentation in case anyone wanted to inquire further on how to implement something like this.

PR #2683, opened 2018-06-06
[v2] Allow fixed (preset) options in multiselect

daggmano:
Adds abililty to render preset (fixed) options in multiselect control. #2675

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -585,7 +585,6 @@ export default class Select extends Component<Props, State> {
     this.focusInput();
   };
   clearValue = () => {
-    console.log('clearValue');
     const { isMulti, onChange } = this.props;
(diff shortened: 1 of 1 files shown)


daggmano:
After giving this more thought overnight, I think it would be wise to get some feedback on this before going much further. Points to consider:

- Is `isFixed` a good property name? What would be better here?
- Rendering of the selected options should display fixed (preset) options first.
- What happens if an option is marked as `isFixed`, but not included in the initial `defaultValue` list?
- Would it be better to provide a separate property on `<Select>` for preset options, rather then expecting preset items to be in both the `options` and `defaultValue` list?
- Any other feedback from the product owners or community?

This is a fairly important use-case for us, but I think we need to take time to get the implementation right.

PR #2708, opened 2018-06-14
[V2] update components prop types

mike1808:
Indicator components can be explicitly set to `null` if the user doesn't want to render (e.g. separator). There were all necessary checks for indicator components so adding `| null` to the type was not an issue.


JedWatson:
Nice, thanks for this @mike1808!

PR #2717, opened 2018-06-18
[V2] Add meta to onChange when isMulti

SimeonC:
In the case of using isMulti and especially with `hideSelectedOptions = false` (for deselect) the action meta should include the option that is being selected/deselected.

This is kind of an extension of https://github.com/JedWatson/react-select/pull/2637. Only applies to `isMulti`.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -486,7 +486,7 @@ export default class Select extends Component<Props, State> {
       focusedValue: null,
     });
   }
-  setValue = (newValue: ValueType, action: ActionTypes = 'set-value') => {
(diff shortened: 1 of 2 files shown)


gwyneplaine:
@JedWatson  this seems like a viable solution to me, but wanted to sanity check potentially loading the actionMeta with conditional values like this with you.

JedWatson:
@gwyneplaine looks good to me, this is a valid use of what action meta was intended for

@SimeonC thanks for the PR, and the new tests as well!

PR #2719, opened 2018-06-18
[v2] Add option for disabling noOptionsMessage entirely

richardadjogah:
This fixes https://github.com/JedWatson/react-select/issues/2720.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -148,7 +148,7 @@ export type Props = {
   menuShouldScrollIntoView: boolean,
   /* Name of the HTML Input (optional - without this, no input will be rendered) */
   name?: string,
-  /* Text to display when there are no options */
+  /* Text to display when there are no options. If set to null, the entire menu won't be rendered when no options are present */
   noOptionsMessage: ({ inputValue: string }) => string,
   /* Handle blur events on the control */
   onBlur?: FocusEventHandler,
@@ -1350,6 +1350,10 @@ export default class Select extends Component<Props, State> {
     };
 
     let menuUI;
+
+    if (!this.hasOptions() && noOptionsMessage === null) {
+        return null;
+    }
 
     if (this.hasOptions()) {
       menuUI = menuOptions.render.map(item => {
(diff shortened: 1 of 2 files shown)


gwyneplaine:
@richardadjogah after looking through this PR, I'm not convinced that forging an assumption between setting null on the noOptionsMessage prop and not showing the menu is the correct way to resolve your issue. 

Would be great for @JedWatson to sanity check this, but I believe supplying you with either the results or the number of results in onInputChange  in conjunction with the menuIsOpen prop that we currently already allow you to control should allow you to do the same thing with a little more effort. 

Let me know how you feel about this, happy to discuss it further 👍

richardadjogah:
I totally understand where you're coming from with regards to the assumption being made here about null. I've got no strong feelings on this particular implementation of the fix. A more thorough solution might be to make the `noOptionsMessage` feature opt-in rather than opt-out, but I didn't want to conflict with any of the existing paradigms of the library for such a small change.

With the exception of this issue, my use case doesn't really require manually managing the menu open state and the documentation of the `menuIsOpen` prop is also fairly sparse at the moment. Is it possible to pass it `undefined` to preserve the default logic except in cases where I want to override it, or would have to manually manage it for the entire component lifecycle?

gwyneplaine:
@richardadjogah you absolutely can, here's a sandbox with your implementation in mind, minus the additional information in regards to available results. 
https://codesandbox.io/s/5vzlv540qx

Setting any of the uncontrolled props to `undefined`, will revert the prop value to being controlled by the selects internal stateManager restoring default functionality

JedWatson:
@richardadjogah thanks for the PR. I implemented similar capability with #2754 which I think resolves @gwyneplaine's concerns about clarity.

We looked into whether we think we should change the default behaviour, and think the way it works now is the clearest default behaviour for users - but as of #2754 if the message to display is `null` for either "loading" or "no options", the menu won't be rendered at all.

To me it makes sense that we shouldn't render the menu if there's nothing to display; at the same time, using the result of the function makes it easier to finesse the situations when this occurs (e.g. you could hide the message if all options have been selected, but keep it if the user has typed a search string and there are no matches)

Issue #2688, opened 2018-06-07
[v2] 2.0.0-beta.6 LoadingDot css prop


mike1808:
The build 2.0.0-beta.6 which is published in NPM have issues with `<LoadingDot/>` component which is in `internal/indicators.js`. I don't why but `css` prop was not transpilied by babel-plugin-emotion and was not changed to `className`.

I tried to build the lib myself it and everything was OK. 

Dear maintainers, can you please try to publish one more time with updated node_modules?

gwyneplaine:
@mike1808  thanks for this, I have a resolution for this bug in here 
https://github.com/JedWatson/react-select/pull/2723 

I'll cut this out with the next release.

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue / pull request is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

PR #2723, opened 2018-06-19
[V2] fix for #2688

gwyneplaine:
Changing the namespace using the import { css as Blah } syntax in the indicators file, meant that babel-plugin-emotion was internally transpiling the css prop using the incorrect namespace.

This has now been addressed, we now normalise calls to emotion's css function in all cases where we use it. styling functions that clash with this naming (in Input, Option and Placeholder) have been renamed inputCSS, optionCSS, placeholderCSS respectively.


JedWatson:
Thanks @gwyneplaine

PR #2724, opened 2018-06-19
[v2] Added makeAnimated factory for constructing animated components

gwyneplaine:
added makeAnimated method, animated/index export default is now an empty invocation of makeAnimated


seancurtis:
LGTM

PR #2733, opened 2018-06-20
[v2] fix className composition

nihgwu:
right now if `isDisabled={true}` the container's className would be `prefix__--is-disabled`


SilencerWeb:
but it should be `prefix__name--is-disabled` if name is passed and `prefix--is-disabled` if name isn't passed

nihgwu:
the problem is when `name='--is-disabled'` it becomes `prefix__--is-disabled`

JedWatson:
Good catch @nihgwu, thanks!

Issue #2713, opened 2018-06-15
[v2.0] Missed dash in className modifier


SilencerWeb:
It should be `react-select__control--is-focused` like `react-select__value-container--is-multi`, not `react-select__control-is-focused`

![image](https://user-images.githubusercontent.com/20713191/41469017-40e758d2-70c5-11e8-8537-1bbe94110f71.png)

P.S. pls, notify me when you fix it because i will have to change it in my styles

SilencerWeb:
Same with `react-select__control-is-disabled`, it should be `react-select__control--is-disabled`

![image](https://user-images.githubusercontent.com/20713191/41496086-3a7ad9ac-7150-11e8-860b-52039c237df8.png)

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues / pull requests.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue / pull request is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

PR #2739, opened 2018-06-22
Fix disabled and focused classNames in Control

SilencerWeb:
Fix for #2713 

Due to BEM methodology, modifiers should be with double dash


nihgwu:
isn't a dup of #2733 ?

SilencerWeb:
yep, it is

Issue #2740, opened 2018-06-22
(v2) React-Select -No Accessibility Support when using option groups with searchable=false


Harini03:
No Accessibility options are passed in the input box in react-select when using option group. So this does not provide any default aria-options and also not able to pass aria tags with custom values.

gwyneplaine:
@Harini03 thanks for this, we're working on landing a PR atm that should alleviate this problem. 
You can track it here https://github.com/JedWatson/react-select/pull/2581

chandawarlokesh:
@gwyneplaine I have updated react-select to `2.0.0-beta.7` but still this issue is not fixed VoiceOver on Mac is not reading the menu list options if searchable is set to false and if it is true then it treats react-select as text-box while reading but when I press space-bar its open the menu and read menu item but this behavior is incorrect cause new person/blind person can't justify this is a select box(due to wrong indications) 

Below is the code which I have written please let me know if I am doing anything wrong
```
        const customStyles = {
            indicatorSeparator: () => {}
        }
        return (
            <ReactSelect
                styles={customStyles}
                classNamePrefix="Select"
                className="Select"
                name={name}
                value={selectedValue}
                options={options}
                onChange={onChange}
                isClearable={false}
                aria-labelledby={name}
                maxMenuHeight={200}                
                placeholder={placeholder}/>
        )
```

**TIA**

TanveerAliS:
According to [https://react-select.com/props](Doc) , this can be set by `inputProps={{ 'aria-labelledby': name}}`

ChristopherBeards:
I'm experiencing the same issue. Is there a fix in the works?

ebonow:
Greetings,

`aria-label` and `aria-labelledby` exist as props on the Select at least as of V2. There are other existing and specific accessibility issues so we will be closing this one to address these more current issues.

PR #2741, opened 2018-06-22
Fix for #2740 aria-options for option groups

Harini03:


gwyneplaine:
Hi @Harini03  thanks for raising this, we've opted for using aria-live regions to communicate this information #2581

PR #2752, opened 2018-06-27
WIP V2 hide selected option

gwyneplaine:
PR to allow hideSelectedOption prop functionality for single select. (previously it only affected isMulti). 
Also changes the default value for this prop to false. 

This is a BREAKING CHANGE, as it changes core assumed functionality for isMulti selects. 
We now have to explicitly set hideSelectedOption to true with isMulti to replicate the previously default behaviour. I'm not really sure about this (hence the WIP), as it's adding more configuration to a core use case. Happy to discuss further @JedWatson  @jossmac


JedWatson:
Resolved by #2753 without breaking / behavioural changes

PR #2753, opened 2018-06-27
V2 hide selected default

jossmac:
Alternative to #2752


JedWatson:
Discussed in person with @jossmac, I'm 👍 for this approach

PR #2754, opened 2018-06-27
[V2] Hide menu when messages return null

JedWatson:
Alternative solution to #2719

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -129,7 +129,7 @@ export type Props = {
   /* Whether to enable search functionality */
   isSearchable: boolean,
   /* Async: Text to display when loading options */
-  loadingMessage: ({ inputValue: string }) => string,
+  loadingMessage: ({ inputValue: string }) => (string | null),
   /* Minimum height of the menu before flipping */
   minMenuHeight: number,
   /* Maximum height of the menu before scrolling */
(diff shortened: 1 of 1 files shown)


gwyneplaine:
@JedWatson @jossmac  leaving the thoughts we discussed circa this PR here. 
We have agreed that this is the most pragmatic way to resolve this issue for the time being, however conflating the implicit responsibility of noOptionsMessage and loadOptionsMessage props with the larger concern of rendering the menu seems not ideal for the following reasons: 

- By tying together two unrelated concerns like this to fulfill a niche use case, we may be setting ourselves up for more perilous api entanglement further down the road. Any additional augmentations to menu behaviour (or the ability to customise menu behaviour) will now need to factor this relationship into consideration. 
- We've built a more explicit event handling api and configuration interface to allow users to replicate these sorts of niche relationships, with the information we provide them, a solution that allows for this would be (alongside the benefits of the first point mentioned) contributing to a more consistent api. 
- We may be blocking other niche cases and customization points here that we have no yet encountered. For example users that would like to not show a results message but still show header and footer elements they've made available in the menu. 
- We now have menu rendering behaviour that sits outside of the core `menuIsOpen` prop, that we expose and document as the core value for controlling menu rendering behaviour. 

**Note the above are reasons are not enough to dissuade us from implementing this solution for now, as it is an easy solution to revert and resolves the use case in a backwards compatible manner, but will be a useful reference should we ever decide to come back this PR.

PR #2756, opened 2018-06-28
Fix mobile scroll blocking

### LINKED ISSUE #2726: Scroll on mobile devices does not work with menuShouldBlockScroll as true
Scroll on mobile devices does not work when "menuShouldBlockScroll" is true.

For my use case I need to block scroll outside the menu when menu is open and to work inside the menu.

I attached the select configuration in codesandbox link bellow
https://codesandbox.io/s/olwy71p73q

Version: "2.0.0-beta.6"

jossmac:
Resolves #2726

All relevant notes are inline comments or against the code.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -250,22 +250,6 @@ let instanceId = 1;
 
 export default class Select extends Component<Props, State> {
   static defaultProps = defaultProps;
-  blockOptionHover: boolean = false;
-  clearFocusValueOnUpdate: boolean = false;
-  components: SelectComponents;
-  commonProps: any; // TODO
-  controlRef: ElRef;
-  focusedOptionRef: ?HTMLElement;
-  hasGroups: boolean = false;
-  input: ?ElRef;
-  inputIsHiddenAfterUpdate: ?boolean;
-  instancePrefix: string = '';
-  initialTouchX: number = 0;
-  initialTouchY: number = 0;
-  menuRef: ?ElRef;
-  openAfterFocus: boolean = false;
-  scrollToFocusedOptionOnUpdate: boolean = false;
-  userIsDragging: ?boolean;
   state = {
     focusedOption: null,
     focusedValue: null,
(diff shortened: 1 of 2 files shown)


src/Select.js
```diff
@@ -250,30 +250,54 @@ let instanceId = 1;
...
   userIsDragging: ?boolean;
-  state = {
-    focusedOption: null,
-    focusedValue: null,
-    inputIsHidden: false,
-    isFocused: false,
-    menuOptions: { render: [], focusable: [] },
-    selectValue: [],
+
+  // Refs
```
jossmac: I did a bunch of (somewhat unrelated) shuffling/cleanup on the Select class to make sense of things:

- `state` is now up with `defaultProps`
- named `ref` properties on the class are separated from miscellaneous properties
- the methods used to create the `refs` are now located with their applicable ref

src/Select.js
```diff
@@ -335,35 +359,18 @@ export default class Select extends Component<Props, State> {
 
     // scroll the focused option into view if necessary
     if (
-      this.menuRef &&
+      this.menuListRef &&
```
jossmac: At one point this referenced the menu. It's now a child container that actually scrolls called `MenuList` -- just updated for cleanliness.

src/Select.js
```diff
@@ -384,12 +391,12 @@ export default class Select extends Component<Props, State> {
   // ==============================
 
   focusInput() {
-    if (!this.input) return;
-    this.input.focus();
+    if (!this.inputRef) return;
```
jossmac: All our other refs have the "Ref" suffix. Again, just tidying up a bit.

src/Select.js
```diff
@@ -486,7 +493,11 @@ export default class Select extends Component<Props, State> {
       focusedValue: null,
     });
   }
-  setValue = (newValue: ValueType, action: ActionTypes = 'set-value', option?: OptionType) => {
+  setValue = (
```
jossmac: No change, just prettier in action.

PR #2772, opened 2018-07-04
Adds a Select prop that prevents unselecting previously-selected item

brentg5:
We're using react-select to allow the user to pick multiple people from a list of contacts, and we're setting the `removeSelected` prop to be false so that the user can see which contacts they've already picked both in the input and in the menu. So far, most everything is working well.

However, during user testing we noticed that users found it unintuitive that hitting tab or enter after filtering the list down would actually unselect an option if it was previously selected. 

After getting a couple more eyes on it around the office, we agree that it's a little weird to hit "enter" and then have things disappear, given that enter is generally a positive input rather than a negative. We think that a better interaction is to either do nothing, or to maintain the previously-selected option but to also move it to be the newest item in the input list (as if they had just selected it for the first time).

This pull request exposes a boolean prop, `unselectOnEnter`, which toggles on the latter behavior described above when set to false (it defaults to true to maintain existing behavior by default). Note that this behavior is only invoked when using enter/tab to select a value; clicking on a selected option will still unselect it even if `unselectOnEnter` is set to false.

Please let me know if there's anything else I can do to assist with this pull request. Thanks for all your hard work!

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -482,13 +482,13 @@ class Select extends React.Component {
 					break;
 				}
 				event.preventDefault();
-				this.selectFocusedOption();
+				this.selectFocusedOption(!this.props.unselectOnEnter);
 				break;
 			case 13: // enter
 				event.preventDefault();
 				event.stopPropagation();
 				if (this.state.isOpen) {
-					this.selectFocusedOption();
+					this.selectFocusedOption(!this.props.unselectOnEnter);
 				} else {
 					this.focusNextOption();
 				}
@@ -610,7 +610,7 @@ class Select extends React.Component {
 		}
 	}
 
-	selectValue (value) {
+	selectValue (value, preventUnselecting) {
 		// NOTE: we actually add/set the value in a callback to make sure the
 		// input value is empty to avoid styling issues in Chrome
 		if (this.props.closeOnSelect) {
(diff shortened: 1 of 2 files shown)


jossmac:
Hi @brentg5,

Thanks for the PR -- Unfortunately this isn't something that belongs in Select core.

I've setup a sandbox example of how you could achieve the same result in you source code https://codesandbox.io/s/n4v943j820

brentg5:
Hey @jossmac,

I appreciate you taking a look at this so quickly, and I especially appreciate you taking the time to put together a sandbox example for me.

In the sandbox example you provided, it does indeed prevent enter/tab from removing previously-selected options. However, clicking the `x` on an item in the input section of `Select` doesn't remove the selected option; in fact there doesn't seem to be a way to remove a selected item. We'd definitely want to provide a way for the user to do so.

I think the crux of the issue is that there is no way for the `onChange` callback to know how it was invoked. At that point we don't know if it was the result of hitting enter/tab, a click on an option in the menu, or even a click on a value's `x` button in the input.

I see that we're getting a second parameter to `onChange` in v2, and I'm wondering if we'll be able to distinguish these actions with the new API. I haven't had the chance to work with it yet, but if `remove-value` is the action associated with clicking a value's `x` button, and `deselect-option` is the action associated with hitting enter/tab/clicking on a previously-selected option in the menu... then we should be able to customize Select sufficiently. I'll have a look after the holiday.

jossmac:
Hi @brentg5,

That's correct, the second argument to `onChange` is "action meta", which you will be able to use for this purpose.

```jsx
onChange = (newValue, actionMeta) => void;

// where actionMeta enum is

'clear' | 'create-option' 'deselect-option' | 'pop-value' | 'remove-value' | 'select-option' | 'set-value'
```

You can read the docs here https://deploy-preview-2289--react-select.netlify.com/upgrade-guide#concepts

PR #2779, opened 2018-07-06
Allow ordering of displayed value

Deiru2k:


JedWatson:
Hey @Deiru2k, thanks for the PR.

I'd prefer not to add this to the API, you can fairly easily manage the `value` prop and handle ordering yourself outside of React Select.

Deiru2k:
The problem is that we specifically wanted to avoid doing any unnecessary work on the value prop, because value comes from our state and API and we would love to preserve it as it is. 

If we sort value prop, whenever something is added or removed via select, the whole order of values in that array would be altered, which is an outcome that my team is strongly against.

Deiru2k:
I understand this is a late reply, but I must clarify something.
The problem is that we specifically wanted to avoid doing any unnecessary work on the value prop, because value comes from our state and API and we would love to preserve it as it is.
If we sort value prop, whenever something is added or removed via select, the whole order of values in that array would be altered, which is an outcome that my team is strongly against.
I think one can argue that there could be some mechanism to controlling the presentation while keeping the underlying data intact. Maybe my approach is just not optimal?
I also haven't seen what's in v2 yet, so maybe I can implement my own component now that might handle this.

PR #2809, opened 2018-07-18
Added `closeMenuOnScroll` prop

Vynlar:
Added a new prop that allows the user to have the menu close when they scroll the window. This is useful when you have a portaled menu _inside_ of a scrollable view. Without this, the dropdown often ends up in strange places.

A few notes:
I was unable to get the `npm start` server to run, which means I can't run the cypress tests.
Here's the error I got:
```
defaultProps = convert(p, { ...context, mode: "value" });
SyntaxError: Unexpected token ...
```
It seems like `transform-object-rest-spread` isn't being enabled despite it being inside of `.babelrc`.

I attempted to add unit tests, but since this feature relies heavily on binding listeners to the document object, I was unable to write any that were meaningful. Considering the `touchdown` et. al. handlers are not covered in the unit tests, I figured it would be reasonable for mine to not be as well.

Flow types + linting check out fine.

If someone is willing to help me get my server to start, I'm happy to add some cypress tests that cover the new functionality.

Great job with the new API, migrating from v1 has been wonderful! I've deleted several hundred lines of code that are no longer necessary because of the new styling/component system.

Keep it up!


JedWatson:
This is a really neat addition, thanks @Vynlar

ghfkui:
1. Menu cannot close when set 'closeMenuOnScroll'. Why not exec 'this.onMenuClose'?
2. If 'stopPropagation' in parent element, listener cannot be fired.

Vynlar:
@ghfkui Could you provide more detail as to what the problem is? Ideally some example code that demonstrates the issue.

PR #2806, opened 2018-07-17
[v2] Add labelProps prop to MultiValue, as an analogue to removeProps

xxx:
... to allow adding behaviors to the label as well. This also opens an avenue for users to implement the now-removed `onValueClick` functionality, which I really kinda need.

Behavior and actual props being passed are identical to removeProps.

PR #2821, opened 2018-07-23
[V2] Remove primitives and refactor MultiValue components

JedWatson:
This primarily makes the `MultiValue{X}` components more consistent with the rest of the customisable components' APIs.

`MultiValueContainer`, `MultiValueLabel`, and `MultiValueRemove` all now receive the following props:

* `data` - the option data for the value being displayed
* `innerProps` - as per other components, these are spread onto the dom component
* `selectProps` - passed from the outer Select component so functionality can be extended

This should be sufficient to recreate the `onValueClick` behaviour in v1.

Inspired by #2806


SimonGodefroid:
Any chance to have someone providing an example of how to replicate the onValueClick with the props you have here ?

The example of Github users in V1 was super explicit, any chance to have such a snippet?

Thanks.

TrevorSayre:
@SimonGodefroid 	

```js
const MultiValueLabel = props => {
  return (
    <components.MultiValueLabel
      {...props}
      innerProps={{
        ...props.innerProps,
        onClick: e => {
          console.log(props.data.label);
        }
      }}
    />
  );
};
```

This seems to be what they're thinking, and it works, but it's not clear to me how to prevent opening the menu when clicking the label as `e.preventDefault()` and `e.stopPropagation()` in the above defined `onClick` do not fix this.

SimonGodefroid:
Thanks for sharing @TrevorSayre, in the end a colleague suggested to use Single Value.

Here's a snippet with how I got around, although I do remember having the menu opening as you mention... however I tanked my project so haven't be cracking the right way to do it.

```js
import React, { Component } from 'react';
import AsyncCreatableSelect from 'react-select/async-creatable';

const Option = ({ innerProps, data, selectProps, isSelected }) => {
  return (
    <div {...innerProps}>
      <span>{data.label}</span>
    </div>
  );
};

const SingleValue = ({ innerProps, data, selectProps, isSelected }) => {
  return (
    <div {...innerProps} onClick={this.handleClick}>
      <a href={data.url}>{data.label}</a>
    </div>
  );
};

class Form extends Component {
  render() {
    return (
      <div>
        <AsyncCreatableSelect
          {...props}
          creatable={true}
          isClearable
          components={{ Option, SingleValue }}
          loadOptions={this.getOptionsArtists}
          value={props.input.value}
          onChange={props.input.onChange}
          onBlur={() => props.input.onBlur(props.input.value)}
        />
      </div>
    );
  }
}
```

PR #2824, opened 2018-07-23
V2 redistribute inner ref

gwyneplaine:


JedWatson:
Great, thanks @gwyneplaine

PR #2836, opened 2018-07-23
[v2][website fix] fix the navigation on website pages

ajaymathur:
Website nav was not styles for the items that are clicked as margin-top was not accounted for and all heading tags use margin to space themselves.


JedWatson:
Thanks @ajaymathur!

PR #2838, opened 2018-07-23
[v2] Passing props to MenuPortal's getStyles

almsh:
I'm using app-level context to determine current z-index (select can be opened in different "layers" of app, like modals, popups, etc.). I'm passing current z-index as custom prop to Select and need to have an access to it from menuPortal's `getStyles()`. Furthermore there is an access to `selectProps` from all `getStyles()` functions, except `menuPortal`, so i think that this will made styling api more consistent.

### DIFF
--- a/src/components/Menu.js
+++ b/src/components/Menu.js
@@ -487,7 +487,7 @@ export class MenuPortal extends Component<MenuPortalProps, MenuPortalState> {
     // same wrapper element whether fixed or portalled
     const menuWrapper = (
       <div
-        className={css(getStyles('menuPortal', state))}
+        className={css(getStyles('menuPortal', state, this.props))}
       >
         {children}
       </div>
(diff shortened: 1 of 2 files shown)


jossmac:
Hi @almsh,

Thanks for the PR. The shape for `getStyles` is:

```jsx
(string, Object) => Object
```
Adding a third argument for a single component would be inconsistent. I think you can resolve your issue without an update to core though, using the `styles` API:

```jsx
// assuming <16 context API
const IndexAwareSelect = (props, context) => {
  const zIndex = context.yourDynamicLayerIndex;
  return (
    <Select styles={{
      menuPortal: (provided) => ({ ...provided, zIndex })
    }} />
  );
}
```

---

I'm going to close this issue, however please comment if you need it reopened.

Issue #2590, opened 2018-05-07
[V2] Theme is not customisable


eemeli:
The `v2` branch exports three `const` values from `src/theme.js`: `borderRadius`, `colors` and `spacing`. The latter two are objects, which allows for the theme to be modified externally, but `borderRadius` is just a `number`, so it cannot. Could it also be wrapped in an object to make this possible?

eemeli:
Ah, bother. As the theme isn't actually exported, the theme as a whole is not customisable from an ES6 context, as the package.json `jsnext:main` and `module` values direct packagers to use `dist/react-select.es.js` rather than `lib/index.js`.

So could the theme be exported, and made customisable? As an alternative to clobbering the base theme, could something similar to the `styles` props be implemented for the theme? This would make react-select v2 much easier to customise.

JedWatson:
I think this would make sense.

I'd like to clean up the `theme` export, to make it easy to apply simple re-skins (e.g some colours, spacing, border radius etc.) then we could pass it down to to components in `commonProps`

Then a new `theme` prop on `Select` could support two modes:

* `theme={themeObject}` ("clobber mode")
* `theme={base => ({ ...base, primaryColor: 'pink' })}` (extend mode, like `styles`)

Open to PRs 😄

eemeli:
I'm not sure if it makes sense to have the theme completely replaceable, rather than just overridable. The core code refers to particular theme values in multiple places, and those could very easily be left undefined by complete replacement.

isitrita:
@JedWatson When do you plan make release with theme? We are really waiting for this!

Dandry:
It looks that this feature has already been merged, however I fail to accomplish custom theming. Even though I pass the theme object as presented, the colors do not change. What is more, [`CodeSandbox link`](https://codesandbox.io/s/1oz7n0wok4?module=/example.js) on the  [`doc's subpage`](https://react-select.com/styles#using-classnames) shows custom theming example, but the result also does not get colors changed. Also, it seems that Typescript type definitions are missing for this `theme` prop.

Dandry:
@JedWatson I also need this desperately! Thanks!

giocodes:
How come this works on the docs:
https://react-select.com/styles#overriding-the-theme

But not on the live code or locally? 
https://codesandbox.io/s/1oz7n0wok4?module=/example.js

Are the docs on a later version than 2.0.0 ?

jkeruzec:
[...]

PR #2839, opened 2018-07-23
[v2] Allow configuring the Theme object

akx:
As discussed in #2590, the global theme object is (was) very hard-coded.

This PR makes it a top-level prop that is then propagated down through `CommonProps` (and you can pass in a ready `Theme` object as well as a mutator function – the default theme is available as a `defaultTheme` export):

```jsx
<Select
    theme={(theme) => ({
      ...theme,
      borderRadius: 0,
      colors: {
      ...theme.colors,
        text: 'orangered',
        primary25: 'hotpink',
        primary: 'black',
      },
    })}
/>
```

Please let me know if I've done something silly – this is my first time working with the `react-select` codebase, and actually my first time working with Flow too :)

As an aside, I think the actual keys within `Theme.colors` should be pared down and formalized; right now there's 37 (!) different colors, most of which are various shades of black.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -45,7 +45,9 @@ import {
   type SelectComponents,
   type SelectComponentsConfig,
 } from './components/index';
+
 import { defaultStyles, type StylesConfig } from './styles';
+import { defaultTheme } from './theme';
 
 import type {
   ActionMeta,
@@ -653,6 +655,10 @@ export default class Select extends Component<Props, State> {
   // Getters
   // ==============================
 
+  getTheme() {
+    return defaultTheme;
+  }
+
   getCommonProps() {
     const { clearValue, getStyles, setValue, selectOption, props } = this;
     const { classNamePrefix, isMulti, isRtl, options } = props;
@@ -674,6 +680,7 @@ export default class Select extends Component<Props, State> {
       selectOption,
       setValue,
       selectProps: props,
+      theme: this.getTheme(),
     };
   }
 
@@ -1318,7 +1325,7 @@ export default class Select extends Component<Props, State> {
       'aria-labelledby': this.props['aria-labelledby'],
     };
 
-    const { cx } = this.commonProps;
+    const { cx, theme } = this.commonProps;
 
     return (
       <Input
@@ -1336,6 +1343,7 @@ export default class Select extends Component<Props, State> {
         onFocus={this.onInputFocus}
         spellCheck="false"
         tabIndex={tabIndex}
+        theme={theme}
         type="text"
         value={inputValue}
         {...ariaAttributes}
--- a/src/Select.js
+++ b/src/Select.js
@@ -47,7 +47,7 @@ import {
 } from './components/index';
 
 import { defaultStyles, type StylesConfig } from './styles';
-import { defaultTheme } from './theme';
+import { defaultTheme, type ThemeConfig } from './theme';
 
 import type {
   ActionMeta,
@@ -213,6 +213,8 @@ export type Props = {
   screenReaderStatus: ({ count: number }) => string,
   /* Style modifier methods */
   styles: StylesConfig,
+  /* Theme modifier method */
+  theme?: ThemeConfig,
   /* Sets the tabIndex attribute on the input */
   tabIndex: string,
   /* Select the currently focused option when the user presses tab */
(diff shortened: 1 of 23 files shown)


src/Select.js
```diff
@@ -653,6 +657,16 @@ export default class Select extends Component<Props, State> {
   // Getters
   // ==============================
 
+  getTheme() {
+    if (this.props.theme) {
+      if (typeof this.props.theme === 'function') {
+        return this.props.theme(defaultTheme);
+      }
+      return this.props.theme;
```
eemeli: Returning instead `Object.assign({}, defaultTheme, this.props.theme)` here would guarantee that all of the default theme values are included, rather than trusting and requiring that a custom theme takes care of them. That would simplify the requirements on a custom theme to just define the changed variables, and would help with backwards-compatibility when the default theme will eventually change.

akx: Do you mean also removing the function form of the prop? I'm okay with that too, but it'd mean people have to import the default theme.

eemeli: No, I meant just replacing line 665 with the `Object.assign` statement. The function form makes sense as a route for "advanced" theming, and with it the default theme itself wouldn't even need to be exported from the package.

akx: Gotcha. Fixed :)

akx:
Hey @JedWatson, sorry to prod, but would it be possible to get some (more) eyes on this? :)

src/components/Menu.js
```diff
@@ -49,7 +53,8 @@ export function getMenuPlacement({
   placement,
   shouldScroll,
   isFixedPosition,
-}: PlacementArgs): MenuState {
+  theme,
+}: PlacementArgs) {
```
jossmac: @akx any reason you removed the return type?

src/components/Menu.js
```diff
@@ -242,6 +249,7 @@ export class Menu extends Component<MenuProps, MenuState> {
   state = {
     maxHeight: this.props.maxMenuHeight,
     placement: null,
+    theme: null,  // TODO: is this okay
```
jossmac: @akx I think I'm missing something; why would the `theme` be in `state`?

Also, what's the TODO in reference to?

jossmac:
> As an aside, I think the actual keys within Theme.colors should be pared down and formalized; right now there's 37 (!) different colors, most of which are various shades of black.

Totally agree, though it would be a breaking change once this is merged. I'll see if I can tidy it up a bit if you're willing to rebase? Shouldn't be too messy

akx:
@jossmac Check, will do tomorrow! :)

src/components/Menu.js
akx: No idea! Restored.

src/components/Menu.js
akx: Yeah, no idea either... Wonder what I'd been thinking there.

akx:
Rebased, cleaned and ready to roll!

akre54:
I just came here looking for this! Excited to see it get merged. Great work @akx

src/components/Group.js
```diff
@@ -2,20 +2,21 @@
...
-import { spacing } from '../theme';
 import type { CommonProps } from '../types';
 
 type ComponentProps = {
   /** The children to be rendered. */
   children: Node,
   /** Component to wrap the label, recieves headingProps. */
   Heading: ComponentType<any>,
+  /** Props to pass to Heading. */
+  headingProps: any,
```
jossmac: This is new. @akx What's the plan here?

akx: It was missing, but referenced in the comment for `Heading` above:

```javascript
/** Component to wrap the label, recieves headingProps. */
```

sahibjotsaggu:
@akx For some reason, and I can't figure out why, the theme customization doesn't work when the same code is brought into CodeSandbox: https://codesandbox.io/s/1q32j2p5m4

Can you please shed some light on this?

Thanks.

jossmac:
@sahibjotsaggu we need @JedWatson to release a new version to npm.

sahibjotsaggu:
@jossmac Is there any way I can get it to work without a new version being released to npm? I need this ASAP.

jlcasto:
is there an expected release date?

jlcasto:
Thanks for the release of this!  If you have a dark theme, the options' text color is still black.  I think perhaps this might need to be a new bug to style them with one of the styles from the theme that will provide appropriate contrast?

Ecksters:
I believe this change may have created a new error that appears in console for anyone that was using 2.0.0 and was upgraded automatically:
```
Warning: Unknown prop `theme` on <input> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop
    in input (created by AutosizeInput)
    in div (created by AutosizeInput)
    in AutosizeInput (created by Input)
    in div (created by Input)
    in Input (created by Select)
    in div (created by ValueContainer)
    in ValueContainer (created by Select)
    in div (created by Control)
    in Control (created by Select)
    in div (created by SelectContainer)
    in SelectContainer (created by Select)
    in Select (created by StateManager)
```

Doesn't seem to break anything, but the error is annoying.

src/components/Group.js
Ecksters: This needs to be updated to pull theme out before passing the props on to the actual HTML input element.

PR #2842, opened 2018-07-23
Fixed rollup warnings about unresolved dependencies

Andarist:


mitchellhamilton:
Thanks for this @Andarist!! We switched to preconstruct here a bit ago so this isn't relevant anymore

PR #2843, opened 2018-07-23
Centralize babelrc config & use loose mode

Andarist:
1. do not duplicate babel config in rollup.config.js, should be easier to keep this in sync now 

2. loose mode - mainly bundle size optiomation. Measuring with `npx terser -cm --toplevel dist/react-select.esm.js | gzip -c | wc -c` from 19047 bytes to 18545 bytes. Should u chose loose mode? Thread - https://twitter.com/brian_d_vaughn/status/1021183132135931905

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,14 +1,3 @@
 {
-  "plugins": [
-    "emotion",
-    "transform-class-properties",
-    "transform-object-rest-spread"
-  ],
-  "presets": ["env", "react"],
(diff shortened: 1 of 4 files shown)


mitchellhamilton:
Thanks for this!! As mentioned in the other PR, react-select switched to preconstruct a while ago so this mostly isn't relevant anymore. I just opened a new PR, #3797, to enable loose mode because the Babel config has moved to babel.config.js

PR #2845, opened 2018-07-24
[v2] Add example for fixed options in multi select

daggmano:
Adds example for fixed (preset) options in multiselect list (#2675).

### DIFF
--- a/docs/examples/FixedOptions.js
+++ b/docs/examples/FixedOptions.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 
-import Select, { components } from '../../src';
+import Select from '../../src';
 import { colourOptions } from '../data';
 
 type State = {
--- a/docs/examples/FixedOptions.js
+++ b/docs/examples/FixedOptions.js
@@ -1,0 +1,75 @@
+import React, { Component } from 'react';
+
+import Select, { components } from '../../src';
+import { colourOptions } from '../data';
+
+type State = {
+  value: [{ [string]: string }]
+};
+
+const TagStyle = {
+  multiValue: (base, state) => {
+    return state.data.isFixed ? { ...base, backgroundColor: 'gray' } : base;
+  },
+  multiValueLabel: (base, state) => {
+    return state.data.isFixed ? { ...base, fontWeight: 'bold', color: 'white', paddingRight: 6 } : base;
+  },
+  multiValueRemove: (base, state) => {
+    return state.data.isFixed ? { ...base, display: 'none' } : base
+  }
+};
+
+const ClearIndicator = (props) => {
+  let value = props.getValue();
+  let isFixed = value && value.reduce((p, v) => p && v.isFixed, true);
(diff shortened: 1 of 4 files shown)


daggmano:
@JedWatson Thanks for the guidance in this, let me know if I can refine this in any way.

docs/examples/FixedOptions.js
```diff
@@ -0,0 +1,75 @@
...
+  },
+  multiValueLabel: (base, state) => {
+    return state.data.isFixed ? { ...base, fontWeight: 'bold', color: 'white', paddingRight: 6 } : base;
+  },
+  multiValueRemove: (base, state) => {
+    return state.data.isFixed ? { ...base, display: 'none' } : base;
+  }
+};
+
+const ClearIndicator = (props) => {
```
jossmac: Rather than providing a conditional component, could you move this logic to `render()` and use the `isClearable` property?

docs/examples/FixedOptions.js
```diff
@@ -0,0 +1,75 @@
...
+  let value = props.getValue();
+  let isFixed = value && value.reduce((p, v) => p && v.isFixed, true);
+  return isFixed ? (<span/>) : (<components.ClearIndicator {...props}/>);
+};
+
+const OrderOptions = (values) => {
```
jossmac: This could be confused as a component, would you mind lowercasing the first letter?

docs/examples/FixedOptions.js
```diff
@@ -0,0 +1,75 @@
+import React, { Component } from 'react';
+
+import Select, { components } from '../../src';
+import { colourOptions } from '../data';
+
+type State = {
+  value: [{ [string]: string }]
+};
+
+const TagStyle = {
```
jossmac: Could this just be `styles` or similar? Pascal case is typically reserved for classes/types within JS.

daggmano:
Thanks for the feedback, always happy to improve!

docs/examples/FixedOptions.js
```diff
@@ -0,0 +1,69 @@
...
+    this.setState({ value: value });
+  }
+
+  render () {
+    return (
+      <Select
+        value={this.state.value}
+        isMulti
+        styles={styles}
+        isClearable={!this.state.value.reduce((p, v) => p & v.isFixed, true)}
```
jossmac: Just noticed this clever use of `reduce()`!

You might be interested in [`some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some), which is arguably better suited to this case:

```jsx
<Select
  ...
  isClearable={this.state.value.some(v => !v.isFixed)}
/>
```

jossmac:
Oh no, wrong branch!

@daggmano could you please open another PR against `master`?

daggmano:
Yep, no problems at all. Done as PR#2941, waiting for tests to run. Also swapped out `reduce` for `some` - didn't even know that existed!

PR #2847, opened 2018-07-24
Fix syntax error in styles docs

jmz7v:
### Problem
An example object for `customStyles` on `docs/styles` is missing a comma, it causes a syntax error.

### Solution
Add missing comma.

### References

Link to example object on the docs website: https://react-select.com/styles#base-and-state

Screenshot of failing example
image


JedWatson:
Thanks for picking that up!

PR #2857, opened 2018-07-26
Fixed bug where portal-ed menu was positioned incorrectly

Vynlar:
We were using the `window.pageYOffset` always to determine scroll distance; however, this caused the menu to be placed in the wrong location on screen if the scrolling container was not the body. Changed the code so it finds the nearest scrollable parent and then uses the `scrollTop` of that (it'll default to the body to maintain existing behavior).

Essentially if your `menuPortalTarget` is positioned in the top left corner of the nearest scrollable element, then the menu should be placed in the correct place now. This helps ease the struggles of putting the dropdown inside of modals and other scrollable areas.

### DIFF
--- a/src/components/Menu.js
+++ b/src/components/Menu.js
@@ -424,6 +424,7 @@ export type MenuPortalProps = CommonProps & {
   controlElement: HTMLElement,
   menuPlacement: MenuPlacement,
   menuPosition: MenuPosition,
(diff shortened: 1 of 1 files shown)


Vynlar:
With more complex situations, my changes create new unexpected behavior in some situations. I'm going to close this pull request and make a new one after making these changes optional.

PR #2861, opened 2018-07-27
[v2] add listener to ScrollCaptor, even when no scroll is available on mount

pyreta:
## Current Behavior
The `onMenuScrollToBottom` callback will only fire if the dropdown menu is scrolling on mount.

## Proposed Changes
The `onMenuScrollToBottom` callback always fires, regardless of whether or not the dropdown menu was scrollable on mount.

## Why Change?
A dropdown menu may render more options after mounting and therefore event listeners may need to be listening for scroll events.

## Use case
In a project I'm working on, I'm creating a dropdown supporting pagination that loads asynchronously.  When I mount my `Select` component, my initial state is empty.  Once the user starts typing, an async request is made loading my state with options, and showing them in the dropdown.  However since the dropdown mounted with no options to start with, the `ScrollCaptor` is not listening for scroll events.  This prevents me from being able to load the next page on bottom scroll after loading options asynchronously.

## CodeSandbox Example
You can see an [example here](https://codesandbox.io/s/1vq0l1p2l4) where opening the dropdown with options already loaded allows for the `onMenuScrollToBottom` callback to log "bottom" in the console.  However if you clear the options before opening the dropdown, the callback will not fire after options have loaded.

### DIFF
--- a/packages/react-select/src/internal/ScrollCaptor.js
+++ b/packages/react-select/src/internal/ScrollCaptor.js
@@ -25,9 +25,8 @@ class ScrollCaptor extends Component<CaptorProps> {
     this.stopListening(this.scrollTarget);
   }
   startListening(el: HTMLElement) {
-    // bail early if no scroll available
+    // bail early if no element is available to attach to
     if (!el) return;
-    if (el.scrollHeight <= el.clientHeight) return;
 
     // all the if statements are to appease Flow 😢
     if (typeof el.addEventListener === 'function') {
@@ -41,8 +40,6 @@ class ScrollCaptor extends Component<CaptorProps> {
     }
   }
   stopListening(el: HTMLElement) {
-    // bail early if no scroll available
-    if (el.scrollHeight <= el.clientHeight) return;
 
     // all the if statements are to appease Flow 😢
     if (typeof el.removeEventListener === 'function') {


jrhalchak:
👍 Got a vote from me.  I'm currently forcing the height to be smaller than I know most results will be.

Brunomm:
+1

chuvikovd:
+1

dpetranek:
+1

jossmac:
@pyreta this is a good idea, though I worry about binding a listener for every menu; especially when many won't have need for it.

What about using a `ResizeObserver` to watch for changes?

Kemosabert:
Hey @jossmac, is there anything we can help with to get this issue resolved? Would investigating the `ResizeObserver` option help here?

jossmac:
@pyreta @Kemosabert if you can resolve the conflicts I'll merge it. DW about the rest.

Kemosabert:
@jossmac Thanks! I created a new PR for this: #3752 because I don't know if @pyreta saw this message. When the PR is merged into master, how long approximately before this fix is available in a version on NPM?

pyreta:
@jossmac  @Kemosabert conflicts have been resolved.

packages/react-select/src/internal/ScrollCaptor.js
```diff
@@ -25,10 +25,6 @@ class ScrollCaptor extends Component<CaptorProps> {
     this.stopListening(this.scrollTarget);
   }
   startListening(el: HTMLElement) {
-    // bail early if no scroll available
-    if (!el) return;
```
Kemosabert: @pyreta I think this check could/should stay in here. Otherwise you'd undo #3419 I'm afraid :)

pyreta: ok fixed!

jossmac:
> how long approximately before this fix is available in a version on NPM?

Not sure, sorry mate. I don't have publishing permission.

Dalitzky:
Is there anyone else experiencing issues with the scroll not working after this merge ?
For me this merge caused an issue when using react-select on top of modal window with menuPortaltarget && menuShouldBlockScroll.

romanfazulianov:
Hi everyone!
I faced vertical scroll lock after update to version 3.0.5 and higher. If I create fully dynamic menu list component, that will be recreated on every render, scroll works. However, there is another issue instead - option hover scrolls menu list to top. 

rollback to 3.0.4 fixes any problems with scroll, but causes warnings about old lifecycle methods

Any thoughts?

PR #2871, opened 2018-07-31
Add name to action meta to allow for more DRY code

SimeonC:
This can be useful to know when you have multiple selects on the same page. It allows us to use a pattern similar to the suggested [multiple input pattern from react docs](https://reactjs.org/docs/forms.html#handling-multiple-inputs).

```jsx
onChange = (newValue, actionMeta) => {
  this.setState({
    [actionMeta.name]: newValue
  });
}

// render
<ReactSelect name="select1" onChange={this.onChange} value={this.state.select1} ... />
<ReactSelect name="select2" onChange={this.onChange} value={this.state.select2} ... />
```

(Assumed I don't need to tag this as v2 anymore as it's been released)

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -561,12 +561,16 @@ export default class Select extends Component<Props, State> {
       focusedValue: null,
     });
   }
+  onChange = (newValue: ValueType, actionMeta: ActionMeta) => {
+    const { onChange, name } = this.props;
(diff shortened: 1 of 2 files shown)


SimeonC:
@JedWatson Any chance of a review?

gwyneplaine:
This is great, thanks @SimeonC, merged

PR #2879, opened 2018-08-01
tweak rendering to display a small coloured box next to each option

jonfreedman:
I am curious if this is the expected way to modify the rendering of a custom shape option?  I came up with this whilst working on the typescript typedef for the library.

You can see the changes in a sandbox [here](https://codesandbox.io/s/kx7oy7j1p3)

### DIFF
--- a/docs/examples/CustomSingleValue.js
+++ b/docs/examples/CustomSingleValue.js
@@ -1,18 +1,24 @@
-import React, { Component } from 'react';
+import React, { PureComponent } from 'react';
 import Select, { components } from '../../src';
 import { colourOptions } from '../data';
 
 
-const SingleValue = ({ children, ...props }) => (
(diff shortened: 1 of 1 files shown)


jossmac:
@jonfreedman that's correct, though the sandbox is broken.

There's also an alternative approach. Let's say for instance, that you want a uniform aesthetic within the option and value, you could take advantage of the `formatOptionLabel` property:

```jsx
<Select
  formatOptionLabel={({ color, label }) => (
    <div style={{ display: 'flex', alignItems: 'center' }}>
      <div style={{ background: color, height: 4, width: 4, marginRight: 4 }} />
      {label}
    </div>
  )}
/>
```

Checkout the [CodeSandbox](https://codesandbox.io/s/014q6zwp8l)

Issue #2870, opened 2018-07-30
[V2] Add function to common props that closes the menu


Vynlar:
I'd like to add a prop to the set of common props available to all custom components that would allow the developer to close the menu.

Use case:
* Developer adds a 'Close' button to their custom `Menu` component that should close the menu.
* A custom component needs to close the menu for any other reason.

I'm willing to submit the PR if this idea jives with the maintainers.

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our limited efforts to maintain the latest version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue  is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

PR #2881, opened 2018-08-01
Added prop that allows custom Menu's to close themselves

Vynlar:
See #2870

I needed to be able to close the menu from within a custom `Menu` component, so I added a prop `closeMenu` that does this.


jossmac:
@Vynlar you can control the state of the menu using the `menuIsOpen` property.

Please see [Controlled Props](http://react-select.netlify.com/advanced#controlled-props) in the docs for more information.

PR #2897, opened 2018-08-06
Fix v1 readme links

giraffesyo:
I updated the v1 docs to accurately point to v1 materials, and updated the info about v2 now that it is released.


jossmac:
This is great, thanks @giraffesyo 🙂

PR #2899, opened 2018-08-06
[v2] Fixed bug where options were not reloaded when cacheOptions is false

Vynlar:
## The Problem

With Async, if `defaultOptions` was set to `true` and `cacheOptions` set to `false`. I would expect the default options to be reloaded every time that the user cleared out the input field; however, as it stands, the options are loaded once on mount and then cached, regardless of the value of `cacheOptions`.

## The Solution

In this bugfix, I made the value of `cacheOptions` affect the behavior of `defaultOptions={true}` meaning the results for `loadOptions('')` are found each time they are needed instead of once at the start.

## The Use Case

This caused a bug in my application where newly-created options (those added on the server were not being populated in the default options, but did show up in search results).

Comments and concerns welcome.

Cheers!

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -44,12 +44,11 @@ export const makeAsyncSelect = (SelectComponent: ComponentType<*>) =>
     constructor(props: Props) {
       super();
       this.state = {
-        defaultOptions: Array.isArray(props.defaultOptions)
+        loadedOptions: Array.isArray(props.defaultOptions)
           ? props.defaultOptions
-          : undefined,
+          : [],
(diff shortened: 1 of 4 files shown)


bladey:
Hi @Vynlar,

Thanks for this pull request.

In an effort to sustain the `react-select` project going forward, we're cleaning up and closing old and unmergeable pull requests.

However - if you feel this pull request is still relevant and you'd like us to review it, we'd appreciate if you updated this pull request use the latest version (`master`) and request us to review it.

Issue #2901, opened 2018-08-07
[V2] Feature request: V2 more flexibility when supplying `value`


dan-weaver:
# Summary
currently the value supplied to Select must be of type Option or Array of Option.  This is fine in all the use cases seen in the examples but it would be convenient to be able to supply a single value or an array of *just* values and not Options.  for example `["value1", "value2"]` instead of `[{value: "value1", label: "value 1"},{ value: "value2", label: "value2" }]`.  The `Select` component has enough information, assuming that `value` is always a unique value to map it to the right option in the list.  It's not always convenient to do that mapping in "user space"

# Use case
when values are derived from means such as query parameters in a url this would be really nice.  example: `?someThing=value1`.  We don't have access to the label there and users of `Select` would have to manually hydrate `value1` into an object that includes the label.

# Propsal
Expand the type for `value` such that a `string | number` or array of `string | number` are allowed.  When something that doesn't fit `Option` is passed, a lookup could occur which hydrates the option into a proper type of {value: any, label: string} based on the values in `options`.

# Example of proposed API
## Non multi
```javascript
<Select options={[{value: "foo", label: "Foo"}]} value={"foo"} />

```
## Multi
```javascript
<Select options={[{value: "foo", label: "Foo"}]} value={["foo"]} />
```

The normal behavior of passing something of type Option would still be respected.
Thoughts? Happy to submit a PR if this makes sense. :)

## Edit:
Maybe the isOptionSelected prop can satisfy this use case enough? But that still requires more "user space" intervention when it seems the component has enough information to derive the selected option.

holmrenser:
I was actually using v1 like this, which worked fine. After upgrading to v2 my components stopped working, and I now have to provide Objects or Arrays of Objects.

jossmac:
This keeps coming up. Part of the problem is documentation, for sure, but if you still really want to use simple value I've created a wrapper. You can get it [from npm](https://www.npmjs.com/package/react-select-simple-value) or take a look at [the source](https://github.com/jossmac/react-select-simple-value/blob/master/index.js) and grab whatever you want.

#### Usage

```jsx
<SimpleValue options={[...]} value="blue">
  {props => <Select {...props} />}
</SimpleValue>
```

#### Supports

- simple `value`
- simple `defaultValue`
- grouped options
- custom `getOptionValue`
- `isMulti` where value/defaultValue is a comma delimited string

---

#### Trivial Alternative

```jsx
[...]

PR #2908, opened 2018-08-08
Allow passing of primitive values as value prop

dan-weaver:
This is very much a first pass I just want to know if this is a feature that's worthwhile and if the approach seems right.

- [ ] add flow types
- [ ] verify examples still work
- [ ] add more test cases
- [ ] squash commits

addresses #2901

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -25,6 +25,7 @@ import {
 import {
   classNames,
   cleanValue,
+  hydrateValues,
   isTouchCapable,
   isMobileDevice,
   noop,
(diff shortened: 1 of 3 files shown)


jossmac:
Hi @dan-weaver,

The decision to limit `value` to an option object makes many things easier for maintainers and consumers. Unfortunately, just not this one case.

I'd like some more input on this, but I think it's not too much to ask consumers to write a helper function to derive the option object from a given key:

```js
const valueFromId = (opts, id) => opts.find(o => o.value === id);
```

Perhaps @JedWatson or @gwyneplaine can weigh in?

jossmac:
Hi @dan-weaver,

Thanks again for the PR. We've discussed and would really rather not to allow "simple value" into core. That being said the preferred pattern would be components that augment react-select.

To get the ball rolling I've created [react-select-simple-value](https://www.npmjs.com/package/react-select-simple-value), a function that receives an object (props) and calls the consumer's `children` function with an augmented version of that object.

```jsx
<SimpleValue options={[...]} value="blue">
  {props => <ReactSelect {...props} />}
</SimpleValue>
```

Issue #1879, opened 2017-07-18
react-select loadOptions question


thisthatother:
In my application, I have an onChangeHandler in the first react-select that makes an AJAX call to get the data for populating a second react-select on the page. It needs to populate the second one automatically, with no typing required to trigger the loadOptions. I have not seen any examples that do it like this-- i.e. in the example here, it seems to require typing into the second select to trigger the load. 

Is an automatic load like I describe even possible?

phamthaibaoduy:
AFAIK, loadOptions can not be trigger manually outside of Async component :(

kinyat:
One of the workaround is using `key` props of a component.

When you want to tigger loadOptions again, give it a different key so it will re-render the whole react-select component.

I hope we can have a proper API to trigger this outside of Async component

oldmill1:
I think @phamthaibaoduy is correct, and loadOptions can not be trigger manually outside of Async component. However, you can always just use Select with `onInputChange` and get the desired result you want: both "type to search" and dynamic loading. 

```
render() {
    <Select
        onInputChange={this.onInputChange.bind(this)}
        value={this.state.value}
        options={this.state.options || []}
    />
}
```

So every time the user types in Select, we call `onInputChange` and pass along the string that the user typed. We then call our api function which returns a Promise. If the Promise resolves successfully, we set this.state.options to be what was returned from the API.

```
onInputChange(option) {
    this.getOptionsAsync(option)
}

// expect results to be in [{value: '', label: ''},...] format
getOptionsAsync(newInput) {
    get({ input: newInput }).then(({results, error}) => {
            this.setState({
                options: results ? results: [],
                value: newInput
            })
        })
        .catch((error) => {
            console.error(error)
        })
    }
```

alexcroox:
@kinyat the key trick was genius. After struggling for an hour to get 1 select to change options in another select (and have the updated options available on first click), your key suggestion worked first time. Thank you!!

aliasgar-choolawala:
@alexcroox Could you please help me out? I am facing an issue similar to urs...
<Select.Async
                    key={this.state.specialitySelected}
                    multi={this.state.multi}
                    value={this.state.value}
                    onChange={this.onChange}
                    valueKey="value"
                    labelKey="label"
                    loadOptions={this.getDoctors}
                    backspaceRemoves={this.state.backspaceRemoves}
                    onValueClick={this.getDoctorDetails}
                    cache={false}
                    ref="listing"
/>
[...]

PR #2915, opened 2018-08-09
Reload defaultOptions when cacheOptions changes in Async

wouterhund:
Previously the default options did not reload when the cache changes, leading to incorrect results.

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -57,17 +57,16 @@ export const makeAsyncSelect = (SelectComponent: ComponentType<*>) =>
       this.mounted = true;
       const { defaultOptions } = this.props;
       if (defaultOptions === true) {
-        this.loadOptions('', options => {
-          if (!this.mounted) return;
-          const isLoading = !!this.lastRequest;
-          this.setState({ defaultOptions: options || [], isLoading });
-        });
+        this.reloadDefaultOptions();
       }
     }
     componentWillReceiveProps(nextProps: Props) {
       // if the cacheOptions prop changes, clear the cache
       if (nextProps.cacheOptions !== this.props.cacheOptions) {
         this.optionsCache = {};
+        if (defaultOptions === true) {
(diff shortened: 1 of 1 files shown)


Guichaguri:
I'm also experiencing this issue in three async selects:
* Country
* State: depends on the country
* City: depends on the state and country

It's working as expected, but only when using the search. The defaultOptions are not being reloaded properly.

bladey:
Hi @wouterhund,

Thanks for this pull request.

oowowaee:
@bladey Seeing this in the latest version as well.  My use case was the options of the select are determined based on the value of another select, which I am passing in as the cache key.  I was expecting that since the cache key had changed, it would refetch the defaultOptions, but that is not the case.  As @Guichaguri mentioned above, it works correctly when typing, as the cache has been cleared, but it looks like the request for loadOptions('') is not reissued when the cache key changes.  Will have to investigate managing the defaultOptions manually.

ebonow:
Greetings @oowowaee ,

Since AsyncSelect will do a fetch on each key down, defaultOptions is meant to provide a first time experience when the input is first focused. After that, `loadOptions` can be used for input (including inputValue === '')

`defaultOptions` can be several things...

First and foremost, a boolean will cause the AsyncSelect to call loadOptions with an empty string for inputValue if truthy. 

However, defaultOptions can also be a list of options. Therefore you could create a function which would return a list of options based on some criteria. Since this is used by the initial state, you may want to use this same type of logic in your loadOptions call when the inputValue === ''.

it's worth noting that "default" in most contexts will only refer to the initial value of something on mount similar to setting default state value in a class component in the constructor.

oowowaee:
@ebonow Thanks!  I found https://github.com/JedWatson/react-select/issues/1879 after commenting, and look forward to when that functionality is added.  For the moment the key workaround solves the issue!

PR #2941, opened 2018-08-16
V2 fixed options example

daggmano:


gwyneplaine:
This is great, thanks for this @daggmano

Issue #2945, opened 2018-08-17
CreatableSelect doesn't create new options if used with filterOption prop


Anuj16:
Hello All,

Thanks a lot for this amazing library.

I am having an issue while using "CreatableSelect" with **filterOption** prop, which has 'matchFrom' value as 'start'. The issue is that I am not able to create new options.

My Component usage looks like the following:-

`const filterConfig = {
    matchFrom: 'start'
 };
<CreatableSelect
                isMulti
                isClearable={false}
                openMenuOnClick={false}
                value={value}
                options={options}
                filterOption={createFilter(filterConfig)}
                className={className}
                classNamePrefix={classNamePrefix}
                onChange={onChange} />` 

If I remove the **filterOption** prop, it works absolutely fine. Am I missing something here? Any help is greatly appreciated.

Thanks in advance.

chrysb:
I have encountered this issue as well — it's not clear whether it's a bug or working as intended.

I also tried to force `isValidNewOption={() => {return true}}` but that didn't work.

Did you ever find a way around this, Anuj?

Rall3n:
You have to wrap the function created by `createFilter` into another function which checks if the option is the `Create ...` option.

```jsx

const customFilter = createFilter({ matchFrom: 'start' });

/* ... */

<CreatableSelect
    { ... }
    filterOptions={(option, rawInput) => {
        // Data says its new so we return true.
        if (option.data.__isNew__) return true;
        
        // Filter using the filter function
        return customFilter(option, rawInput);
    }}
/>
```
[CodeSandbox](https://codesandbox.io/embed/505ppxw7jx)

Anuj16:
@Rall3n Yes you are right. I also did exact same way.

Rall3n:
@Anuj16 Please don´t forget to close this issue if the problem you mentioned here is solved.

Anuj16:
@Rall3n I had raised a pull request to fix this in repository properly instead of having a workaround. That's why kept this issue open.

PR #2946, opened 2018-08-17
Create new options with CreatableSelect and filterOption

Anuj16:
Solves #2945 

Issue:-
- While using CreatableSelect, If **filterOption** prop is provided and that has 'matchFrom' property as 'start'.
- In that case, we will not be able to create new options, because the **createFilter** function will search in the label from start and that label will have text as "Create New "xxx"" which doesn't start with xxx, hence will **return false** and the option will not be displayed.

Fix:-
- The fix is very simple.
- If __isNew__ flag is true, ignore all the filter config and return true.

### DIFF
--- a/src/filters.js
+++ b/src/filters.js
@@ -17,6 +17,9 @@ export const createFilter = (config: ?Config) => (
   option: { label: string, value: string, data: any },
   rawInput: string
 ) => {
+  if (option.data.__isNew__) {
+    return true;
+  }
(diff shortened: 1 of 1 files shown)


byjoh:
+1... would very much like for this to be merged @JedWatson.  Thanks!

Anuj16:
Hello guys....I was wondering if there is any plan to merge this pull request. @JedWatson. Thanks!

bladey:
Hi @Anuj16,

Thanks for this pull request.

PR #2951, opened 2018-08-20
filterOptions -> filterOption

arty-name:
The "Advanced" section of the docs still refers to filterOptions and uses a wrong url


gwyneplaine:
thanks for this @arty-name

Issue #2953, opened 2018-08-20
[V2] Creatable throws with custom value / label fields


akre54:
## Description: 
Go to https://codesandbox.io/s/qj18wmroj and start typing in the select box. Throws `TypeError: Cannot read property 'toLowerCase' of undefined`.

## Cause: 
Creatable.js's [https://github.com/JedWatson/react-select/blob/v2.0.0-beta.7/src/Creatable.js#L41-L42](`compareOption`) function incorrectly assumes the `options` object has `label` and `value` properties.

## Possible Fix
`compareOption` should use `getOptionLabel` / `getOptionValue` before referencing the properties.

akre54:
Ping @JedWatson

webzfactory:
compareOption() also falsely assumes that option.value is a string....

plxel:
Ping again @JedWatson

mdzakir:
I had this issue. 

> I found an issue with compareOption function when changes getOptionValue and getOptionLabel to a different (value|label)

**Reason why it is throwing this error:** compareOption function is trying to first `toLowerCase()` the option value and label and then compare it with the characters (string) we are typing. So if you are passing anyone of them (value and label) as number or other than string, it will throw an error.

### Here is what I did & it worked for me.
### And hope it works for everyone here as well. 😄

**1. Mapped the array before hand:**
```
let states = statesArray.map(m => {
      return {
        value: m.id.toString(),
        label: m.stateName
      }
    });
```

2. Passed the mapped **states** to `Creatable` Component
```
<Creatable
            className="basic-single"
            classNamePrefix="select"
            name="states"
            options={states}
            isMulti
          />
```

koutsenko:
Problem is that overrided Option component may contain complex layout and display more that one string value, so 'object'-like label support is necessary.

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our efforts towards the current major version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

JedWatson:
For completeness, the props `formatCreateLabel`, `isValidNewOption` and `getNewOptionData` props were introduced to `Creatable` to fix this and other related issues - [docs here](https://react-select.com/props#creatable-props)

[...]

PR #2956, opened 2018-08-20
Fix creatable not supporting Int values and value accessors #2955 #2953

### LINKED ISSUE #2955: [V2] Creatable crashes with values without `toLowerCase` in prototype
This issue is very similar to https://github.com/JedWatson/react-select/issues/2953 and the same lines are affected here.

When using Creatable you're limited to only using values who's type's prototype has the toLowerCase method which causes the entire component/application to crash if you're using int's for values which i can only imagine im not the only one doing this.

Example: https://codesandbox.io/s/18w2ymppql

Reproduce steps:
1. visit example
2. open dropdown
3. type anything
4. watch the fireworks

th3fallen:
Resolves #2955 and #2953

### DIFF
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -10,6 +10,10 @@ import Select, { type Props as SelectProps } from './Select';
 import type { OptionType, OptionsType, ValueType, ActionMeta } from './types';
 import { cleanValue } from './utils';
 import manageState from './stateManager';
+import {
+  getOptionLabel,
+  getOptionValue,
+} from './builtins';
 
 export type CreatableProps = {
   /* Allow options to be created while the `isLoading` prop is true. Useful to
@@ -36,10 +40,10 @@ export type CreatableProps = {
 export type Props = SelectProps & CreatableProps;
 
 const compareOption = (inputValue, option) => {
-  const candidate = inputValue.toLowerCase();
+  const candidate = inputValue.toString().toLowerCase();
   return (
-    option.value.toLowerCase() === candidate ||
-    option.label.toLowerCase() === candidate
+    getOptionValue(option).toString().toLowerCase() === candidate ||
+    getOptionLabel(option).toString().toLowerCase() === candidate
   );
 };
 


akre54:
Just saw this PR. Unfortunately this doesn't fix #2953, as that will still use the default `option.label` and `option.value` instead of the custom fields. I created #2964 for that use case.

I'm wondering for this bug if we can just check for the presence of `toLowerCase` instead of casting to a string? Since '2345' doesn't have a lowercase representation, and will allow for other things like passing objects as values.

The alternative is the library can expose `compareOption` as an overridable option instead of needing to replace the entire `isValidNewOption` logic.

th3fallen:
i thought about checking their types but that seemed like a lot of overhead and i wasnt super familiar with the codebase so i didnt want to get too far down a rabbit hole

akre54:
I would do 

```
(option.value.toLowerCase ? option.value.toLowerCase() : option.value === candidate) ||
(option.label.toLowerCase ? option.label.toLowerCase() : option.label === candidate)
```

This only solves this bug. To solve both #2955 and #2953 you should save value and label to temp variables:

```
const value = getOptionValue(option);
const label = getOptionLabel(option);
(value.toLowerCase ? value.toLowerCase() : value) === candidate ||
(label.toLowerCase ? label.toLowerCase() : label) === candidate
```

webzfactory:
Any ideas when this will be merged? It looks like a standard use case.

bladey:
Hi @th3fallen,

As pointed out by @Alphonse-Indpro, it appears this PR is no longer required.

akre54:
Thank you for the more thoughtful response here. That pull request doesn't fix the main issue, though, which is that the library is looking for `label` and `value` props on each option and ignoring the `getOptionLabel`/`getOptionValue` functions which are supposed to help in this case.

bladey:
Thanks for letting me know @akre54.

ebonow:
Greetings @th3fallen  ,

Sorry for the extremely late reply to your PR submission. There is quite a backlog of user submitted PR's and the community deserves better than waiting months or years for a collaborator to reply to the work they've done.

I wanted to let you know that the issues that this PR addresses has been resolved:
https://github.com/JedWatson/react-select/pull/3316/files
https://github.com/JedWatson/react-select/pull/4444/files

As such, I will be closing this, thought I feel frustrated for you that this took so long to get resolved and wanted to add that I truly appreciate the time you put into coding a solution. Thank you for being a part of the community and contributing to react-select.

Issue #1086, opened 2016-07-13
Positioning react select to a specific item when opening


andreareginato:
Hi there.

I've been looking for a way to pre-select the position of the list to a specific item when opening it the first time and when an option is not selected.

The use case where I need it is one where there is a list of years (from 1900 to 2016). What I want to do is to set the opening list to 1970, the year that is selected the most as a mean.

I may be missing something, but both reading the docs, looking into the issues and checking the source code, I was not able to come up with a solution.

Below an image showing the initial state I would love to have (not current year)
<img width="754" alt="screen shot 2016-07-13 at 14 52 51" src="https://cloud.githubusercontent.com/assets/37480/16803907/5369822c-4909-11e6-8f0e-f50969016c17.png">

Thanks for your help and your time.

willdavidow:
Been poking around on this issue for the last 30 minutes, and arrived at the conclusion that there are some bugs in the default `renderMenu` function. I believe I'm using the latest - version `1.0.0-beta12`:

Checking out this block of code, lines 715-767 in `src/Select.js`:

```
renderMenu (options, valueArray, focusedOption) {
		if (options && options.length) {
			if (this.props.menuRenderer) {
				return this.props.menuRenderer({
					focusedOption,
					focusOption: this.focusOption,
					labelKey: this.props.labelKey,
					options,
					selectValue: this.selectValue,
					valueArray,
				});
			} else {
				let Option = this.props.optionComponent;
				let renderLabel = this.props.optionRenderer || this.getOptionLabel;

				return options.map((option, i) => {
					let isSelected = valueArray && valueArray.indexOf(option) > -1;
					let isFocused = option === focusedOption;
					let optionRef = isFocused ? 'focused' : null;
					let optionClass = classNames(this.props.optionClassName, {
						'Select-option': true,
						'is-selected': isSelected,
						'is-focused': isFocused,
						'is-disabled': option.disabled,
					});

					return (
						<Option
							className={optionClass}
							isDisabled={option.disabled}
							isFocused={isFocused}
							key={`option-${i}-${option[this.props.valueKey]}`}
							onSelect={this.selectValue}
							onFocus={this.focusOption}
							option={option}
							isSelected={isSelected}
							ref={optionRef}
							>
							{renderLabel(option)}
						</Option>
					);
				});
			}
		} else if (this.props.noResultsText) {
			return (
				<div className="Select-noresults">
					{this.props.noResultsText}
				</div>
			);
		} else {
			return null;
		}
[...]

PR #2957, opened 2018-08-21
[v2] Allow to focus an arbitrary option on menu open

malykhinvi:
The changes in this PR introduce a new property: `focusedOption`. It allows to focus the specified option when menu opens.

For example:
```
//years is an array of years, and we want to focus on some particular year
<Select
  options={years}
  focusedOption={years[100]}
/>
```

In v1 rc, we used to use `focusOption` method via refs. But looks like it doesn't exist anymore. See the similar issue https://github.com/JedWatson/react-select/issues/1086, which is related to v1 though.

Let me know, it you think that this change should be a part of react-select, or if you think that there is a better (more useful) solution to the problem. I'll be happy to work on it!


malykhinvi:
The PR got behind too much. Tests are not passing after conflicts resolution. Let's close PR for now and reopen later.

PR #2969, opened 2018-08-24
Fix "Replacable" typo in docs header

derekstavis:
This fixes a typo in Header's Select.


gwyneplaine:
thanks for the fix @derekstavis 👍

PR #2971, opened 2018-08-26
Typo: isMult -> isMulti

tony:


gwyneplaine:
Thanks for this @tony

PR #2973, opened 2018-08-27
Fix creatable options with integer values

guilatrova:
If I give options like
```
{
   label: "Apple",
   id: 1
}
```

The code breaks since `1` doesn't have a `toLowerCase` function.


akre54:
Dupe of #2956

PR #2981, opened 2018-08-28
fix changed prop typo

timothypage:
just a quick fix to upgrade guide


gwyneplaine:
Thanks for this @timothypage, good catch!

PR #2964, opened 2018-08-22
Fix for create with custom value / label fields

akre54:
Fixes #2953

PR #2990, opened 2018-08-30
Update Creatable.js

dextermb:
- Alter `compareOption` to not break if the option value is not a `string`

### DIFF
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -37,8 +37,16 @@ export type Props = SelectProps & CreatableProps;
 
 const compareOption = (inputValue, option) => {
   const candidate = inputValue.toLowerCase();
+  
+  if (typeof option.value === 'string') {
+    return (
+      option.value.toLowerCase() === candidate ||
+      option.label.toLowerCase() === candidate
+    );
+  }
+  
   return (
-    option.value.toLowerCase() === candidate ||
+    option.value === candidate ||
     option.label.toLowerCase() === candidate
   );
 };


src/Creatable.js
```diff
@@ -37,8 +37,16 @@ export type Props = SelectProps & CreatableProps;
...
+  if (typeof option.value === 'string') {
+    return (
+      option.value.toLowerCase() === candidate ||
+      option.label.toLowerCase() === candidate
+    );
+  }
+  
   return (
-    option.value.toLowerCase() === candidate ||
+    option.value === candidate ||
```
dextermb: Perhaps this should be `==` rather than `===` as I assume `inputValue` will always be a string, even if its numeric

DrMeers:
This should be more flexible and retain case-insensitive string comparisons? : https://github.com/JedWatson/react-select/pull/3000

dextermb:
@DrMeers I found this to be an issue when my value was numeric. It attempted to `toLowerCase` it which obviously won't work.

gamalielwkr:
Please merge and publish this asap. 😬 
Thanks!

IanVS:
@dextermb are you using `getOptionValue` to convert your integer to a string?  If so, then perhaps https://github.com/JedWatson/react-select/pull/2964 would also solve your issue?

bladey:
Hi @dextermb,

Thanks for this pull request.

JedWatson:
For completeness, the props `formatCreateLabel`, `isValidNewOption` and `getNewOptionData` props were introduced to `Creatable` to fix this and other related issues - [docs here](https://react-select.com/props#creatable-props)

If anybody comes across this issue and is experiencing something similar to the problems described here, please open a new issue with details so we can track down anything we missed.

Thanks!

PR #2996, opened 2018-09-01
fix CustomGetOptionLabel path

nomuyoshi:
Fix component path at custom getOptionLabel function example.

`docs/examples/CustomSingleValue.js` → `docs/examples/CustomGetOptionLabel.js`


bladey:
Thanks for this @nomuyoshi! @JedWatson this looks good to go.

PR #3001, opened 2018-09-03
Add custom getOptionValue function example

nomuyoshi:
I added `custom getOptionValue function example`.


gwyneplaine:
This is great @nomuyoshi, especially the doggos array 😄, thanks for the contrib!

PR #3003, opened 2018-09-03
Fix redirection link in Creatable doc

sushant24:
Fix the redirection link in  Creatable doc (api -> props)


gwyneplaine:
@sushant24  thanks for this contribution!

Issue #2998, opened 2018-09-01
Issue with compareOption Uncaught TypeError: Cannot read property 'toLowerCase' of undefined


vrbarros:
I found an issue with compareOption function when changes getOptionValue and getOptionLabel to a different (value|label) structure.= using with AsyncCreatable.

Follow the example to replicate.

The options are loaded from an endpoint using callback handler, with the JSON structure: `[{ id: 1, name: 'teste1' }, { id: 2, name: 'teste2' }, { id: 3, name: 'teste3' } ]`

```
const customFilter = createFilter({
      ignoreCase: false,
      ignoreAccents: false,
      trim: false,
      matchFrom: 'any',
    });

<AsyncCreatableSelect
        name="tag"
        id="tag"
        isMulti
        isClearable
        placeholder="Marque algumas tags para este registro"
        getOptionValue={option => option.id}
        getOptionLabel={option => option.id}
        filterOption={customFilter}
        loadingMessage={e => `Carregando opções próximas de '${e.inputValue}'`}
        noOptionsMessage={e => `Sem opções disponíveis para escolha`}
        formatCreateLabel={e => `Criar uma nova tag com rótulo '${e}'`}
        components={{ Option: Option, ...makeAnimated() }}
        loadOptions={loadOptions}
       
      />
```

Uncaught TypeError: Cannot read property 'toLowerCase' of undefined
    at compareOption (Creatable.js?067d:40)
    at eval (Creatable.js?067d:51)
    at Array.some (<anonymous>)
    at isValidNewOption (Creatable.js?067d:50)
    at Creatable.componentWillReceiveProps (Creatable.js?067d:133)
    at callComponentWillReceiveProps (react-dom.development.js?cada:11528)
    at updateClassInstance (react-dom.development.js?cada:11720)
    at updateClassComponent (react-dom.development.js?cada:13154)
    at beginWork (react-dom.development.js?cada:13825)
    at performUnitOfWork (react-dom.development.js?cada:15864)

When I got the function:

var compareOption = function compareOption(inputValue, option) {
  var candidate = inputValue.toLowerCase();
  return option.value.toLowerCase() === candidate || option.label.toLowerCase() === candidate;
};

vrbarros:
The workaround was creating a function to put the structure in the same value|label model that the compoment was made for.

```
const organizeOptions = array => {
      var items = [];
      for (var index in array) {
        items.push({ value: array[index].name, label: array[index].name });
      }

      return array;
    };
```

[...]

PR #3004, opened 2018-09-03
Fix creatable issue with user defined value and label;

jkeruzec:
Use select parent style to implement method.
TODO: Find a way that user provides property name for new creatables. Maybe add option.

Linked issue : #2998

### DIFF
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -35,31 +35,10 @@ export type CreatableProps = {
 
 export type Props = SelectProps & CreatableProps;
 
-const compareOption = (inputValue, option) => {
-  const candidate = inputValue.toLowerCase();
-  return (
-    option.value.toLowerCase() === candidate ||
-    option.label.toLowerCase() === candidate
-  );
-};
-
 const builtins = {
-  formatCreateLabel: (inputValue: string) => `Create "${inputValue}"`,
-  isValidNewOption: (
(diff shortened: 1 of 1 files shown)


src/Creatable.js
```diff
@@ -141,6 +122,51 @@ export const makeCreatableSelect = (SelectComponent: ComponentType<*>) =>
...
+    	    } else {
+    	return !(
+    	      !inputValue ||
+    	      selectValue.some(option => this.compareOption(inputValue, option)) ||
+    	      selectOptions.some(option => this.compareOption(inputValue, option))
+    	    )
+    	    }
+    }
+	 
+ compareOption(inputValue, option){
```
nafg: Inconsistent indentation

src/Creatable.js
```diff
@@ -141,6 +122,51 @@ export const makeCreatableSelect = (SelectComponent: ComponentType<*>) =>
...
+	 if (typeof this.props.compareOption === 'function') {
```
nafg: It seems wasteful to check and delegate here, since this will be called on every element.

Instead I would leave `compareOption` (perhaps as `defaultCompareOption`) in the outer scope and put the check in isValidNewOption.

bladey:
Hi @jkeruzec,

Thanks for this pull request.

PR #3008, opened 2018-09-04
Fix redirection in docs/advanced

sushant24:
Redirect to props#prop-types instead of api#prop-types.

I also searched for any other occurrences of /api in docs/pages but couldn't find any


gwyneplaine:
thanks @sushant24!

Issue #2994, opened 2018-08-31
async handleInputChange doesn't receive new value 


alegaagafonov:
You can see the issue right inside one of the live examples [here](https://codesandbox.io/s/p7m40vl7wq?module=/example.js)
state's inputValue is always empty because handleInputChange doesn't actually receive newValue.

sahibjotsaggu:
I'm not entirely sure but I think this may be it:

https://github.com/JedWatson/react-select/blob/master/src/Select.js#L575

Vitzkrieg:
I ran into this same issue. The `onInputChange `needs to be `onChange`. The sandbox is updated with the change.

Rall3n:
> Also, what is the difference between `onChange` and `onInputChange` because replacing `onInputChange` with `onChange` makes it work if you change

@sahibjotsaggu The prop `onChange` is used if you change the select value (e.g. selecting an option), which should be obvious because the same name is used for the standard `onChange` prop on vanilla html inputs.

The prop `onInputChange` is used if you type something into the search input of the select component.

You should also not confuse the props `value` and `inputValue`: `value` is for the actual select value and `inputValue` for the text input (the search).

@alegaagafonov Could you specify your error and usecase? I could not find an error with `onInputChange`. It receives the value which is being typed into the input.

Also, [Line 575 (now 607 as of current master) in Select.js](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L607) is not the culprit, as it only changes the `inputValue` to an empty string **after an option from the select has been selected**.

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our efforts towards the current major version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue  is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

PR #3009, opened 2018-09-04
pass the value to onInputChange

alegaagafonov:
Fix for #2994 Issue


yairEO:
when clicking outside of the field, the `value` passed to `onInputChange` callback is an empty string.. any ideas?

ebonow:
Greetings and thank you for the submission. Also apologies for the delay in a response. 

The outcome of this PR seems to contradict the designed behavior of react-select in that the input selection is intended to be cleared when the value is changed and thus would cause breaking changes for uses, without any opportunity for opt-in (as some other similar PRs have made available). 

Issue #2816, opened 2018-07-20
[V2] IE11 vertical scrollbar


rogier2012:
Opening the first select component of the demo page in Internet Explorer 11 shows a scrollbar, but clicking on this scrollbar closes the selector menu.

4dom2dom:
I have the same issue in IE11. Hope it will be get soon carried out.

4dom2dom:
The issue is discussed here https://github.com/JedWatson/react-select/issues/1020. Unfortunately, it is not fixed.

andrey99z:
Here is quick patch for current version based on Rakid's code: 
```
   componentDidMount() {
        const select = this.rs.select;
        if (!select.onInputBlurPatched) {
            const originalOnInputBlur = select.onInputBlur;
            select.onInputBlur = e => {
                if (select.menuListRef && select.menuListRef.contains(document.activeElement)) {
                    select.inputRef.focus();
                    return;
                }
                originalOnInputBlur(e);
            }
            select.onInputBlurPatched = true;
        }
    }
```
(assuming that this.rs is ref to React-select component)
HTH

pohsoon:
Nice. This quick patch works in my IE 11.

M-Yankov:
> 
> 
> Here is quick patch for current version based on Rakid's code:
> 
> ```
>    componentDidMount() {
>         const select = this.rs.select;
>         if (!select.onInputBlurPatched) {
>             const originalOnInputBlur = select.onInputBlur;
>             select.onInputBlur = e => {
>                 if (select.menuListRef && select.menuListRef.contains(document.activeElement)) {
>                     select.inputRef.focus();
>                     return;
>                 }
>                 originalOnInputBlur(e);
>             }
>             select.onInputBlurPatched = true;
>         }
>     }
> ```
> 
> (assuming that this.rs is ref to React-select component)
> HTH

The code works good, but in my case I am using `react-window` to replace the MenuList component. In that case `select.menuListRef` is always null. Any idea how to fix this ?

andrey99z:
Upgrade to the current version? This seem to be fixed now and the patch is not needed anymore.

M-Yankov:
> 
> 
> Upgrade to the current version? This seem to be fixed now and the patch is not needed anymore.

Yes it fixed, I was editing my previous comment with the solution I found, but obviously I forgot to click save 🤓

bladey:
Hello -

[...]

Issue #1020, opened 2016-06-10
Dropdown closes when click on scroll in IE11


tknysh:
![](http://g.recordit.co/H95HgaDpoz.gif)

twschiller:
Potential regression of #495?

Related issues:
- #848
- #846

acromarco:
According to my research, this bug (re-)appeared with IE11 in Win7 after this security update https://support.microsoft.com/en-us/help/4040685/cumulative-security-update-for-internet-explorer 
With the new IE11 version 11.0.9600.18816 this bug is reproducible, but everything worked fine with the older version 11.0.9600.17420.

acromarco:
selectize is also affected by this IE11 bug and they have probably found a fix https://github.com/selectize/selectize.js/issues/1004

acromarco:
I wonder if this is already fixed with https://github.com/JedWatson/react-select/commit/2da6ff6a352c8720fe59ca037524dbff9109af96 because adding tabIndex={-1} should fix it.

sushant-at-nitor:
Code on [this line](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L347) should add one more condition to fix this issue.

`if (this.menu && (this.menu === document.activeElement || this.menu.contains(document.activeElement))) {`

`document.activeElement.classList.contains(<configured class name for menu renderer>)` 

OR

Let [this line](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L353) decide whether to set new state to hide menu options or to continue to show them, stop further execution of code.

```
const result = this.props.onBlur(event);

if(result === false){
  this.focus();
  return;
}
```

stephanebarbey:
Is this issue still open? (or any chance that the fix by sushant-at-nitor gets integrated?)

jojo-tutor:
just add this code: 
```
handleOnBlur() {
    const e = null
    throw e
  }
```

AO17:
if this has been fixed (see [merge](https://github.com/sushant-at-nitor/react-select/commit/791ca1deee611b1e618db3b66b8cbd9ec77400c9) above):

1. Why is this issue still open?
2. I still have the same issue, if the fix has been merged, why is the bug still occuring?

maeln:
I can confirm that I have the same issue with the latest version of react-select.

maeln:
To expand on @jojo-tutor "fix-hack", you can do this if you still want your select to close when it lose focus:
```
 onBlur={() => {
[...]

PR #3013, opened 2018-09-05
fix Select close when scroll on menu in IE11

rakid:
related to #1020 and #2816

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -1039,6 +1039,10 @@ export default class Select extends Component<Props, State> {
     this.openAfterFocus = false;
   };
   onInputBlur = (event: SyntheticFocusEvent<HTMLInputElement>) => {
+    if(this.menuListRef && this.menuListRef.contains(document.activeElement)) {
+      this.inputRef.focus();
+      return;
+    }
     if (this.props.onBlur) {
       this.props.onBlur(event);
     }


DaveCole:
This fixes the issue but it would be nice if clicking outside of the element closed the drop-down.  Hopefully @JedWatson sees this soon.

rakid:
Hey @DaveCole! did you test it on differents browsers, especially IE11? 
the url of fiw is : https://deploy-preview-3013--react-select.netlify.com/home

DaveCole:
Thanks @rakid - I did test it on IE11, as well as Chrome; the fix works very well; there's just one case where it didn't quite work for me, and that was if I opened the select, scrolled a little bit, then went to click outside of the expanded window -specifically- to the right of the select.  If I clicked anywhere above or below it, it would close.  I'm wondering if maybe there's a cropped div catching the event or something in that case.  Anyhow the fix makes things way, way better.  Thanks again.

rakid:
Ok I just figure out the bug you have, I will try to fix it.
Maybe I need to add this code
`inputRef.focus()` before return?

piotrwitek:
Hey @JedWatson @gwyneplaine are you willing to merge this PR? Thanks!

sunil4444:
Waiting for this fix

chris-herring:
Also need this fix.

sbaergen:
@JedWatson @gwyneplaine Any possibility of this being merged soon?

gwyneplaine:
thanks for this @rakid, merged

PR #3016, opened 2018-09-06
chore: typo in test comment (porps -> props)

theneva:
Hehe. I misspelled it in a search, and found this one.


gwyneplaine:
hahah thanks for this @theneva, good catch 😄

PR #3027, opened 2018-09-11
classNamePrefix documentation improvements

TristanJM:
The documentation lacks clarity around the `classNamePrefix` prop, making it difficult to integrate react-select with popular libraries such as [styled components](https://www.styled-components.com/) and [CSS modules](https://github.com/css-modules/css-modules).

This PR introduces:
- clear descriptions to the Props API docs for `classNamePrefix`
- an explanation that classNames are added to inner elements when the `classNamePrefix` is provided
- the removal of a redundant example in the Styles section
- a reference to other component styling methods to bolster react-select's integration capabilities

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ Common props you may want to specify include:
 
 * `autoFocus` - focus the control when it mounts
 * `className` - apply a className to the control
+* `classNamePrefix` - apply classNames to inner elements with the given prefix
 * `isDisabled` - disable the control
 * `isMulti` - allow the user to select multiple values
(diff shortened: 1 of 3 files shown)


gwyneplaine:
Thanks for this @TristanJM  looks great, definitely makes the prop behaviour much easier to understand.

jeromeraffin:
In which case it ensures compatibility with CSS Modules? If someone find a solution to override the styles of the select with CSS modules I'm interested in and it would be good to have it on the doc...

gegham1:
what about this issue ?
https://github.com/JedWatson/react-select/issues/3754

PR #3034, opened 2018-09-13
fix for control disabled state

stanislav-ermakov:
There was a check in onDropdownIndicatorMouseDown, but not in onControlMouseDown, so it was possible to open dropdown still even if disabled.


gwyneplaine:
Closing in favor of  #3036

PR #3036, opened 2018-09-13
fix for control disabled statev2

stanislav-ermakov:
There was a check in onDropdownIndicatorMouseDown, but not in onControlMouseDown, so it was possible to open dropdown still even if disabled.


e-serebriakov:
@JedWatson 
Could you merge this fix?

theskillwithin:
+1

gwyneplaine:
merged, thanks @stanislav-ermakov

Issue #2669, opened 2018-06-01
[V2] Setting initial value of select component with a string value no longer works


jbgarr:
Maybe I'm missing something, but from what I can tell the way that used to work to set the initial value of the select component does not work anymore in version 2.0. Here's what used to work (simplified for readability):
    
    <ReactSelect
        value={field.value}
        options={field.options.map((option) => ({
            ...option,
            label: option.name,
            value: option.name,
            clearableValue: true
        }))}
    />

In the above example the `field.value` was a string value coming back from the server that matched the name of one of the options.

Is it no longer possible to set the initial value (and all subsequent values) with a string? That was a very nice feature and I would like to make a request for it to make a comeback if it's no longer available. If it's still there, can someone show me the light? I can't seem to get it working with v2.

billmers:
I ran into this as well. The workaround is to pass the full option object and not just the string value, for example: `value: { label: 'foo', value: 'bar' }`.

Not sure if it's a bug or not, but it'd be convenient to set the value just using a string.

That being said I've spent the last two weeks building a reasonably advanced searching and filtering solution all powered by react-select v2 beta, and have run into almost zero issues. Reading through the source code is a nice experience as well, super well factored.  Thanks @JedWatson and friends!

jbgarr:
Thanks for the reply @billmers.

I've noticed that passing in the object works as well, but that means that instead of using a simple string value (which is what I assume most people will use to save values as when passing data between the client and the server) we need to first lookup the proper object in the options array and then pass in a reference to that. It's not the end of the world, but it's definitely not as convenient as using the string value.

Hopefully this will make a comeback in V2.

TheSharpieOne:
The issue comes from `cleanValue` which only expects an array or an object. `cleanValue` is used get a value for comparing to determine which option is selected.
https://github.com/JedWatson/react-select/blob/94df8a7ac19ec76beb8c8ccbc7fbbde86f26b5fb/src/utils.js#L53-L61

sys13:
I also second having the value string functionality. App went from:
```
value={this.state.contactInfoTypeCandidate}
```
to: 
```
value={
  this.state.contactInfoTypeCandidate && {
    label: getEnumsAsSelectOptions({ enums: ContactInfoType }).find(
       ({ value }) => value === this.state.contactInfoTypeCandidate
       ).label,
    value: this.state.contactInfoTypeCandidate,
   }
}
```

Parader:
+1

[...]

PR #3050, opened 2018-09-18
Find correct option when value is passed as a string

niekert:
Aims to resolve: https://github.com/JedWatson/react-select/issues/2669. 

I believe this is a pretty common use case so I think it's nice to support this by default.

Our main use case for this is that we set the `value` of a `<Select />` based on a string value from an API response. Instead of having to map the string value to one of the `options` object passed to the `<Select />` I think it would be nice to be done automatically because it seems to be a common use case.

The implementation is pretty straightforward so maybe (likely) I'm missing some cases.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "javascript.validate.enable": false
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-  "javascript.validate.enable": false
-}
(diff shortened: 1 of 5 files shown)


jossmac:
Hi @niekert,

Thanks for the PR. We've discussed and would really rather not allow "simple value" into core. That being said the preferred pattern would be components that augment react-select.

To get the ball rolling I've created [react-select-simple-value](https://www.npmjs.com/package/react-select-simple-value), a function that receives an object (props) and calls the consumer's `children` function with an augmented version of that object:

```jsx
<SimpleValue options={[...]} value="blue">
  {props => <ReactSelect {...props} />}
</SimpleValue>
```

niekert:
Awesome, that makes sense. Love that there's a package for it now 👍

baleeds:
Hi @jossmac,

Could you elaborate on why allowing simple values is not preferred?  Just curious if there are negative side-effects of implementing it, even as a wrapper?

Issue #986, opened 2016-05-24
Search is weird (e.g. "u" brings up "Hungary")


callumlocke:
It's a very weird default, to search in the middle of words. And I can't find any way to change this behaviour.

hedaro:
From the README:

> matchPos: "start" or "any": whether to match the text entered at the start or any position in the option value

Just pass the prop `matchPos="start"` and that should do what you want.

callumlocke:
thank you! that solves my problem, I had missed that option.

but even with `matchPos="any"` is it really supposed to search inside words? I would understand if it searched from the start of _any_ word (e.g. typing <kbd>b</kbd> could bring up `Great Britain`), but it seems a really weird option to want to search for any letter _inside_ any word.

hedaro:
That is because each value/label are whole strings, so "start" matches the beginning of the string and "any" searches anywhere inside the string.

callumlocke:
Thanks, I get how to configure it now. But I still think react-router should change its default setting for this. In my opinion `matchPos="start"` would be a saner default

PR #3073, opened 2018-09-26
Filter: Add support for a mode where a match at the start of any word is accepted

papandreou:
Hey, I ran into the same thing as https://github.com/JedWatson/react-select/issues/986#issuecomment-221432678 and thought I'd give it a shot.

Essentially I want to look for a match in any word of each item, which is not quite the same as `start` (which only looks in the first word) or `any` (which looks for any substring).

Come to think of it, I'd actually like for multi word needles to be matched in an order-insensitive way so that a search for "Brit Gre" would find "Great Britain". It feels like something that should be implemented as an orthogonal `ignoreOrder` mode. Happy to give that a shot in a separate PR if there's any interest, but let's start here :)


papandreou:
Here's a quick POC of the `ignoreOrder` mode suggested above: https://github.com/papandreou/react-select/commit/f176472854fc40fb9bbe04b9149fb7a72fd09cf3

papandreou:
Rebased. Is there any interest in this?

papandreou:
Ping @JedWatson?

packages/react-select/src/filters.js
```diff
@@ -35,7 +35,17 @@ export const createFilter = (config: ?Config) => (
...
+  if (matchFrom === 'start') {
+    return candidate.substr(0, input.length) === input;
+  } else if (matchFrom === 'start-word') {
+    const matchIndex = candidate.indexOf(input);
+    // Accept a match at the start of the string or one preceded by space:
+    return (
+      matchIndex === 0 ||
+      (matchIndex > 0 && /\s/.test(candidate.charAt(matchIndex - 1)))
+    );
+  } else {
```
Methuselah96, line 47: A little bit of an edge case that might not be worth handling, but I don't think `filter('foo ooze', 'oo')` would work with this code.

Methuselah96 (changes requested):

packages/react-select/src/filters.js
ebonow: I was thinking maybe a better use-case might be a `match-words` type. Similar to a google search where "react select" could return the same results as "select react". It could be used in a customFilter to search against keywords, aliases, metadeta, etc...

Thoughts?

Methuselah96: Yeah, I agree that `match-words` might be more helpful than `start-word`. But it's worth noting that this can all be customized with the `filterOption` prop (from what I can tell), so this falls in the question of whether this should be "batteries included" or whether it just adds to the bloat.

ebonow: Agreed. I personally also have a filterOption that includes all group options if the GroupHeading matches. 

Choose a region:

ie: North => 
```
North Korea
+ Chagang
+ North Hamgyong
+ South Hamgyong
+ North Hwanghae
+ South Hwanghae
+ Kangwon
+ North Pyongan
+ South Pyongan
+ Ryanggang
...
United States
+ North Carolina
+ North Dakota
 ```

I think things like these are useful and I like the idea of creating a place for frequently used/requested features, but these all seem like use cases we may want to splinter off into its own place independent of the general plumbing. 

I'm good to close this until we can better organize this outside the primary package. Agree?

Methuselah96: Yeah, sounds good.

papandreou: I fixed the `foo ooze` case, but I guess that means NFA from me for now :)

ebonow: I genuinely appreciate and like the PR and I would love to continue to add more functionality, so wanted to let you know a bit more about what's going on behind the scenes...

There are discussions about what the future of react-select and it's looking like there is going to be a big rewrite to embrace more functional components and hook implementation. The idea is to allow users to add more functionality on a plugin basis. We're not quite sure what this is going to look like, but creating a place specifically for common filters for users to import would help us start to add more functionality while allowing us to decouple more of the "opinionated" decisions.

I think we'd like to naturally organize and promote these into one place to make it easier for others to find this functionality and this like many other feature request PR's that exist right now would qualify as candidates for this. We're still in the planning stage and we're quite a bit away from there, so it's more of a hold tight because we want to make this fit as we scale the project.

papandreou: Thanks for the update! It's great to learn that there are active maintainers of this important package now!

In my own component library I've just copied in a filter implementation that does the "start of word" thing -- luckily the library was already flexible enough that I could do that without forking. So I'm already unblocked, I just felt that these filtering semantics would be a better default :)

I can leave this PR around as a reminder, or I can just close it?

PR #3083, opened 2018-09-28
fix: add sideEffects: false to package.json

SimenB:
Make tree-shaking with webpack work better.

See example project for justification and example: https://github.com/SimenB/select-side-effect


theneva:
Hooray! :tada: This makes it possible for us to use this component. @gwyneplaine will this be released in the foreseeable future? :smile:

PR #3086, opened 2018-10-01
Fix dropdown position in portal: use container boundaries

estambakio-sc:
In 2.0 `Select` can render dropdown using `ReactDOM.createPortal()` which is useful. But its position doesn't take into account container boundaries. 

I believe this is because it was tested with `document.body` as a portal target. But there are cases when we'd want portal to be rendered inside a different element. For example, if `Select` is emdebbed somewhere in a `react-bootstrap/Modal` then dropdown gets hidden under modal, which has `z-index: 1050`. Solution: render portal into an element which is inside `Modal`. 

`Modal` is centered on a page and dropdown is positioned like this: 

react-select-2-dd-portal-position-bug

The reason is `Modal` has `top`, `left`, `offsetTop` and `offsetLeft` which are not taken into account currently (`document.body` has these values equal to `0`).

After applying this PR it is fixed: 

react-select-2-dd-portal-position-bug-fixed

**Please review if you can merge & release it.** Let me know if something doesn't conform to code of conduct, because this is my first PR to `react-select`, though I passed all steps in this doc https://github.com/JedWatson/react-select/blob/master/.github/CONTRIBUTING.md.

### DIFF
--- a/src/components/Menu.js
+++ b/src/components/Menu.js
@@ -12,7 +12,6 @@ import PropTypes from 'prop-types';
 import {
   animatedScrollTo,
   getBoundingClientObj,
-  type RectType,
   getScrollParent,
   getScrollTop,
   scrollTo,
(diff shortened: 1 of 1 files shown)


jossmac:
Interesting PR, thanks @estambakio-sc 

Just thought I'd mention that you can change the `z-index` using the styles API. Something like:

```jsx
<Select
  ...
  styles={{
    menuPortal: provided => ({ ...provided, zIndex: 2000 })
  }}
/>
```

estambakio-sc:
@jossmac sure, but it's not a bullet-proof solution in case if modal called in another modal or other case, if someone picks a higher number for `z-index`.

gwyneplaine:
@estambakio-sc  this looks good, could you also create an example for this in the docs please, additionally an integration test with cypress would be much appreciated. 

thanks 🙏

vishakaw1:
Hi, I have same kind of problem, when my multi select use menuPortalTarget and place it in 'document.body' it is not responsive in mobile devices.
The dropdown is inside a modal and when open the dropdown the keyboard is also opened up and modal get centered. But the dropdown still stays in the originally created place.

bladey:
Hey @estambakio-sc,

As per @gwyneplaine's request, are you able to also add a example for the docs and an integration test?

Let me know if you have moved on, and we'll re-assess this PR. Thank you!

smeijer:
It's been two years. I think it's safe to say that OP has moved on.

This fix would be nice to have though!

estambakio-sc:
@smeijer I don't have time to spend on Cypress tests for it; if someone does it then maintainers will probably merge PR. In my projects I use this [fix mentioned above](https://github.com/JedWatson/react-select/pull/3086#issuecomment-428443568).

vsorrokin (approved):

vsorrokin (approved):

dcousens:
@atMari, if I may ask, what version of `react-select` are you using?

dcousens:
Marking as stale, if this is still relevant, we can re-open

PR #3089, opened 2018-10-01
More reliable way to disable autocomplete

methyl:
See https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -1355,7 +1355,7 @@ export default class Select extends Component<Props, State> {
     return (
       <Input
         autoCapitalize="none"
-        autoComplete="off"
+        autoComplete="nope"
         autoCorrect="off"
         cx={cx}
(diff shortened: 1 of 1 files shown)


bladey:
Hey @methyl,

Thanks for the PR - I can't find anything on the linked documentation about the `nope` value that can be used in `autoComplete`.

The page may have been updated since the PR was created, could you provide more information if this is still relevant?

Thanks!

methyl:
Well, at this point I'm not sure if it's still needed. There was a time that autocomplete needed to be set to some random, not supported value (hence nope), but it's _probably_ no longer relevant.

bladey:
Thanks for your feedback @methyl. From what I can see it was once a thing, but doesn't appear to be relevant anymore. If you see otherwise in the future, please do let me know!

PR #3090, opened 2018-10-02
Add aria attributes to dummy input

akiselev:
When `props.isSearchable` is `false`, aria attributes like `aria-label` and `aria-labelledby` are not passed through to the `DummyInput` control that takes place of the input text box. Since the dummy component can't be replaced, making a non-searchable select breaks accessibility even though there are standards patterns for non-editable combo boxes.

Since `react-select` consumers have control over both the accessibility props and `isSearchable` without control over `DummyInput`, they should remove the accessibility props from the dummy input themselves by changing `<Select>` props.


scolacur:
@JedWatson is it possible to get this added into the library?

jrparish:
Can we revisit this fix? It would be great to get this in to solve accessibility issues with this component.

darian-trama:
Just wondering if this update will make it into the next release. Also wondering when next release would be?

PR #3100, opened 2018-10-09
fixed indicators flow types

iseredov:
Flow types for indicator components was described incorrectly. An error occurred while redefining indicators as described in the documentation. In order for the type check in the children to work correctly, the default value define in defaultProps.


gwyneplaine:
@iseredov much appreciated, thanks!  merged

PR #3108, opened 2018-10-10
Update docs to add the single-value stored as string use case

### LINKED ISSUE #3107: [Docs] Add specific use-case for single value react-select storing string instead of the entire object.
This is a follow-up from #2669, where we said to update the docs to reflect how to properly define a react-select with single values, storing its value instead of the react-select object.

~~~javascript
const options = [
  {label: 'Whatever', value: 'whatever'},
  {label: 'Two', value: 2},
]
return (
  <ReactSelect
    isMulti={false}
    options={options}
    value={options.filter(({value}) => value === this.state.value)}
    getOptionLabel={({label}) => label}
    getOptionValue={({value}) => value}
    onChange={({value}) => this.setState({value})}
  />
~~~

Please note that I'm already creating a PR for this. This issue is just to properly track of it.

elboletaire:
closes #3107

### DIFF
--- a/docs/pages/upgradeGuide/index.js
+++ b/docs/pages/upgradeGuide/index.js
@@ -264,6 +264,32 @@ const customFilter = createFilter({
 
 See the [Advanced Guide](/advanced) for more details and examples.
 
+## Storing strings with non multi select
+
+React-select v1 allowed to use strings for the \`value\` prop, but with v2 it
+uses arrays or objects in all the cases. If you still want to use strings you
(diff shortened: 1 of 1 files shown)


gwyneplaine:
Thanks for this @elboletaire

elboletaire:
Hhmm.. apparently the title broke the sidebar

imatge

elboletaire:
Filled new issue #3134

Issue #2682, opened 2018-06-06
[v2] Backspace on single select results in [] rather than {}


ernestopye:
# Are you reporting a bug or runtime error?
Bug. To reproduce, you can use the link below. Note: This is NOT in isMulti mode. [] makes sense for isMulti.
[Code Sandbox](https://codesandbox.io/s/x977z5p7wo)

1. Select a value.
2. Click on the select to expand the options menu.
3. While the menu is still open, hit backspace.
4. Actual: onChange is called with []
5. Expected: onChange is called with {}

This seems to be an issue that has happened in the past.

Let me know if I can provide any other details. The code sandbox above is simply one of your examples, but with an added console.log call from onChange, nothing else.

Darep:
Also ran into this 🙂 Solved for now with a wrapping onChangeHandler:

```javascript
  onChangeHandler = (value, actionMeta) => {
    if (this.props.onChange) {
      // fix react-select returning [] on clear
      if (Array.isArray(value) && value.length === 0) {
        this.props.onChange({}, actionMeta);
      } else {
        this.props.onChange(value, actionMeta);
      }
    }
  };
```

vincentjames501:
Just ran into this one today. Definitely doesn't make sense to return an empty array if it's a single select. The item or null is what makes sense to return IMO.

someden:
https://github.com/JedWatson/react-select/blob/18219f5271705ef150ad16177db43f0c44f0413d/src/Select.js#L802-L810

`single select, by default, IS NOT clearable`, but `backspaceRemovesValue` by default is `true`, so single select can be cleaned by backspace, it is very confusing.

grassick:
I've created a PR (https://github.com/JedWatson/react-select/pull/3110) and put a temporary corrected version up at `github:mWater/react-select#fix_2682_built` for anyone else who is blocked by this bug.

Thebarda:
Here is a workaround that works with search input and clearable select :
```javascript
onSelectChange = item => {
    // Prevent backspace in input (item.length === undefined)
    if (item === null || item.length === undefined) {
      // Write your code here
    }
  }
```

I hope it help someone

kstratis:
Basically if that’s the case, then why fire onChange in the first place?

PR #3110, opened 2018-10-10
Prevent backspace from returning [] in non-multi select (issue #2682)

grassick:
The backspace key causes onChange to be called with `[]` on a non-multi select control, which breaks untold things.

This patch changes the backspace behavior to only pop values when `isMulti` is true. This prevents the issue with the incorrect call to onChange.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -1114,7 +1114,7 @@ export default class Select extends Component<Props, State> {
         if (focusedValue) {
           this.removeValue(focusedValue);
         } else {
-          if (!backspaceRemovesValue) return;
+          if (!backspaceRemovesValue || !isMulti) return;
           this.popValue();
         }
         break;


IanVS:
Shouldn't the `onChange` still be called, but with `null` instead of `[]` in the case of a non-multi select?

IanVS:
I didn't see your reply, and made another PR: https://github.com/JedWatson/react-select/pull/3216

As part of that, I added some tests, and if I do what you've done here, I find that [this one](https://github.com/JedWatson/react-select/pull/3216/files#diff-b261a148a668ffb134e0042bc200c90dR1419) fails.  Am I missing something?

grassick:
How right you are... backspace doesn't work in the single case. Oops. I'll close this PR in favor of yours. Thanks!

Issue #2872, opened 2018-07-31
[V2] Disable default styles


burtek:
There is no easy way to disable all default styles to use CSS styling. I need to pass:

`styles={clearIndicator: () => ({}), container: () => ({}), control: () => ({}), ...}`

Which is (needless to say) very ugly. Please, provide a prop (i.e. `noDefaultStyles`) to be able to disable default styles.

vikasraok:
+1

Andries-Smit:
There is pull request open for this #3113

Zacaria:
+1

iJoyCode:
+1

MobileD2:
+1

ytkonaft:
+1

ymazhar:
+1

at-cuongtran:
I almost never use default style ever, maybe a theme, so it would be very helpful if we can just disable it

iJoyCode:
+1

Zacaria:
Hello guys, don't stay stucked here.
Patch yourself the lib and use https://www.npmjs.com/package/patch-package.

This will be much faster 😄

Rall3n:
If you don't want to use an object with all the styles set to an empty function, why not use a [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy):

```jsx

const styleProxy = new Proxy({}, {
    get: (target, propKey) => () => {}
});

<Select styles={styleProxy} {...} />
```

It's much cleaner than a complete object.

Methuselah96:
I've started [a fork](https://github.com/react-select-oss/react-select) of react-select. Feel free to resubmit this issue on the fork and/or submit a PR to resolve this issue on the fork and we can get it merged and released.

ebonow:
@Rall3n's answer appears to resolve this question.

Here is a working [codesandbox demo](https://codesandbox.io/s/react-select-unstyled-tx8z9?file=/src/App.js)

As a result, I will be closing this now, but if there are any other concerns or questions regarding this, please feel free to reply and this can be reopened if necessary.

studnik18:
Hello. I just used the Proxy solution, but the style tags are still injected - I mean the styles for `Svg` and `a11yText`. I can't figure out how to disable them.

unloved:
+1

PR #3113, opened 2018-10-11
feat: Disable inline styles

jschill:
To disable the inline styles in 2.x you need to pass an object with a bunch of empty functions. With this PR I suggest the style prop to accept a null-value for these use cases.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -740,9 +740,14 @@ export default class Select extends Component<Props, State> {
     return this.props.getOptionValue(data);
   };
   getStyles = (key: string, props: {}): {} => {
+    const { styles } = this.props;
+    if (styles === null) {
+      return {};
+    }
+
     const base = defaultStyles[key](props);
     base.boxSizing = 'border-box';
-    const custom = this.props.styles[key];
+    const custom = styles[key];
     return custom ? custom(base, props) : base;
   };
   getElementId = (element: 'group' | 'input' | 'listbox' | 'option') => {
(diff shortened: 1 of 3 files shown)


Andries-Smit:
Would be nice to have a CSS file with all default style so it could be bundled in an other way

ChrisBrownie55:
Any updates on if this can get merged or if there's already a way to do this?

ramraphael:
Would love to work on this feature to help it get merged in. Any updates @jschill ?

ramraphael:
Would love to help get this merged in. Any updates @jschill ?

jschill:
@ramraphael I haven't done anything since the PR. Feel free to take ownership :-)

ebonow (approved):

ebonow:
Greetings @jschill and my apologies for receiving no response from any of the collaboration team in over two years. Our contributors deserve better than this level of communication and I am sorry that so much time has passed since this PR was submitted without any communication.

I think having the ability to easily remove styles without worrying about overrides would be a benefit to those who want more control over the styling. I would recommend an approach like this and will recommend this for consideration for v3.3

ebonow:
After some further discussion, I think there will be momentum for an unstyled Select which will likely be a preferred option as it would ship with no dependency on Emotion. I will try to keep you in communication with how this conversation goes especially once we are able to get a sense of priority and a milestone behind such a feature to see if this would be a good stopgap, though it should be mentioned that there already exists a one line solution to this without impacting existing source code: https://github.com/JedWatson/react-select/issues/2872#issuecomment-654251799

```jsx

const styleProxy = new Proxy({}, { get: (target, propKey) => () => {} });

<Select styles={styleProxy} {...} />
```

rcketscientist:
AFAICT that "near future version" never emerged, unless I missed somethign in the notes.  If I missed something, please let me know.  I'd love to get rid of an ancient bug in the backlog and update the doc here to make it clearer.

Issue #2755, opened 2018-06-27
[V2] Invalid attempt to destructure non-iterable instance


ppascualv:
It only happens in Android 4.3

### Traceback
TypeError: Invalid attempt to destructure non-iterable instance
- File "webpack:///./node_modules/react-select/dist/react-select.es.js" line 145 col 1 in E
throw new TypeError("Invalid attempt to destructure non-iterable instance");
- File "webpack:///./node_modules/react-select/dist/react-select.es.js" line 3445 col 1 in HTMLDocument.onTouchStart
var _ref2$touches = slicedToArray(_ref2.touches, 1),

ppascualv:
I'm using `^2.0.0-beta.6`

amytych:
FWIW, we experience exactly the same error with the same traceback in our app. As reported by Rollbar, it happens on iOS, in Mobile Facebook and Safari

![image](https://user-images.githubusercontent.com/998924/45146452-2efef500-b1c3-11e8-821b-2d9065ab157c.png)

We're using `2.0.0` version.

kuongknight:
Confirmed this issue on safari - iOS 10
We're using 2.0.0 version.

xakep139:
@JedWatson @gwyneplaine do you plan to fix this issue? Any workaround for this?

xakep139:
Maybe consider using [`item()`](https://developer.mozilla.org/en-US/docs/Web/API/TouchList/item) instead of destructuring `touches` property of [TouchEvent](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent) in [Select.js#L990](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L990) and [Select.js#L995](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L995)

siod:
Can confirm we're seeing this on Safari - iOS 10 with version 2.1.2.

PR #3126, opened 2018-10-18
Fixed issue #2755

kuongknight:
TouchList is not iterable on safari 9
https://github.com/babel/babel/issues/545


kuongknight:
uncertain ???

gwyneplaine:
Thanks for this @kuongknight, we've merged in a similar solution with an additional null check for safety, this bug should now be resolved in v2.2.0

PR #3151, opened 2018-10-29
Docs: Updated getOptionValue and getOptionLabel examples

cutterbl:
Added a new `data` variable and updated examples for the `getOptionValue` and `getOptionLabel` examples, as well as linking to the examples from the Select doc meta.

### DIFF
--- a/docs/data.js
+++ b/docs/data.js
@@ -116,3 +116,9013 @@ export const groupedOptions = [
     options: flavourOptions,
   },
 ];
+
+export const movieOptions = [
+  {
+    id: 1,
+    first_name: 'Heida',
+    last_name: 'Matcham',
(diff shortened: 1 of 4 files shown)


ebonow:
Greetings @cutterbl and thank you for the submission. Also apologies for the delay in a response. 

The documentation currently includes code examples for both `getOptionValue` and `getOptionLabel`. While in some cases more is better, it's fair to say that adding more complexity to the documentation may make it more difficult for users to find what they are looking for especially in regards to all of the props, behaviors, and nuances, some of which are still likely undocumented. The team would like to err on the side caution before adding anything that might be considered redundant and make navigation any more difficult than it already might be considered for some.

PR #3154, opened 2018-10-30
Add missing `data` prop type to `Option`

kirillku:
The `data` prop type is missing in the docs here https://react-select.com/props, it exists in `Option` 's props.


gwyneplaine:
@kirillku  thanks for this

PR #3157, opened 2018-10-31
Tidy styles docs

jossmac:
A few things here:

- The term `base` is not very clear, I've opted for `provided` as the variable that represents the default styles provided by react-select.
- Having the style keys as a "cloud" was hard to parse, moved to an unordered list.
- Removed references, like `isFullscreen` and `<View/>`, which were copy/pasta from react-images.


gwyneplaine:
sweet, thanks @jossmac 
merged

PR #3161, opened 2018-10-31
AsyncSelect: Initialize state with `inputValue` when `defaultOptions` is true

### LINKED ISSUE #3160: AsyncSelect does not respect inputValue prop on initial load
When working with the AsyncSelect, and attempting to apply in `inputValue` prop to the component, the `inputValue` is not passed to the `loadOptions` method on the initial load with `defaultOptions` is `true`. An example of this can be [seen in this codesandbox](https://codesandbox.io/s/z64y6k9xlm).

cutterbl:
Passes the `inputValue` to initial `loadOptions` request when `defaultOptions` is set to `true`.

Resolves #3160


gwyneplaine:
merged!

PR #3163, opened 2018-11-01
Disable focusable when isDisabled=true and isSearchable=false

gm0t:
When props.isSearchable is false, isDisabled property is not passed to the DummyInput control. As a result user can focus Select even though it is disabled.


driesd:
@JedWatson Can we get this merged please?

gwyneplaine:
merged, thanks @gm0t

gm0t:
@gwyneplaine thanks!

driesd:
Thanks idd! @gwyneplaine

PR #3166, opened 2018-11-01
Update All Async Examples

cutterbl:
Once PR#3161 was brought in it caused all Async documentation to fail. This fixes the issue.


gwyneplaine:
thanks @cutterbl

PR #3170, opened 2018-11-02
Docs: Remove references to theme `colors.text`

acrawford13:
## Reason for this PR
`colors.text` is no longer defined in the [theme](https://github.com/JedWatson/react-select/blob/master/src/theme.js) (removed in https://github.com/JedWatson/react-select/pull/2934/)

## Changes proposed in this PR
- Remove `text` property from the theme customisation [example](https://react-select.com/styles#overriding-the-theme).

- Remove reference to `colors.text` in `ExampleWrapper`  
(It was supposedly being used for the color of action icons on hover, but it looks fine without)
peek 2018-11-02 14-33


gwyneplaine:
thanks @acrawford13

PR #3172, opened 2018-11-02
Docs: Add theme colors to docs

acrawford13:
## Reason for this PR
- When trying to override theme colors, I had to check the source code to find out which keys to use. I think adding the keys into the documentation would make things easier for users. Also open to other suggestions :slightly_smiling_face: 

## Changes proposed in this PR
- Expose keys and default values of all `theme.colors` in documentation, in order to make it easier for users to take advantage of the `theme` prop.
- Added `ColorSample` component

## Screenshot of changes
image


gwyneplaine:
This is great, thanks @acrawford13 👍

PR #3185, opened 2018-11-07
Docs: Update example for the Provided Styles

berikiushi:
Added an example of using pseudo-classes in the provided styles.

### DIFF
--- a/docs/pages/styles/index.js
+++ b/docs/pages/styles/index.js
@@ -79,6 +79,10 @@ export default function Styles() {
         borderBottom: '1px dotted pink',
         color: state.isSelected ? 'red' : 'blue',
         padding: 20,
+
+        '&:hover': {
+          borderBottom: '1px solid pink',
+        },
       }),
       control: () => ({
         // none of react-select's styles are passed to <Control />


sunglux:
is anything stopping this from getting merged?

JedWatson:
Thanks @berikiushi, and sorry for the delayed reply.

This is actually the wrong way to achieve hover styling for most cases, options can be "selected" by either hovering with the mouse, or navigating with the keyboard (important for accessibility).

Styling hovered options is already demonstrated; the colour will change from blue to red because of this line:

```
color: state.isSelected ? 'red' : 'blue',
```

So I don't want to merge this because it's misleading to demonstrate using the `:hover` pseudo-class. But it sounds like the current docs don't explain what's going on sufficiently - maybe an inline comment or additional paragraph above the example would help?

Open to other PRs to address this concern.

jota12x:
> Thanks @berikiushi, and sorry for the delayed reply.
> 
> This is actually the wrong way to achieve hover styling for most cases, options can be "selected" by either hovering with the mouse, or navigating with the keyboard (important for accessibility).
> 
> Styling hovered options is already demonstrated; the colour will change from blue to red because of this line:
> 
> ```
> color: state.isSelected ? 'red' : 'blue',
> ```
> 
> So I don't want to merge this because it's misleading to demonstrate using the `:hover` pseudo-class. But it sounds like the current docs don't explain what's going on sufficiently - maybe an inline comment or additional paragraph above the example would help?
> 
> Open to other PRs to address this concern.

As mentioned by @JedWatson , I could replicate the hover property using the isSelected propety. Thank you for helping 👍

PR #3194, opened 2018-11-10
repairs the isOptionDisabled function.

givethemheller:
The previous implimentation set isOptionDisabled to a function via default props. The function would check if the function was defined in this.props and then call itself. It looks like an infinite loop - not sure how it wasn't breaking.
*created an internal function isOptionDisabled internal
* calls the internal function if the external prop is not set
* calls the prop if the external prop has been set

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -37,7 +37,6 @@ import {
   formatGroupLabel,
   getOptionLabel,
   getOptionValue,
-  isOptionDisabled,
 } from './builtins';
 
 import {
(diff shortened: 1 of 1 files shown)


bladey:
Hi @givethemheller,

Thanks for this pull request, we appreciate the time you've taken to contribute.

If you feel this pull request is still relevant for the latest version (v3), let us know and we'll determine if it should be reopened and brought up to date to get it merged.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our efforts towards the current major version (v3) and beyond.

PR #3195, opened 2018-11-10
Adds a primitive typing to String in the compareOption function.

givethemheller:
This allows the compare function to work on any typing that might be provided to the selectOptions interface.

number typing fails in typescript

### DIFF
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -36,10 +36,10 @@ export type CreatableProps = {
 export type Props = SelectProps & CreatableProps;
 
 const compareOption = (inputValue, option) => {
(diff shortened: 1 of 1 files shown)


cutterbl:
@givethemheller Looks like you missed the unit tests on this PR

bladey:
Hi @givethemheller,

Thanks for this pull request.

If you feel this pull request is still relevant for the latest version (v3), let us know and we'll determine if it should be reopened and brought up to date to get it merged.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our efforts towards the current major version (v3) and beyond.

PR #3205, opened 2018-11-14
Menu: constrain to a maximum of maxHeight

benmosher:
When automatically flipping, the menu constraint math may result in a "constrained" height that exceeds the original max height.

This change uses `Math.min` to choose the constrained height only when it is _less_ than the original `maxHeight`.

I made these changes with the web UI so I have not had a chance yet to run the tests/lints, will do that when I get a chance.


gwyneplaine:
thanks for this @benmosher  merged

PR #3216, opened 2018-11-20
Respect `isClearable` and `isMulti` in `backspaceRemovesValue`

### LINKED ISSUE #1215: Pressing backspace when item is selected already
Hello, i think i have found a bug. Its important to have some **option selected first** to reproduce the behavior!

My component:

``` javascript
<Select
options={datasource}
value={this.state.value}
    onChange={value=>{
        let isArray = value == null ? false : value.constructor == Array
        this.setState({value: isArray ? null : value})
    }}
/>
```

Scenario A:
1. select an option ( onChange param `value == { value: X, label: Y}` )
2. click clearing cross ( onChange param `value == null` ) - **which is expected result**

Scenario B:
1. select an option ( onChange param `value == { value: X, label: Y}` )
2. press backspace ( onChange param `value == []`) - **expected** `value == null`

Explanation:
I mean, when you click clearing cross OR press backspace when the Select has **MULTI** enabled, `value` param is the same in both cases - an empty array. Thats correct, the result should be 0-N items, an **array**.

When you do this with **SINGLE**-value Select, clicking cross sets `value` param to null BUT pressing backspace sets it to empty array instead of null, which seems inconsistent and a bit confusing to me. Thats why there is the isArray condition in my onChange method. The result should be **null/object**, not an array.

Is this a bug or not?
Thanks for your answer in advance.

Good job & keep it up!
HK

### LINKED ISSUE #1344: I have multi=false but react-select emits a change event with an array for the value.
I have multi={false}. But sometimes, react-select emits a change event with an empty array for the value.

I am using:
AsyncCreatable
react-select v1.0.0-rc.2
Chrome Version 54.0.2840.71 (64-bit)
Mac OS X

Here is the stack trace from the Chrome console:

Jed emitted an array []
Uncaught Error: IllegalState(…)
	CompanyPicker._this.onChange			@	CompanyPicker.jsx?9896:64
	setValue								@	Select.js?407b:654
	popValue								@	Select.js?407b:689
	handleKeyDown							@	Select.js?407b:509
	ReactErrorUtils.invokeGuardedCallback	@	ReactErrorUtils.js?6b31:70
	executeDispatch							@	EventPluginUtils.js?0958:89
	executeDispatchesInOrder				@	EventPluginUtils.js?0958:112
	executeDispatchesAndRelease				@	EventPluginHub.js?c6b1:44
	executeDispatchesAndReleaseTopLevel		@	EventPluginHub.js?c6b1:55
	forEachAccumulated						@	forEachAccumulated.js?b898:25
	processEventQueue						@	EventPluginHub.js?c6b1:231
	runEventQueueInBatch					@	ReactEventEmitterMixin.js?8a20:18
	handleTopLevel							@	ReactEventEmitterMixin.js?8a20:29
	handleTopLevelImpl						@	ReactEventListener.js?2365:73
	perform									@	Transaction.js?6dff:138
	batchedUpdates							@	ReactDefaultBatchingStrategy.js?ef70:63
	batchedUpdates							@	ReactUpdates.js?ce09:98
	dispatchEvent							@	ReactEventListener.js?2365:150

### LINKED ISSUE #1560: Select with `clearable={false}` can still be cleared with backspace
Here's a Plunker:
https://plnkr.co/edit/ZaFyPW0zGqcf2oWeVnaL?p=preview

Note that although it has `clearable={false}`, you can still clear the value by focusing the field and pressing Backspace. To disable this behavior, you have to also explicitly set `deleteRemoves={false}`. 

This is surprising to me; I would expect that `clearable={false}` would mean that you can never get a null ([or empty array, to be more precise](https://github.com/JedWatson/react-select/issues/1215)) value from the field.

Maybe `deleteRemoves` should default to the value of `clearable`? Or require _both_ `deleteRemoves` and `clearable` to be true for the backspace behavior?

### LINKED ISSUE #2682: [v2] Backspace on single select results in [] rather than {}

### LINKED ISSUE #2804: onChange return value is inconsistent
https://codesandbox.io/s/jp38mx52x9

Select an item using the component.
Clear the value using the clear (`X`) button.
Notice the console logs (`null`).

Select an item using the component.
Clear the value using the backspace button.
Notice the console logs (`[]`).

I believe as a single-select component the value returned should either always be `null` or an empty object `{}`, while in a multi-select component the value returned should either always be `null` or an empty array `[]`

### LINKED ISSUE #3129: Erasing value issue
when I erase the value using backspace I'm getting empty array as the value, I think it would be better if I receive the same value as I get for isClearable (I think it's null)

IanVS:
Currently, `backspaceRemovesValue` seems to assume that the select component is both clearable, and a multi-select.  This PR will change the following behavior:

- The `clear` action will be used when `isMulti` is `false.
- Nothing will happen when backspace or delete is pressed if `isMulti` is false _and_ `isClearable` is also `false`.
- When `isMulti` is true, this PR changes no behavior, and `isClearable` still has no impact.  I'm not sure if it should or not, but wanted to limit the scope of this PR to only non-multi components.

Fixes #2682
Fixes #1560

Seemingly duplicates:
Closes #2804
Closes #3129
Closes #1344
Closes #1215

### DIFF
--- a/cypress/integration/select_spec.js
+++ b/cypress/integration/select_spec.js
@@ -133,6 +133,17 @@ describe('New Select', function() {
             .get(selector.noOptionsValue)
             .should('contain', 'No options');
         });
+
+        it('Should not clear the value when backspace is pressed ' + view, function() {
+          cy
+            .get(selector.singleBasicSelect)
+            .click()
+            .find('input')
+            .type('{backspace}', { force: true })
+            .get(selector.singleBasicSelect)
+            .find(selector.placeholder)
+            .should('not.be.visible');
+        });
       });
     });
 
--- a/src/Select.js
+++ b/src/Select.js
@@ -1137,7 +1137,7 @@ export default class Select extends Component<Props, State> {
           if (!backspaceRemovesValue) return;
           if (isMulti) {
             this.popValue();
-          } else {
+          } else if (isClearable) {
             this.clearValue();
           }
         }
--- a/src/Select.js
+++ b/src/Select.js
@@ -1135,7 +1135,11 @@ export default class Select extends Component<Props, State> {
           this.removeValue(focusedValue);
         } else {
           if (!backspaceRemovesValue) return;
-          this.popValue();
+          if (isMulti) {
+            this.popValue();
+          } else {
+            this.clearValue();
+          }
         }
         break;
       case 'Tab':
--- a/src/__tests__/Select.test.js
+++ b/src/__tests__/Select.test.js
@@ -1416,6 +1416,40 @@ test('should not call onChange on hitting backspace even when backspaceRemovesVa
   expect(onChangeSpy).not.toHaveBeenCalled();
 });
 
+test('should call onChange with `null` on hitting backspace when backspaceRemovesValue is true and isMulti is false', () => {
+  let onChangeSpy = jest.fn();
+  let selectWrapper = mount(
+    <Select
+      {...BASIC_PROPS}
+      backspaceRemovesValue
+      isClearable
+      isMulti={false}
+      onChange={onChangeSpy}
+    />
+  );
+  selectWrapper
+    .find(Control)
+    .simulate('keyDown', { keyCode: 8, key: 'Backspace' });
+  expect(onChangeSpy).toHaveBeenCalledWith(null, { action: 'clear' });
+});
+
+test('should call onChange with an array on hitting backspace when backspaceRemovesValue is true and isMulti is true', () => {
+  let onChangeSpy = jest.fn();
+  let selectWrapper = mount(
+    <Select
+      {...BASIC_PROPS}
+      backspaceRemovesValue
+      isClearable
+      isMulti
+      onChange={onChangeSpy}
+    />
+  );
+  selectWrapper
+    .find(Control)
+    .simulate('keyDown', { keyCode: 8, key: 'Backspace' });
+  expect(onChangeSpy).toHaveBeenCalledWith([], { action: 'pop-value' });
+});
+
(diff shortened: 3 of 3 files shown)


IanVS:
I'm having trouble updating the e2e tests, because of the way they're structured.  I actually think that some of them are actually incorrect.  I may need to submit a separate PR to clean those up a bit, first.  😞

IanVS:
Done.  Thanks.

IanVS:
wait one sec, actually

Ecksters:
Would be nice if when a user explicitly sets `backspaceRemovesValue` if it would ignore `isClearable`

IanVS:
I'd suggest that the setting whether an input is clearable or not should not be used to allow some kinds of clearing and not others.  This resulted in a lot of confusion amongst users in the past.

Instead, I'd suggest that if you want to not include a clearable icon, to use:

```
components={
	ClearIndicator: null,
}
```

That way, no clear icon is rendered, but the form is still marked as clearable, which it is.

kstratis:
@IanVS That trick was awesome. Was looking everywhere for exactly that! Thanks.

Issue #2600, opened 2018-05-14
[V2] TODOs


gwyneplaine:
## Before RC 

- [x]  screen-reader accessibility with Aria Live Regions.
- [x] implementing cx function to support a base classname. 
- [ ] select v2 as monorepo.
- [x] examples and docs improvements. 
- [ ] cypress tests refactor.

IanVS:
Can you explain what you intend by `cypress tests refactor.`?  I'm trying to work on a PR, but struggling with the way the cypress tests are structured, they seem very brittle.  I have my own ideas of how to improve them, and would be willing to submit a PR, if there's any chance it can be merged.

naheeyahh:
Will there be plans to make V2 selenium testable? Because "Select-option" and "Select-control" are no longer accessible by the web driver it is impossible to test V2 with selenium.

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our efforts towards the current major version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue  is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

PR #3218, opened 2018-11-21
Refactor and expand end-to-end tests

IanVS:
Refs: #2600

This is a start towards adding a bit of organization into the cypress tests, cleaning up the selectors used, and making them less brittle.  So far only the single-select components are addressed, but if this pattern is approved, I can take the same approach for the mutli-select component.

What I tried to do here is:
- Encapsulate tests down to a single Select component on the test page.  Previously there were tests asserting on the total number of components on the page, which made it difficult to add any further components without breaking and fixing unrelated tests.
- Added `describe` blocks for the various types of Select components on the test page, and moved relevant test cases inside them.  This is mostly just for improved organization and to make it easier to spot any missing tests that we should be making.
- Break apart selectors into those targeting an id of a particular type of select component on the test page and those targeting specific _types_ of sub-components inside the main `Select`.  This, combined with the use of `.get().find()`, allows us to have a smaller number of overall selectors, and still combine them in a way that targets specific DOM elements for the component under test.
- Ensure that all of the tests pass both when the cypress browser is focused as well as when it is backgrounded.  I don't know why, but some of the tests were passing only when it was backgrounded, and would fail when running the tests live.  I _think_ it was because they were faulty tests, but I'm not entirely sure.  I'll try to call out tests I've removed below.
- Added a new clearable single-select component, along with some tests that clearing works as expected.

I haven't worked with cypress before, so I'd love to hear if there are better approaches than this, but hopefully it's at least a step in the right direction.


cypress/fixtures/selectors.json
```diff
@@ -1,30 +1,30 @@
...
-  "firstMultiValueRemove":
-    "#multi-select .react-select__multi-value__remove:first",
-  "groupColor": "#cypress-container-single #cypress-single-grouped .react-select__group",
-  "menuGrouped": "#grouped-options-single .react-select__menu",
-  "menuMulti": "#multi-select .react-select__menu",
+  "singleBasic": "#cypress-single",
+  "singleBasicSelect": "#basic-select-single",
+  "singleClearable": "#cypress-single-clearable",
+  "singleClearableSelect": "#clearable-select-single",
+  "singleGroupedSelect": "#grouped-options-single",
```
IanVS: This section is for the various types of single select components (and their containers) that we'll want to make assertions about.

cypress/fixtures/selectors.json
```diff
@@ -1,30 +1,30 @@
...
+  "checkboxEscapeClearsValue": ".escape-clears-value-checkbox",
+
+  "groupHeading": ".react-select__group-heading",
+  "indicatorClear": ".react-select__clear-indicator",
+  "indicatorDropdown": ".react-select__dropdown-indicator",
+  "menu": ".react-select__menu",
   "menuOption": "[role='option']",
+  "noOptionsValue": ".react-select__menu-notice--no-options",
+  "placeholder": ".react-select__placeholder",
+  "singleValue": ".react-select__single-value",
```
IanVS: These are sub-components that might be found inside any of the components under test.  Typically we'll want to `.find()` them after getting one of the main components by its id.

cypress/fixtures/selectors.json
```diff
@@ -1,30 +1,30 @@
...
+
   "menuSingle": "#basic-select-single .react-select__menu",
+  "singleSelectSingleInput": "#react-select-basic-select-single-input",
+  "toggleMenuSingle": "#basic-select-single .react-select__dropdown-indicator",
```
IanVS: These are technically some old ones that I think should go away, but I kept them around because they're used in a `skip`ed test.  If we remove that test, we can remove these three selectors.

cypress/fixtures/selectors.json
```diff
@@ -1,30 +1,30 @@
...
-  "singleSelectFirstValue": "#basic-select-single .react-select__single-value",
-  "singleSelectGroupedInput":
-    "#grouped-options-single .react-select__input input",
-  "singleSelectSingleInput": "#react-select-basic-select-single-input",
-  "toggleMenus": ".react-select__dropdown-indicator",
-  "toggleMenuGrouped":
-    "#grouped-options-single .react-select__dropdown-indicator",
-  "toggleMenuMulti": "#multi-select .react-select__dropdown-indicator",
-  "toggleMenuSingle": "#basic-select-single .react-select__dropdown-indicator"
+  "toggleMenuMulti": "#multi-select .react-select__dropdown-indicator"
```
IanVS: This last group are for the multi-select component, which I haven't changed so far.

cypress/integration/select_spec.js
```diff
@@ -1,6 +1,7 @@
 const selector = require('../fixtures/selectors.json');
 
 const viewport = ['macbook-15', 'iphone-6'];
+// const viewport = ['macbook-15'];
```
IanVS: Whoops, I'll remove this.

cypress/integration/select_spec.js
```diff
@@ -10,89 +11,18 @@ describe('New Select', function() {
...
-          .get(selector.menuGrouped)
-          .should('be.visible')
-          .get(selector.groupColor)
-          .should('be.visible');
-      });
-      it(
+
+      // This test seems to fail when cypress tab is focused.
+      // Also, manual testing does not confirm the desired behavior
+      it.skip(
```
IanVS: It's not entirely clear to me what this test is trying to assert, and it doesn't seem to work properly.  I'd like to remove it, if that's okay with the maintainers.  Otherwise, we should probably fix it so that it passes when the cypress browser is focused during the tests.

gwyneplaine: I believe the intention of this is to ensure that on mobile the menu isn't triggered on a drag event, if you can fix it within this PR that would be great, if not we can leave it skipped for now and come back to it in a subsequent body of work

gwyneplaine:
this looks great, thanks @IanVS, left a comment regarding the skipped test; that aside this is looking much more maintainable than the previous fixture

cypress/integration/select_spec.js
IanVS: Yeah to be honest I'm really not sure how to get it working, and would love if we could leave it skipped for now and allow someone with more knowledge of testing mobile behavior in cypress to step in and get it working correctly.

IanVS:
Thanks for the review, @gwyneplaine!

If you're happy with this, do you think it could be merged so that https://github.com/JedWatson/react-select/pull/3216 can also get in?  I'm happy to tackle multi-select in a separate PR if we like this general approach.

Issue #2862, opened 2018-07-27
[v2] onMenuScrollToBottom does not fire if dropdown is not scrollable on mount


pyreta:
@JedWatson I wasn't sure how to test but you can see a [codesandbox example here](https://codesandbox.io/s/1vq0l1p2l4) where opening the dropdown with options already loaded allows for the `onMenuScrollToBottom` callback to log "bottom" in the console.  However if you clear the options before opening the dropdown, the callback will not fire after options have loaded.

I've submitted [this PR](https://github.com/JedWatson/react-select/pull/2861) as a fix.

Thanks!

vtaits:
```
// src/internal/ScrollCaptor.js

  startListening(el: HTMLElement) {
    // bail early if no scroll available
    if (el.scrollHeight <= el.clientHeight) return; // <------- listening stops here

    // all the if statements are to appease Flow 😢
    if (typeof el.addEventListener === 'function') {
      el.addEventListener('wheel', this.onWheel, false);
    }
    if (typeof el.addEventListener === 'function') {
      el.addEventListener('touchstart', this.onTouchStart, false);
    }
    if (typeof el.addEventListener === 'function') {
      el.addEventListener('touchmove', this.onTouchMove, false);
    }
  }

```

What about add prop for allow listen scrolls always?

KarolinaPru:
Hi, is @pyreta 's change going to be merged anytime soon? I'm having the same issues.

macobo:
@JedWatson What's the next step to resolve this bug?

There's two existing PRs addressing this usecase, neither with clear feedback on what needs to be done differently to accept the PRs:
- https://github.com/JedWatson/react-select/pull/3219
- https://github.com/JedWatson/react-select/pull/2861

bogdanpetru:
Any update on this issue? 

To me this fix seems ok: https://github.com/JedWatson/react-select/pull/3219

We have this issue with paginated select component. If the select is opened before the first page is loaded the pagination is broken.

The fix I've used it is kinda hacky and I would like to get rid of it. I change the key on select-menu when first page is loaded. This makes the menu remount, menu as scroll and it listens to scroll events.

I am open to contribute to fix this issue. 

Thank you!

<img width="831" alt="Screenshot 2019-06-18 at 11 05 38" src="https://user-images.githubusercontent.com/6267231/59664405-aed9d700-91b9-11e9-9e13-60c40abf2d1d.png">

ebonow:
[...]

PR #3219, opened 2018-11-21
add isMenuDynamic prop

vtaits:
Workaround for #2862


Pragatheesh:
Any update on the PR? Will this be merged any sooner?

ebonow:
Greetings @vtaits  and my apologies for such a long wait until receiving a response to this PR.

After reviewing the code, it seems that this attempts to patch the behavior by simply providing an escape hatch prop. This feels very much like a hack to work around the issue as opposed to resolving the core issue of the scrollTarget ref being null at the time of mount.

I added a reproducible codesandbox demo to illustrate the issue as well as some comments about the root cause and some solutions. Adding a prop that will be deprecated once the issue is fixed would add confusion and complexity so it would be in the best long term interest to close this PR without merging it. 

Myself, the team and the community appreciate any and all efforts to improve react-select and again thank you for taking the time to contribute. Bonus thank you for clearly taking the time to run the unit tests.

PR #3245, opened 2018-11-30
Add a line to the docs homepage explaining what Creatable is used for

justingrant:
I wasted time trying to figure out how to enable users to add new options before realizing that this capability was already there in Creatable, but the docs didn't actually explain what Creatable was. This PR adds a line to the docs to explain.


gwyneplaine:
manually rebased and merged, thanks @justingrant

PR #3253, opened 2018-12-05
fix: update broken tests

theneva:
Looks like these tests were broken in 6079e9e77a261b3f239b48530fda65e450cda423—I assume they were meant to be updated.

BTW, the CircleCI build is currently failing with:

```
Build-agent version 0.1.1073-1f69f340 (2018-11-20T18:07:03+0000)
Starting container cypress/base:8
  image is cached as cypress/base:8, but refreshing...

Error response from daemon: unauthorized: authentication required
```


theneva:
_nudge_ :smile:

This seems to be blocking #3260 now, and it burned me again when I tried to submit another fix.

Would someone please have a look at this PR? :heart:

gwyneplaine:
Thanks for this @theneva, merged

PR #3259, opened 2018-12-06
Pass selectProps to Input component

maxmarchuk:
I ran into an issue with this a few months when attempting to create a custom `Input` component, and not having any way to pass props to that custom component. I've seen a few mentions of this in other issues and thought I'd just create a PR for it.

I noticed that a handful of other components don't have `selectProps`, and was curious if there is a reason for not passing `selectProps` to every component that can be customized.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -1381,7 +1381,7 @@ export default class Select extends Component<Props, State> {
       'aria-labelledby': this.props['aria-labelledby'],
     };
 
-    const { cx, theme } = this.commonProps;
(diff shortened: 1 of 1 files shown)


gwyneplaine:
thanks for this @maxmarchuk 
I've rebased master ontop of your changes in this branch here 
https://github.com/JedWatson/react-select/pull/3303

It also includes a similar solution for GroupHeading which is the other main component that hasn't been passed selectProps. 

Closing this PR  in lieu of this, thanks again!

PR #3272, opened 2018-12-13
Use isOptionSelected to find selectedIndex

StefanWerW:
The selectedIndex is not always consistent when using Array.indexOf(). indexOf() will not work correctly when the isOptionSelected is overriten, or the selectValue is a clone of the Option object and is not in Options.


GustavoEdinger:
+1

bladey:
Hi @StefanWerW,

In an effort to sustain the `react-select` project going forward, we're cleaning up and closing old pull requests.

However, if you feel this issue is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

PR #3273, opened 2018-12-13
Add aria-describedby attribute

### LINKED ISSUE #1570: Adding support for aria-describedby attribute

sreahard:
Added aria-describedby attribute support in order to link elements to the react-select select input such as help text and validation errors.

The prop was added with pr #1571 but somehow got removed in the current version. This adds it back.

Per the upgrading docs:

`aria-describedby | unchanged`

Fixes #1570


DBjelovuk:
Still not able to set `aria-describedby`. 🙁 Has no one else looked into this?

bladey:
Hi @sreahard, thanks for your contribution, I've since taken your changes and updated them to support master and the new test framework, see PR #4082.

PR #3275, opened 2018-12-14
fixing the toLowerCase error

tavareshenrique:
Correction suggestion for file **creatable.js**.
Converting the fields to String before transforming into LowerCase eliminates the previous errors.


gwyneplaine:
thanks for this @tavareshenrique, rebased and merged into master

PR #2767, opened 2018-07-03
[v2] Pressing <Enter> in IME mode should not select a option

shamabe:
In composing mode, enter key is for selecting a suggest of conversion.
So that, the behavior of enter key is collision in "react-select" with IME.

Like this...
react-select-before

It is necessary to check whether it is in composing-mode or not.
Following is the preferred behavior.

react-select-after

PR #3296, opened 2018-12-20
Pressing <Tab> in IME mode should not enter a option

### LINKED ISSUE #3295: [v2] Creatable Multiselect creates value every tab push
Version: 2.x

## Issue 

I found that Creatable Multiselect has bug on IME input.

When I push tab after input words, it is inputed into input field every time I press a tab..

※ It does not happen in V1

## Refs

- https://react-select.com/creatable
`Creatable Multiselect Example`

yshr446:
Fix the tab behavior with IME.

Fixes #3295 

## ref

https://github.com/JedWatson/react-select/pull/2767

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -1158,6 +1158,8 @@ export default class Select extends Component<Props, State> {
         }
         break;
       case 'Tab':
+        if (isComposing) return;
+
         if (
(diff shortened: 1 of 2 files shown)


gwyneplaine:
this is great, thanks @yshr446

Issue #3125, opened 2018-10-18
[V2] Input commonProps / selectProps


samueldepooter:
Hi!

I'm currently developing a custom `Creatable`. With it I'm trying to maintain the placeholder on the input when I have options selected.

- Current behaviour: placeholder is **not** shown when options are selected
- Wanted behaviour: placeholder is persisted

I tried doing it with a custom `Input` element but noticed that the element is not receiving `selectProps` (from `commonProps`). [This line is where Input is created & as you can see it's not passed.](https://github.com/JedWatson/react-select/blob/18219f5271705ef150ad16177db43f0c44f0413d/src/Select.js#L1317)

I was wondering what's the reasoning behind it? I could make a PR to fix this.

Mentioned before in #2680

gwyneplaine:
thanks @samueldepooter , this issue is now resolved in this PR https://github.com/JedWatson/react-select/pull/3303 released as part of v2.2.0

Issue #3052, opened 2018-09-19
Input component does not have selectProps as a prop


amirnafei:
All the react-select components (like ValueContainer, Menu, ...) have selectProps included among their props (as it is described in the documentation) while this is not happening for the Input component.

gwyneplaine:
thanks for raising this @amirnafei, Input component has now been updated to be passed selectProps as of v2.2.0. 
Closing this issue now

thefran1412:
Hi! I'm having the same problem in version `3.1.0`. This is my code:

``` javascript
const styles = {
	input: (provided, state) => {
		console.log(state.selectProps) // this outputs: undefined
		return {
			...provided,
			color: state.selectProps.txtColor
		}
	},
}
```

trevorglick:
just ran into this, we're on react-select 3.0.8 this is what the state object shows in the input api:
`aria-autocomplete: "list"
aria-label: undefined
aria-labelledby: undefined
autoCapitalize: "none"
autoComplete: "off"
autoCorrect: "off"
id: "react-select-19-input"
onBlur: ƒ (event)
onChange: ƒ (event)
onFocus: ƒ (event)
spellCheck: "false"
tabIndex: "0"
theme: {borderRadius: 4, colors: {…}, spacing: {…}}
type: "text"
value: ""`

cchoiniere:
Can this issue be re-opened? Looks like as of 3.0.X that state no longer has selectProps. This makes it difficult to control styles as needed.

tavurth:
I'm having an issue with this now, was wondering for the last hour why my `react-hook-form` wasn't working, turned out to be this component.

tavurth:
I'm having an issue with this now, was wondering for the last hour why my `react-hook-form` wasn't working, turned out to be this component.

stcalica:
selectProps does not work for my component when passing into the custom styles object. 
In my case, I'll need to create different variants just for width alone.

markovone:
looks like `styles.menuPortal` cant access selectProps either

tavurth:
What's the reason this issue is closed?

ebonow:
Greetings, 

It was closed because it has been merged into master. I would have to do a bit more digging to find where it was added in, but it was a recent addition such that selectProps should be accessible to the Input component. 

```jsx
  const styles = {
    input: (css, state) => ({ ...css, color: state.selectProps.color })
  };

  return (
    <>
[...]

Issue #2680, opened 2018-06-05
[v2] common props for <Input />


mike1808:
Hi all! I was migrating my old code to the v2 and it is amazing. However, I have several obstacles making to hack and patch.

What is the motivation not to include `commonProps` to `<Input>` component?

marcselman:
I noticed the same thing.

samueldepooter:
Looking for the same thing! No update on this yet? Can take a look to create a PR for it.

mike1808:
@samueldepooter I didn't create PR because I thought the author had some intentaions for it, but I guess, they just didn't need that for that time.

samueldepooter:
> @samueldepooter I didn't create PR because I thought the author had some intentaions for it, but I guess, they just didn't need that for that time.

@mike1808 yes it's a quick fix. Not sure why you wouldn't want these props present. Hopefully it can be answered soon 👍

gwyneplaine:
@mike1808 we've passed a majority of the needed props from common props into the Input component now.
selectProps being the latest to be added to the set of props to be passed down as of v2.2.0

Closing this issue now

Feel free to reopen if this is still an issue.

PR #3303, opened 2018-12-28
Maxmarchuk add select props to input

gwyneplaine:
Rebasing master onto #3259 as well as stripping selectProps from internal input DOM element and passing selectProps to GroupHeading component.


maxmarchuk:
Thanks @gwyneplaine.
I believe issues #3125, #2680, and #3052 are related and can be closed

PR #3308, opened 2019-01-01
Fix title of Advanced docs page

nafg:


gwyneplaine:
thanks for this @nafg, merged

Issue #3269, opened 2018-12-12
Clearing Input does not focus the option defined in focusedOption prop.


mike391:
When clearing the input of react-select, the `focused option` is still the previously selected value, it would make more sense to be the default `focusedOption` or the first option in the list instead.

vnallamhawk:
is this issue still open ?

mike391:
Update: PR Opened: [#3314](https://github.com/JedWatson/react-select/pull/3314)

lukebennett88:
Hi @mike391
I was unable to reproduce this, but I think it may have been resolved already.
Please re-open if needed.

PR #3314, opened 2019-01-02
Allow 'focusOption' to define menu item to focus when input is cleared

mike391:
Adds the behaviour to auto focus the 'focusedOption' (if defined in props) when the input field if cleared. 

Issue:
[#3269](https://github.com/JedWatson/react-select/issues/3269)


lukebennett88:
Hi @mike391
I was unable to reproduce this, but I think it may have been resolved already.
Please re-open if needed.

PR #3318, opened 2019-01-03
logging updated state in readme example

dallin-christensen:
Just a suggestion, but the readme example makes a bit more sense if you're logging the updated state after setState, rather than the value used to update the state. This addition would fit a more of the unidirectional data flow paradigm that React is so supportive of. A side note, react-select is awesome 👍🙂 great work


JedWatson:
Thanks @dallin-christensen, I'm on board with this!

PR #3319, opened 2019-01-03
Fix for manually setting 'isLoading' on async select box

### LINKED ISSUE #2986: [v2] No way to manually set isLoading on async select box
So I'm using the isLoading prop on the normal Select to show a loader in the select box after you select something while it is fetching data to populate some fields below.

Somewhere else in the app I am trying to do the same with the async select component and it doesn't seem to work, I think it should be possible to override the isLoading prop.

mnpreetkaur:
Closes #2986

### DIFF
--- a/src/Async.js
+++ b/src/Async.js
@@ -148,7 +148,7 @@ export const makeAsyncSelect = (SelectComponent: ComponentType<*>) =>
       return inputValue;
     };
     render() {
-      const { loadOptions, ...props } = this.props;
+      const { loadOptions, isLoading: isLoadingProp, ...props } = this.props;
       const {
         defaultOptions,
         inputValue,
@@ -169,7 +169,7 @@ export const makeAsyncSelect = (SelectComponent: ComponentType<*>) =>
             this.select = ref;
           }}
           options={options}
-          isLoading={isLoading}
+          isLoading={(isLoadingProp === undefined) ? isLoading : isLoadingProp}
           onInputChange={this.handleInputChange}
         />
       );


orrybaram:
Would love to get this merged in!!

danichim:
Would love to get this merged in!!

veiko:
@gwyneplaine Is there anything holding this up?

jacarma:
I think it makes more sense to show the loading indicator when the `isLoading`  prop is true **OR** when the `state.isLoading`is true 
```isLoading={isLoadingProp || isLoading}```
So it is shown when the Async select is waiting for the promises and when the additional requests the parent component may be doing

JonathanGuo:
I think the commit missed the type definition and a default value for the props.

I am not sure if this is the reason the PR is not merged yet. But it would be nice to see the type is fixed and PR is merged.

It's gonna be another piece of work to update the async definition for @types/react-select too.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-select/lib/Async.d.ts

davidholyko:
I'm working on a project and was wondering if anyone can help move this through the merging process. What needs to be done to get it approved?

JedWatson:
Thanks for the PR @mnpreetkaur, closing in favour of #3690 (merged)

orrybaram:
@JedWatson I was at your talk today at react conf and I sincerely apologise for my above comment. Thanks for all your hard work 😅.

alfechner:
Do I get something wrong? Setting `isLoading` on `AsyncSelect` to true should make the component display the loading indicator, right? 

Did I miss something, it doesn't work that way in my [codesandbox](https://codesandbox.io/s/priceless-perlman-wdsr7?fontsize=14&hidenavigation=1&theme=dark).

PR #3320, opened 2019-01-04
Update Expiramental.js docs to include emotion

theskillwithin:
fix for 2.2.0

thanks @gwyneplaine


gwyneplaine:
@theskillwithin this is no longer necessary with the work in #3346
thanks for this though, always great to have contributions

Issue #3317, opened 2019-01-03
SSR 2.2.0 - createEmotion


marudor:
Would it be possible to get the option to pass a custom emotion instance to Select?
Currently you will always create one - and it's hardly possible to get to that instance to ServerRender styles.
=> I can't render those styles on the server anymore.

theskillwithin:
I found when upgrading to 2.2.0 i got error with emotion.   emotion.css did not exist.  so i reverted back to 2.1.2   (idk if related but worth noting just in case)

gwyneplaine:
@theskillwithin it's because within Group, you're not passing the emotion prop down to the Heading component. Doing this resolves the error 
https://codesandbox.io/s/5zk5j74k4k

gwyneplaine:
@marudor https://github.com/JedWatson/react-select/pull/3321
the issue mentioned above should be resolved in this PR here, moving up to emotion 10 to enable zero config SSR, and an easier time passing through contextual values like nonces.

PR #3260, opened 2018-12-07
Add CSP Nonce option to Select component

Avaq:
## Motivation

To close #2917

## Changes

```diff
 <Select
+  nonce={MY_SECRET_NONCE}
 >
```

- The `emotion` dependency was replaced by `create-emotion`.
- An instance of `emotion` is created in the Select constructor, passing in the `nonce` if the user provided it.
- The `emotion` instance is passed to all components via `commonProps`.
- All components were updated to make use of the custom `emotion` instance instead of the global one.

## Checklist

- [x] ensure all the examples work correctly after your change
- [x] ensure you do not introduce any new [linting] errors or warnings
- [x] ensure that you do not introduce any new type errors
    <details><summary>I was unable to run <code>yarn flow</code>:</summary>

    ```console
    $ yarn flow
    yarn run v1.12.3
    $ ./react-select/node_modules/.bin/flow
    events.js:167
          throw er; // Unhandled 'error' event
          ^

    Error: spawn ./react-select/node_modules/flow-bin/flow-linux64-v0.72.0/flow ENOENT
    ```
    </details>
- [x] make sure there's an issue open for any work you take on
- [x] make sure you do not add regressions
    <details><summary>Two tests were failing before, and after the changes I made (see #3253):</summary>
    
    ```console
    ● should call onChange with `null` on hitting backspace when backspaceRemovesValue is true and isMulti is false

      expect(jest.fn()).toHaveBeenCalledWith(expected)

      Expected mock function to have been called with:
        {"action": "clear"}
      as argument 2, but it was called with
        {"action": "clear", "name": "test-input-name"}.

      Difference:

      - Expected
      + Received

        Object {
          "action": "clear",
      +   "name": "test-input-name",
        }

        1433 |     .find(Control)
        1434 |     .simulate('keyDown', { keyCode: 8, key: 'Backspace' });
      > 1435 |   expect(onChangeSpy).toHaveBeenCalledWith(null, { action: 'clear' });
             |                       ^
        1436 | });
        1437 | 
        1438 | test('should call onChange with an array on hitting backspace when backspaceRemovesValue is true and isMulti is true', () => {

        at Object.toHaveBeenCalledWith (src/__tests__/Select.test.js:1435:23)

    ● should call onChange with an array on hitting backspace when backspaceRemovesValue is true and isMulti is true

      expect(jest.fn()).toHaveBeenCalledWith(expected)

      Expected mock function to have been called with:
        {"action": "pop-value"}
      as argument 2, but it was called with
[...]

PR #3321, opened 2019-01-04
emotion 9 --> emotion 10

gwyneplaine:
Refactor our usage of create-emotion as of v2.2.0 to emotion 10, and reverting the addition of the custom emotion instance passed into each component. 

Please see the following PR for context on the create-emotion change:  #3260 
This is also an indirect response to the following issue #3317  caused by the above PR 😄 

## There's a few valid reasons for the upgrade:
- Zero configuration SSR is very appealing, especially for users who are using css-in-js solutions other than emotion. 
- nonce support for CSP is trivial, and doesn't involve creating a custom emotion instance. 
- Users on latest version of emotion, don't need to worry about pulling in multiple versions of emotion because of react-select. 
- Not passing in a custom emotion instance to every component that needs styling is actually kind of ok, it means the api for our components aren't bound to a specific css-in-js solution, making future changes to our styling opinions less obtrusive.

## Possible arguments against emotion 10: 
- Support for react 15 users who aren't able to use the new context api
- This could be a breaking change for users using the components api ** (see rebuttal below) 

** this is only true for users who are completely rewriting components, and doing their own style composition. Users on emotion 10 who are doing this should be familiar with this syntax already, users who are using any other css-in-js solution should have the necessary tools to create functioning custom components without having to use emotion. 

## Things I want to validate before this PR goes through: 
- That support for react 15 is a non-issue @JedWatson 
- That we're leveraging the correct patterns in our use of emotion 10 @mitchellhamilton
- The assertion that while this is a non-trivial change to our internal css-in-js implementation, it is ultimately a non-breaking change for users of the custom components api.


package.json
```diff
@@ -48,7 +50,7 @@
     "css-loader": "^0.28.7",
     "cypress": "^1.4.1",
     "dotenv": "^5.0.1",
-    "enzyme": "^3.3.0",
```
gwyneplaine: upgrade enzyme so it plays well with emotion 10 and new context api

src/Select.js
```diff
@@ -1,9 +1,9 @@
 // @flow
 
 import React, { Component, type ElementRef, type Node } from 'react';
-
 import memoizeOne from 'memoize-one';
-import createEmotion, { type Emotion } from 'create-emotion';
```
gwyneplaine: removed createEmotion in favor of emotion 10s CacheProvider and createCache for adding nonce support.

src/Select.js
```diff
@@ -1703,7 +1696,7 @@ export default class Select extends Component<Props, State> {
               onTopArrive={onMenuScrollToTop}
               onBottomArrive={onMenuScrollToBottom}
             >
-              <ScrollBlock emotion={this.emotion} isEnabled={menuShouldBlockScroll}>
```
gwyneplaine: we don't want to be doing this anymore, it locks us into emotion as an intrinsic  part of our components API, while this is true of our own implementations, this should ultimately not be true for power users.

src/components/Control.js
```diff
@@ -53,20 +54,24 @@ export const css = ({
...
-      }, className)}
-      {...innerProps}
-    >
-      {children}
-    </div>
+    <ClassNames>
+      {({ css }) => (
+        <div
+          ref={innerRef}
+          className={cx(css(getStyles('control', props)), {
```
gwyneplaine: Here we're using the ClassNames fn from emotion 10 to access the css method so we can pass the processed style objects from css(getStyles()) to our custom cx implementation. 

@mitchellhamilton I just wanted to sanity check the fact that we're using a custom cx implementation her instead of the one provided to us by ClassNames is ok.

src/components/indicators.js
```diff
@@ -1,29 +1,33 @@
...
+const Svg = ({ size, ...props }: { size: number }) => (
+  <ClassNames>
+    {({ css }) => (
+      <svg
+        height={size}
+        width={size}
+        viewBox="0 0 20 20"
+        aria-hidden="true"
+        focusable="false"
+        className={css({
```
gwyneplaine: @mitchellhamilton We technically don't need to be using ClassNames here to access the css method, since we aren't generating a custom className using our own cx implementation, however I noticed that using the css prop with the jsx pragma generated a style tag without a nonce attribute. Not sure why this is the case, but using the ClassNames renderProp to circumnavigate this for now

src/components/indicators.js
```diff
@@ -165,24 +183,24 @@ export const loadingIndicatorCSS = ({
...
-      borderRadius: '1em',
-      display: 'inline-block',
-      marginLeft: offset ? '1em' : null,
-      height: '1em',
-      verticalAlign: 'top',
-      width: '1em',
-    })}
-  />
+type DotProps = { color: string, delay: number, offset: boolean };
+const LoadingDot = ({ color, delay, offset }: DotProps) => (
```
gwyneplaine: @mitchellhamilton same as with the svg example above, technically not necessary to use <ClassNames/> here, but noticed that nonces were not being added, and the keyframes`` call above was not working.

src/components/Control.js
mitchellhamilton: Using your own cx function is totally fine, it just means that emotion styles won't be merged together which in this case, I don't believe it matters. I'm not really sure about this whole approach though, ClassNames is intended to be a workaround when a component doesn't accept a className prop, I'd recommend an approach like this:
```diff
diff --git a/src/components/Control.js b/src/components/Control.js
index a964547..c38a9ad 100644
--- a/src/components/Control.js
+++ b/src/components/Control.js
@@ -1,6 +1,7 @@
 // @flow
-import React, { type Node, type ElementRef } from 'react';
-import { ClassNames } from '@emotion/core';
+import { type Node, type ElementRef } from 'react';
+/** @jsx jsx */
+import { jsx } from '@emotion/core';

 import type { CommonProps, PropsWithStyles } from '../types';

@@ -56,22 +57,19 @@ export const css = ({
 const Control = (props: ControlProps) => {
   const { children, cx, getStyles, className, isDisabled, isFocused, innerRef, innerProps, menuIsOpen } = props;
   return (
-    <ClassNames>
-      {({ css }) => (
-        <div
-          ref={innerRef}
-          className={cx(css(getStyles('control', props)), {
-            'control': true,
-            'control--is-disabled': isDisabled,
-            'control--is-focused': isFocused,
-            'control--menu-is-open': menuIsOpen
-          }, className)}
-          {...innerProps}
-        >
-          {children}
-        </div>
-      )}
-    </ClassNames>
+    <div
+      ref={innerRef}
+      css={getStyles('control', props)}
+      className={cx({
+        'control': true,
+        'control--is-disabled': isDisabled,
+        'control--is-focused': isFocused,
+        'control--menu-is-open': menuIsOpen
+      }, className)}
+      {...innerProps}
+    >
+      {children}
+    </div>
   );
 };

diff --git a/src/utils.js b/src/utils.js
index 0787bfd..d37f037 100644
--- a/src/utils.js
+++ b/src/utils.js
@@ -41,11 +41,10 @@ function applyPrefixToName(prefix, name) {

 export function classNames(
   prefix?: string | null,
-  cssKey?: string | null,
   state?: ClassNamesState,
   className?: string,
 ) {
-  const arr = [cssKey, className];
+  const arr = [className];
   if (state && prefix) {
     for (let key in state) {
       if (state.hasOwnProperty(key) && state[key]) {
```

src/components/indicators.js
mitchellhamilton: That's very strange, gonna look into it.

src/components/Control.js
gwyneplaine: @mitchellhamilton This is objectively much nicer, only issue is that it seems converting to 
```javascript 
<div css={someemotionstyles} className={react-select-generated-classnames}/>
```
generates the following selector `react-select-classname emotion-classname' 
Its important that we generate the selector as `emotion-classname generated-react-select-classname' as we need to preserve css specificity for the react-select classname  for users not using css-in-js to easily override emotion styles without having to resort to !important. 

If there's a good way around this,, I'd love to refactor the code to the pattern you've specified above, as it's considerably cleaner.

gwyneplaine:
@JedWatson  after talking to @jossmac  we agreed that this is a breaking change, and should be a 3.0.0.

williamtstanley:
With this update how would I style custom Options? I used to get emotion as a prop but nolonger... I also don't seem to be getting `css` as a replacement any longer...

Faradey27:
@gwyneplaine are we plan to merge this?

gwyneplaine:
@Faradey27 plan is to merge this as a 3.x, we're doing a bit of clean up to release another 2.x.x before this happens, thanks for the patience

ematipico:
This is critically needed. 
We can't afford to install install emotion only to support SSR. Thanks!

gwyneplaine:
hey @ematipico, thanks for raising interest in this body of work. 
I've been a little swamped over the past few weeks, but am intending to ship 3.0.0 with emotion 10 towards the end of the month. Thanks for your patience :)

leerob:
Would love to see this get released! We've had to do a bunch of workarounds to get `react-select` working with Next.js and this would prevent that.

jossmac:
@gwyneplaine give me a yell if you need a hand with anything. Happy to pair on this.

lukehamilton:
Any ETA on when this will be merged in?

senorgeno:
@leerob I am using this with Gatsby and having issues after SSR. Any advice on what workarounds you used to get it to work with Next.js? Looking forward to v3.0.0 👍

vaunus:
@senorgeno I am just forcing client-side only rendering for now until this is released by wrapping the select with `<NoSSR>`. I'm importing `NoSSR` from MaterialUI as I am using that in my project, but I imagine any of the other implementations of it would work as well.

senorgeno:
Thanks @vaunus `<NoSSR>` worked for me in the mean time

PR #3323, opened 2019-01-05
Added primitive data type check in cleanValue util function

### LINKED ISSUE #3048: Doesn't display the selected option (Single Select & getOption)
I have an array of strings and using `getOptionValue` and `getOptionLabel` I want to use them in Select component.
If Select is in "multi" mode, it is working properly, if not, it doesn't display the selected option.

React v15 +
React Select v2

live example: https://codesandbox.io/s/v0kp0607x7

mnpreetkaur:
Closes #3048

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -62,7 +62,9 @@ export function classNames(
 
 export const cleanValue = (value: ValueType): OptionsType => {
   if (Array.isArray(value)) return value.filter(Boolean);
-  if (typeof value === 'object' && value !== null) return [value];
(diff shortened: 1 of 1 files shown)


ebonow:
**Case for PR:**
When using an array of options that are of type String or Number and using `getOptionValue` and `getOptionLabel`, the result of the onChange value returns no value.

Since the select can render these options and that this is not an issue when `isMulti`, it should be expected that the value should be the same value as the selected option when isMulti is false.

PR #3326, opened 2019-01-07
Fix focus issue on iOS when hitting clear indicator

stevemao:
`input.focus()` only works with touch events on iOS. Wrapping it in `setTimeout` makes it async and you won't bring up the virtual keyboard.


JedWatson:
Thanks @stevemao

matanstorr:
The keyboard now comes up after hitting the clear indicator on iOS(thanks @stevemao), but the input is not focused(I can't type immediately, I need to click the input first), can you offer any help or fix?

stevemao:
@matanstorr this PR should fix the problem you are talking about.

Issue #3352, opened 2019-01-15
Aria messages cannot be localized and one of the messages contains wrong statement


dmarcautan:
The messages returned by 'instructionsAriaMessage' from accessibility/index are in plain English and there's no option to localize them. Also, when a value in multi-option setting is focused the messages returned states that item can be removed by pressing Enter that is not true.

Kashkovsky:
#3358 Here's the PR for this issue.

PR #3358, opened 2019-01-16
Issue #3352: Aria messages cannot be localized and one of the message…

Kashkovsky:
The messages returned by 'instructionsAriaMessage' from accessibility/index are in plain English and there's no option to localize them. Also, when a value in multi-option setting is focused the messages returned states that item can be removed by pressing Enter that is not true.


src/Select.js
```diff
@@ -242,6 +257,10 @@ export type Props = {
   tabSelectsValue: boolean,
   /* The value of the select; reflected by the selected option */
   value: ValueType,
+  /* A CSP Nonce which will be used in injected style sheets */
+  nonce?: string,
```
gwyneplaine:
```suggestion

```

src/Select.js
```diff
@@ -242,6 +257,10 @@ export type Props = {
   tabSelectsValue: boolean,
   /* The value of the select; reflected by the selected option */
   value: ValueType,
+  /* A CSP Nonce which will be used in injected style sheets */
```
gwyneplaine:
```suggestion
```

Kashkovsky:
@gwyneplaine Removed nonce

bb-sdailey:
I was wondering if this is still being worked on? We currently have to hide the provided aria-live region in order to provide our own localized messaging which leads to a large amount of duplication on our end. Having the ability to provide our own strings would be a huge improvement. Any info is appreciated.

mohit181191:
Any plans to merge this pull request. I did lock one issue, seems like It should be closed as it is similar to it. https://github.com/JedWatson/react-select/issues/4001

ouelletalexandre:
any infos in this ?

bladey:
Closing this in favor of @gwyneplaine's PR #3550 which is based off this PR with further improvements.

If a mistake has been made and this PR is still relevant, please let me know.

Issue #2612, opened 2018-05-15
[V2] focusedOption when rendering custom MenuList


PengyuanZhao:
In V2, I want to completely replace the MenuList including its children (the options). But then the option loses the default highlighted color when it is "focused". How could I keep that in V2?

In V1, there is a `focusedOption` prop which could be used to highlight the option manually. 

Thanks!

hrnik:
@gwyneplaine Can you help with the answer, or clarify place where we can ask this question?

I also have this problem, need `focusedOption` for integrate react-virtualized or react-window.

dzbiku:
Anyone fixed it?

mike391:
+1 this is important

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our efforts towards the current major version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue  is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

PR #3360, opened 2019-01-16
Pass focusedOption to the MenuList

sunniejai:
This is useful for when you're overriding MenuList with a virtualized list (react-virtualized). You will need to know the focusedOption so that you can handle scrolling the virtualized list.

Now it's possible to do something like this in MenuList:
```
class MenuList extends React.PureComponent {
    componentDidUpdate(prevProps) {
        const { focusedOption } = this.props
        const previousFocusedIndex = getIndex(prevProps.focusedOption)
        const focusedOptionIndex = getIndex(focusedOption)
        if (previousFocusedIndex != focusedOptionIndex && this.listRef) {
          this.list.scrollToRow(focusedOptionIndex)
        }
    }
    ...
}
```

Also useful here: https://github.com/JedWatson/react-select/issues/2612


ebonow:
Quick question: Would this cause any additional re-renders passing in this option to the menu?

JedWatson (approved):

ebonow (approved):

Issue #3362, opened 2019-01-17
Incorrectly calculated code coverage


DragorWW:
Jest wraps Istanbul, and therefore also tells Istanbul what files to instrument with coverage collection.

[Istanbul is built into Jest](https://istanbul.js.org/docs/tutorials/jest/), so configuration is handled directly by Jest. The report can be generated by triggering jest with the --coverage flag.

**Istanbul is not needed when using jest**

**Example of problem**: `src/components/Menu.js`
```js
49 [√] |export function getMenuPlacement(...) {
58 [√] |    const { spacing } = theme;
       |   // ...
63 [√] |    if (!menuEl || !menuEl.offsetParent) return defaultState;
67 [X] |    const { height: scrollHeight } = scrollParent.getBoundingClientRect();
       |    // ...
90 [X] |    switch (placement) {
       |        // ...
152    |        case 'top':
153    |        // 1: the menu will fit, do nothing
154[√] |        if (viewSpaceAbove >= menuHeight) {
155[√] |            return { placement: 'top', maxHeight };
156    |        }
       |     // ...
       |    }
       |    // ...
       |}
```

`Switch` *(line 154)* cannot be covered because the code after the first `return` *(line 67)* is not used

So now coverage is calculated correctly in the version of `babel@7`

**real coverage is 64%**

DragorWW:
@gwyneplaine issue can be closed

PR #3363, opened 2019-01-17
Issues #3362 fix incorrect calculation of code coverage

DragorWW:
fix problem #3362

`nyc` and `Istanbul` is not needed when using jest


gwyneplaine:
lgtm cc @JedWatson  @Noviny, this is the root of our code coverage issue

PR #3366, opened 2019-01-18
update npm deps: babel@7 rollup@1

DragorWW:
I selected dependent modules from [PR update flow](https://github.com/JedWatson/react-select/pull/3350)

update:
- babel@6 to babel@7
- rollup@0 to rollup@1
- extract-react-types-loader
- pretty-proptypes


.babelrc
```diff
@@ -1,9 +1,18 @@
 {
-  "plugins": [
-    "emotion",
-    "transform-class-properties",
-    "transform-object-rest-spread"
+  "plugins": ["emotion", "@babel/plugin-proposal-class-properties"],
+  "presets": [
+    ["@babel/preset-env", { "modules": false }],
```
gwyneplaine: Is there a reason for changing this to `{modules: false}`  here?

For context,  our build script uses rollup to compile to UMD and ESM, we use the babel-cli to build our lib folder as CJS. the implicit `{ modules: auto }` accommodates this behaviour see [here](https://github.com/babel/babel/blob/master/packages/babel-preset-env/src/normalize-options.js#L150)

.babelrc
```diff
@@ -1,9 +1,18 @@
...
+    "@babel/preset-react",
+    "@babel/preset-flow"
   ],
-  "presets": ["env", "react"],
-  "ignore": ["node_modules"]
+  "ignore": ["node_modules"],
+  "env": {
```
gwyneplaine: If the answer to the above query is no, then we can remove this additional configuration

rollup.config.js
```diff
@@ -19,19 +19,6 @@ const globals = {
 };
 // $FlowFixMe This should be inferred by Flow and manual casting does not work inside of this config.
 const external = Object.keys(globals);
-const babelOptions = () => {
-  let result = {
```
gwyneplaine: Is this removal a necessary part of the rollup 1.x upgrade, or is this an  adjacent optimisation related to the [.babelrc](https://github.com/JedWatson/react-select/pull/3366/files#diff-e56633f72ecc521128b3db6586074d2c) changes? If it's the latter, then we should revert this

gwyneplaine:
@DragorWW  thanks for this PR, added questions / actions where relevant.

.babelrc
TrySound: `modules: false` will be passed by rollup-plugin-babel. You may not specify it. And remove the following `env`

.babelrc
```diff
@@ -1,9 +1,18 @@
...
-    "transform-object-rest-spread"
+  "plugins": ["emotion", "@babel/plugin-proposal-class-properties"],
+  "presets": [
+    ["@babel/preset-env", { "modules": false }],
+    "@babel/preset-react",
+    "@babel/preset-flow"
   ],
-  "presets": ["env", "react"],
-  "ignore": ["node_modules"]
+  "ignore": ["node_modules"],
```
TrySound: ignore in rollup itself

rollup.config.js
```diff
@@ -3,7 +3,7 @@
 import babel from 'rollup-plugin-babel';
 import commonjs from 'rollup-plugin-commonjs';
 import resolve from 'rollup-plugin-node-resolve';
-import uglify from 'rollup-plugin-uglify';
+import { uglify } from 'rollup-plugin-uglify';
```
TrySound: I recommend to use `rollup-plugin-terser` instead. It has less bugs.

rollup.config.js
```diff
@@ -3,7 +3,7 @@
...
 import { minify } from 'uglify-es';
```
TrySound: It's dead project. Burn it.

rollup.config.js
TrySound: You may keep here only `plugins: ['emotion']`. Other plugins will be merged from root config. External helpers are added out of the box now.

TrySound: Add `exclude: /node_modules/`

rollup.config.js
```diff
@@ -53,7 +40,7 @@ export default [
       globals,
     },
     external,
-    plugins: [babel(babelOptions()), resolve(), commonjs()],
+    plugins: [babel(), resolve(), commonjs()],
```
TrySound: Add `include: /node_modules/` to commonjs plugin

rollup.config.js
```diff
@@ -41,7 +28,7 @@ export default [
       format: 'es',
```
TrySound: You may use `esm` alias now. It's more descriptive.

.babelrc
DragorWW: @TrySound Why do it?

rollup.config.js
DragorWW: @gwyneplaine revert this

rollup.config.js
```diff
@@ -65,6 +67,6 @@ export default [
       globals,
     },
     external,
-    plugins: [babel(babelOptions()), resolve(), commonjs(), uglify({}, minify)],
+    plugins: [babel(babelOptions()), resolve(), commonjs(), terser()],
```
TrySound: Add `include: /node_modules/` to commonjs plugin to skip parsing for source files.

rollup.config.js
```diff
@@ -19,16 +18,19 @@ const globals = {
...
+      '@babel/preset-flow'
+    ],
     plugins: [
       'emotion',
-      'transform-class-properties',
-      'transform-object-rest-spread',
-      'external-helpers',
-    ],
+      '@babel/plugin-proposal-class-properties'
+    ]
```
TrySound: You may remove `babelrc: false` and just add emotion plugin for simplification.

DragorWW: good idea, but this an adjacent optimisation related to the .babelrc changes ?)

TrySound: I don't understand what is the problem with this. It's tooling and build scripts upgrade. @gwyneplaine Why can't this be changed here?

gwyneplaine: @TrySound  @DragorWW  This is fine, the main point with the previous .babelrc change comments was to make sure we don't break the lib build via babel-cli, this doesn't have any effect on that part of the build step 👍

gwyneplaine:
thanks for resolving those comments @DragorWW, once @TrySound's feedback has been addressed I'll merge this in 👍

DragorWW:
gwyneplaine @TrySound All is well? Are there any reasons not to merge? If you need to fix something else, write 👍

gwyneplaine:
@DragorWW did another smoke test of this work, our PropType documentation is now no longer working. Having a look at this now, if you want to dig into it as well that would be helpful, once that's resolved we can merge.

DragorWW:
@gwyneplaine I forgot about in new version `extract-react-types-loader` this display of PropTypes  only work there are environment variables `WEBSITE_ENV == 'staging' | 'production'` or `FORCE_EXTRACT_REACT_TYPES`

I think the variable `FORCE_EXTRACT_REACT_TYPES` will be a more explicit choice.

DragorWW:
> @DragorWW did another smoke test of this work, our PropType documentation is now no longer working. Having a look at this now, if you want to dig into it as well that would be helpful, once that's resolved we can merge.

fix it

gwyneplaine:
Merged, thanks @DragorWW

Issue #3369, opened 2019-01-19
react-select > memoize-one@4.1.0: New custom equality api does not play well with all equality helpers. Please use v5.x


adam187:
When installing (using Yarn) the following warning is shown:

> react-select > memoize-one@4.1.0: New custom equality api does not play well with all equality helpers. Please use v5.x

gwyneplaine:
pr released as part of v2.4.0, thanks @adam187

PR #3370, opened 2019-01-19
Upgrade memoize one to v5

adam187:
Should resolve issue: #3369

Mayor change for v5 was signature for equality function:

https://github.com/alexreardon/memoize-one/tree/v4.1.0#equality-function-type-signature
https://github.com/alexreardon/memoize-one/tree/v5.0.0#custom-equality-function

`react-select` doesn't use it in [compare function](https://github.com/JedWatson/react-select/blob/master/src/internal/react-fast-compare.js#L7) so it can be safely upgraded.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "dependencies": {
     "classnames": "^2.2.5",
     "emotion": "^9.1.2",
-    "memoize-one": "^4.0.0",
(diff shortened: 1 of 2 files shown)


gwyneplaine:
thanks for this @adam187, merged

PR #3371, opened 2019-01-21
Precommit improvement

DragorWW:
After working with the react-select codebase, I really missed this, it will be useful to all contributors in the future.

add to lint-staged:
- `prettier --write`: auto fix code style
- `eslint --fix`: auto fix eslint error
- `git add`: and add all fix before commit

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "moment": "^2.20.1",
     "nps": "^5.7.1",
     "nps-utils": "^1.3.0",
+    "prettier": "1.16.0",
     "pretty-proptypes": "^0.2.1",
     "raf-schd": "^2.1.0",
     "raw-loader": "^0.5.1",
@@ -137,6 +138,10 @@
     ]
   },
   "lint-staged": {
-    "*.js": "eslint"
+    "*.js": [
+      "prettier --write",
+      "eslint --fix",
+      "git add"
+    ]
   }
 }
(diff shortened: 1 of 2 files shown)


JedWatson:
@DragorWW won't the `git add` step stage all changes, and make it impossible to commit partial changes to files?

That's not ideal, I personally will sometimes work through a series of changes to a single file then commit them in parts so my process is easier to follow in code reviews. It can also lead to unexpected changes being committed, which I think is worse than needing to run the linter and formatter yourself.

FWIW, I've got my editor set up to format and lint --fix on save, which is less opinionated on git and achieves the same result.

DragorWW:
@JedWatson I believe that the all code of project must follow the code style (eslint, prettier)

The question of the automatic fix code - a developer experience

Any contributor should be able to create the code according to the project standards (eslint, prettier) as simply as possible.


**Perhaps**, to start,  add to the prepush check of the project code without fix, what do you think about it?

SimpleCookie:
I agree, I always got this on my projects, it makes life so much better

Methuselah96:
I personally prefer not having pre-commit hooks. I don't like "automatic" things when happening when I go to commit code. I would rather just adding `prettier --check` to the CI and leaving it up to the developer to set it up Prettier with their editor or just run the command separately.

Just my voicing my personal preference, I'm willing to go with whatever the majority opinion (or Jed) decides.

Methuselah96:
Thank you for the PR! As Jed pointed out above, you can now configure editors to automatically apply formatting changes automatically on save. Now that https://github.com/JedWatson/react-select/pull/4338 is merged, the formatting of the code is enforced by the CI. The team's preference on this one is to rely on editors and not precommit hooks since it allows developer's more flexibility on when to apply formatting and keep it separate from the Git process.

PR #3372, opened 2019-01-22
async-select-refactor

### LINKED ISSUE #3032: [v2] Loading indicator disappears once the menu is opened
If we open the `menu` while Async component is loading the options, `loadingIndicator` disappears even though `loadOptions` is not yet completed. `noOptionsMessage` displayed until the options are loaded but no more `loadingIndicator`(Even if you open/close the `menu`, it's same). There is no indication if the options are still loading.

Although it's a simple scenario, I couldn't find a similar bug report. Simple test case forked from the web site: https://codesandbox.io/s/14919k7q4

Would there be an alternative way to fix this problem?

gwyneplaine:
# Async Select Refactor
## Overview 
Previous Async Select was a composition of makeAsyncSelect, manageState and the base Select component in the following order.
`makeAsyncSelect(manageState(Select)); `

This meant the async logic from the exported Async Select was not able to leverage the helpful abstraction of controlled properties like onChange and inputValue that manageState provides. 

This PR aims to rectify this by changing this composition order, to the following:
`manageState(makeAsyncSelect(Select))`

In doing so we are afforded a bunch of niceties: 
* we no longer need to control the inputValue within the state of the Async component, this gets handled by the wrapping logic in manageState. 
* because of the change above, we also no longer need custom onChange logic within the Async component, the most recent inputValue gets read from props; and onChange logic is pushed up as a responsibility of the manageState function. Removing this custom logic means we also resolve #3032 

AFAIK the api also doesn't change, so this might even be possible to ship as either 2.3.1 or 2.4.0 @JedWatson


SimpleCookie:
I'm not really into this project, but this PR looks good to me.

ebonow:
@Methuselah96 thoughts on closing this in light of the refactor to TypeScript?

dcousens:
This is significantly out of date since `4.0.0` and `5.x.y`; the changes may still be welcome, but the pull request needs to be significantly updated.
Happy for @gwyneplaine to re-open if they want to rebase the work.

PR #3380, opened 2019-01-24
replace event.target to event.currentTarget

DragorWW:
fix flow error in Select.js, related to #3350

The reason why you want to use event.currentTarget is that event.target may be the wrong element due to [event propagation](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Examples#Example_5:_Event_Propagation).
see more in:
- https://github.com/DefinitelyTyped/DefinitelyTyped/issues/11508#issuecomment-256045682
- https://flow.org/en/docs/react/events/

flow types from react
```
declare class SyntheticEvent<+T: EventTarget = EventTarget, +E: Event = Event> {
  bubbles: boolean;
  cancelable: boolean;
  +currentTarget: T;
  defaultPrevented: boolean;
  eventPhase: number;
  isDefaultPrevented(): boolean;
  isPropagationStopped(): boolean;
  isTrusted: boolean;
  nativeEvent: E;
  preventDefault(): void;
  stopPropagation(): void;
  // This should not be `T`. Use `currentTarget` instead. See:
  // https://github.com/DefinitelyTyped/DefinitelyTyped/issues/11508#issuecomment-256045682
  +target: EventTarget;
  timeStamp: number;
  type: string;
  persist(): void;
}
```

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -886,13 +886,11 @@ export default class Select extends Component<Props, State> {
     } else if (!this.props.menuIsOpen) {
       this.openMenu('first');
     } else {
-      // $FlowFixMe HTMLElement type does not have tagName property
-      if (event.target.tagName !== 'INPUT') {
+      if (event.currentTarget.tagName !== 'INPUT') {
         this.onMenuClose();
       }
     }
-    // $FlowFixMe HTMLElement type does not have tagName property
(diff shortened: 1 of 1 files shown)


gwyneplaine:
lgtm thanks @DragorWW

PR #3384, opened 2019-01-26
Make MultiValueContainer's heading consistent

tylercollier:
The heading is currently an h4 but should be an h3 like all the rest. Yup, I'm doing this because it matters; I missed it the first time through the doc due to that.


gwyneplaine:
@tylercollier, thanks for this, merged

PR #3393, opened 2019-01-29
Correct aria text for removing options

msharkeyiii:
Hi! I am working through an accessibility review and noticed the wording here was off. Pressing Enter does nothing when you're focused on a selected item. You need to press Backspace to remove it. Just thought I'd pass this along. Thanks!


gwyneplaine:
thanks for this @msharkeyiii, merged

PR #3400, opened 2019-02-02
Feature: cx-function as a Prop

psineur:
Pretty small change should be self-explanatory.

Motivation: I'm using css-modules & would like to keep my styling for react-select together with component's style I use it with.

I've found it the easiest to simply pass my own cx function (that is already bound with my styles loaded) to the Select component, and thus, to everything else.

This allows me simply target 'control', 'indicator' and all other classes in my css files.

Any chance this can be a useful feature for react-select? If so - what documentation / tests/ anything you want me to add to this PR?

Thank You!

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -63,6 +63,7 @@ import type {
   OptionsType,
   OptionType,
   ValueType,
+  ClassNamesState,
 } from './types';
 
 type MouseOrTouchEvent =
@@ -244,6 +245,9 @@ export type Props = {
   tabSelectsValue: boolean,
   /* The value of the select; reflected by the selected option */
   value: ValueType,
+
+  /* External cx-function. */
+  cxx?: (?string | null, ClassNamesState | void, string | void) => string | void,
 };
 
 export const defaultProps = {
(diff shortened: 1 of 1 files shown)


gwyneplaine:
@psineur Thanks for the PR! This seems interesting, allowing for custom style composition for non emotion / non css-in-js users seems valuable. 

I do want to validate this usecase though, could you provide us with an example of this being used with the css-modules case you've described? 

If this ends up looking like something we'll want to pull in as a feature, you'll also need to alter the [cx](https://react-select.com/styles#cx-and-custom-components) section in the styling docs to reflect this change 

A basic unit test validating the custom cx mechanism should be enough for now. 

- [ ] Add an Example of the CSS-modules usecase defined above, so we can validate if its applicable to a larger cross-section of users
- ? Add docs 
- ? Add tests

bladey:
Hi @psineur, I'm now going to close this PR due to inactivity.

If you have any further feedback as requested by @gwyneplaine, please do let us know and we'll re-open and assess.

Anyone else who'd like to take control of this PR, please also let me know and we'll re-open it.

PR #3402, opened 2019-02-02
fix MenuList scrolling issue in IE11 with openMenuOnFocus option

### LINKED ISSUE #3342: [V2] IE11 bug: Scrollbar arrows scroll menu to the top when using openMenuOnFocus={true}
# Issue:
When using the prop `openMenuOnFocus={true}`, we cannot use the down arrow under the scroll bar to scroll the menu list down in IE11 (Chrome works fine). I am on react-select version 2.2.0.

My suspicion is that the menu gets refocused after hitting the scrollbar arrows, and with `openMenuOnFocus={true}`, it's reopening the menu, scrolling to the top. 

# Repro steps
1. Create a react-select dropdown with enough options to have a dropdown arrow
2. Give that react select component a prop `openMenuOnFocus={true}`
3. Open the app in IE11 (Chrome works fine)
4. In the dropdown, try scrolling down hitting the down arrow icon under the scroll bar
5. Notice that the react-select dropdown keeps scrolling back up to the top

Codesandbox does not seem to work in IE11 right now, so I am attaching my code and a GIF showing the behavior instead. Hope that's ok!

## Code:
```
import React, { Component } from "react";
import "./App.css";

const options = [
  { value: "chocolate", label: "Chocolate" },
  { value: "strawberry", label: "Strawberry" },
  { value: "vanilla", label: "Vanilla" },
  { value: "chicken", label: "Chicken" },
  { value: "candycorn", label: "CandyCorn" },
  { value: "whiskey", label: "Whiskey" },
  { value: "chocolatewhiskey", label: "ChocolateWhiskey" },
  { value: "strawberrywhiskey", label: "StrawberryWhiskey" },
  { value: "vanillawhiskey", label: "VanillaWhiskey" },
  { value: "coke", label: "Coke" },
  { value: "clock", label: "Clock" },
  { value: "snow", label: "Snow" },
  { value: "nachoflavorbeanchips", label: "nachoflavorbeanchips" }
];

class App extends Component {
  state = {
    selectedOption: null
  };

  handleChange = selectedOption => {
    this.setState({ selectedOption });
  };

  render() {
    const { selectedOption } = this.state;

    return (
      <div className="App">
        <Select
          value={selectedOption}
          onChange={this.handleChange}
          options={options}
          openMenuOnFocus={true}
        />
      </div>
    );
  }
}

export default App;
```

## GIF:
ie11 scroll arrows refocus menulist and wont scroll

timothypage:
seems related to 30ead6d, a fix for the MenuList closing when the
scrollbar is clicked in IE11.  When the select input is refocused, the
openMenu method re-scrolls MenuList to the currently selected value,
causing the user to be unable to scroll to other values via mouse click

fixes #3342

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -480,7 +480,7 @@ export default class Select extends Component<Props, State> {
   blur = this.blurInput;
 
   openMenu(focusOption: 'first' | 'last') {
-    const { menuOptions, selectValue } = this.state;
+    const { menuOptions, selectValue, isFocused } = this.state;
     const { isMulti } = this.props;
     let openAtIndex =
       focusOption === 'first' ? 0 : menuOptions.focusable.length - 1;
@@ -492,7 +492,8 @@ export default class Select extends Component<Props, State> {
       }
     }
 
-    this.scrollToFocusedOptionOnUpdate = true;
+    // only scroll if the menu isn't already open
+    this.scrollToFocusedOptionOnUpdate = !(isFocused && this.menuListRef);
     this.inputIsHiddenAfterUpdate = false;
 
     this.onMenuOpen();


timothypage:
codesandbox.io seems to be having some IE11 issues today too haha, so I put up an example of the issue here:

https://tzwolak.com/ie11-scroll-issue/index.html (open in IE11)

running this code:

```javascript
import React, { Component } from 'react';
import './App.css';

const options = 'abcdefghijklmnop'.split('').map(letter => ({label: letter, value: letter}));

class App extends Component {
  render() {
    return (
      <div className="App">
        <Select
          options={options}
          openMenuOnFocus
        />
      </div>
    );
  }
}

export default App;
```

Issue #3355, opened 2019-01-15
Critical accessibility issues: aria roles and aria-label


geoffrich:
When running the [Axe accessibility tester](https://www.deque.com/axe/) against the [react-select v2 code-sandbox](https://k5rvn9z3lv.codesandbox.io/), two critical issues are identified:
* [Form elements must have labels](https://dequeuniversity.com/rules/axe/3.1/label?application=AxeChrome). While you can pass `aria-label` as a prop to the `Select` component, it is not assigned as an attribute to the `#react-select-2-input` HTML element. Adding this attribute satisfies the Axe tester.
* [Certain ARIA roles must be contained by particular parents](https://dequeuniversity.com/rules/axe/3.1/aria-required-parent?application=AxeChrome). Each `.select__option` has `role="option"`; however, the parent div `.select__menu-list` does not have `role="listbox"`. Adding this attribute satisfies the Axe tester. Note that I had to use React Developer Tools to set `menuIsOpen` on the `StateManager` to see this violation.

Below are screenshots of the Axe test results.
![Form elements must have labels](https://user-images.githubusercontent.com/4992896/51203234-6b4f0a00-18b5-11e9-996d-04681f6ca195.png)

![Certain ARIA roles must be contained by particular parents](https://user-images.githubusercontent.com/4992896/51203195-4d81a500-18b5-11e9-83d4-1acb6ca2d1d8.PNG)

gwyneplaine:
thanks for this @geoffrey1218, we'll have a closer look at these to harden accessibility in selects. 
If you have the time and thoughts on how to do so, PRs are warmly welcome.

sarahbethfederman:
Hi! I'm not able to reproduce the first issue. May be that it was fixed in react-autosize-input. When providing an aria-label prop to the Select, it seems to be applied to the input. Here's a sandbox demo-ing this with a label value of "this is a label!!!" https://ryw61w7rnn.codesandbox.io/
Running axe on this sandbox produces no valid issues. 

Looking into second issue now :)

sarahbethfederman:
I think this issue can be closed

sarahbethfederman:
(someone yell at me if I'm not supposed to close issues)

bmsuseluda:
doesn't work for me if isSearchable is false. Can you check the reason behind this?

sarahbethfederman:
@bmsuseluda can you be a bit more specific about what part isn't working for you? Happy to look into this.

[...]

PR #3407, opened 2019-02-05
menu aria-roles

sarahbethfederman:
Adds role listbox to the select__menu-list element, to satisfy axe error "Certain ARIA roles must be contained by particular parents" as demonstrated in https://github.com/JedWatson/react-select/issues/3355


sarahbethfederman:
change of plans. Chatted with our a11y expert Sean and it turns out that we don't need any roles within the select because we are using an aria live region instead. Removing the extra aria roles also removes the error.

PR #3409, opened 2019-02-05
fix custom filter function example in docs

felixzhuologist:
There are [two custom custom filter examples](https://react-select.com/advanced#custom-filter-logic) on the advanced page of the docs. The `ExampleWrapper` of the second one has the same label and raw code as the first one, instead of the actual example component that gets rendered.


JedWatson:
Oooh thanks @felixzhuologist, good catch!

PR #3410, opened 2019-02-05
Remove dead code in example

jkjustjoshing:
The code I removed seems to not be referenced at all in the example. It's just dead code - confusing to a beginner trying to figure out how it's relevant!


gwyneplaine:
thanks for this @jkjustjoshing

PR #3432, opened 2019-02-18
changed event.currentTarget back to event.target for ontouchEnd logic

gwyneplaine:
Regression added as part of #3350. 
Logic change toe assert on currentTarget as a part of a flow refactor. 
In this case  event.target is the correct property to assert on due to the assertion being carried out 
on an eventListener attached to the document. 

See comments in code for details.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -1019,11 +1019,13 @@ export default class Select extends Component<Props, State> {
     if (this.userIsDragging) return;
 
     // close the menu if the user taps outside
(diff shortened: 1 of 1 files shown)


karlingen:
This unfortunately introduces this issue: https://github.com/JedWatson/react-select/issues/3607

karlingen:
And I've created a PR to fix the issue! @gwyneplaine 
PR: https://github.com/JedWatson/react-select/pull/3617

PR #3436, opened 2019-02-20
Allow Node as placeholder

### LINKED ISSUE #1378: Allow `placeholder` to be a node
I would like to allow the placeholder to be a node so I can put an icon.  If I have this method in my class
`
renderPlaceholder() {
    return (
      <span>
        <i className="icon mdi tag" />
        Add tags...
      </span>
    );
  }
`

and call it as a `prop` like
`placeholder={this.renderPlaceholder()}`

it works well but I get a React Props warning.   I think we should allow this so perhaps we can change the allowable props for `placeholder` to include `node`

eemeli:
Fixes #1378

Even though Placeholder [accepts `Node` children](https://github.com/JedWatson/react-select/blob/a23e9cdd115e024ce9afb69082e5628ed2f0471e/src/components/Placeholder.js#L8), the `placeholder` prop on Select [expects `string` type](https://github.com/JedWatson/react-select/blob/a23e9cdd115e024ce9afb69082e5628ed2f0471e/src/Select.js#L230). This causes an unnecessary dev mode warning when using e.g. a React element as a placeholder, even though it already works.

This PR relaxes the prop type to allow any node as placeholder. I also fixed the placeholder tests to actually test what they say they test.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -226,8 +226,8 @@ export type Props = {
   options: OptionsType,
   /* Number of options to jump in menu when page{up|down} keys are used */
   pageSize: number,
-  /* Placeholder text for the select value */
-  placeholder: string,
+  /* Placeholder for the select value */
+  placeholder: Node,
   /* Status to relay to screen readers */
   screenReaderStatus: ({ count: number }) => string,
   /*
--- a/src/__tests__/Select.test.js
+++ b/src/__tests__/Select.test.js
@@ -1837,19 +1837,28 @@ cases(
     'single select > should display default placeholder "Select..."': {},
     'single select > should display provided string placeholder': {
       props: {
+        ...BASIC_PROPS,
         placeholder: 'single Select...',
       },
       expectPlaceholder: 'single Select...',
     },
     'single select > should display provided node placeholder': {
       props: {
+        ...BASIC_PROPS,
         placeholder: <span>single Select...</span>,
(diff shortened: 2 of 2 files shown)


Rall3n:
https://github.com/JedWatson/react-select/issues/1378#issuecomment-465684538

Original issue was for version 1.x. In version 2.x you can replace every component that react-select uses. So this merge request is pretty unnecessary imo.

eemeli:
@Rall3n Custom components don't really help with the restricting prop type on Select. Yes, they technically allow for an alternative way of producing a custom placeholder, but the ergonomics of doing that are really rather clumsy, especially when compared to what's possible already with the default Placeholder -- provided that you ignore the prop type warning that I'm seeking to get rid of here.

Rall3n:
Prop type `Node` is there so you can overwrite the Placeholder but keep its default layout, like this:
```jsx
import { components } from 'react-select';

const Placeholder = ({children, ...props}) => {
    return <components.Placeholder {...props}>
        <div></div>
    </div>
}
```

eemeli:
@gwyneplaine Any chance of getting this looked at? It's literally a single-line fix, plus associated tests.

eemeli:
Rebased on latest master.

JedWatson:
I can't really see any downside to this. Thanks @eemeli 👍

PR #3438, opened 2019-02-20
Fix documentation code link for Custom Filter Options

mgmarlow:
The "Custom Filter Options" section in the advanced usage guide displays the incorrect code sample. This PR replaces that code sample with the one corresponding to "Custom Filter Options".


JedWatson:
Thanks @mgmarlow!

I just merged #3409 which also fixes this.

Issue #2770, opened 2018-07-03
[v2] hover style


berikiushi:
How to change styles on hover without use class names?

berikiushi:
issue solved

```
const customStyles = {
  control: base => ({
    ...base,
    borderColor: 'gray',

    '&:hover': {
      borderColor: 'red'
    }
  })
}
```

brainbrian:
@berikiushi would it be good to add this to the [docs site](https://react-select.com/styles)? I had to dig to find this.

berikiushi:
just added this in PR #3185

NSExceptional:
This is still not on the docs site…

gtdeng:
for V3, remove the `&`

```js
const customStyles = {
  control: base => ({
    ...base,
    borderColor: 'gray',

    ':hover': {
      borderColor: 'red'
    }
  })
}
```

PR #3441, opened 2019-02-24
issue #2770 docs for hover style was not present

ashu8912:
a pr was made #3185 but for some reason it was not merged so i am making another pr


JedWatson:
Thanks for bringing this to attention @ashu8912, I'm declining for the same reason explained in #3185

Let me know if you think there's a better way to document the example that's there.

PR #3442, opened 2019-02-25
Ignore keydown events on `Enter` in IME

sat0yu:
This PR is a bugfix and is helpful for people using an [input method editor(IME)](https://en.wikipedia.org/wiki/Input_method
)

e.g. `google input method`
https://www.google.com/intl/en/inputtools/services/features/input-method.html
screen_shot_0031-03-03_at_11_45_26

## Problem
- There's a browser dependent difference in the way handling keydown events on Enter 
  - In Safari (confirmed with `Version 12.0.3 (12606.4.5.3.1)`), the event triggers `onChange` and selects an option consequently
  - In Chrome(confirmed with `Version 72.0.3626.119 (Official Build) (64-bit)`), the event does nothing on the browser, not affect the selected value. it just determines the IME value

## Expected behavior
-  Make browsers act in the same way aginst keydown events, especially the ones on Enter key
  - In my opinion, in this case, Safari is wrong. it should behave as like Chrome

## Proposal
- Ignore keydown events on Enter in IME
  - The specification on the key codes is mentioned in https://www.w3.org/TR/uievents
    - `If an Input Method Editor is processing key input and the event is keydown, return 229.`
    - https://www.w3.org/TR/uievents/#determine-keydown-keyup-keyCode
  - We can ignore events which satisfy `event.key === 'Enter'  && event.keyCode === 229` 

Please review and merge this one, thanks 🙇

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -1179,6 +1179,11 @@ export default class Select extends Component<Props, State> {
         this.selectOption(focusedOption);
         break;
       case 'Enter':
+        if (event.keyCode === 229) {
+          // ignore the keydown event from an Input Method Editor(IME)
+          // ref. https://www.w3.org/TR/uievents/#determine-keydown-keyup-keyCode
+          break;
+        }
         if (menuIsOpen) {
           if (!focusedOption) return;
           if (isComposing) return;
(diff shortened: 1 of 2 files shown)


JedWatson:
Thanks for this @sat0yu, really appreciate the fix and comprehensive explanation!

PR #3457, opened 2019-02-28
New controlled inputValue example + ability to click/select inputValue text

ansonkao:
## Summary of Changes
- I've added a new condition around the `.preventDefault()` call as shown below.
- Added an example of controlled inputValue which helps showcase the motivation for this change

## Motivation for Change
In some cases, developers will control the prop `inputValue` as shown in the example below, so that it can be displayed and/or controlled by another component elsewhere in the application. This makes `react-select` behave more like a plain 'ol `<input>` that "also does autocomplete". This is often a compelling use case. 

The problem here is that the user now expects to be able to click around and highlight text within the `react-select` just like they can with a normal `<input>`, but they aren't able to. This change enables them to.

See this video showcasing what I mean: [See Video](https://www.youtube.com/watch?v=Lh71yoR2Pu0)

I've tested in Chrome on macOS and ran the tests and the change appears to be backwards compatible with all existing use cases. 

There certainly is a risk however, of introducing this change. Not sure what 2nd order consequences there will be to other developers.

And, I don't fully understand what role this `event.preventDefault()` plays in the code (can anyone comment who is intimately familiar with the internals?). I know that without it, the menu falls into some weird corner case... so it's necessary - just not when `inputValue` is present, it seems.

### DIFF
--- a/docs/examples/ControlledInputValue.js
+++ b/docs/examples/ControlledInputValue.js
@@ -1,0 +1,37 @@
+import React, { Component, Fragment } from 'react';
+
+import Select from '../../src';
+import { colourOptions } from '../data';
+
+type State = {
+  inputValue: string,
+};
+
+export default class controlledInputValue extends Component<*, State> {
+  state = {
+    inputValue: 'Default Input Value',
+  };
+  select: ElementRef<*>;
+  onInputChange = (inputValue, { action }) => {
+    if (action === 'input-change') {
+      this.setState(() => ({ inputValue }));
+    }
+  }
+  render() {
+    const { inputValue } = this.state;
+    return (
+      <Fragment>
+        <p><code>inputValue:</code> {inputValue}</p>
+        <Select
+          ref={(ref) => { this.select = ref; }}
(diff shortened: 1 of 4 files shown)


vertic4l:
@JedWatson where are you ?

chunkjs:
Hello :) are there plans to  incorporate  this?

ebonow:
Appreciate the attempt here to standardize the behavior to be something more in line with user expectations. I think this is a case of adding more code to a repository to enable a user to do something that is already possible simply by using a controlled input on their own for which there are plenty of examples. I would sooner support reinstating the prop `clearInputOnBlur` than a HOC like this.

Working demo: [codesandbox](https://codesandbox.io/s/react-select-editable-and-selectable-6nfdv)

I have created this as an example of how to have an editable input with hopes to create more visibility and perhaps have something like this added to the code examples.

Given that this PR adds no new functionality and only serves as a controlled input for one state variable, this would not qualify as a PR that justifies increasing the package size for all react-select users.

PR #3460, opened 2019-03-04
add explanations for what the various actionTypes are

### LINKED ISSUE #3451: ActionTypes need explanations
# Are you making a feature request?

ActionTypes are listed in the docs but no documentation about what they do. What's the difference between `pop-value` vs `remove-value`, for example?

This is needed so I can implement the `onChange` handler properly for a multi value `Creatable`.

mikekellyio:
This fixes #3451 and scratches my current need for explanations of these types as well.

### DIFF
--- a/docs/pages/advanced/index.js
+++ b/docs/pages/advanced/index.js
@@ -136,13 +136,13 @@ export default function Advanced() {
 
       The action argument is a string with the following possible values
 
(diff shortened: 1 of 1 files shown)


gwyneplaine:
amazing  @mikekellyio, thanks for the contribution

PR #3466, opened 2019-03-07
Fixed incorrect prop in upgrade guide

kumarharsh:
The upgrade guide was mentioning `className` as the prop which would supply classes to sub-components, when in fact, it was `classNamePrefix` which did the job.


JedWatson:
Nice catch, we changed that just before the final 2.0.0 release and must have missed it in the upgrade guide.

Thanks!

PR #3471, opened 2019-03-12
Update badge for build status to use CircleCI instead of TravisCI

pavog:
The old badge for the build status still uses TravisCI but react-select changed to CircleCI around 11 months ago


JedWatson:
🤔 thanks for picking up on that @pavog!

Issue #2848, opened 2018-07-24
Disable auto focus for option


khawerrind:
First of all thank you for the great library. 

I wanted to ask if there is any option to not select any option by default? I want the option to be selected only when user use arrow keys or click on an option using mouse. Why? Because when i enter some text and there is a matching item pressing enter uses the default selected option.

tung-dang:
In this code line https://github.com/JedWatson/react-select/blob/master/src/Select.js#L458, look like 
`focusOption` is always the first or last menu option. 
I am looking for a way to disable this behaviour - "auto select an option when menu open".

software-project:
Anyone found a solution for this?

yurii-lubynets:
Hello, @JedWatson 
Thank you for the flexible and useful tool. Could you provide some hardcoded solution for this issue? And what about the `autoFocus` prop?
Thanks,
Yurii

alisman:
While it would be great for this PR to be merged, I believe I managed to accomplish this with the following hackery:

onInputChange={(value: string) =>{ this.selectRef.select.getNextFocusedOption = ()=>null; }}

There is no longer any default selection, but arrow keying still works.

I then also bind to onKeyDown to intercept submission from input and fire submission callback with text directly from input.

manvydasu:
There is an approved PR for this - https://github.com/JedWatson/react-select/pull/3010

bladey:
This will be resolved when #4080 is merged.

augusto-proano:
@khawerrind, quick question, are you planning to merge this pr at some point?

PR #3010, opened 2018-09-04
Add prop to make "automatically focusing to the first option" optional

trkaplan:
Related to #2653 

This PR:

- Add autoFocusFirstOption prop to make "automatically focusing to the first option" optional.
- Update "Select" test.
- Update snapshots

PR #3484, opened 2019-03-22
Added new props : autoFocusOption

CharlieBreval:
Hello,

Recently, I have been facing a problem : Using React Select without preselection of the first available (focusable) option.
This option didn't exist at all. The problem has been raised in multiples issues : 
#2848 

Behavior with this option is simple : 
1 : Click on the select : No option is focused
2 : Your user starts typing a word, no option is focused until the user do an action for that.

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -402,13 +402,13 @@ export default class Select extends Component<Props, State> {
     ) {
       const selectValue = cleanValue(nextProps.value);
       const menuOptions = this.buildMenuOptions(nextProps, selectValue);
-      let focusedValue = this.getNextFocusedValue(selectValue);
-      let focusedOption = this.getNextFocusedOption(menuOptions.focusable);
-
-      if (nextProps.autoFocusFirstOption === false  && this.state.focusedOption === null) {
-        focusedValue = null;
-        focusedOption = null;
-      }
+      const { focusedOption, focusedValue } =
+        nextProps.autoFocusFirstOption === false && this.state.focusedOption === null
+        ? { focusedValue: null, focusedOption: null }
+        : {
+          focusedValue: this.getNextFocusedValue(selectValue),
+          focusedOption: this.getNextFocusedOption(menuOptions.focusable)
+        };
 
       this.setState({ menuOptions, selectValue, focusedOption, focusedValue });
     }
--- a/src/Select.js
+++ b/src/Select.js
@@ -82,6 +82,8 @@ export type Props = {
   'aria-labelledby'?: string,
   /* Focus the control when it is mounted */
   autoFocus?: boolean,
+  /* Focus first focusable option on menu open and when user is typing */
+  autoFocusFirstOption: boolean,
   /* Remove the currently focused option when the user presses backspace */
   backspaceRemovesValue: boolean,
   /* Remove focus from the input when the user selects an option (handy for dismissing the keyboard on touch devices) */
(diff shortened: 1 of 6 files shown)


src/Select.js
```diff
@@ -399,8 +402,14 @@ export default class Select extends Component<Props, State> {
     ) {
       const selectValue = cleanValue(nextProps.value);
       const menuOptions = this.buildMenuOptions(nextProps, selectValue);
-      const focusedValue = this.getNextFocusedValue(selectValue);
-      const focusedOption = this.getNextFocusedOption(menuOptions.focusable);
+      let focusedValue = this.getNextFocusedValue(selectValue);
+      let focusedOption = this.getNextFocusedOption(menuOptions.focusable);
+
+      if (nextProps.autoFocusFirstOption === false  && this.state.focusedOption === null) {
```
Blightwidow: It is usually better to consult the state directly in the `this.setState()` method as it is asynchronous. Something like 

```js
this.setState((prevState) => ({
  focusedOption: nextProps.autoFocusFirstOption === false  && prevState.focusedOption === null ? null : focusedOption,
  ...
}));
```

src/Select.js
```diff
@@ -82,6 +82,8 @@ export type Props = {
   'aria-labelledby'?: string,
   /* Focus the control when it is mounted */
   autoFocus?: boolean,
+  /* Focus first focusable option on menu open and when user is typing */
+  autoFocusFirstOption: boolean,
```
Blightwidow: I would maybe see something more like `autoFocusOption` with a value of `null`, `'first'`, or `'last'` for more flexibility.

CharlieBreval: I will rename the prop, but 'first' or 'last' is decided in another part of the library, not the concern of this option.

Blightwidow: Ok fine by me then.

danishmughal:
+1 for this, sorely needed.

alisonhall:
Any idea when this will be merged and released?

shepitko:
> Any idea when this will be merged and released?

+1, useful functionality, I would like to get information about the merge and release.

LillaEli:
Any chances this could be merged in the near future! 🙏

evgenyilin:
+1 When?)

johannesdold:
+1 We would urgently need this option - any plans when this pr will be merged?

rmertens:
we're also waiting for this. is there any news on this?

alisman:
While it would be great for this PR to be merged, I believe I managed to accomplish this with the following hackery:

`onInputChange={(value: string) =>{
     this.selectRef.select.getNextFocusedOption = ()=>null;
}}`

There is no longer any default selection, but arrow keying still works.   

I then also bind to onKeyDown to intercept submission from input and fire submission callback with text directly from input.

paulpiseckiy:
@alisman you have a great but incomplete solution here 🙂The problem is that it doesn't work when you open menu for the first time. Here's the complete:
```javascript
const ref = React.useRef();
  
React.useEffect(() => {
  ref.current.select.getNextFocusedOption = () => null;
}, []);

const handleInputChange = () => {
  ref.current.select.getNextFocusedOption = () => null;
};

const options = [
  {label: 'Option 1', value: '1'},
  {label: 'Option 2', value: '2'},
];

return (
  <ReactSelect
    ref={ref}
    options={options}
    onInputChange={handleInputChange}
  />
);
```

bladey:
PR to be reviewed, duplicate of #3010

bladey:
Hi all,

Thanks for your contribution @CharlieBreval!

There are a number of pull requests adding this feature - I've since created an up to date PR - #4080, and closed similar pull requests.

If I've missed anything, please let me know.

Closing this PR in favor of #4080.

PR #3492, opened 2019-03-30
Add labels to fields to make inspection easier

destructobeam:
I was trying to inspect elements and was having trouble figuring out which was what, thought labels on the elements might be helpful.

Cheers.

### DIFF
--- a/src/__tests__/__snapshots__/Async.test.js.snap
+++ b/src/__tests__/__snapshots__/Async.test.js.snap
@@ -161,7 +161,7 @@ exports[`defaults - snapshot 1`] = `
         }
       >
         <div
-          className="css-10nd86i"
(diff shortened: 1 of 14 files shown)


Rall3n:
@destructobeam The [react-devtools](https://github.com/facebook/react-devtools) from Facebook itself provide an excellent component inspector which makes identifiying elements of React components much easier.

destructobeam:
@Rall3n True, but often I like to see the  rendered DOM, especially when styling content as that’s where you can manipulate styles on the fly. Debug in Safari, etc.

Fine to close this if that’s a pretty uncommon case though.

JedWatson:
This seems like a nice quality of life improvement, thanks @destructobeam

PR #3508, opened 2019-04-11
Add the list of customizables components to the Components Docs

shug0:
List of components extracted from `react-select/dist/react-select.esm.js` - Lines **2636** - v2.4.2


JedWatson:
Thanks @shug0

Issue #3330, opened 2019-01-08
Name is not returned in Creatable 'select-option' action


kino90:
I'm using both `Creatable` and normal `Select` elements in my app, but when I input a "new" value to a Creatable one, the "actionMeta" object has only the `action: 'select-option'` and not the name of the select field that triggered the action. I think they should behave in the same way.

I've made a sandbox here:
https://codesandbox.io/s/nw1pklvykj

If you choose an existing option (from both of the select) it outputs on the console: 
(notice: ' name: "standard" and name: "creatable")

```
Object {value: "ocean", label: "Ocean", color: "#00B8D9"}
Object {action: "select-option", option: undefined, name: "standard"}
Object {value: "ocean", label: "Ocean", color: "#00B8D9"}
Object {action: "select-option", option: undefined, name: "creatable"}
```

But if you input a "new value" on the Creatable it outputs : 
(notice: no name here!)

```
Object {label: "DarkGreen", value: "DarkGreen", __isNew__: true}
Object {action: "create-option"}
```

TallOrderDev:
Running into this same issue. Would love to see consistency with the Meta object.

chaitanyapilaka:
Curious to know if any work was done on this yet?

sahilsharmafrank:
Any pointers or reference if this is tracked in an different ticket or an alternate solution to this

amruthsrepo:
@sahilsharmafrank, how do I use that file in my react project?

arjitaM:
Facing the same issue.

ebonow:
Greetings,

It would seem that this bug was patched in v3.0.8 and this should no longer be issue.
<img width="791" alt="Screen Shot 2020-12-03 at 11 08 39 PM" src="https://user-images.githubusercontent.com/1007579/101314754-d3e35f00-381e-11eb-839c-94f29c426533.png">

PR #3520, opened 2019-04-23
Added name param to actionMeta issue #3330

sahilsharmafrank:
`Creatable`
There is an `actionMeta` missing `name` param of an input. 
Case: if i have multiple `Creatable` in an page there should be an identifier for that particular field.

### DIFF
--- a/src/Creatable.js
+++ b/src/Creatable.js
@@ -37,6 +37,7 @@ export type CreatableProps = {|
   value: ValueType,
   isLoading?: boolean,
   isMulti?: boolean,
+  name?: string,
   onChange: (ValueType, ActionMeta) => void,
 |};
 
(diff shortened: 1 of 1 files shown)


SimeonC:
It's not documented but this should already be released - as `Creatable` extends `Select`.
https://github.com/JedWatson/react-select/pull/2871

JedWatson:
@SimeonC is correct, this should be extending the action meta from the base `Select` which includes the `name`.

@sahilsharmafrank are you seeing a different behaviour? if so, could you include a reproduction case please?

bladey:
Hi @sahilsharmafrank, I'm now going to close this PR due to inactivity. If you have any further feedback as requested by Jed, please do let me know and we'll re-open and assess, thank you!

PR #3528, opened 2019-04-28
Use preconstruct

mitchellhamilton:
Things here other than adding preconstruct:
- Remove `nps` because after adding preconstruct it was only being used to specify the build command for the site
- Add a `netlify.toml` which tells Netlify how to build the site, right now it's specified in the Netlify UI but that means that changes to the site's build command or dist folder have to be changed in the UI and therefore all branches have to have the same build config, with this though, different branches can have different build configs since it lives in git.


package.json
```diff
@@ -121,11 +120,6 @@
     "select",
     "ui"
   ],
-  "bundlesize": [
-    {
-      "path": "./dist/react-select.js"
```
mitchellhamilton: This doesn't really make sense now since we don't have a UMD build, there's probably a tool to do a bundle size check that bundles dependencies and minifies but I wasn't sure if having a bundle size check is a thing we care about having exactly right now so I just removed it.

PR #3531, opened 2019-04-29
Add support for menuScrollIntoViewElement

skovhus:
### Current behavior

Currently, `react-select` tries to find the appropriate element to scroll by traversing parents and finding a parent with `overflow=scroll / auto` and defaults to the root document. See  https://github.com/JedWatson/react-select/blob/292bad3298f2cafad6767f2134bd79a9c27e4073/src/utils.js#L127-L143

This doesn't work a `react-select` element is positioned inside a static positioned parent (example using CSS Grid). 

### Suggestion

This PR adds a new `menuScrollIntoViewElement` to support this custom behavior. You can try this PR out here: https://codepen.io/skovhus/pen/JVVGdO

It does not solve all use cases, as there is still code that checks for available view height on window (`viewSpaceBelow >= menuHeight`), and not inside the container. Here `menuScrollIntoViewElement` does not have any effect (yet). Demo https://codepen.io/skovhus/pen/LvvNJE 

### Related issues
https://github.com/JedWatson/react-select/issues/810

### TODO

- [x] proposal
- [ ] get feedback from maintainers
- [ ] see if I can add some unit tests
- [ ] more documentation

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -194,6 +194,8 @@ export type Props = {
   menuPortalTarget?: HTMLElement,
   /* Whether to block scroll events when the menu is open */
   menuShouldBlockScroll: boolean,
+  /* Custom element used when scrolling into view */
+  menuScrollIntoViewElement?: HTMLElement,
   /* Whether the menu should be scrolled into view when it opens */
   menuShouldScrollIntoView: boolean,
   /* Name of the HTML Input (optional - without this, no input will be rendered) */
@@ -1633,6 +1635,7 @@ export default class Select extends Component<Props, State> {
       menuPosition,
       menuPortalTarget,
       menuShouldBlockScroll,
+      menuScrollIntoViewElement,
       menuShouldScrollIntoView,
       noOptionsMessage,
       onMenuScrollToTop,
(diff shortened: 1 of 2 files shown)


skovhus:
@JedWatson are you interested in this suggestion for v3? In that case I can update this PR for version 3, it was build for version 2 back in April.

prateek0227:
@skovhus, regarding this line,
> It does not solve all use cases, as there is still code that checks for available view height on window (`viewSpaceBelow >= menuHeight`), and not inside the container. Here `menuScrollIntoViewElement` does not have any effect (yet). Demo https://codepen.io/skovhus/pen/LvvNJE

Is there any update on adding the functionality of checking the height of the container instead of the window? I'm facing an issue because of this and can only bypass it using some kind of hack by overriding the `componentDidUpdate` of `Select` and then changing the menu's `top` and `bottom` accordingly if it doesn't fit inside the container.

charrondev:
Just because it wasn't cross-linked over: https://github.com/JedWatson/react-select/pull/3830

laukstein:
Why this isn't still released? It is blocking `menuPosition="fixed"` support when scrolled parent element (the dropdown gets dislocated).

ebonow:
@laukstein and everyone else, sorry for the long wait on this.

Currently version 5 is being finalized to replace Flow with TypeScript, so we had to put PR's on hold to avoid further merge conflicts. This and other PR's will be revisited in the coming weeks. Please hold tight.

Issue #3353, opened 2019-01-15
aria-live="assertive" blocks label from being read


rheydt:
Issue:
If you associate a form label with a react-select input, the label is never announced by a screen reader. 

Replication:
Example here: [https://codesandbox.io/s/n0q28p50xp](https://codesandbox.io/s/n0q28p50xp)
Tested using react-select v2, VoiceOver and Safari.

Expected  behavior:
When native select element is associated with a label, a screen reader will read the label when the select receives focus.

Actual behavior:
Instead, when the react-select input receives focus, the screen reader is interrupted and begins immediately reading the contents of the `A11yText` component (which are manually generated by react-select and don't support including label text) because that component uses `aria-live="assertive"`.

Impact:
Without being able to associate a label with the input in a form, this component is not compliant with WCAG 2.0 accessibility guidelines (https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-cues.html)

Possible solutions include:
- Using `aria-live="polite"` instead of `"assertive"` (better practice, anyway, since you don't know what else in the application/website the screen reader might be needing to keep track of)
- Adding a new optional prop where form label text could be passed to the `A11yText` component
- Allowing an `id` prop to be passed to the native input element

afercia:
Worth noting the assertive aria-live region can't be rendered "on the fly". Instead, it needs to be in the DOM since the beginning and stay there to allow screen readers to "monitor" changes within it.

Especially on Windows, screen readers "buffer" the content of a page. Instead, VoiceOver on macOS has a different behavior. This is very likely the reason why testing on different operating systems with different browser / screen reader combinations highlights inconsistent behaviors in the way the live messages are announced:

- macOS Safari VoiceOver: I can reproduce what @rheydt reported:
  - focus the input field
  - the aria-live assertive region is rendered in the DOM
  - because of the "assertive" value, the aria-live region content is announced interrupting any other announcement
  - the input role and its label are not announced

This behavior is more similar to the one of an ARIA `role="alert"` and seems triggered by the way Safari / VoiceOver perceive an aria-live region that is rendered "on the fly" in the DOM.
Instead, aria-live regions are supposed to work differently and only _changes within their content_ should be announced.

- Windows Firefox NVDA: I can reproduce what @zebdelk reported:
  - focus the input field
  - the aria-live assertive region is rendered in the DOM
[...]

PR #3532, opened 2019-04-29
Issue #3353: change aria-live assertive to polite

alisonhall:
The aria-live assertive block the field label from being read, causing it to not be accessible. Changing the aria-live property to polite will allow the label to still be read


alisonhall:
Related issue: https://github.com/JedWatson/react-select/issues/3353

alisonhall:
@gwyneplaine, which branch should I raise this PR to?

JedWatson:
Thanks @alisonhall 

Reading through #3353 we have a lot of work to do, but hopefully this is a good step in the right direction.

PR #3537, opened 2019-05-03
Flow updates

jdelStrother:
I was running into flow difficulties using Async & AsyncCreatable - their HOC functions were losing the defaultProps of the components, so, eg:

```jsx
<Async loadOptions={loadOptions} />
```
would complain with the following:

```
Cannot create Async element because:
 • property cacheOptions is missing in props [1] but exists in AsyncProps [2].
 • property defaultOptions is missing in props [1] but exists in AsyncProps [2].
 • property onInputChange is missing in props [1] but exists in AsyncProps [2].
```

WDYT to this?

I've also changed a couple of string types to React$Node - the children of SingleValue, and the return value from noOptionsMessage.  There shouldn't be any functional code changes, just type changes.


mufasa71:
@jdelStrother I did almost the same here #3538 :smile:

jdelStrother:
@JedWatson anything that me/mufasa71 can do to help get either of these merged?  I'll fix my merge conflicts if you're intending to merge this.

PR #3538, opened 2019-05-04
Fix flow typing

mufasa71:


mitchellhamilton:
Thanks for the PR!! We're going to merge #3537 because it seems like it does the same things in terms of external usage as this PR but this PR makes code changes and we're planning to rewrite some internals soon so avoid unnecessary code changes would be good.

PR #3540, opened 2019-05-04
Fix issue with disabled options and add some docs examples for :active

risenforces:
Fix issue with disabled options

When you hold the mouse on the disabled option, you can see that its background color is being redefined by the ":active" selector. Logically, the user should not have any ability to interact with the disabled options.

Add examples with ":active" to the documentation

At first glance, it is not very clear how to customize active options. In my first experience, I began to look for the "isActive" option in the state, but it was not there. Since the current implementation uses the "hardcoded" selector ":active", it would be good to add the appropriate code to the custom styling examples.


gwyneplaine:
Thanks for this @risenforces, merged

PR #3542, opened 2019-05-05
Replace react-fast-compare with a shallow equal function

mitchellhamilton:
> Note: this is based on the monorepo branch so #3530 should be merged first

I was looking around the codebase and saw that react-fast-compare was only used to compare the components objects so a shallow equal would be just as good and it's smaller.


mitchellhamilton:
Gonna do this later in a new PR because this is too big of a merge conflict

PR #3548, opened 2019-05-10
Misc package.json fixes

mitchellhamilton:


gwyneplaine:
Unit tests appear to be choking on emotion related snapshots @mitchellhamilton, once that's resolved we can merge this in  🎉

PR #3554, opened 2019-05-13
Update Select.js

matanstorr:
If the select component is cleared via the clear indicator, you can't type in the textfield anymore.


JedWatson:
Thanks @matanstorr 

Can you include some information on how this solves the issue you're seeing, or how to reproduce it?

Thanks!

JedWatson:
Aah, I think I've found another PR with the information: #3326

I'll close this as a duplicate and get that one landed

Issue #3559, opened 2019-05-15
Incorrect Korean string composed on IE11


migbot:
With the Korean Microsoft IME set, I'm seeing incorrect strings composed only on IE11. As an example, try typing `gkwjd` in any inputs of https://react-select.com/creatable. On Chrome, the correct string `하정` is composed while on IE11 `핮정` is composed.

PR #3563, opened 2019-05-17
Fix IME composition for non-Chrome browsers

jwilander:
When composition starts the component's state was being updated to indicate that composition was occuring. This seemed to be causing either unnecessary renders or re-calculation that would affect the composition of CJK characters in the input box, resulting in characters or words not matching the input entered by the user on non-Chrome browsers. The fix is to not use component state to track `isComposing` preventing attempts at re-rendering when the composition starts and ends. This is also the proper way to store such data as it is not used in the rendering of the component.

You can reproduce the issue by changing your keyboard to be Korean on Windows and using IE11 to type `gkwjd` into a react-select input. The correct result should be `하정`.

Fixes https://github.com/JedWatson/react-select/issues/3559

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -296,7 +296,6 @@ type State = {
   ariaLiveContext: string,
   inputIsHidden: boolean,
   isFocused: boolean,
-  isComposing: boolean,
   focusedOption: OptionType | null,
   focusedValue: OptionType | null,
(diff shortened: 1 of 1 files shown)


gwyneplaine:
thanks for this @jwilander, merged

jwilander:
Glad to help!

PR #3567, opened 2019-05-20
Fix for incorrect compile check

vadimka123:
Fix check for ensure focus. Now code compiled as (isFocused && !isDisabled && prevProps.isDisabled || isFocused && menuIsOpen && !prevProps.menuIsOpen). See prettier/prettier#187


JedWatson:
Thanks @vadimka123 but I'm not seeing the problem here. With the latest version of prettier, that code is preserved as expected, and the logic will be the same before and after your change.

Are you seeing a different behaviour?

vadimka123:
@JedWatson, I've seen this behavior before
Perhaps it is already fixed by updating prettier to latest version
I will retest and if not repeat, then close this PR

vadimka123:
@JedWatson, it's issue is not fixed 
I'm check on local machine build (version of prettier - 1.17.1), also check builded code from package 
Results is same, boolean logic is damaged

Снимок экрана от 2019-06-03 09-57-54

JedWatson:
@vadimka123 Sorry for the delay, I've worked out what's going on here.

This isn't actually a prettier issue but a babel optimisation, and it's going to behave correctly because the **Logical AND** operator has a higher precedence (6) than the Logical OR operator (5) in JavaScript. Which means this code is safe to leave as-is, and the parenthesis grouping the logic are really just helping with code readability (as discussed on that prettier issue you linked)

reference https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

So I'm going to close this, but I personally learned something in the process, so thank you!

PR #3568, opened 2019-05-20
Fixed isValidNewOption when value is null

vadimka123:


JedWatson:
I'm not clear on why / when this is a problem.

The call to `isValidNewOption` passes the `value` prop through `cleanValue` which will always return an Array.

Implementation here (from `utils.js`)

```js
export const cleanValue = (value: ValueType): OptionsType => {
  if (Array.isArray(value)) return value.filter(Boolean);
  if (typeof value === 'object' && value !== null) return [value];
  return [];
};
```

Is there a logical path I'm missing that allows a falsey value to get through here? What problem is checking for `!selectValue` solving?

vadimka123:
@JedWatson, I have seen this problem for quite some time
Perhaps she fixed
I will recheck and write about results

vadimka123:
Yes, it's fixed

PR #3569, opened 2019-05-20
BuildMenuOptions only when menu is opened

vadimka123:
Very big performance issue: buildMenuOptions only when menu is opened
I'm have page with ~1000 selects with a ~1000 options. When render this page it takes ~ 40-50s and while page either hangs terribly or crashed. After this update page render takes 10s and there are no hangs. It's very need update.


JedWatson:
Thanks @vadimka123 this is a very helpful change. Will be included in the next release.

yuri-scarbaci-lenio:
For anyone having troubles updating from 2.4.4 to latest releases,

this looks like it could be a breaking change depending on how you interacted with the menu ( for example if your logic had a custom menu and a reference to it, your `reference.current` would now be `null` instead of your menu ref as it used to, which is indeed a breaking change )

PR #3589, opened 2019-05-28
Make onMenuScrollToBottom fire on desktop when the scrollbar is mouse-dragged to the bottom

### LINKED ISSUE #3232: onMenuScrollToBottom not fired on desktop when the scrollbar is mouse-dragged to the bottom
Greetings!

Thank you so much for taking time out to maintain `react-select`

One of our users reported that the dropdowns in our product (the ones powered by react-select) were not loading further options when she scrolled to the bottom. On further investigation, we realized that this was because `react-select` does not fire `onMenuScrollToBottom` when the scrolling is performed by dragging the scrollbar to the bottom using a mouse-drag. Unfortunately for us and for this user, her mouse lacks a scroll wheel and so this is the only way she can scroll to the bottom.

Here's a code sandbox demonstrating this issue: https://codesandbox.io/s/x20ppz92xw Note that when scrolled using the mousewheel, and alert dialog is displayed when the menu is scrolled to the bottom. However, when the scrollbar is dragged to the bottom using the mouse, the alert dialog isn't displayed.

Looking at the source code, I can see that the `ScrollCaptor` is only listening for the `wheel`, `touchstart` and `touchmove` events. The use-case of dragging the scrollbar with a mouse is ignored.

I've confirmed that this works in the previous major version (v1.2.1)

cdax:
Closes #3232


M-Yankov:
@cdax  Why this functionality was not integrated ?

Duarte10:
Created PR https://github.com/JedWatson/react-select/pull/4970 with this fix.

PR #3599, opened 2019-06-02
Remove bolt config

mitchellhamilton:
This isn't necessary now that we're using the new version of changesets that works with yarn workspaces


JedWatson:
👍 thanks @mitchellhamilton

PR #3611, opened 2019-06-07
openMenu: Compare option values instead of objects

raphaelbadia:
When using the Select as controlled, it frequently happens that `value` object's reference is different than the options references, leading to this piece of code at line 493 returning -1, while the objects are identical.

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -485,11 +485,12 @@ export default class Select extends Component<Props, State> {
   openMenu(focusOption: 'first' | 'last') {
     const { menuOptions, selectValue, isFocused } = this.state;
     const { isMulti } = this.props;
+    const focusableValues = menuOptions.focusable.map(option => this.getOptionValue(option));
     let openAtIndex =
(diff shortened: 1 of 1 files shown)


raphaelbadia:
I saw that the test fail, if you agree with my pr I'd gladly fix them

Methuselah96:
Thanks for the PR! Unfortunately comparing the `value` (or the `value` received from `getOptionValue`) is not consistent with the rest of the library, since we usually compare the options themselves using strict equality. If you have more specific examples of what breaks in the status quo, feel free to open an issue to explore these problems.

Methuselah96:
EDIT: I'm actually not sure what our policy is on using `value` vs. strict equality so I'm going to reopen this so that I can come back and investigate this some more.

raphaelbadia:
Sorry, I dont recall my case since I opened this 1.5 year ago, and have moved to downshift since then ! 
from what I remember, I had an array of object as options, and the selected item was a similar object (same id, label etc) but from redux so the strict equality failed

raphaelbadia:
Very kind of you to reopen this and investigate further

PR #3630, opened 2019-06-16
fix: Remove unused classnames dependency

eps1lon:
`classnames` isn't used (anymore?) but still listed as a dependency. This might be confusing to people using `clsx` and thinking they pull in (basically) duplicate dependencies.


TrySound:
Friendly ping @JedWatson

JedWatson:
Oops 😊 you're 100% correct, this should have been removed with v2! thanks for the PR @eps1lon

PR #3639, opened 2019-06-20
Implemented a refresh token and fixed  #1581 and fixed #1879

### LINKED ISSUE #1581: How to force reload of options?
I want to reload default async options , when use enter some data in other part of app,but I do not know how to force reload of options.

### LINKED ISSUE #1879: react-select loadOptions question

whizsid:
I want to refresh options in async select relating to another input value. This is the usage:-

```
<Async
                        defaultOptions
                        value={value}
                        onChange={onChange}
                        refreshToken={JSON.stringify(otherValues)}
/>
```

### DIFF
--- a/packages/react-select/src/Async.js
+++ b/packages/react-select/src/Async.js
@@ -23,6 +23,9 @@ export type AsyncProps = {
   cacheOptions: any,
   onInputChange: (string, InputActionMeta) => void,
   inputValue?: string,
+  /* If refreshToken is set and defaultOptions is truthy, All options will
+     refreshing when refreshToken is changed. */
+  refreshToken?: string
 };
 
 export type Props = SelectProps & AsyncProps;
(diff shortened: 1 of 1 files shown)


osahner:
the selected value(s) should be reseted

whizsid:
@osahner Fixed it.

ruaneg:
Would be great to get this merged!

dcousens:
Marking as stale, I don't this pull request could be merged as-is, and significant discussion needs to follow whether this approach is appropriate.

PR #3669, opened 2019-07-06
Expand and scroll container when opening menu

### LINKED ISSUE #1267: Options menu appear outside the viewport for some layouts

### LINKED ISSUE #3473: menuShouldScrollIntoView does not work well inside scroll container
A bug, here is a codesandbox: https://codesandbox.io/s/w0ymoxq827

The issue is that `viewHeight` [is calculated](https://github.com/JedWatson/react-select/blob/71df767105e3af737d17254e2bd524dc6423e005/src/components/Menu.js#L75) from the `window`, which causes problems for select elements inside of other scroll containers.  This can be seen in the provided codesandbox, by changing the height of the embedded browser window, and seeing that the scroll behavior changes depending on that height.  

I've played around in the code enough to find what seems to be a solution, causing the scroll container to scroll when needed, and also constraining the height of the menu to the total visible height of the scroll container, to avoid scrollbars within scrollbars.  I'll work on cleaning up my code and will submit a PR, but I wanted to at least open an issue to explain the problem.

Oh and one other question while I'm at it, why does the scroll container need to be `position: relative;` for `menuShouldScrollIntoView` to work?  It's not immediately obvious that it's necessary, and it was only by digging through the 
source that I figured out why my container was not scrolling.

I believe this is similar to the case reported back in https://github.com/JedWatson/react-select/issues/1267, but that plnkr is broken, so I can't tell for sure.

IanVS:
Fixes #3473
Fixes #1267 

This PR does three main things:

1. Fixes the above-mentioned issues by correctly identifying and using the height of a scroll parent when the select component is inside one.  This means that the menu will no longer be only partially visible and force the user to scroll the parent in order to see the rest.

2. Allows the menu to be sized a bit smaller in cases where it can fit inside the scroll parent without causing the scroll parent to grow.  This is preferable to expanding the scroll parent and causing the select component to jump to a new position, caused by the open menu.

3. Updates cypress to the latest version (I was having a hard time getting the existing version to run correctly in newer versions of chrome), and adds a new e2e test file for menus.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,5 +1,4 @@
 {
   "baseUrl": "http://localhost:8000/cypress-tests",
-  "screenshotOnHeadlessFailure": false,
-  "videoRecording": false
+  "video": false
 }
--- a/cypress/integration/menu_spec.js
+++ b/cypress/integration/menu_spec.js
@@ -1,0 +1,64 @@
+import baseSelectors from '../fixtures/selectors.json';
+
+const selector = {
+  caseDefault: '[data-cy="case-default"]',
+  caseExpandDown: '[data-cy="case-expand-down"]',
+  scrollContainer: '[data-cy="scroll-container"]',
+};
+
+Cypress.Screenshot.defaults({
+  screenshotOnRunFailure: false,
+});
+
+describe('Menus', () => {
+  before(function() {
+    cy.visit('http://localhost:8000/cypress-menu-tests');
+  });
+
+  it('test page renders successfully', () => {
+    cy.get('h1').should('contain', 'Test Page for React-Select Menus');
+  });
+
+  describe('Inside scroll containers', () => {
+    describe('by default', () => {
+      it('menu should be visible when select is clicked', () => {
+        cy.get(selector.caseDefault)
+          .find('.react-select__control')
+          .click()
+          .get(selector.caseDefault)
+          .find(baseSelectors.menu)
+          .should('be.visible');
+      });
+    });
+
+    describe('At bottom of container', () => {
+      it('menu should be visible when select is clicked and should expand container', () => {
+        cy.get(selector.caseExpandDown)
+          .find(selector.scrollContainer)
+          .then(scrollContainer => {
+            expect(scrollContainer).to.have.length(1);
+            const scrollContainerElement = scrollContainer[0];
+            // Start out at top
+            expect(scrollContainerElement.scrollTop).to.equal(0);
+            // Save height so we can tell if it grows later
+            const originalHeight = scrollContainerElement.scrollHeight;
+
+            cy.get(selector.caseExpandDown)
+              .find('.react-select__control')
+              .click()
+              .get(selector.caseExpandDown)
+              .find(baseSelectors.menu)
+              .should('be.visible')
+              .then(() => {
+                // Make sure the container has expanded
+                expect(scrollContainerElement.scrollHeight).to.be.gt(
+                  originalHeight
+                );
+                // And that it's not scrolled to top anymore
+                expect(scrollContainerElement.scrollTop).to.be.gt(0);
+              });
+          });
+      });
+    });
+  });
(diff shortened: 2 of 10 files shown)


IanVS:
Here's the message from CI for the e2e failure.  Any help in fixing this would be appreciated!

```
yarn run v1.9.4
$ concurrently --kill-others --success=first --names 'SERVER,E2E' 'yarn start --progress=false --no-info' 'yarn test:cypress'
[SERVER] $ cd docs && yarn start --progress=false --no-info
[E2E] $ cypress run --spec ./cypress/integration/select_spec.js
[E2E] The cypress npm package is installed, but the Cypress binary is missing.
[E2E] 
[E2E] We expected the binary to be installed here: /root/.cache/Cypress/3.3.2/Cypress/Cypress
[E2E] 
[E2E] Reasons it may be missing:
[E2E] 
[E2E] - You're caching 'node_modules' but are not caching this path: /root/.cache/Cypress
[E2E] - You ran 'npm install' at an earlier build step but did not persist: /root/.cache/Cypress
[E2E] 
[E2E] Properly caching the binary will fix this error and avoid downloading and unzipping Cypress.
[E2E] 
[E2E] Alternatively, you can run 'cypress install' to download the binary again.
[E2E] 
[E2E] https://on.cypress.io/not-installed-ci-error
[E2E] 
[E2E] ----------
[E2E] 
[E2E] Platform: linux (Debian - 8.11)
[E2E] Cypress Version: 3.3.2
[E2E] error Command failed with exit code 1.
[E2E] info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
[E2E] yarn test:cypress exited with code 1
```

IanVS:
If anyone wants to try this out, you can use this commit from our fork, which includes the generated files: https://github.com/nutshellcrm/react-select/commit/c2319432106e956884d9796f5ecd87781a18bc02

IanVS:
If there's a chance this could get merged in, I can spend some time rebasing and getting it ready.  Let me know if y'all will have time to review at some point.

ebonow:
@IanVS Thanks for circling back on this. We'll be exploring some long overdue menu positioning behavior fixes following the release of the TypeScript refactor.

Methuselah96:
@IanVS Hey, we're finally circling back to this. We apologize it took so long to get to this after the work you've put into it. Right now we're planning to use Popper for the menu positioning. The changes you've made to the positioning logic are great, however there are a few things I would like to tweak about it. Given that we're switching to Popper anyway, would you be willing to scale back this PR or create a new PR to just include the new Cypress tests so that we can use those for testing with Popper?

Methuselah96:
@IanVS Also, if you don't have time to work on this at the moment, no worries. I've started https://github.com/JedWatson/react-select/pull/4913 which is inspired by the Cypress tests added in this PR. I can incorporate your tests into that PR and close out this PR if that's easier, but if you want credit for it I'd be happy to review/merge this PR with just the tests when it's ready.

IanVS:
Hi, thanks for the consideration, but it if you're able to take what I started here and run with it, I'm more than happy with that.  I'm stretched a bit thin at the moment, and not using react-select at the moment.  I'm glad that the PR was useful, even if only as a guide or starting point.  Feel free to copy whatever's useful and close this PR.  Best of luck with everything!

PR #3676, opened 2019-07-10
noOptionsMessage

wiesys:
It should be `noOptionsMessage`, not `noOptionsMessageCSS`, because otherwise it throws Flow error:
```Cannot create `Select` element because property `noOptionsMessage` is missing in  `StylesConfig` [1] but exists in  object literal [2] in property `styles`.```

### DIFF
--- a/.changeset/old-radios-hug.md
+++ b/.changeset/old-radios-hug.md
@@ -1,0 +1,5 @@
+---
+"react-select": patch
+---
+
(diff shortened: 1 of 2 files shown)


VasiliKubarka:
@wiesys maybe you can change `loadingMessageCSS?: StyleFn,` to `loadingMessage?: StyleFn,` either?

wiesys:
Sure, no problem, @VasiliKubarka.

PR #3682, opened 2019-07-16
Fix a bug when replacing the input component to be a textarea

lmvco:
When using react-select, we can provide a custom input component [here](https://react-select.com/components). If this custom component is a TEXTAREA, some events will not work properly since react-select is expecting the input component to be an INPUT.
This pull request fixes this behavior.


JedWatson:
Thanks @lmvco

PR #3690, opened 2019-07-23
Manually set 'isLoading' state for async selects

### LINKED ISSUE #2986: [v2] No way to manually set isLoading on async select box

csilkwor:
Fixes #2986

### DIFF
--- a/.changeset/nine-goats-care.md
+++ b/.changeset/nine-goats-care.md
@@ -1,0 +1,5 @@
+---
+"react-select": minor
+---
+
+Add `isLoading` prop support to the AsyncSelect component (see #3690)
--- a/packages/react-select/src/Async.js
+++ b/packages/react-select/src/Async.js
@@ -23,6 +23,7 @@ export type AsyncProps = {
   cacheOptions: any,
   onInputChange: (string, InputActionMeta) => void,
   inputValue?: string,
+  isLoading: boolean
 };
 
 export type Props = SelectProps & AsyncProps;
@@ -31,6 +32,7 @@ export const defaultProps = {
   cacheOptions: false,
   defaultOptions: false,
   filterOption: null,
+  isLoading: false
 };
 
 type State = {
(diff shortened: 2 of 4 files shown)


csilkwor:
There is a related pr already open for this issue, but it has been dormant and sitting with a merge conflict for a few months.

It is necessary for my use case to be able to explicitly set the loading property of an asynchronous select. This change keeps the current functionality intact while allowing the loading indicator to be displayed at the discretion of the developer.

comfroels:
@JedWatson Is it possible for someone to maybe take a look at this?

oudoulj:
I have an Async select and would like to **always** hide the _"Loading..."_ and the _3 dots_ on the right while typing and there is no result yet.
How do I do that ?
Thanks!
Async select

oudoulj:
> I have an Async select and would like to **always** hide the _"Loading..."_ and the _3 dots_ on the right while typing and there is no result yet.
> How do I do that ?
> Thanks!
> Async select

I am replying to myself.
I was able to hide both the Loading indicator and its 3 dots on the right by using the styles mentioned [here](https://react-select.com/styles)
Something like that :
```
<AsyncSelect
        ...
        styles={myStyles}
/>

export const myStyles = {
        ...
        loadingIndicator: () => {
		return {
			display: "none"
		};
	},
	loadingMessage: () => {
		return {
			display: "none"
		};
	}
};
```

RobertDaleSmith:
@JedWatson Please merge. I could really use this so that I can show the loading indicator on AsyncCreatableSelect while onCreateOption is fetching.

JedWatson:
Looks good, thanks @csilkwor! sorry for the delay merging. Will ship with the next release!

Mujaddadi:
I am still facing this issue. Not sure it is released yet or not.

sirjuan:
Could this please be released @JedWatson?

laznic:
@JedWatson could you do a release with this fix included please?

makker:
It would be greatly appreciated if this could be be released. Thanks for the awesome work with the project!

mika-kaki:
This fix would be greatly appreciated! Thanks for a great project!

yannbriancon:
I agree with all the previous comments, I'd love to see this released! 
Always cool to work with this project!

PR #3694, opened 2019-07-25
Confine group type, prevent confusion of group and option

KyoungWan:
- Problem
When I use data like 
`
tempOptions: [
    id: number
    ...
    options: { id: number, type: 'temp', ... }
]
`

Function 'bulidMenuOption' in src/Select.js confuse my 'option' as 'group' data.'
'bulidMenuOption' detects tempOptions's attribute 'options' and it regard tempOptions as 'group' not as 'option'
So it maps tempOptions.options and raise Error 'items.map is not a function'

- Reproduce of error
https://codesandbox.io/s/react-codesandboxer-example-e62ro

- Suggestion
I think it's because GroupType is subset of OptionType
So Confining GroupType with 'label' attibute can would make it more safe.

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -1312,7 +1312,7 @@ export default class Select extends Component<Props, State> {
 
     return options.reduce(
       (acc, item, itemIndex) => {
-        if (item.options) {
+        if (item.options && item.label) {
           // TODO needs a tidier implementation
           if (!this.hasGroups) this.hasGroups = true;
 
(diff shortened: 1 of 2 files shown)


JedWatson:
Hi @KyoungWan sorry the detection of `options` is causing you difficulties.

Unfortunately, this is the only required key that determines the difference between a group and an option; switching to have an expectation of the `label` prop is not safe because options explicitly may not have a `label` prop. Users can change the properties of options that are used to generate the label and value determined by the shape of their data.

For better or worse we decided that the only key we *would* have an opinion about is `options`

Effectively, you can consider this a reserved keyword. As a workaround, I would suggest changing your data so that your options contain a `_options` property (or similar) by transforming the data before it is provided to the `<Select />` (and after the value is changed, if necessary)

PR #3707, opened 2019-08-07
Changing singleValue to not use absolute positioning

seancurtis:
This allows the parent elements to make use of flex-grow/shrink (absolute positioning is removed from the flow)

### DIFF
--- a/docs/examples/BasicMulti.js
+++ b/docs/examples/BasicMulti.js
@@ -11,5 +11,6 @@ export default () => (
     options={colourOptions}
     className="basic-multi-select"
     classNamePrefix="select"
+    placeholder="Choose a colour"
   />
 );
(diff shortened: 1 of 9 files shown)


seancurtis:
I have a codesandbox showing the changes. It can be achieved using the "styled" prop, but I think this would make more sense as the default so that this works more like a normal block of content. https://codesandbox.io/s/react-codesandboxer-example-ug6ci

artooras:
What is the status with this update? I just raised a [question on SO](https://stackoverflow.com/questions/62324546/react-select-to-stretch-to-fit-value) which I think is related to this proposal.

ebonow:
Greetings @seancurtis and @artooras ,

One of the things holding this up is that it presents possible breaking changes since it impacts differences in styling. With the TypeScript rewrite underway, we will be launching a new major version which affords the opportunity to address small possible breaking changes such as this.

I would say that this (or some similar implementation) will likely be a looked at as a version 5 candidate.

PR #3722, opened 2019-08-19
Updated Option Component for better styling

KushalKatta:
Changed Color and Background Color for Option Menu when an option isFocused.


JedWatson:
Thanks for the PR @KushalKatta 

I appreciate the design input but this would be an expected, breaking change for anyone who's using the component without custom styling, and it's mainly a design opinion rather than a fix so I'm going to have to decline it.

PR #3645, opened 2019-06-23
Add a working example of drag-drop sortable MultiSelect values

gunn:
This uses the `react-sortable-hoc` package and is the same approach as #2284. It only makes two small line change tos `react-select` itself.

I think with this we could close #1606, #2284, and #3223.

PR #3727, opened 2019-08-21
Add index for multi select values

tonytangau:
This change is for users with drag and drop requirements.

### DIFF
--- a/.changeset/khaki-balloons-behave.md
+++ b/.changeset/khaki-balloons-behave.md
@@ -1,0 +1,5 @@
+---
+"react-select": patch
+---
+
+Adding an `index` prop to `MultiValue` components
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -1480,7 +1480,7 @@ export default class Select extends Component<Props, State> {
     }
 
     if (isMulti) {
-      const selectValues: Array<any> = selectValue.map(opt => {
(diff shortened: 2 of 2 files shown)


tonytangau:
Created a drag and drop multi select example based on this PR: [sortable-select](https://github.com/tonytangau/sortable-select)

ttttsai:
This looks great, could we merge this? @JedWatson @gwyneplaine

kroeberk:
Please merge this. This is a great and simple implementation.

JedWatson:
This looks good, thanks @tonytangau (and sorry for the delayed review!)

If you're up for it, adding your example from your [sortable-select](https://github.com/tonytangau/sortable-select/blob/master/src/App.js) repo to the docs would be a great addition to help people understand how to use this feature.

Please open a PR for the "advanced" page if you're up for it!

gunn:
Glad we have this in now.
@JedWatson my PR, which this one comes from, has an example added to the docs too - #3645

JedWatson:
Hey @gunn yeah, I found that after merging this :) working to pull your PR into master now, thanks for that!

gunn:
Awesome. Thank you @JedWatson.

loopmode:
Hey guys. Thanks for adding the index, it really helps! (And not only in drag+drop scenarios :))

Not sure whether this is the right place but it would be great to add this to `@types/react-select` as well! (in `MultiValue.d.ts`)
Is somebody here affiliated with the types package?

PR #3742, opened 2019-08-29
clearInputOnSelect prop

chuvikovd:
### clearInputOnSelect prop

This PR introduces new prop `clearInputOnSelect`. Used alone it does pretty much nothing, but with `isMulti` and `closeMenuOnSelect={false}` it allows user to easily select multiple entries, matching his search and not typing query after each selection.

Example GIF


mitchellhamilton:
Thanks for the PR! Unfortunately we can't accept this because we try to avoid adding props for behavior that can be implemented in userland since it makes react-select harder to maintain. Here's an example implementation: https://codesandbox.io/s/react-codesandboxer-example-t43q0

PR #3749, opened 2019-09-05
Double-tap is needed to select an option in mobile

sgwanlee:
Thanks for building great things. With react-select, I can reduce a lot of works.
In react-select@3.0.4, I found that double tab is needed on mobile to select an option.
I used fastclick.js to resolve this issue, but there's some side-effect with fastclick.
So I added onTouchEnd event handler just the same as onClick event except it is disabled when a user is dragging. It resolved my issue.

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -1296,6 +1296,7 @@ export default class Select extends Component<Props, State> {
         innerProps: {
           id: optionId,
           onClick: onSelect,
+          onTouchEnd: () => {!_this3.userIsDragging && onSelect()},
           onMouseMove: onHover,
           onMouseOver: onHover,
           tabIndex: -1,


JedWatson:
@sgwanlee thanks for the PR

Can you provide some more information about this? I can't reproduce it (iOS 13) but the issue may be specific to other mobile OS / browser combinations.

I'm hesitant to merge this as-is because it could lead to `onChange` being called twice with an identical value and action meta, which could lead to unexpected behaviour for users (at least, could be considered a breaking change)

May need some additional workarounds to prevent that happening and make it safe

pachuka:
@JedWatson - I actually just ran into this scenario, seems specific to when `menuPosition="fixed"`. Removing that prop fixed the issue for me so might just be the combination of that interaction.

Tested on iPhone 11 (13.4.1)

bladey:
Hi @sgwanlee, I'm now going to close this PR due to inactivity. If you have any further feedback as requested by Jed, please do let me know and we'll re-open and assess, thank you!

PR #3752, opened 2019-09-11
Add listeners to ScrollCaptor, even when no scroll is available on mount

Kemosabert:
This is the same PR as #2861.

The issue is that when you work with async data, when opening the list of options, the options list might be empty and data may arrive later. This results in the `onMenuScrollToBottom` listener not firing, or being attached.


jossmac:
Closing in favour of #2861

Issue #1530, opened 2017-02-06
Add an offset to logic of onMenuScrollToBottom


omriavigdor:
I suggest exposing a prop to the user to set a pixel-offset for when onMenuScrollToBottom is called.
Meaning, suppose I want it to fire 20 pixels before the user scrolls to the bottom of the menu, then I set:
`menuScrollToBottomOffset={20}`

It will be very helpful if I can do my logic before the user gets to the bottom because if they do it fast enough the whole page is scrolled down (my case is implementing infinite scroll).

brayhoward:
Would pay money for this feature, literally

lukebennett88:
Thanks for opening this issue.
In an effort to stay on top of open issues, we're moving feature requests to discussions.

PR #3757, opened 2019-09-12
Add an optional offset for onMenuScrollToBottom

jackspiva:
Adds an optional `onMenuScrollToBottomOffset` prop that will cause `onMenuScrollToBottom` to fire a certain amount of pixels before the user scrolls to the bottom fo the menu. 

I think this makes the `onMenuScrollToBottomProp` way more useful. I am currently using this change in a fork of `react-select` in a work project. 

My use case is that I am using `onMenuScrollToBottom` to load more options for the select menu. Without an offset, the user has to hit the bottom of the menu and then wait for more options to load. However, with an offset, the options can be loaded before they reach the bottom, which is a lot smoother.

Addresses #1530.


anushaNemilidinne:
@jackspiva I am using a similar approach where if the user scrolls to the bottom of the drop down, it fetches more data from the server. But the problem is the scroll bar goes to top instead staying at the bottom. I know it's because the react component re renders on new state change. But is there a way to achieve this functionality. I tried lot of different ways but no luck. Any help will be greatly appreciated! Thanks in advance!

Issue #2937, opened 2018-08-15
Select accessiblity issues for JAWs and NVDA


jquense:
Great job ya'll with the recent efforts to make react-select more accessible. It performs great in VoiceOver! Unfortunately the strategy doesn't seem to work for the two most commonly used SR's JAWs and NVDA. 

For both, the live region alerts are not read aloud when you enter or select items in the list. This ends being mostly a silent black box for these SR's, and the input doesn't seem usable :/

I will admit i'm not the best SR user, so some of the issue may be me, but i am not a complete novice either. I'm also trying to com up with a good a11y strategy for these sorts of inputs for the next iteration of react-widgets!

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our efforts towards the current major version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue  is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

jquense:
Still relevant last I checked and important :) 

Thanks!

bladey:
Thanks @jquense!

ebonow:
Greetings @jquense ,

Would you kindly verify if this this is still an issue for you in version 4.2.1?

I spent a bit of time refactoring the aria-live implementation and was sure to render the region on mount, but was not personally able to do any testing outside of VoiceOver.

jquense:
@ebonow thanks for working on this, I unfortunately don't have time to verify this, but if you or anyone else wants to test with NVDA, Microsoft provides free VM's for testing stuff (like old internet explorer), In the past i've used them to test with NVDA from my Mac

ebonow:
@jquense I've been using virtual box but it seems laggy so I may have to reorganize some things to get the space for Bootcamp. There are two developers who have begun taking on accessibility testing that did verify the changes work on JAWS and NVDA, but wanted to verify with you in case you had a different experience.

I'll mark this as resolved especially since the linked PR which would have closed this PR was implemented in a similar way.

jquense:
awesome thanks for all the hard work 👍

Watso196:
Hey there! Curious if there are any plans to address the issue listed here? My team is still seeing this issue in the React Select and are wondering if this non-live region content should still be included with the update to role="combobox"?

Issue #3675, opened 2019-07-09
NVDA has issues reading options


kylehalleman:
NVDA can read options very oddly, even breaking apart words or omitting parts of words altogether. Using the [first example on the homepage](https://react-select.com/home) ("Single"), here is the NVDA / Firefox output from initial focus through using <kbd>↓</kbd> to navigate through the list.

```
blank
blank
option Ocean focused, 1 of 10. 10 results available. Use Up and Down to choose options, press Enter to select the currently focused option, press Escape to exit the menu, press Tab to select the option and exit the menu.
blank
Blue focused disabled, 2 of 10. 10 results available. Use Up and Down to choose options
blank
Purple focused, 3 of 10. 10 results available. Use Up and Down to choose options, press Enter to select the currently focused option
blank
Red
4
blank
Orange
5
blank
Yellow
6
blank
Green
7
blank
Fo
st
8
blank
Slate
9
blank
i
ver
10
```

VoiceOver / Safari reads this sequence as expected.

agonzalezcr:
did you find a work around for this  @kylehalleman  ?

kylehalleman:
@agonzalezcr Unfortunately no. I did try shortly after filing this issue to fix it in the library to eventually make a PR. Didn't get too far though.

ebonow:
Greetings @agonzalezcr  and @kylehalleman , 

I am working on a fairly large PR to address many of the accessibility and aria-live issues.

It would be appreciated if someone could test these changes against JAWS or NVDA as all of my testing has been against OSX VoiceOver.

PR: https://github.com/JedWatson/react-select/pull/4414
Git branch: https://github.com/ebonow/react-select/tree/aria-messages-v4

kylehalleman:
Unfortunately I've changed jobs since I created this issue and no longer have access to NVDA to be able to test this. I'll see if there's some way I can get a setup going to test this and get back to you.

ebonow:
@kylehalleman thank you and no pressure. Jed has mentioned that there will be another internal member added to the team soon that will be focusing on accessibility, but again just happy to address any issues and welcome any assistance the community is able to provide.

mellis481:
@ebonow Your work in #4414 did not address the "blank" problem reported in this issue and [others](https://github.com/JedWatson/react-select/issues/5121). Is there any update?

PR #3781, opened 2019-09-27
Fixes for aria-live region

Malgalad:
A11y implementation fixes based on [issue 3353](https://github.com/JedWatson/react-select/issues/3353#issuecomment-480823396).

- aria-live region is rendered in the DOM since the beginning
- aria-live region is not re-rendered, only its content changes
- aria-atomic="true" is required for NVDA+FF so that screen reader reads the whole region and not just changed characters
- aria-relevant="additions text" is, [in theory](https://www.w3.org/TR/wai-aria-1.1/#aria-relevant), a default value if not specified, however, it doesn't work unless actually specified
- unique keys for paragraphs are for fixing a bug with screen reader reading updates twice. After looking through the similar issues, I've come to the following possible explanation: browser updates text in 2 nodes, so two "text changed" updates are dispatched; with unique keys React remounts nodes instead of updating text inside, so only one update "two new nodes added" is dispatched (and "two old nodes removed" is ignored due to aria-relevant)

Also may fix https://github.com/JedWatson/react-select/issues/2937, https://github.com/JedWatson/react-select/issues/3675

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -1787,11 +1787,17 @@ export default class Select extends Component<Props, State> {
   }
 
   renderLiveRegion() {
-    if (!this.state.isFocused) return null;
+    const { ariaLiveSelection, isFocused } = this.state;
+    const ariaLiveMessage = this.constructAriaLiveMessage();
+
     return (
-      <A11yText aria-live="polite">
-        <p id="aria-selection-event">&nbsp;{this.state.ariaLiveSelection}</p>
-        <p id="aria-context">&nbsp;{this.constructAriaLiveMessage()}</p>
+      <A11yText aria-live="polite" aria-relevant="additions text" aria-atomic="true">
+        {isFocused && (
+          <React.Fragment>
+            <p id="aria-selection-event" key={ariaLiveSelection}>&nbsp;{ariaLiveSelection}</p>
+            <p id="aria-context" key={ariaLiveMessage}>&nbsp;{ariaLiveMessage}</p>
+          </React.Fragment>
+        )}
       </A11yText>
     );
   }
--- a/packages/react-select/src/__tests__/Select.test.js
+++ b/packages/react-select/src/__tests__/Select.test.js
@@ -584,7 +584,7 @@ cases(
     focusedOption,
   }) => {
     let onChangeSpy = jest.fn();
-    props = { ...props, onChange: onChangeSpy, menuIsOpen: true, hideSelectedOptions: false, isMulti: true, menuIsOpen: true };
+    props = { ...props, onChange: onChangeSpy, hideSelectedOptions: false, isMulti: true, menuIsOpen: true };
     let selectWrapper = mount(<Select {...props} />);
 
     let selectOption = selectWrapper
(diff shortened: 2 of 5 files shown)


forgo:
Looking forward to this fix! =)

anthonyvialleton:
@Malgalad Any news about this PR?

Actually I also found that `renderLiveRegion` from : https://github.com/JedWatson/react-select/blob/3c7de0de52826fe74d303a01475c43fe88256156/packages/react-select/src/Select.js#L1800-L1803 is responsible for W3C validation to break because of a `span` element wrapping the 2 `p`.

Could we change the `p` by `span`s instead?

Malgalad:
@anthonyvialleton I'm assuming the ball is in the @JedWatson's court. Updated the PR with your suggestions.

anthonyvialleton:
@Malgalad It was not needed, as mentioned I already created a PR for that specifically ;)

AaronV:
Hi @Malgalad, I'm really grateful that you've put this work in on this! I had been trying to wrap custom-components with the necessary pieces to properly support aria and wcag, but no luck so far. Getting proper JAWS support is important to me.

I'd like to try out your PR in my project but am having trouble installing the module and getting it to run.

I've installed the module using yarn, and it appears in my `dependencies` as `"@react-select/monorepo": "https://github.com/Malgalad/react-select.git",`.
I'm trying to import the package using `import Select, { components } from '@react-select/monorepo/packages/react-select';`

But I'm guessing I've missed some sort of build-step. I didn't see, in the docs, how to go about doing that. Can you help point me in the right direction?

agonzalezcr:
is there another possible solution for this issue besides waiting until this merge? 
Thanks

asharron:
Is there anything I can do to help get this through? 🙂 

I think this may include some fixes for some issues I'm seeing with Jaws, so would be happy to help in anyway I can

Methuselah96:

ebonow:
Greetings, 
Thank you for contributing this and your patience. It is on our radar and we're also looking at  https://github.com/JedWatson/react-select/pull/4161 for consideration and it appears there may be some merge conflicts. 

Does anyone have concerns or an idea about what a proper merge would look like so we can include this in a near future release?

JedWatson (approved):

Methuselah96 (changes requested):
@Malgalad Thanks for the PR! The unit tests are failing at the moment because they use `enzyme` which was removed in https://github.com/JedWatson/react-select/pull/3989. Can you update the tests to use `@testing-library/react` instead of `enzyme` when you get a chance?

JedWatson:
Moving this to 3.3 so we can get the tests updated, 3.2 is about to be released

ebonow:
Greetings @Malgalad ,

I have tested your changes and the proposed aria-relevant and aria-atomic changes suggested here introduces a new behavior in that a selected option is now prepended to every new aria live message.

Screen Shot 2021-02-01 at 4 43 59 PM

It's likely that two aria-live regions, may need to be present to replicate the desired behavior. I am working on a new PR that moves all of the aira-live messaging to its own component with the hopes of better customizing the accessibility of this component. For instance, some may want to remove all aria live messaging while others may want to set the aria text to be "assertive" so that the user is not blocked by the default contextual information about "You are currently on a text input..."

hprobotic:
> Hi @Malgalad, I'm really grateful that you've put this work in on this! I had been trying to wrap custom-components with the necessary pieces to properly support aria and wcag, but no luck so far. Getting proper JAWS support is important to me.
> 
> I'd like to try out your PR in my project but am having trouble installing the module and getting it to run.
> 
> I've installed the module using yarn, and it appears in my `dependencies` as `"@react-select/monorepo": "https://github.com/Malgalad/react-select.git",`.
> I'm trying to import the package using `import Select, { components } from '@react-select/monorepo/packages/react-select';`
> 
> But I'm guessing I've missed some sort of build-step. I didn't see, in the docs, how to go about doing that. Can you help point me in the right direction?

Hey have you managed to install?

PR #3794, opened 2019-10-02
Convert class components that don't have to be class components to function components to reduce bundle size

mitchellhamilton:


JedWatson:
Thanks @mitchellhamilton

PR #3797, opened 2019-10-03
Enable Babel loose mode to improve bundle size

mitchellhamilton:
It's super safe, I use it on Emotion and things. The savings are about 900 bytes.


JedWatson:
This looks good @mitchellhamilton, thanks

Issue #3822, opened 2019-10-21
Position of select menu not based on scrollable parent


prateek0227:
As mentioned in PR [3531](https://github.com/JedWatson/react-select/pull/3531), 
> It does not solve all use cases, as there is still code that checks for available view height on window (viewSpaceBelow >= menuHeight), and not inside the container.

I'm facing the similar issue when the menu is present near the bottom of the view area inside a scrollable div, in my case a slide panel, the menu's positioning is being based on the window height and not the scrollable parent. 
The following are the issues that I'm facing:-
# The menu is being truncated if there is some other container at the bottom of the page but the menu can fit inside the window and is being placed at the bottom.
# The scrollable div is not being scrolled to bring the menu into the view.
# If I use `menuPosition = "fixed"`, I'm able to remove the truncation problem, but that brings another issue as the position of menu is not being changed if the parent div is scrolled up or down.

I'm working on a fix for the `getMenuPlacement` function inside `Menu.js`, and trying to change the calculations based on the scrollable parent instead of using the `window` every time. If there is no scrollable parent, then the default thing will be working as is.

prateek0227:
I'm getting this error while running `yarn start`

```
ERROR in ./node_modules/react-select/src/utils.js 3:14
Module parse failed: Unexpected token (3:14)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| // @flow
|
> import { type ElementRef } from 'react';
| import type {
|   ClassNamesState,
 @ ./App/PageNav.js 13:0-58 81:8-24
 @ ./App/index.js
 @ ./index.js
```

Can anyone help me with this?

prateek0227:
@JedWatson @mitchellhamilton do you guys have any plan of adding this functionality if I send a PR? I'm still working on a generic fix that works for a scrollable div as well as the default html element but it would not make sense if it won't be merge. Will be waiting for your reply.

bzalasky:
This is pretty similar to the issue I described in [3810](https://github.com/JedWatson/react-select/issues/3810) (not quite the same though). We're using fixed menu positioning and configured the dropdown to close on scroll.

prateek0227:
[...]

PR #3830, opened 2019-10-24
Fixed menu position based on scrollable parent

prateek0227:
This fixes [3822](https://github.com/JedWatson/react-select/issues/3822), as previously the positioning of the menu was based on the window height and the menu was being truncated if it could fit inside the window but not the scrollable parent. 
But now, the position will be based on the height and boundaries of the closest scrollable parent element, like in the case of a slide panel or a dialog box, where the panel/dialog can be placed anywhere in the window which limits the area where the menu can be viewed.

### DIFF
--- a/packages/react-select/src/components/Menu.js
+++ b/packages/react-select/src/components/Menu.js
@@ -20,9 +20,9 @@ import {
 } from '../utils';
 import type {
   InnerRef,
-  MenuPlacement,
-  MenuPosition,
-  CommonProps,
+    MenuPlacement,
+    MenuPosition,
+    CommonProps,
 } from '../types';
 import type { Theme } from '../types';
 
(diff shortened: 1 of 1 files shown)


prateek0227:
@charrondev yeah I forgot to mention that we have to add `position: relative` to the parent. I too had to use this in order to fix the issue on my end as well. I've updated the description now as well.

prateek0227:
@charrondev I've added the changes to `getScrollParent`. Let me know if anything else needs to be changed.

IanVS:
I also opened a PR with a fix and tests for this issue months ago, and it got no attention from the maintainers.  I'd be curious to compare our two approaches, but I don't really have the time or energy right now.  https://github.com/JedWatson/react-select/pull/3669

charrondev:
If @JedWatson is unavailable is there another active maintainer that could take a look at this? Going by the commits, the next person seems to be @gwyneplaine

prateek0227:
@flexdinesh I saw that you have started to look into stuff in this repo, so, whenever you get some time, can you please take a look at this PR or ask someone else to merge this? Would really appreciate your help. Thanks!

htech9 (approved):

htech9:
The branch contains conflicts, if some with write access could help here.
(This is a very old PR but the content seems relevant)

PR #3850, opened 2019-11-10
Add selectProps documentation

codercodingthecode:
Add an explanation/example in the documentation on how to use ```customStyles``` in a clear way for building dynamic custom Select components and how to access those props inside the second argument ```state``` .

https://github.com/JedWatson/react-select/issues/3851


JedWatson:
Thanks @codercodingthecode! Much appreciated.

Issue #3865, opened 2019-11-22
make <Select>'s Component defaultProps public


katalonne:
It would be nice if [defaultProps](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.js#L248) would be exported globally.

Use case: 
- Using this module as a third-party and therefore need to extend its DefaultProps

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our efforts towards the current major version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue  is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

PR #3866, opened 2019-11-24
Exported SelectDefaultProps

katalonne:
#3865


mitchellhamilton:
Thanks for the PR @katalonne! Unfortunately we can't accept this PR because you can already access the default props from `Select.defaultProps` and we try very hard to not increase the already huge API surface area of react-select as much as possible.

Issue #3656, opened 2019-06-27
focused option starting at the selected option


csantos1113:
Seems that on **v3.0.0** there was a change introduced in the way the focused option is determined.

✅ **Here is an example of how it worked on v2.4.4: https://codesandbox.io/s/react-codesandboxer-example-sywkk**

![2019-06-27 15 26 48](https://user-images.githubusercontent.com/9648559/60294750-13024680-98f0-11e9-9e05-a201b41b376c.gif)

- Selected option ==> Blue background ==> "Purple" option
- Focused option ==> Light blue ==> Starts in "Purple" option and you could move from there.

Which is the same as the behavior of the native select inputs:
![2019-06-27 15 33 24](https://user-images.githubusercontent.com/9648559/60295145-03373200-98f1-11e9-9583-dcf98c8df418.gif)

----

❌ **Here is the same exact example running on v3.0.4: https://codesandbox.io/s/react-codesandboxer-example-3cutx**

![2019-06-27 15 30 52](https://user-images.githubusercontent.com/9648559/60294999-a471b880-98f0-11e9-96d5-828ad62514b8.gif)

- Selected option ==> Blue background ==> "Purple" option
- Focused option ==> Light blue ==> Starts in "Ocean" option.

----

Not sure if this is an issue, or if it was a change intentional made.
But, how can I get the focused option starting at the selected option?

ankush29:
solved #bb9c1d7

csantos1113:
Thanks @ankush29 - will be fixed on PR https://github.com/JedWatson/react-select/pull/3652

guiyep:
Hello, 

I still this error happening in 3.0.4. Any idea when this is going to be merged?

Thanks for the good work!
Guiyep

dicrescenzoa:
> Thanks @ankush29 - will be fixed on PR #3652

Waiting for this fix 👍 as well

imnikij:
Hey @csantos1113 I am getting same error on v2.4,4, are you sure this occurs only for v3.0.0 and above ? Any fix for this ?

guiyep:
Hello, Any update on this?

I still see this happening in 3.0.8.

Thanks!

carlosriveroib:
yes, still happening in 3.0.8. Will you update this?

nsobolev:
Go update, please! :)

Gabriellvs60:
property theme defines the option colors...

export const StyledDropdown = styled(Select).attrs({
  menu: true,
[...]

Issue #3375, opened 2019-01-22
defaultMenuIsOpen doesn't trigger onMenuOpen


lucasveigaf:
When setting `defaultMenuIsOpen`, `onMenuOpen` isn't triggered on the initial opening. See: https://codesandbox.io/s/lyzx95oo37

I need this because I do some fetching and calculations based on the opening of the menu. Perhaps this behaviour is as intended, but if so, it should probably be documented.

Workaround: set both `autoFocus` and `openMenuOnFocus` props, which has a similar effect and triggers `onMenuOpen`.

lecstor:
I have a button displaying a value which when clicked should be replaced by select with the menu open.
I expected `defaultMenuIsOpen` to do this for me but it doesn't. Or maybe it partially does bc it then takes 2 clicks on the select to open the menu. Not sure if this is possibly related to use of `menuPortalTarget` somehow.

Regardless, @lucasveigaf's workaround works for this too.

akrigline:
This also affects somehow whether the menuPlacement prop is taken into affect with `defaultMenuIsOpen` set to true.

I had a `menuPlacement` of auto but with the `defaultMenuIsOpen`, it wasn't working. Using the above workaround does work though.

codercodingthecode:
My issue is that I render 2 selectors after depending on a previous selector option.
The intended functionality is to auto open the first of the two new rendered selectors, however, somehow it doesnt work with either of the options above neither with the default menu is open prop.

lukebennett88:
Unfortunately changing this behaviour now would be a breaking change, so if this is still an issue for people, I'd recommend controlling the component yourself.
We will take this into consideration for a future release.

PR #3868, opened 2019-11-25
Fix not focusing the selected value on menu open

Tirzono:
**Bug**

When opening the react select menu, the focused option is not scrolled into the view:

https://codesandbox.io/s/autumn-wind-l3ed8

**Fix**

The first effort was made in PR #3813, but since the author of the PR is not responding anymore, I decided to have a look myself and try to fix this issue. It turns out that commit 8ee3cea73620b9ddff36f6e57f694c3b19d9f05a in PR #3569 was breaking the autofocus on menu opening, as the `menuOptions` in the `openMenu` was empty, because it was only set **after** `onMenuOpen` was called (and thus `UNSAFE_componentWillReceiveProps`). Now we set the `menuOptions ` in `openMenu` already.

I've made `buildMenuOptions` memoized, so that it does not run the function again if the props that the function uses didn't change.

### DIFF
--- a/.changeset/angry-cherries-nail.md
+++ b/.changeset/angry-cherries-nail.md
@@ -1,0 +1,5 @@
+---
+'react-select': patch
+---
+
+Fix for not focusing the selected value when the menu opens
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -367,6 +367,17 @@ export default class Select extends Component<Props, State> {
       'react-select-' + (this.props.instanceId || ++instanceId);
 
     const selectValue = cleanValue(value);
+
+    this.buildMenuOptions = memoizeOne(
+      this.buildMenuOptions,
+      (newArgs: any, lastArgs: any) => {
+        const [newProps, newSelectValue] = (newArgs: [Props, OptionsType]);
+        const [lastProps, lastSelectValue] = (lastArgs: [Props, OptionsType]);
+
+        return isEqual(newSelectValue, lastSelectValue)
+          && isEqual(newProps.inputValue, lastProps.inputValue)
+          && isEqual(newProps.options, lastProps.options);
+      }).bind(this);
     const menuOptions = props.menuIsOpen
       ? this.buildMenuOptions(props, selectValue)
       : { render: [], focusable: [] };
@@ -434,8 +445,8 @@ export default class Select extends Component<Props, State> {
       this.scrollToFocusedOptionOnUpdate
     ) {
       scrollIntoView(this.menuListRef, this.focusedOptionRef);
+      this.scrollToFocusedOptionOnUpdate = false;
     }
-    this.scrollToFocusedOptionOnUpdate = false;
   }
   componentWillUnmount() {
     this.stopListeningComposition();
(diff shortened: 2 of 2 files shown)


Tirzono:
I am not experienced enough with Flow and I have no clue how to solve the flow check. It's probably something simple, so let me know how to fix this issue.

Tirzono:
I can remove the memoizing part to make the build pass, if that would speed up the merging and reviewing of this PR.  It would be a bit more inefficient, but it would fix the bug.

mathieulj:
The flow errors can be resolved as follows (* note this may not follow best practices):

```diff
-      (newArgs: [Props, OptionsType], lastArgs: [Props, OptionsType]) => {
-        const [newProps, newSelectValue] = newArgs;
-        const [lastProps, lastSelectValue] = lastArgs;
+      (newArgs: any, lastArgs: any) => {
+        const [newProps, newSelectValue] = (newArgs: [Props, OptionsType]);
+        const [lastProps, lastSelectValue] = (lastArgs: [Props, OptionsType]);
```

```diff
-  buildMenuOptions(props: Props, selectValue: OptionsType): MenuOptions {
+  buildMenuOptions = (props: Props, selectValue: OptionsType): MenuOptions => {
...
-  }
+  };
```

Tirzono:
I can see why it would not follow the best practices, but it at least it passes the build now. Up to @JedWatson to decide if he can live with this. Thank you for your help @mathieulj.

Tirzono:
Or maybe @mitchellhamilton wants to have a look at it? :)

Tirzono:
Hi @flexdinesh, as per your comment in #3910, would you mind have a look at this PR? :)

berghall:
This is a must-have!

akomatsu-indeed:
@JedWatson @Tirzono , any movement here? this bug is blocking us from upgrade to 3.0.8. And 3.0.3 is using componentWillReceiveProps() which is bad too.

Tirzono:
@akomatsu-indeed unfortunately not. The PR is ready to be merged. @flexdinesh mentioned in #3910 that he will have a look at it, but since then it's silent again. For now you can use [react-select-reborn](https://github.com/marcelltoth/react-select-reborn) (as discussed in #3910 too):

```
npm i react-select-reborn
```

The fix has been merged to that fork. Until the maintainers will be back, please use that fork.

akomatsu-indeed:
Thanks a lot @Tirzono

flexdinesh:
@Tirzono Thanks for the PR. As mentioned in the other page. We're still ramping up the backlog and you can see more movement in the repo going forward. I've merged this now. I'll add a note here once the patch is released to npm. Thanks again for your patience.

flexdinesh:
@Tirzono This fix has been released in version `3.1.0`. Thanks again!

https://github.com/JedWatson/react-select/blob/master/packages/react-select/CHANGELOG.md#310

Tirzono:
Thank you @flexdinesh, much appreciated!

andreme:
Not working when using defaultMenuIsOpen or menuIsOpen: https://codesandbox.io/s/zealous-leaf-58ett?file=/src/index.js

It is working when using openMenuOnFocus (and  autoFocus).

Tirzono:
@andreme You are using an old version in your Codesandbox that doesn't contain this fix.

andreme:
@Tirzono Sorry, I didn't realise I need to save manually. Updated and showing the issue: https://codesandbox.io/s/zealous-leaf-58ett?file=/src/index.js

Tirzono:
I see. Do you mind opening a new issue for this (if there is none already)? I'll have a look if I can find a fix.

asazernik:
Bookkeeping note: GH issue for original problem is #3656.

zzhcch:
@andreme i think the issue in your example is caused by your dataSource which is populating each time along with re-render.
try to move this out of the function or use `useMemo` would solve the problem.

andreme:
@zzhcch No, that's not it, I've updated the example
#4013

zzhcch:
@andreme Ok, I got your point. that seems like an underlying issue.
here is some workaround: https://github.com/JedWatson/react-select/issues/3375

PR #3875, opened 2019-12-04
Creatable: Add option meta field to onChange for 'create-option' action

andreme:
...


mitchellhamilton:
Thanks for the PR @andreme! Unfortunately we can't accept this because we already pass the option to the onChange handler and making sure there is only a single way to do a particular thing in react-select is very important because it's already so large and complex

andreme:
@mitchellhamilton That is exactly what happens for select-option (setValue): https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.js#L636
Why would you want create-option to behave differently?

mitchellhamilton:
Oh, sorry, I didn't realise that it's passed for select-option, I'm not 100% sure on this so I'll defer to @JedWatson

ebonow (approved):
Seems like a nice addition and better standardizes expectations of `onChange`

JedWatson (approved):

PR #3892, opened 2020-01-04
Fix react-select ignoring HTML5 "form" attribute

MrLeebo:
Issue: react-select ignores the HTML5 "form" attribute, which you can use to specify which form element to submit when the user presses "Enter" while the Select has focus.

Repro: https://codesandbox.io/s/jovial-montalcini-7dppv

Steps:
1. Using a modern browser, focus the "Name" field and press "Enter", a green toast should appear.
2. Focus the "Role" field and press "Enter" twice, once to close the menu and again to submit. Nothing happens.
3. Check the checkbox to "fix" the Input component.
4. Repeat step #2, the toast should appear.

Reference for the "form" attribute
https://devdocs.io/html/element/input#form

### DIFF
--- a/.changeset/sixty-forks-fry.md
+++ b/.changeset/sixty-forks-fry.md
@@ -1,0 +1,5 @@
+---
+"react-select": patch
+---
+
+Fix react-select ignoring HTML5 "form" attribute
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -243,6 +243,8 @@ export type Props = {
   tabSelectsValue: boolean,
   /* The value of the select; reflected by the selected option */
   value: ValueType,
+  /* Sets the form attribute on the input */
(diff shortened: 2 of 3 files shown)


flexdinesh:
Hey @MrLeebo! Thanks for the PR. We also need to add `form` as an optional param to this prop types here —

https://github.com/JedWatson/react-select/blob/122708c0619298d2b754127fa15be224fd85b49b/packages/react-select/src/components/Input.js#L9

Lemme know if you need more info here.

MrLeebo:
@flexdinesh Like so?

flexdinesh:
That's it. Thanks @MrLeebo!

Reflex-Gravity:
After we add `form` attribute to `Select`, and when we get all the field values in a form via `FormData`. It doesn't consider the `Select` field.

Issue #3620, opened 2019-06-13
Why trigger onMenuOpen when handle input change?


Cedcn:
https://github.com/JedWatson/react-select/blob/98f469eeb34c689f2697ffabd43c920077130bbc/packages/react-select/src/Select.js#L1086  

#3335

cutterbl:
Pretty standard, really. User experience expectation here would be to begin filtering the list of options, if you were to type directly in the Select.

ktarasov-clearscale:
I don't agree. 
Event should be triggered once the menu is opened. 
Why trigger the event on input?
Regarding the `onInputChange` handler, it is useless and confusing.

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our efforts towards the current major version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

However, if you feel this issue  is still relevant and you'd like us to review it - please leave a comment and we'll do our best to get back to you!

PR #3897, opened 2020-01-07
Fixes #3335 - removes the call to onMenuOpen within handleInputChange

### LINKED ISSUE #3335: onMenuOpen triggers on each input
When a select (or creatable) is open, writing triggers onMenuOpen callback on each keystroke. 

Demo: https://codesandbox.io/s/k0l8r2z6yr

I think that it should only trigger when opening the menu, right?

lorisdev:
As discussed in https://github.com/JedWatson/react-select/issues/3335 and https://github.com/JedWatson/react-select/issues/3620 , the expected behavior of `onMenuOpen` is to be triggered when the Menu opens.
Currently it is called on every input change as well which is quite confusing.

Removing the call of `onMenuOpen` within `handleInputChange` gives us the expected behavior.

### DIFF
--- a/.changeset/thick-eyes-walk.md
+++ b/.changeset/thick-eyes-walk.md
@@ -1,0 +1,8 @@
+---
+'react-select': patch
+---
+
+Removes the call to `onMenuOpen` on every input change
+
+If you were relying on this undesired behavior it may be a breaking change.
+Please upgrade accordingly.
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -1089,6 +1089,9 @@ export default class Select extends Component<Props, State> {
     const inputValue = event.currentTarget.value;
     this.inputIsHiddenAfterUpdate = false;
     this.onInputChange(inputValue, { action: 'input-change' });
+    if (!this.props.menuIsOpen) {
+      this.onMenuOpen();
+    }
   };
   onInputFocus = (event: SyntheticFocusEvent<HTMLInputElement>) => {
     const { isSearchable, isMulti } = this.props;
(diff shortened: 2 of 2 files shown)


pruik:
Although I would agree with you that the `onMenuOpen` should only be called once when the menu opens, your proposed change would break the following scenario:
- Open dialog
- Press escape. Dialog will close, but selector will still have focus
- Type anything.
- Expected behaviour would be to reopen the menu and perform a search based on the typed input
- However, with your proposed change the menu won't reopen.

I would suggest wrapping the `onMenuOpen` in a check:

```
if (!this.props.menuIsOpen) {
	this.onMenuOpen();
}
```

lorisdev:
Thanks for your input !
I've seen that some test were broken but I didn't have time to understand in which scenario.

Cmoen11:
@pruik suggestion should be implemented. We've faced this issue and needed to implement this solution in our code. I think this should be prioritized..

bladey:
Can verify this is an issue, PR looks good, @JedWatson to confirm the original change wasn't intentional.

vvitto:
Any updates regarding this bug?

JedWatson:
This looks good - thanks @lorisdev and @pruik! apologies for the delay on merge, we'll have it released shortly.

andrewmtam:
Hey all, just wanted to check in on this thread -- any idea when the release for this bug fix will happen? I saw the Version Packages PR was not yet merged:

https://github.com/JedWatson/react-select/pull/3980

Methuselah96:
I've started [a fork](https://github.com/react-select-oss/react-select) of react-select. This PR has been released in [`react-select-oss@3.1.1`](https://github.com/react-select-oss/react-select/releases/tag/react-select-oss%403.1.1).

PSoltes:
Hello, i am using 4.3.1 version of react-select and this is still happening... when input is blurred onMenuClose is called even though menu is closed

Methuselah96:
@PSoltes Can you a new issue with a CodeSandbox that clearly demonstrates the incorrect behavior? Thanks!

Issue #3354, opened 2019-01-15
Disabled options should be highlightable for accessibility


zkuzmic:
When navigating with a keyboard, disabled options should still be highlightable (but not selectable obviously) so that they can be read out by a screen reader. Without the ability to highlight the option, a screen reader user won't have any idea the disabled option exists.

An example of a Select with a disabled option:
https://codesandbox.io/s/232x2yzjkn

sarahbethfederman:
@JedWatson was discussing this issue with @gwyneplaine and we agree that disabled options should be focusable as per @zkuzmic's comment.  This reflects the built-in <select> behavior. Generally disabled options are focusable (and announced they are "dimmed") but when attempting to select them, the <select> closes w/ the previously selected value still selected. 

It sounds like we will essentially be reversing <https://github.com/JedWatson/react-select/pull/1008>(this) change + blocking the selection of a disabled option. I will get started on this issue if we're all on board with this.

PR #3913, opened 2020-01-22
Add skipDisabled prop

### LINKED ISSUE #3902: Feature Request: Skip keyboard navigation of disabled options
**Current results**:
- When `isOptionDisabled` prop is present, it disables all selecting option value events (click, keyboard), but does not skip it's focus via keyboard when clicking arrow keys up and down.

**Expected results**:
- If option is disabled, when navigating via keyboard it should skip it. The focus should go to another available option.  Example, default select behavior: https://jsfiddle.net/zm7uanxo/1/ 
- It worked like this at least in v1 of the plugin, when I tried it.

luke-webdev:
### skipDisabled prop

This allows the user to specify if a disabled option in the menu list should be skipped upon keyboard navigation. Skipped meaning that the focus highlight will be applied to the next non-disabled option.

This should prevent unnecessary keyboard strokes on options that cannot be selected.

Fixes #3902


packages/react-select/src/__tests__/Select.test.js
```diff
@@ -812,6 +812,26 @@ cases(
...
+      props: {
+        menuIsOpen: true,
+        options: OPTIONS_DISABLED,
+        skipDisabled: true
+      },
+      keyEvent: [{ keyCode: 40, key: 'ArrowDown' }],
+      selectedOption: OPTIONS_DISABLED[0],
+      nextFocusOption: OPTIONS_DISABLED[2],
+    },
+    'single select > disabled options should be skipped, ArrowUp': {
```
bladey, line 825: Hey @luke-webdev! Thanks for this PR, it looks good, however the second test is failing:

```
  ● focus on options > keyboard interaction with Menu › single select > disabled options should be skipped, ArrowUp

    expect(received).toEqual(expected) // deep equality

    Expected: "2"
    Received: "3"

      847 |     expect(
      848 |       container.querySelector('.react-select__option--is-focused').textContent
    > 849 |     ).toEqual(selectedOption.label);
          |       ^
      850 |
      851 |     for (let event of keyEvent) {
      852 |       fireEvent.keyDown(container.querySelector('.react-select__menu'), event);

      at toEqual (packages/react-select/src/__tests__/Select.test.js:849:7)
      at Object.cb (node_modules/jest-in-case/index.js:40:20)
```

For some reason it thinks that the focused item is the 3rd option, not the 2nd. I've attempted to debug this but I can't see anything obvious. Any ideas?

luke-webdev: Hi @bladey, thanks for taking the time to review the code. I have fixed the test that was failing in this branch.

Let me know if there are any other issues with the pull request. Thanks!

bladey:
Thanks @luke-webdev!

Taaqif:
Would be great to see this feature pushed to the main package

Travis-Kirton:
Are there any plans for this to be merged into main package? It's causing some issues for me & having to sacrifice UX for getting this working.

Methuselah96:

ebonow:
Some of my personal thoughts about the merit and approach of this PR.

The topic of disabled options been discussed before. To be clear, the functionality was originally changed to the existing behavior in order to properly support accessibility and not the other way around. https://github.com/JedWatson/react-select/issues/3354

Regarding native select behavior, here is the provided example of a native select when arrowing down to a disabled option with VoiceOver enabled:
Screen Shot 2021-01-14 at 8 02 54 AM

It's reasonable to say that `disabled` is meant to be a visual clue indicating that some criteria disqualifies it from being selectable and that this indication should also be supported by assistive technology. The team would like to encourage best practices when it comes to accessibility, so it would seem that this feature request is an _anti-pattern_ for accessibility.

This criteria likely disqualifies this as a candidate for inclusion into a future release. Please not that it is a difficult balance to add new features that people have put the time into developing and testing vs the added complexity and code weight of adding and supporting a new prop. I defer to the argument that this library is not meant to be everything to everyone but rather to provide the tools to be anything to anyone.

With that said, I would be open to discussing how a developer might be able to implement this functionality without changing the source code and using the existing apis. If there are gaps, we can explore adding some more universal api methods that would be beneficial to a larger group of developers.

One such feature might be an internal method such as `setFocusedOption` that could use some criteria to identify a visible option based using the data and index. This seems relevant to user requests who also do not want to select an option by default or may wish to include some other focus criteria so that focusing an option is not dependent on the sort order.

I'm open to more conversation or thoughts about this, but wanted to add my $.02. I'll try to work out an approach to this without modifying the source code in the original issue in #3902

PR #3930, opened 2020-02-11
fixing regressed change - tab should move to next dialog

srevenant:
https://github.com/JedWatson/react-select/pull/708

Conventional UX is that a tab switches to a new input.  I don't know if this was reverted intentionally, or by accident, but this fixes the problem.


srevenant:
What do we need to do to expedite something like this?  Is there an issue with this PR, or will it eventually go in?

nicholaai:
can we get this in @JedWatson ?

ebonow:
Greetings @srevenant, and my apologies that so much time has passed since you have submitted this PR. Our contributors deserve better than months or even years of silence in some cases. We all have a common interest in providing a 


This behavior has been identified as an issue in my own professional workplace, and given the relatively straightforward nature of this, I will raise this as a candidate for v3.3,

ebonow (approved):

ebonow:
> Looks like a minimal change, but I actually wonder why Tab selects the value.
> 
> I don't see that in the aria examples: https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.0pattern/combobox-autocomplete-both.html
> 
> And for a native select element, tab doesn't do anything.

/shrug 

I'm actually not entirely sure, though from professional experience, I had set `tabSelectsValue` to false for an in initial implementation at my company, only to have our business team send a request to select options with tab.

I'm curious about the backstory on this one, and am trying to think when we would want to preventDefault(). I'm not sure that we want to prevent the user from tabbing to the next field even if they are selecting an option unless `closeMenuOnSelect` is set to false.

bozdoz:
Perhaps an unintended event with this PR is with a multi-select component: 
1. user selects an option with ENTER or SPACE
2. user presses TAB
3. selected option is unselected and loses focus

caleb:
You could require two tabs on a multi select component. I tested a change which checks if `isMulti` and breaks or returns based on the value.

To me though, just having the tab key select the current item and move you forward even if it's a multi select seems like an acceptable behavior.

I would love this PR to be merged because my users have also specifically asked for this feature as well.

srevenant:
I'm curious if anybody knows if this is going to get resolved or has it been resolved some other way?  I'm still running on a 2-year old branch because this feature is critical to my behavior.  It'd be nice to get current again.  I'm rather unsure why such a simple fix is taking YEARS.

Rall3n:
@srevenant I apologize if this is taking longer than expected.

IMO this change is not needed. The current behavior matches how a default `<select>` element behaves.
The correctness of this behavior is also mentioned in a [combox example](https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html#kbd_label_listbox) of the "WAI-ARIA Authoring Practices" documentation:
> Note: the native `<select>` element closes the listbox but does not move focus on tab.

bozdoz:
I agree with @Rall3n. As I mentioned above, this is not native behaviour. It also isn't referenced in the w3 accessibility best practices.

caleb:
@srevenant yeah, if you want usability for your users I would go with something like Downshift.js where you can customize the behavior with a bit of work. It’s a react hook/HOC where you can heavily control how your combo boxes work.

I’m in the same boat. My users need to use a combo box to code hundreds of transactions, so being fast is their priority.

Good luck!

nightpool:
@Rall3n I believe you're misinterpreting that document. You're quoting from the document on correct _combobox_ (autocomplete input) behavior, which says (with more context): 

> **Tab:**
>   - Sets the value to the content of the focused option in the listbox.
>   - Closes the listbox.
>   - Performs the default action, **moving focus to the next focusable element.**
>    Note: the native <select> element closes the listbox but does not move focus on tab. This pattern matches the behavior of the other comboboxes, **rather than the native element in this case.**

In other words, it's explicitly saying that for *comboboxes*—that is, autocomplete dropdowns where a text input is used to control a list of elements—the correct behavior is to move to the next field, **unlike** a `<select>` element.

React-select is an _autocompelete_ input, where a visible text field is used to control an associated list of elements. So it's inappropriate to base its UX on the `<select>` element as this case, as the ARIA documented you quoted explicitly outlines. Instead, the Tab behavior given here is much more obvious and understandable to users. 

You can see a DOM-native example of an autocomplete input with the `<datalist>` element. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist for an example. When using a `<datalist>` element, it has the exact same behavior outlined here—`Tab` moves to the next form input. In contrast, at least in my browser, the DOM native `<select>` input doesn't do *anything* when users press `Tab` in the middle of arrowing through a select list.  This makes sense, because the pop-over is a modal that captures focus until it's dismissed, rather then being controlled by a separate text input. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select for an example. Because `react-select` serves as a *combination* of a text input form with a select form, the `combobox` behavior quoted in the document is the appropriate model to base the UX on

nightpool:
@Rall3n I believe you're misinterpreting that document. You're quoting from the document on correct _combobox_ (autocomplete input) behavior, which says (with more context): 

> **Tab:**
>   - Sets the value to the content of the focused option in the listbox.
>   - Closes the listbox.
>   - Performs the default action, **moving focus to the next focusable element.**
>
> Note: the native `<select>` element closes the listbox but does not move focus on tab. This pattern matches the behavior of the other comboboxes, **rather than the native element in this case.**

(emphasis added)

In other words, it's explicitly saying that for *comboboxes*—that is, autocomplete dropdowns where a text input is used to control a list of elements—the correct behavior is to move to the next field, **unlike** a `<select>` element.

React-select is an _autocompelete_ input, where a visible text field is used to control an associated list of elements. So it's inappropriate to base its UX on the `<select>` element as this case, as the ARIA documented you quoted explicitly outlines. Instead, the Tab behavior given here is much more obvious and understandable to users. 

You can see a DOM-native example of an autocomplete input with the `<datalist>` element. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist for an example. When using a `<datalist>` element, it has the exact same behavior outlined here—`Tab` moves to the next form input. In contrast, at least in my browser, the DOM native `<select>` input doesn't do *anything* when users press `Tab` in the middle of arrowing through a select list.  This makes sense, because the pop-over is a modal that captures focus until it's dismissed, rather then being controlled by a separate text input. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select for an example. Because `react-select` serves as a *combination* of a text input form with a select form, the `combobox` behavior quoted in the document is the appropriate model to base the UX on

nightpool:
> I don't see that in the aria examples: https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.0pattern/combobox-autocomplete-both.html

I'm not sure if this has changed since the time this comment was posted, but I just looked at the combobox ARIA example now, and I can confirm that as of 2022, pressing "tab" while in the combobox text input field will move the browser focus to the next focusable element (in this case, a link somewhere else on the page but in most forms probably a "submit" button or another form input)

Rall3n:
@nightpool I apologize, I might have misunderstood the documentation at the time of writing my previous comment.

Now the only thing I think needs some more consideration is selection of multiple options, as this is not specified in the pattern guidelines.
There are 3 behaviors that could possibly be adopted:
1. Double-Tab. <kbd>Tab</kbd> once to select, <kbd>tab</kbd> a second time to progress in the tab order.
2. Ignore selection by pressing <kbd>Tab</kbd>. Will not select anything and just progresses in the tab order.
3. Same behavior as singular selection. Could result in confusion by the user if the user wanted to select another option.

Also this PR was created pre TS-conversion and needs to be updated to target our current `master` branch. I will see if conflict resolution is possible, else the PR would have to be re-created against our current `master`.

lukebennett88:
Hi @srevenant.
As far as I can tell, what exists now is the expected behaviour. Pressing the tab key whilst the menu is open closes the menu, and pressing it a second time shifts your focus to the next selectable item.

alexw23:
I'm quite confused why I have to double tab as an end user.

When tabbing between traditional `<select>` fields, initial focus allows to type what you need then once you see it selected tab will move to the next field.

Try it for yourself: https://codesandbox.io/s/cranky-fermat-hxfdke?file=/index.html

Focus name, input name, tab, type 'marc', tab. Input is now on day entry.

Now yes if you click into the select, then type the option tabbing doesn't actually close the menu, you need to hit enter. At which point yes focus is still on the field.

The simple solution it seems, if focus source was via tab, then upon tab select the selected option and move on to next field.

leikoilja:
@alexw23, did you find any workaround to make it work, so when the user hits 'tab' it selects the option and focuses the next element

alexw23:
Hi @leikoilja, I just checked our app and it appears we did, as the functionality is different the comment from above.

The basic premise is to store refs `useRef()` for the next fields, then when the current field changes, focus to the next field.

Here's a basic example:
```jsx
import Select from 'react-select'

const dayRef = useRef()
const monthRef = useRef()
const yearRef = useRef()

<Select
  ...
  ref={ref}
  onChange={(option) => {
    setDay(option ? option.value : null);
    monthRef.current.focus();
  }}
/>

<Select 
  ...
  ref={monthRef}
  onChange={(option) => {
    setMonth(option ? option.value : null);
    yearRef.current.focus();
  }}
/>;


<Select
  ...
  ref={yearRef}
  onChange={(option) => {
    setYear(option ? option.value : null);
  }}
/>;
```

PR #3943, opened 2020-02-24
React Testing Library

tarang9211:


JedWatson:
Context for anyone watching this, it's been a huge refactoring effort by @tarang9211 and @mitchellhamilton to switch from enzyme to [react-testing-library](https://github.com/testing-library/react-testing-library) so that we can start using hooks in react-select.

Long story short I tried to make some updates late last year and clean up the codebase, and realised that (not for the first time) our tests were a major blocker to progress, so this is a really exciting thing to have out of the way 🎉

PR #3945, opened 2020-02-24
Fix the problem when you hit Enter, it submits the forms

diogobas:


goremukin:
Faced the same issue. Any updates?

packages/react-select/src/Select.js
```diff
@@ -1206,18 +1206,25 @@ export default class Select extends Component<Props, State> {
         this.selectOption(focusedOption);
         break;
       case 'Enter':
+        const targetIsButton = event.target && event.target.type === 'button';
+
+        if (targetIsButton) {
+          return;
```
ebonow, line 1212: What is the purpose of checking for `event.target.type === 'button'`

packages/react-select/src/Select.js
```diff
@@ -1206,18 +1206,25 @@ export default class Select extends Component<Props, State> {
...
+
         if (menuIsOpen) {
-          if (!focusedOption) return;
+          if (!focusedOption) break;
           if (this.isComposing) return;
           this.selectOption(focusedOption);
           break;
         }
-        return;
+        break;
```
ebonow, line 1227: Why change `return` to `break` here? This seems like a breaking change should a user decide to use react-select with a `textarea`.

packages/react-select/src/Select.js
```diff
@@ -1206,18 +1206,25 @@ export default class Select extends Component<Props, State> {
...
+
         if (event.keyCode === 229) {
           // ignore the keydown event from an Input Method Editor(IME)
           // ref. https://www.w3.org/TR/uievents/#determine-keydown-keyup-keyCode
           break;
         }
+
         if (menuIsOpen) {
-          if (!focusedOption) return;
+          if (!focusedOption) break;
```
ebonow, line 1222: Since the purpose of break is simply to fall through, why not instead simply call `return event.preventDefault()`

ebonow (commented):
I could be overlooking some things, but I have questions regarding some of the implementation here. I could be over simplifying this, but it seems like replacing 
```jsx
if (!focusedOption) return;
```
with
```jsx
if (!focusedOption) return event.preventDefault(); 
```

or simply break as you have it would fix this, so unsure why there is extra logic to identify if the target is a button or to preventDefault on all Enter clicks which are not selecting an option. This is purely assumptive on my part so looking to better understand the use-case here.

usamamashkoor:
Did you find any solution for this?

PR #3968, opened 2020-03-12
Fix focus disable option when menu open

SakamotoRyoutaMF:
Hello!

I fix a bug(may be).
when I open the Select, Select focus a first option.
But I don't this work, if a first option is `isDisabled`.
Because, `isDisabled option` can't be hovered and selected on default setting.
`isDisabled option` is disable but, we see that enabled option.(Please see gif ↓↓↓↓)
So, I omit disabled option from `options.focusable`.

Let me apologize for my poor English. 

---

|before|after|
|-----|-----|
|before|after|

※ I custom style


SakamotoRyoutaMF:
oops!
I overlooked this issue https://github.com/JedWatson/react-select/issues/3354 .

Issue #3988, opened 2020-03-27
onCreateOption is not called, fails at `valueArray[valueArray.length - 1] === newOption`


euvs:
Getting a weird bug at this line.

https://github.com/JedWatson/react-select/blob/c8d74bd5710b1db6736837fb4334a59e46614a27/packages/react-select/src/Creatable.js#L148

It seems that the `===` comparison fails on this line. 

```
valueArray[valueArray.length - 1] === newOption
```

I put debugger breakpoint and inspected the values. Please see the console output. 

```bash
> valueArray[valueArray.length - 1] === newOption
false
> newOption === newValue 
false
> newOption
{label: "Create "sd"", value: "sd", __isNew__: true}
> newValue
{label: "Create "sd"", value: "sd", __isNew__: true}
> valueArray
[{…}]0: {label: "Create "sd"", value: "sd", __isNew__: true}length: 1__proto__: Array(0)
> valueArray.length
1
```

It looks like the `newOption` and `newValue` have the same content by they fail when compared using `===` operator. As the result the `onCreateOption` is not called.

nikitaindik:
I can confirm. Works in 3.0.8, stopped working in 3.1.0.

nikitaindik:
Created a reproducible: https://codesandbox.io/s/magical-margulis-xm9x3

nikitaindik:
@gwyneplaine Please check out the fix and tell me what you think. I've ran the tests and did some manual testing on the docs page - everything seems to be fine. I'd be happy to add a test, but I'll need a little guidance.

nikitaindik:
@JedWatson @gwyneplaine Friendly ping :)

brandonweiss:
I also just hit this. `onCreateOption` doesn’t seem to be being called for Creatable on 3.1.0. Although it does seem to be called for AsynCreatable. Rolling back to 3.0.8 fixes it.

nikitaindik:
@JedWatson @gwyneplaine Hi there! How should we proceed on this PR? :)

sirjuan:
We also encountered this bug, and can confirm @nikitaindik PR will fix the problem.

bladey:
Thanks all, I'll try and get the PR reviewed as soon as I can.

hamfz:
For anyone struggling with this in the interim, you can use the `getNewOptionData` prop to bust the equality check and properly pass through to ` onCreateOption`.


Ex:
```
[...]

PR #3990, opened 2020-03-28
Fix onCreateOption is not always called for Creatable

nikitaindik:
**Issue**: https://github.com/JedWatson/react-select/issues/3988

The issue is caused by "newOption" being recreated as a new object on every render.


johannesleander:
Any update on plans to release this fix?

bcalvignac:
Yes, this was causing me problems too

sirjuan:
This issue (comparing with reference) sometimes causes `onCreateOption` to not be called. Instead `onChange` with `select-option` action is called with new option object.

Merging and releasing this PR would fix it.

laznic:
Could this fix get merged please? Thanks in advance!

makker:
Would be greatly appreciated if this fix could be released! 🙏

mika-kaki:
Looking forward for this fix. Thanks for your great work!

bladey:
@JedWatson this looks good to me - pending your review.

nikitaindik:
@bladey @JedWatson Hi guys! Do you think we can merge this? :)

Methuselah96:

Methuselah96 (dismissed):
@nikitaindik Thanks for the PR!

1. Can you add a test?
2. My preference would actually be to change the behavior in `Select.js` to only memoize the options when they are strictly equal by changing `isEqual(newProps.options, lastProps.options)` to `newProps.options === lastProps.options` [here](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.js#L381) (might as well change the other ones from `isEqual` to `===` as well). This allows people to use strict equality in their own `onChange` methods and not have to worry about the implementation detail that the memoization depends on an internal `isEqual` method. This memoization was introduced in https://github.com/JedWatson/react-select/pull/3868 and I don't believe the memoization was essential to the fix in that PR.

nikitaindik:
Hi @Methuselah96! I'll have some spare time during the weekend. I'll try to add a test see if I can resolve this isEqual thing.

Methuselah96:
@nikitaindik Let me know if you don't have time to make the changes. I can do it myself if you don't have the time.

nikitaindik:
Yep, sure. I'm still a little busy. Sorry. Please feel free to improve this
PR.

JedWatson (approved):

PR #4034, opened 2020-04-30
Rewrite stripDiacritics to be faster

sophiebits:
We now loop over the string once instead of... 84 times. :)

In my tests, this makes rerendering a Select with ~8,000 options about 5x as fast.

We could probably make filtering even faster by transforming the search string into a regex that matches the diacritics so we don't have to transform each option, but this alone is a big improvement.


JedWatson:
Thanks @sophiebits! much appreciated.

PR #4058, opened 2020-05-24
Update cleanValue() to allow non-object values(string, number, etc.)

jimsleon:
React-select is great and easy to use. I had a simple wrapper around react-select to allow non-object options like string, number, etc. for simple use case so that I don't need to construct object with 'label' & 'value' from simple type by overriding getOptionValue() and getOptionLabel(), but found it doesn't work.(display works fine but can't select a value), **it does work for multiple selection**. After looked into code I found it is caused by clearValue() which assuming 'option' is object. cleanValue() is only used for processing 'value' property, so please review if this can be merged.

### DIFF
--- a/packages/react-select/src/__tests__/Select.test.js
+++ b/packages/react-select/src/__tests__/Select.test.js
@@ -2065,7 +2065,7 @@ cases(
     expect(
       container.querySelector('.react-select__placeholder')
     ).not.toBeInTheDocument();
-    rerender(<Select {...props} value="" />);
+    rerender(<Select {...props} value={null} />);
     expect(
       container.querySelector('.react-select__placeholder')
     ).toBeInTheDocument();
--- a/packages/react-select/src/utils.js
+++ b/packages/react-select/src/utils.js
@@ -62,8 +62,8 @@ export function classNames(
 // ==============================
 
 export const cleanValue = (value: ValueType): OptionsType => {
-  if (Array.isArray(value)) return value.filter(Boolean);
-  if (typeof value === 'object' && value !== null) return [value];
+  if (Array.isArray(value)) return value.filter(v => v !== undefined && v !== null);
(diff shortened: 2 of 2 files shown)


L3au:
@jimsleon Nice work. The bad news is I don't think the collaborator would accept this change after I checked the early related issues and PR.

- Find correct option when value is passed as a string https://github.com/JedWatson/react-select/pull/3050
- Allow passing of primitive values as value prop https://github.com/JedWatson/react-select/pull/2908
- `simple value` in upgrade guide https://react-select.netlify.app/upgrade-guide#simple-value
- some explanations https://github.com/JedWatson/react-select/issues/2669#issuecomment-423017909 https://github.com/JedWatson/react-select/issues/2669#issuecomment-427700804 https://github.com/JedWatson/react-select/pull/2908#issuecomment-413102198
- a wrapper for this case was created. https://github.com/jossmac/react-select-simple-value

jimsleon:
> @jimsleon Nice work. The bad news is I don't think the collaborator would accept this change after I checked the early related issues and PR.
> 
> * Find correct option when value is passed as a string #3050
> * Allow passing of primitive values as value prop #2908
> * `simple value` in upgrade guide https://react-select.netlify.app/upgrade-guide#simple-value
> * some explanations [#2669 (comment)](https://github.com/JedWatson/react-select/issues/2669#issuecomment-423017909) [#2669 (comment)](https://github.com/JedWatson/react-select/issues/2669#issuecomment-427700804) [#2908 (comment)](https://github.com/JedWatson/react-select/pull/2908#issuecomment-413102198)
> * a wrapper for this case was created. https://github.com/jossmac/react-select-simple-value

Probably, but my change is a little bit different, I didn't change any contract of react-select. 🤞

bladey:
@JedWatson to review and comment.

Issue #3910, opened 2020-01-21
Current status of`react-select`, how you can contribute and what the future looks like.


marcelltoth:
Right now we have _97_ open pull requests, some important bugfixes, some interesting new features.

Among these pull requests there are numerous that **have been open for more than about half a year, without any sort of feedback**. See #3666 (feature), #3633 (bugfix), #3669 (bugfix) just to name a few. In the meanwhile, there was a single (refactoring) PR accepted in December, and we **didn't get even a patch release out for 4 months** (see [npm](https://www.npmjs.com/package/react-select)).

These pull requests would also close down many of the 1200+ open issues, some of which are repeated bug reports. (About issues that we already have fixes for, standing in the PR queue for months.)

**I understand that maintaining high-usage open source projects is hard, and it takes a lot of time.**
Therefore I would like ask the question: ***How can we help?***

It looks like PRs are not helping much, comments and reactions on them neither. I really would like to get this project going as it is essential for myself, my company, and in my opinion the React community as a whole.

I am trying to avoid having to fork the project, because that creates a hard division line, but right now it seems like this is our only option to move forward, hence my question here. If you are in the need of a maintainer, I can happily help, and I'm sure other experienced engineers can, too. Or if we can provide support in a different way, let us know, I believe many of us are open to contribute, just not finding the means right now.

BrandonKowalski:
I know its less than ideal, but given the radio silence on this and other issues do we as the community make a temporary fork?

I absolutely love this project as it has been a life saver at work and I'd hate to see it rot.

Tirzono:
There are already a lot of people that [fork the library](https://www.npmjs.com/search?q=react-select%20A%20Select%20control%20built%20with%20and%20for%20ReactJS) to get their features in. For example [one of the forks](https://www.npmjs.com/package/@torkelo/react-select) is actually used by Grafana and therefore has a lot of weekly downloads.

I think making a temporary fork is the good way to go for now. It would stop people having to fork the library themselves and it would continue the development of the project. When the current maintainers start being active again we could contribute back to this original project and stop the temporary fork. It should be made clear from the start that the fork is temporary.

How can we make a temporary fork work for everyone? Are there examples of libraries that have been at this stage before so we can learn from their experiences? Who wants to contribute to this?

marcelltoth:
Sadly I have to agree with you, there does not seem to be another way.

[...]

PR #4064, opened 2020-05-29
Enable capibility to write docs in TypeScript

Methuselah96:
See https://github.com/JedWatson/react-select/issues/3910#issuecomment-636045525 for context. It wasn't too hard to allow both Flow and TypeScript types at the same time, so I went ahead and did that.


Methuselah96:
Nevermind, the Flow types aren't building. I'll look into it.

Methuselah96:
I'll just do all the docs at once.

PR #4069, opened 2020-06-03
Use pointer cursor for dropdown and options

brandonparis:
Issue: #3831 

This PR adds `cursor: pointer` CSS to the indicator icons, the option items (that are not disabled), and the container wrapping the select to indicate to the user that all of these items are clickable.

### DIFF
--- a/packages/react-select/src/__tests__/__snapshots__/Async.test.js.snap
+++ b/packages/react-select/src/__tests__/__snapshots__/Async.test.js.snap
@@ -38,6 +38,7 @@ exports[`defaults - snapshot 1`] = `
 
 .emotion-7:hover {
   border-color: hsl(0,0%,70%);
+  cursor: pointer;
 }
 
 .emotion-2 {
@@ -126,6 +127,7 @@ exports[`defaults - snapshot 1`] = `
 
 .emotion-5:hover {
   color: hsl(0,0%,60%);
+  cursor: pointer;
 }
 
 .emotion-4 {
--- a/packages/react-select/src/__tests__/__snapshots__/AsyncCreatable.test.js.snap
+++ b/packages/react-select/src/__tests__/__snapshots__/AsyncCreatable.test.js.snap
@@ -38,6 +38,7 @@ exports[`defaults - snapshot 1`] = `
 
 .emotion-7:hover {
   border-color: hsl(0,0%,70%);
+  cursor: pointer;
 }
 
 .emotion-2 {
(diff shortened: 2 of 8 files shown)


bladey:
@JedWatson to review and comment.

ebonow:
This PR does not address the use-case when an option is disabled, as I have a similar styling in my own personal implementations, and I recently ran into this.

ebonow:
I edited the title of this PR to reflect that this is not a "fix" but rather a request to change the default styling of the cursor when hovering over an option. 

It is worth stating that the default styling for a pointer hovering over an <option> element within a <select> is `default` and not `pointer`. 

Screen Shot 2021-03-05 at 11 10 42 PM
[Reference MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select)

The styles api also makes this very easy to implement without the need to change the source code and impact the default styling for all users.

ebonow:
Greetings @brandonparis ,

Sorry for the extremely late reply to your PR submission. There is quite a backlog of user submitted PR's and the community deserves better than waiting months or years for a collaborator to reply to the work they've done.

I wanted to let you know this PR was brought forward in a recent meeting and it was decided that the proposed changes here would be considered a breaking UX change for many. It would also introduce a styling decision different than the default styling of a native select element.

Fortunately, styling the cursor remains straight forward enough through any of the existing styling methods so closing this PR will not hinder any developer from applying this desired behavior in their own implementations. I expanded a bit more in depth on the original issue if you would like to read more about the decision and why this will be closed so feel free to add anything to discussion there and if there is a compelling enough reason, this could be reconsidered.

brandonparis:
> Greetings @brandonparis ,
> 
> Sorry for the extremely late reply to your PR submission. There is quite a backlog of user submitted PR's and the community deserves better than waiting months or years for a collaborator to reply to the work they've done.
> 
> I wanted to let you know this PR was brought forward in a recent meeting and it was decided that the proposed changes here would be considered a breaking UX change for many. It would also introduce a styling decision different than the default styling of a native select element.
> 
> Fortunately, styling the cursor remains straight forward enough through any of the existing styling methods so closing this PR will not hinder any developer from applying this desired behavior in their own implementations. I expanded a bit more in depth on the original issue if you would like to read more about the decision and why this will be closed so feel free to add anything to discussion there and if there is a compelling enough reason, this could be reconsidered.

All good! Thanks for the deep consideration.

PR #4075, opened 2020-06-05
fixed issue there menu scrolls into view even though menuShouldScroll…

bladey:
…IntoView is false, happens when parent container is a scrollable div further down the page. The bug casues the menu bottom to be scrolled into view so that the select input/search is not visible


bladey:
Testing only, closing.

PR #3705, opened 2019-08-05
focusDefaultOption prop

Francismb:
### focusDefaultOption prop

This prop enables or disables the automatically applied focus to the first or last option in the drop down menu of options.

The use case I found is that if you want a free text search select input but to also have auto completing options, much like a google search box.

https://github.com/JedWatson/react-select/issues/2848
https://github.com/JedWatson/react-select/pull/3324
https://github.com/JedWatson/react-select/pull/3484

Issue #2653, opened 2018-05-28
Do not auto select first option and populate field


martincarstens:
Current behavior: 

-- start typing charachers
-- dropdown shows partial matches and first result is automatically highlighted
-- hit enter, and first result populates search box

Desired behavior (via attribute):
-- start typing charachers
-- dropdown shows partial matches and first result is not automatically highlighted
-- press down arrow
-- first result is highlighted
-- hit enter, and first result populates search box

Basically the behavior prior to the bellow issue
https://github.com/JedWatson/react-select/issues/2

Is this possible currently?

trkaplan:
"Auto selecting first option" must be optional.
Because I use Select component mainly as a Search Box. 
I want to get the typed input, after hitting enter key.  It'll be still possible to select one an option from the list.

noahzweben:
@martincarstens did you figure this out?

martincarstens:
I did, but had to edit source.

It looks like you won't have to wait too long for this option however:
https://github.com/JedWatson/react-select/pull/3484

The source to be merged is here:
https://github.com/JedWatson/react-select/pull/3484/files/3cdaf4669d649714852a7dca560dd782cc19076f

m

PR #4080, opened 2020-06-10
focusDefaultOption prop

bladey:


bladey:
@JedWatson looks good to me, for your review.

bladey:
Based off PR #3705

rockymountainhigh1943:
Looking forward to this fix getting through! Any chance this is getting close?

bladey:
@rockymountainhigh1943 as soon as possible, I'm doing my best to get a release together.

rockymountainhigh1943:
> @rockymountainhigh1943 as soon as possible, I'm doing my best to get a release together.

👏 👏 👏  Thank you!

danielstecki:
Are you still planning on releasing this?

zbuttram-versium:
If anyone else is looking for a temporary fix until this is landed in a release, you _can_ monkeypatch `<Select>`'s internals via its ref prop. Something like the hook below works if you pass the returned ref to your `<Select>` component's `ref` prop:

```javascript
function useReactSelectFocusFix() {
  const selectRef = useRef()
  useEffect(() => {
    if (selectRef.current && selectRef.current.select) {
      selectRef.current.select.getNextFocusedOption = () => null
    }
  }, [selectRef.current])
  return selectRef
}
```

zbuttram:
If anyone else is looking for a temporary fix until this is landed in a release, you can monkeypatch `<Select>`'s internals via its ref prop. Something like the hook below works if you pass the returned ref to your `<Select>` component's ref prop:

```javascript
function useReactSelectFocusFix() {
  const selectRef = useRef()
  useEffect(() => {
    if (selectRef.current && selectRef.current.select) {
      selectRef.current.select.getNextFocusedOption = () => null
    }
  }, [selectRef.current])
  return selectRef
}
```

b3rg3t:
@zbuttram can you add a codesandbox on how to solve it? doesn't get it to work.

luknl:
This will be a great addition to our project, can't wait for this to be merged!

ayrtongomes:
@JedWatson approve this pls 🙏🙏

arunpresdo:
Any update on when this will be merged?

luknl:
@JedWatson, do you approximately know when this will be merged? We can't wait for this 🙏

Methuselah96:

675r:
Can this be merged please

kylemh (approved):

packages/react-select/src/Select.js
```diff
@@ -415,9 +418,21 @@ export default class Select extends Component<Props, State> {
...
-      const focusedOption = this.getNextFocusedOption(menuOptions.focusable);
-      this.setState({ menuOptions, selectValue, focusedOption, focusedValue });
+
+      let focusedValue = null;
+      let focusedOption = null;
+
+      if (nextProps.focusDefaultOption) {
+        focusedValue = this.getNextFocusedValue(selectValue);
+        focusedOption = this.getNextFocusedOption(menuOptions.focusable);
+      }
```
Methuselah96, line 428: I'm not sure that this prop should affect `focusedValue`, I think it should only affect `focusedOption`. `focusedValue` refers to the focused value in a multi-select (if you hover over a selected value or expand the menu of a multi-select and press the left arrow) and his little relation to the `focusedOption`. By default there is no `focusedValue` (see [implementation](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.js#L769-L788)) unless you perform one of the above actions.

Methuselah96 (changes requested):

packages/react-select/src/Select.js
Methuselah96, line 428: A cleaner way to do this would be to check `focusDefaultOption` within `getNextFocusedOption` instead of in this method.

Methuselah96 (changes requested):

packages/react-select/src/Select.js
Methuselah96: And actually, now that I think it about, this logic is incorrect if we don't move it into `this.getNextFocusedOption`.

The situation where this would make a difference is if the menu is open and the `options` prop changes. What `this.getNextFocusedOption` does is it checks to see if the currently selected option is still a focusable option, and if it is, then it keeps it as the focused option (which is still what we want even if `focusDefaultOption` is `false` because there is already a selected option).

So instead we should change `this.getNextFocusedOption` from:
```
getNextFocusedOption(options: OptionsType) {
  const { focusedOption: lastFocusedOption } = this.state;
  return lastFocusedOption && options.indexOf(lastFocusedOption) > -1
    ? lastFocusedOption
    : options[0];
}
```
to:
```
getNextFocusedOption(options: OptionsType, focusDefaultOption: boolean) {
  const { focusedOption: lastFocusedOption } = this.state;
  if (lastFocusedOption && options.indexOf(lastFocusedOption) > -1) {
    return lastFocusedOption;
  }
  return focusDefaultOption ? options[0] : null;
}
```

packages/react-select/src/Select.js
```diff
@@ -516,7 +531,7 @@ export default class Select extends Component<Props, State> {
     this.setState({
       menuOptions,
       focusedValue: null,
-      focusedOption: menuOptions.focusable[openAtIndex],
+      focusedOption: focusDefaultOption ? menuOptions.focusable[openAtIndex] : null,
```
Methuselah96, line 534: This might have the interesting effect of calling `this.announceAriaLiveContext({ event: 'menu' });` twice if the user uses the keyboard to navigate the Select component. The first time will be in the callback of this call to `setState` (see a few lines below this one). The second time will be in `focusOption` where it calls `this.announceAriaLiveContext({ event: 'menu' });` again [if there is no selected option](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.js#L585-L589).

I believe we can remove the second case because there is not a time where we call `focusOption` when the menu is closed (https://github.com/JedWatson/react-select/pull/3427 removed the last case of this situation) and should do that as part of this PR in order to avoid the above scenario.

Methuselah96 (changes requested):

packages/react-select/src/Select.js
```diff
@@ -138,6 +138,8 @@ export type Props = {
   formatGroupLabel: typeof formatGroupLabel,
   /* Formats option labels in the menu and control as React components */
   formatOptionLabel?: (OptionType, FormatOptionLabelMeta) => Node,
+  /* Applies focus to the first option in the menu when opened */
+  focusDefaultOption: boolean,
```
Methuselah96, line 142: nit and bike shedding: I prefer `autoFocusOption` or `autoFocusFirstOption` as the prop name, but that's just a personal preference.

1. `autoFocus` vs `focus`: in my mind reinforces that this is something that happens once when the menu opens, but that the focus can change with user input.
2. `option` or `firstOption` vs `defaultOption`: `defaultOption` makes it sound like to me that the the default option is something that is customizable when it's not. Using `option` still leaves room for future props that could change which `option` is focused by default, but that's not the case at the moment.

But I'm fine with `focusDefaultOption` if that's easier.

Methuselah96 (changes requested):

packages/react-select/src/Select.js
kylemh: Hmmm I thought `autoFocus` would be like [input#autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefautofocus) where the field snatches a user's focus.

Methuselah96: Yeah, there's a little bit of overloading with the word "focus" which makes it hard to differentiate, since we have to talk about both focus of the Select component and the focus of individual options when the menu is open.

I think maybe the most descriptive (but probably too long) prop name would be `autoFocusFirstOptionOnMenuOpen`  where the `onMenuOpen` disambiguates when the auto-focusing would happen.

kylemh: Knowing the full context, I prefer your `autoFocusOption`; however, could we not change the type signature to allow people to choose a different item to receive first focus? Who's to say maybe somebody would not want to focus the last option, or some middle one 🤷 . Regardless, I like that naming best. If it will be constrained to just boolean, I like your longer variable name: `autoFocusFirstOptionOnMenuOpen` - very crystal clear.

Methuselah96: Yeah, it's hard to know where to stop as far as customization. The options ordered from most customizable to least customizable are:

1. Provide a `getDefaultFocusedOption` prop which is a function where the user is passed information and they can return the option they want to be focused by default.
2. Make `autoFocusOption` a prop that can be `first`, `last`, or false (can be expanded later, but I'm not sure what else you would need.
3. Make `autoFocusOption` a prop that can be `true` or `false`.

Since `react-select` is already very complex and hard to keep straight with all its edge cases, I think we want to be conservative and avoid complexities.

That's why I suggested `autoFocusOption` instead of `autoFocusFirstOption` as my top choice (if we go with `true | false` because it leaves open the possibility of adding `'first' | 'last'` options later on.

But you make a good point. Maybe `autoFocusOptionOnMenuOpen` is the best of both worlds.

herbertpimentel:
Does anybody has a workaround to that ? I also could not make @zbuttram suggestion get to work.

gunn:
Hi @herbertpimentel, I'm currently using this as a workaround:

```tsx
React.memo(props=> {
  const selectRef = React.useRef(null)

  React.useEffect(() => {
    let select = selectRef.current?.select?.select
    select ??= selectRef.current?.select // <-- for multi input

    if (select) {
      const original = select.getNextFocusedOption.bind(select)

      select.getNextFocusedOption = options => {
        const inputValue = selectRef.current?.state?.inputValue
        if (inputValue) return original(options)
      }
    }
  }, [selectRef.current])

  return (
    <ReactSelect
      ref={selectRef}
      { ...props }
    />
  )
})
```

uragecz:
Can be this MR finally merged ? (can someone resolve conflicts?) i would really appreaciate this feature..

JavierPDev:
Any ETA on when this will be merged? [This has been around for almost two years now](https://github.com/JedWatson/react-select/pull/3484) and [almost a year ago it was said that this would be fixed soon.](https://github.com/JedWatson/react-select/issues/2653#issuecomment-492825054)

I see that most of the workarounds involve hooks and functional components. If anyone else is using class-based components and wants to hack this issue away while waiting for this merge here's what worked for me:

```javascript
  componentDidMount() {
    if (this.selectRef) {
      this.selectRef.select.select.select.getNextFocusedOption = () => null;
    }
  }

  render() {
    return (
      <AsyncCreatableSelect
        ref={ref => this.selectRef = ref}
        // Rest of your props here...
      />
    );
  }
```

All those `.select`s must be from me using the async creatable component rather than the base select component. You'll have to console.log the ref and check for the existence of `getNextFocusedOption` in the prototype of each `.select` to see how far down you need to nest depending on which select component you use.

Bosch-Eli-Black:
Just FYI:

As far I can tell, the newest version of react-select (4.3.0, or maybe earlier) no longer exposes `getNextFocusedOption()`, so the posted workarounds no longer work.

Bosch-Eli-Black:
@bladey, I noticed that you have permissions in this repository to close issues. Do you also have permissions to merge pull requests?

grzennio:
Will this be ever merged? it's like one and half year now.

Bosch-Eli-Black:
I think they're not merging any PR's until v5 is out: https://github.com/JedWatson/react-select/issues/4559#issuecomment-860808329

Annie-Huang:
Not sure why this is not merged yet. But does anyone got a walk around solution for version @4.3.1, which is the latest version now (Sep 9th, 2021). The `getNextFocusedOption` doesn't seem to be exposed for this version.

Or please let me know which version I can use to use for the `getNextFocusedOption` work around solution?

Many thanks...

grzennio:
> Not sure why this is not merged yet. But does anyone got a walk around solution for version @4.3.1, which is the latest version now (Sep 9th, 2021). The `getNextFocusedOption` doesn't seem to be exposed for this version.
> 
> Or please let me know which version I can use to use for the `getNextFocusedOption` work around solution?
> 
> Many thanks...

After many tries, we just decided that this is not an issue. 
Our selectors started just using search functionality.
And the only issue we found, is that if:
* 2 values available in selector, it will hide current selected ( so the list will contain only one - the current selected is not shown ) which is not that bad
* 1 values available in selector, it should basically can't event be changed to another
For 3 and more values available in selector, it will hide current selected ( so the list will contain many items except current selected which looks good. After click on select, you can just type to filter out options and it works like google autocomplete (kind-off) which is still nice UX

I know it's still not what we want by this PR but sometimes it's easier to re-think solutions.

Looking for you opinion :)

Annie-Huang:
> > Not sure why this is not merged yet. But does anyone got a walk around solution for version @4.3.1, which is the latest version now (Sep 9th, 2021). The `getNextFocusedOption` doesn't seem to be exposed for this version.
> > Or please let me know which version I can use to use for the `getNextFocusedOption` work around solution?
> > Many thanks...
> 
> After many tries, we just decided that this is not an issue.
> Our selectors started just using search functionality.
> And the only issue we found, is that if:
> 
> * 2 values available in selector, it will hide current selected ( so the list will contain only one - the current selected is not shown ) which is not that bad
> * 1 values available in selector, it should basically can't event be changed to another
>   For 3 and more values available in selector, it will hide current selected ( so the list will contain many items except current selected which looks good. After click on select, you can just type to filter out options and it works like google autocomplete (kind-off) which is still nice UX
> 
> I know it's still not what we want by this PR but sometimes it's easier to re-think solutions.
> 
> Looking for you opinion :)

I am not sure I followed, @grzennio, when you said `After many tries, we just decided that this is not an issue.` Are you talking about your company trying to resolve the similar issue or you are speaking on behalf of the reac-select team?

Because I can see this is a very common request from the community. Google also not auto high-line on the first item that match the search term until user press down arrow key (see screen recording) 

https://user-images.githubusercontent.com/26481566/132780074-dfaede11-e8b9-4d9f-bdd7-4b19418b5430.mov

All I want is when i use AsyncSelect here and when I type `a`, don't auto focus on the `Ocean` option until i press the `down` key. 


https://user-images.githubusercontent.com/26481566/132780339-1192f015-da22-4e2a-9acd-379512cddb44.mov

raulhinojosa:
is this going to be merged?

Weswwn:
Following up on this PR

augusto-proano:
are you guys planning to merge this?

Bosch-Eli-Black:
@UlyssesInvictus Thank you so much for your example code! ^_^

It looks like your sample didn't account for when the user hovered over an option using the mouse, so I've modified your code to handle that case as well and put the code into a reusable component:

`SelectDoNotAutoselectFirstOption.jsx`:
```javascript
import React, { useState, useCallback } from 'react';
import Select, { components as ReactSelectComponents } from 'react-select';

/*
 * This component modifies React-Select so that when the React-Select dropdown is opened, it doesn't automatically select the first option.
 */
export const SelectDoNotAutoselectFirstOption = React.memo((props) => {
    const {
        onKeyDown: originalOnKeyDown,
        onMenuClose: originalOnMenuClose
    } = props;

    // Begin code from https://github.com/JedWatson/react-select/pull/4080#issuecomment-944563481
    const [hasUserSelectedOption, setHasUserSelectedOption] = useState(false);

    const onKeyDown = useCallback((e, ...otherArgs) => {
        if ((e.key === 'ArrowDown' || e.key === 'ArrowUp') && !hasUserSelectedOption) {
            e.preventDefault();

            setHasUserSelectedOption(true);
        }

        return originalOnKeyDown?.(e, ...otherArgs);
    }, [originalOnKeyDown, hasUserSelectedOption, setHasUserSelectedOption]);

    const onMenuClose = useCallback(
        (...args) => {
            setHasUserSelectedOption(false);

            return originalOnMenuClose?.(...args);
        },
        [originalOnMenuClose, setHasUserSelectedOption]
    );
    // End code from https://github.com/JedWatson/react-select/pull/4080#issuecomment-944563481

    const originalOptionComponent = props.components?.Option ?? ReactSelectComponents.Option;

    const OptionComponent = useCallback(
        (props) => {
            return originalOptionComponent({
                ...props,
                innerProps: {
                    ...props.innerProps,
                    onMouseOver: (...args) => {
                        setHasUserSelectedOption(true);

                        return props.innerProps.onMouseOver(...args);
                    }
                }
            });
        },
        [originalOptionComponent, setHasUserSelectedOption]
    );

    return (
        <Select
            {...props}
            components={{
                ...props.components,
                Option: OptionComponent
            }}
            onKeyDown={onKeyDown}
            onMenuClose={onMenuClose}
            styles={{
                ...props.styles,
                // The base of this function is from https://github.com/JedWatson/react-select/pull/4080#issuecomment-944563481
                option: (provided, selectProps) => {
                    const useUnselectedColors = (selectProps.isFocused && !hasUserSelectedOption);

                    return {
                        ...provided,
                        backgroundColor: useUnselectedColors ? 'transparent' : provided.backgroundColor,
                        // NOTE: You may also need to set 'color', etc.
                        // Ideally we'd do something like this: `useUnselectedColors ? selectProps.getStyles('normal', ...) : selectProps.getStyles('hover', ...)`, but I'm not sure what to pass to `getStyles()`
                    };
                }
                // End code from https://github.com/JedWatson/react-select/pull/4080#issuecomment-944563481
            }}
        />
    );
});

SelectDoNotAutoselectFirstOption.displayName = 'SelectDoNotAutoselectFirstOption';

export default SelectDoNotAutoselectFirstOption;
```

In your component, just use `SelectDoNotAutoselectFirstOption` instead of `Select`.

Instead of this:
```javascript
<Select
    name={...}
    options={...}
    onChange={...}
    value={...}
    theme={theme => ... }
    menuPlacement="auto"
    ...
/>

```

Use the `SelectDoNotAutoselectFirstOption` component:
```javascript
<SelectDoNotAutoselectFirstOption
    name={...}
    options={...}
    onChange={...}
    value={...}
    theme={theme => ... }
    menuPlacement="auto"
    ...
/>
```

Hope that's helpful to everyone, and thanks again to @UlyssesInvictus for the concept and base code! 🙂

augusto-proano:
> @UlyssesInvictus Thank you so much for your example code! ^_^
> 
> I've fixed/added a couple of things to your code and put everything into a reusable component 🙂
> 
> `SelectDoNotAutoselectFirstOption.jsx`:
> 
> ```js
> import React, { useState, useCallback } from 'react';
> import Select, { components as ReactSelectComponents } from 'react-select';
> 
> /*
>  * This component modifies React-Select so that when the React-Select dropdown is opened, it doesn't automatically select the first option.
>  * See https://github.com/JedWatson/react-select/pull/4080 for context.
>  */
> export const SelectDoNotAutoselectFirstOption = React.memo((props) => {
>     const {
>         onKeyDown: originalOnKeyDown,
>         onMenuClose: originalOnMenuClose
>     } = props;
> 
>     // Begin code from https://github.com/JedWatson/react-select/pull/4080#issuecomment-944563481
>     const [hasUserStartedSelectingOption, setHasUserStartedSelectingOption] = useState(props.value != null);
> 
>     const onKeyDown = useCallback((e, ...otherArgs) => {
>         if ((e.key === 'ArrowDown' || e.key === 'ArrowUp') && !hasUserSelectedOption) {
>             e.preventDefault();
> 
>             setHasUserSelectedOption(true);
>         }
> 
>         return originalOnKeyDown?.(e, ...otherArgs);
>     }, [originalOnKeyDown, hasUserSelectedOption, setHasUserSelectedOption]);
> 
>     const onMenuClose = useCallback(
>         (...args) => {
>             setHasUserSelectedOption(false);
> 
>             return originalOnMenuClose?.(...args);
>         },
>         [originalOnMenuClose, setHasUserSelectedOption]
>     );
>     // End code from https://github.com/JedWatson/react-select/pull/4080#issuecomment-944563481
> 
>     const originalOptionComponent = props.components?.Option ?? ReactSelectComponents.Option;
> 
>     const OptionComponent = useCallback(
>         (props) => {
>             return originalOptionComponent({
>                 ...props,
>                 innerProps: {
>                     ...props.innerProps,
>                     onMouseOver: (...args) => {
>                         setHasUserSelectedOption(true);
> 
>                         return props.innerProps.onMouseOver(...args);
>                     }
>                 }
>             });
>         },
>         [originalOptionComponent, setHasUserSelectedOption]
>     );
> 
>     return (
>         <Select
>             {...props}
>             components={{
>                 ...props.components,
>                 Option: OptionComponent
>             }}
>             onKeyDown={onKeyDown}
>             onMenuClose={onMenuClose}
>             styles={{
>                 ...props.styles,
>                 // The base of this function is from https://github.com/JedWatson/react-select/pull/4080#issuecomment-944563481
>                 option: (provided, selectProps) => {
>                     const useUnselectedColors = selectProps.isFocused && !hasUserSelectedOption && props.value == null;
> 
>                     return {
>                         ...provided,
>                         backgroundColor: useUnselectedColors ? 'transparent' : provided.backgroundColor,
>                         // NOTE: You may also need to set 'color', etc.
>                         // Ideally we'd do something like this: `useUnselectedColors ? selectProps.getStyles('normal', ...) : selectProps.getStyles('hover', ...)`, but I'm not sure what to pass to `getStyles()`
>                     };
>                 }
>                 // End code from https://github.com/JedWatson/react-select/pull/4080#issuecomment-944563481
>             }}
>         />
>     );
> });
> 
> SelectDoNotAutoselectFirstOption.displayName = 'SelectDoNotAutoselectFirstOption';
> 
> export default SelectDoNotAutoselectFirstOption;
> ```
> 
> In your component, just use `SelectDoNotAutoselectFirstOption` instead of `Select`.
> 
> Instead of this:
> 
> ```js
> <Select
>     name={...}
>     options={...}
>     onChange={...}
>     value={...}
>     theme={theme => ... }
>     menuPlacement="auto"
>     ...
> />
> ```
> 
> Use the `SelectDoNotAutoselectFirstOption` component:
> 
> ```js
> <SelectDoNotAutoselectFirstOption
>     name={...}
>     options={...}
>     onChange={...}
>     value={...}
>     theme={theme => ... }
>     menuPlacement="auto"
>     ...
> />
> ```
> 
> Hope that's helpful to everyone, and thanks again to @UlyssesInvictus for the concept and base code! 🙂
> 
> One remaining issue: When the user presses enter, it still select the first option in the dropdown.

Do you guys have a solution for `Async Select`

Bosch-Eli-Black:
@augusto-proano I've never worked with `Async Select`, sorry. I imagine that probably the same principal would work, though: Once the results are loaded, track whether the user has interacted with the dropdown (pressed the up/down arrow or hovered over an option with the mouse) and call `e.preventDefault();` if they haven't yet.

But... ya, it's going to be complicated 😛

augusto-proano:
Thanks for the info @Bosch-Eli-Black. Do you know about any other way to accomplish this expected behavior without this pr since it hasn't been merged yet?

UlyssesInvictus:
> Thanks for the info @Bosch-Eli-Black. Do you know about any other way to accomplish this expected behavior without this pr since it hasn't been merged yet?

I spent a decent amount of time looking into this when I first posted my original comment, and I really don't think there is.

I'm going to reiterate that I think the PR itself is vastly out of date. I don't think it's anywhere close to being merged in its current state.

As for async select, it should be completely identical to the regular select version, the only possible difference would be in not firing any effects until after the options load, but the sample code above doesn't really make any assumptions on that front. I'm sure with some trial and error it should be possible.

Bosch-Eli-Black:
Yep 🙂  @UlyssesInvictus or my code samples should get you off to a good start 🙂

I'm at work, so I'm unable to contribute a PR, due to company policy.

Rall3n:
Hello @Bosch-Eli-Black, @UlyssesInvictus, @augusto-proano,

we appreciate you offering workarounds for problems, but also see the need for something like this PRs feature being configurable.
Proper menu interaction is still a part of our road map, as is this PR. But as you already mentioned, it is heavily outdated, having unaddressed reviews and needing to be rewritten in TypeScript (as we switched to it for version `5.0.0`).

I will try to contact @bladey to see if he has time to update this PR, but you are welcome to fork, update and submit a new PR based on this one any time.

Yair-Ohana:
Any estimation for that one? extremely necessary ..

dcousens:
Related
- https://github.com/JedWatson/react-select/pull/4080

flippidippi:
I'm working on an autocomplete search and I was able to get around this with a custom `Option` component and sending the current input as an option. It is the least nasty way that I've found, and I like that using the arrow keys will "select" the input option if you go past the first or last real option.

```tsx
interface OptionType {
  value: string
  label: string
}

function Option ({ children, ...props }: OptionProps<OptionType>): ReactElement | null {
  if (props.data?.label === '') {
    return null
  }

  return (
    <components.Option {...props} className='fw-normal'>
      {children}
    </components.Option>
  )
}

function Search (): ReactElement {
  const [input, setInput] = useState('')
  const [value, setValue] = useState<OptionType | null>(null)
  const [menuIsOpen, setMenuIsOpen] = useState(false)
  const options = useMemo<OptionType[]>(() => ([
    { value: input, label: '' },
    { value: 'value1', label: 'label 1' },
    { value: 'value2', label: 'label 2' }
  ]) , [input])

  const doWork = useCallback((term: string) => {
    console.log('doWork', term)
  }, [])

  const handleInputChange = useCallback((value: string, actionMeta: InputActionMeta) => {
    if (actionMeta.action === 'input-change') {
      setMenuIsOpen(true)
      setInput(value)
    }
  }, [])

  const handleValueChange = useCallback((option: OptionType, actionMeta: ActionMeta<OptionType>) => {
    setMenuIsOpen(false)
    setValue(option)
    setInput(option?.label ?? '')
  }, [])

  const handleBlur = useCallback(() => {
    setMenuIsOpen(false)
  }, [])

  const handleFocus = useCallback(() => {
    setMenuIsOpen(true)
  }, [])

  const handleSubmit = useCallback((event: FormEvent<HTMLFormElement>) => {
    event.preventDefault()
    doWork(value?.value ?? input)
  }, [doWork, input, value])

  return (
    <form onSubmit={handleSubmit}>
      <Select
        value={value}
        onChange={handleValueChange}
        options={options}
        inputValue={input}
        onInputChange={handleInputChange}
        filterOption={() => true}
        components={{ Option }}
        menuIsOpen={menuIsOpen}
        onBlur={handleBlur}
        onFocus={handleFocus}
        blurInputOnSelect={true}
        closeMenuOnSelect={false}
      />
      <button type='submit'>Search</button>
    </form>
  )
}
```

dding-g:
In my case, this is work.
I think, hover is work but first element is always focus when first opened.
So, make select item option background `transparent` when focused but not selected.

[reference](https://github.com/JedWatson/react-select/pull/4080#issuecomment-944563481)

```
    <ReactSelector
      ...
       styles={{
        option: (base, selectProps) => ({
          ...base,
          padding: 0,
          backgroundColor:
            selectProps.isFocused && !selectProps.isSelected
              ? 'transparent'
              : base.backgroundColor,
        }),
       ...
      />
```

abdullah-ch-empg:
Need this feature ASAP, PLEASE. My Job depends on this.

francoexequielschuler:
Hey guys! Usually I don't really comment in this kind of things but I managed to code a workaround for this. I just added these styles to the Option component (based on [this response](https://github.com/JedWatson/react-select/pull/4080#issuecomment-1488064761) which was't really working fine for me)

```
option: (provided, { isSelected, isFocused }) => ({
    ...provided,
    backgroundColor: 'transparent',
    _hover: {
        backgroundColor:
            !isFocused || isSelected
                ? isSelected
                    ? 'your_hover_color'
                    : 'transparent'
                : 'your_hover_color',
    },
}),
```

I know it is not the best solution but I works pretty fine for me! Hope this helps anyone, at least until this PR is merged 👍

EamonNerbonne:
I ran into this problem too while trying to use `match-sorter` with `react-select`.  When doing so, match-quality according to match-sorter can change whenever the user changes their search-query. In particular, what was previously a decent match can stay a reasonable match after an extra letter is typed (or removed), but be eclipsed by a clearly superior match.  When that happens, react-select retains focus on the already-focused option, even though that's confusing; there's a much better match now.

This workaround suffices to reset the focus for me if-and-only-if matchSorter finds a better one.  One (unfortunate) side-effect is that the focused item is moved to the top of the list of options when there are multiple options that are equally good matches.

```JSX
const originalOptions = reactSelectProps.options;
const [options, setOptions] = useState(originalOptions);
const selectRef = useRef<SelectInstance<MyOptionType>>(null);
//...
return <Select
  ref={selectRef}
  filterOption={() => true}
  onInputChange={(search) => {
      let toFilter = originalOptions;
      if (selectRef.current?.state.focusedOption) {
          const oldIdx = toFilter.indexOf(selectRef.current.state.focusedOption);
          if (oldIdx > 0) {
              //give the old focused item the best chances of retaining focus...
              toFilter = [toFilter[oldIdx], ...toFilter.slice(0, oldIdx), ...toFilter.slice(oldIdx + 1, toFilter.length)];
          }
      }
      const filtered = search
          ? matchSorter(originalOptions, search, matchSorterConfig)
          : originalOptions;
      setOptions(filtered);

      if (filtered[0] && selectRef.current && selectRef.current.state.focusedOption) {
          //...but if there's truly a new best match, then do focus that instead.
          selectRef.current.state.focusedOption = filtered[0];
      }
  }}
//[...]
```


I suspect there are better workarounds, but I'm all ears...

RachelElysia:
Figured out some work around: 
1. allow keyboard accessibility to show styling border when tabbing through app, but not when hovering/clicking dropdown
2. allow keyboard accessibility to show focus on first option when tabbing through app, but not when clicking menu 

```
  const [isKeyboardFocus, setIsKeyboardFocus] = useState(false);

  useEffect(() => {
    const handleKeyDown = (e: KeyboardEvent) => {
      if (e.key === "Tab") {
        setIsKeyboardFocus(true);
      }
    };

    const handleMouseDown = () => {
      setIsKeyboardFocus(false);
    };

    document.addEventListener("keydown", handleKeyDown);
    document.addEventListener("mousedown", handleMouseDown);

    return () => {
      document.removeEventListener("keydown", handleKeyDown);
      document.removeEventListener("mousedown", handleMouseDown);
    };
  }, []);
```

Updating control custom styles to only show a border on the dropdown when tabbing through app (like in my case on a user menu)
```
  const customStyles: StylesConfig<IDropdownOption, false> = {
    control: (provided, state) => ({
      ...provided,
      border: "2px solid transparent", // So tabbing doesn't shift dropdown
      borderRadius: "6px",
      ...(state.isFocused &&
        isKeyboardFocus && {
          border: `2px solid ${COLORS["ui-blue-25"]}`,
        }),
      }),
    }),
```

Create a custom option component that will only be focused when tabbing through the app, so the first option is not focused on clicking on the dropdown:
```
const CustomOption: React.FC<
  OptionProps<IDropdownOption, false> & { isKeyboardFocus: boolean }
> = (props) => {
  const { innerRef, data, isFocused, isKeyboardFocus } = props;

  return (
    <components.Option
      {...props}
      isFocused={isKeyboardFocus ? isFocused : false} // work around to not preselect first option unless keyboarding
    >
      <div
        className={`${baseClass}__option`}
        ref={innerRef}
        // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
        tabIndex={0}
        role="menuitem"
      >
        {data.label}
      </div>
    </components.Option>
  );
};
```

```
   <Select<IDropdownOption, false>
        options={dropdownItems}
        placeholder={renderPlaceholder()}
        styles={customStyles}
        components={{
          Option: (props) => (
            <CustomOption {...props} isKeyboardFocus={isKeyboardFocus} />
          ),
          // ... rest of code
        }}
        // ... rest of code
      />
```

Lots of extra lines of code and time coding this. Wondering if I regret trying to move all my dropdowns to react-select 5.4 

Open source work: https://github.com/fleetdm/fleet/pull/24281

PR #4082, opened 2020-06-11
Add aria-describedby attribute

bladey:
As per PR #3273, updated again master and new tests.

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -75,6 +75,8 @@ type FormatOptionLabelMeta = {
 };
 
 export type Props = {
+  /* HTML ID of an element that should be used to describe this input (for assistive tech) */
+  'aria-describedby'?: string,
   /* Aria label (for assistive tech) */
   'aria-label'?: string,
   /* HTML ID of an element that should be used as the label (for assistive tech) */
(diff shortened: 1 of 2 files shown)


darkowic:
Nice! I just run into an issue with it. 🚀  Waiting for it to be released! @JedWatson

darkowic:
Btw what about support for other aria attributes like `aria-required` etc?

kylemh (approved):
Small adjustment to comments, but this looks great.

packages/react-select/src/Select.js
```diff
@@ -75,6 +75,8 @@ type FormatOptionLabelMeta = {
 };
 
 export type Props = {
```
kylemh, line 77: An alternative suggestion:

```tsx
import { AriaAttributes } from 'react';

// ...

// aria attributes makes the JSX "noisy", separated for clarity
const ariaAttributes = {
  'aria-autocomplete': 'list',
  'aria-describedby': this.props['aria-describedby'],
  'aria-label': this.props['aria-label'],
};

export type Props = Pick<AriaAttributes, keyof typeof ariaAttributes> & {
  // ... the other props
};

```

```tsx
type Demo = keyof typeof ariaAttributes;
// Demo --> 'aria-autocomplete' | 'aria-describedby' | 'aria-label' | 'aria-labelledby'
```

noelleleigh:
@bladey Any chance this PR could be rebased and reviewed? Our team would love to use React Select, but the lack of support for directly specifying `aria-describedby` is a real concern.

ryanscottaudio:
@bladey same here, would love some 👀 on this

Issue #3055, opened 2018-09-20
All Option components being re-rendered due to cx and getValue props changing at each render


adam1658:
Hi there, thanks for a great select component!

I am using a custom Option component and have noticed poor performance when moving through the options with the keyboard or mouse hover.
My Option component extends React.PureComponent, so shouldn't update unless it's props change. So when moving focus from one option to the next, only 2 Options should re-render (the one losing focus, and the one gaining focus).
I have noticed however, that all my Options are re-rendering, because the `cx` and `getValue` props are changing.
This really slows the render down in my application when moving focus - it takes ~60ms to re-render the list when re-rendering all my options (~150 of them)

I can work around it by implementing shouldComponentUpdate on my custom Option like this:
```js
shouldComponentUpdate(nextProps) {
    return !shallowEqual( // true if each key value pair in the provided objects is equal (===)
        omit(this.props, ['cx', 'getValue']), // omit returns a copy of the provided object, without the 'cx' or 'getValue' keys
        omit(nextProps, ['cx', 'getValue']),
    );
}
```

Here's a codesandbox which logs to the console when an Option is updating and `cx` or `getValue` changes seemingly unnecessarily: https://codesandbox.io/s/ypy2p2vqp9?module=%2Fexample.js
Open the console, open the select then move focus up and down with the keyboard. This codesandbox doesn't feel laggy like it does in my real world application, but does show the props changing.

Thanks,
Adam

hrafaelveloso:
Does this issue have one workaround? I'm facing the same problem.

PR #4084, opened 2020-06-11
[3055] prevent from passing different cx and getValue functions to children components when the results of those functions remain the same as previous render.

manvydasu:
During the render of Select component, functions cx and getValue are recreated as new functions everytime and are passed to children components. Sometimes the result of cx and getValue is same as previous render, but the functions are still re-created thus causing unnecessary re-renders as result. The issue was originally described here: https://github.com/JedWatson/react-select/issues/3055

Would love if someone could suggest a cleaner way to fix this.


Rall3n:
I think a better/cleaner solution for both would be to create the functions on class level and just add them to the `commonProps` object.

```jsx
getValue = () => this.state.selectValue;

cx = (...args) => classNames(this.props.classNamePrefix, ...args);

/* ... */
getCommonProps() {
    const { /* ... */ cx, getValue, props } = this;

    /* ... */

    return {
        cx,
        getValue,
        /* ... */
    };
}
```

No need for memoization and no need for "creation"-functions.

manvydasu:
@Rall3n 
Yeah, you're right. I over engineered that originally. Updated just as you suggested.

bladey:
Look good to me, @JedWatson to review.

packages/react-select/src/Select.js
```diff
@@ -742,14 +742,15 @@ export default class Select extends Component<Props, State> {
     };
   }
 
+  getValue = () => this.state.selectValue;
+
+  cx = (...args: any) => classNames(this.props.classNamePrefix, ...args);
```
Methuselah96, line 747:
```suggestion
  cx = (...args: Array<any>) => classNames(this.props.classNamePrefix, ...args);
```

Methuselah96 (changes requested):

JedWatson (approved):

JedWatson:
Thanks @manvydasu

Issue #4096, opened 2020-06-23
Long word in options is overflowing the select box


annamatveev:
When adding a long value with no spaces for an option, it doesn't break the word to multiple lines.

https://imgur.com/a/ZqCYtZh

ebonow:
I'd say this is more of a feature than a bug. 

Splitting a word at non-naturally occurring white-spaces is not the default behavior of html elements. Fortunately, this is accessible via the `word-break` CSS property ([reference](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break))

If you would like to implement this in your own solution, you do have access to the styling of the options so that you could implement this in a fairly straight forward manner.

```
const styles = {
  option: inlineCss => ({ ...inlineCss, wordBreak: 'break-word' })
};

<Select styles={styles} ...>
```

Please feel free to close this issue if this resolves your use-case.

annamatveev:
Thanks! I created a PR to solve this issue, if you'd like to merge this solution.

ebonow:
I believe you misunderstand. This is not an issue and not something that should be merged into master.

The solution creates an unexpected result as the default behavior of any HTML element is that it should not wrap to new line if there are no spaces. 

My biggest concern with this is that it could interfere with use-cases where a user would want to instead trigger `text-overflow: ellipsis` instead.

That said, it is not a bad idea in principle (especially for those supporting German language) and I appreciate contributions, but wanted to be mindful of negative side effects.

PR #4097, opened 2020-06-23
Break a long work in an option

annamatveev:
Solves issue #4096


ebonow:
I disagree with this PR. It violates the natural expected behavior of HTML elements to not wrap if there are no spaces and it could interfere with use-cases where a user would want to instead trigger `text-overflow: ellipsis`.

bladey:
Thanks for the PR @annamatveev - @ebonow has a good point, if you'd like to go ahead with this change, please fork the repository if you haven't already.

Issue #4094, opened 2020-06-18
Warnings: componentWillReceiveProps in strict mode, Legacy context API, findDOMNode is deprecated


bladey:
_This issue is a round-up of multiple past issues documenting the same bug. Feel free to check out the linked issues below for more information. This issue is the source of truth going forward to investigate the issue, report findings, and implement a bug fix._

**Issue:**

<img width="715" alt="W35yYs" src="https://user-images.githubusercontent.com/737821/84984363-ffde2b80-b17d-11ea-8842-9cc8a2f0eb5c.png">

```
Warning: Using UNSAFE_componentWillReceiveProps in strict mode is not recommended and may indicate bugs in your code. See https://fb.me/react-unsafe-component-lifecycles for details.
```

```
Warning: Legacy context API has been detected within a strict-mode tree.

The old API will be supported in all 16.x releases, but applications using it should migrate to the new version.
```

```
Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of NodeResolver which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://fb.me/react-strict-mode-find-node
```

**Issues:**

https://github.com/JedWatson/react-select/issues/3751
https://github.com/JedWatson/react-select/issues/3720
https://github.com/JedWatson/react-select/issues/3915
https://github.com/JedWatson/react-select/issues/3745
https://github.com/JedWatson/react-select/issues/4068
https://github.com/JedWatson/react-select/issues/3703

If anyone would like to share any further information, please do so.

giangr21:
Same issue here

Sine99:
same issue

manvydasu:
So basically someone would have to create a PR and replace legacy component lifecycle methods with newer version: https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#migrating-from-legacy-lifecycles

stak99:
same issue

corysimmons:
For anyone else going insane over this, you can disable those warnings (and other, actually helpful ones) by commenting out `<React.StrictMode>` in `index.js`. Just remember to uncomment it a few times a day to get the helpful warnings.

martijnhalekor:
Same issue here.

roberto-souza:
Hello! The problem is not solved yet? :(

OlegPuzankin:
Can anyone say  - are these warnings critical and can we expect the developers to fix that in the future?

[...]

PR #4102, opened 2020-06-27
Replace NodeResolver with Ref usage

Rall3n:
With React 16.x and `StrictMode`, the usage of `findDOMNode` has been deprecated.

This PR replaces the `NodeResolver` component, which used the deprecated function with an already existing Ref to the nearest DOM node of the `MenuList` component.

Related to #4094.


bladey:
Thank you so much for this PR @Rall3n.

bladey:
Hey @Rall3n,

It looks like the tests are failing (9 errors):

```
Cannot call this.startListening with this.props.targetRef bound to el because null [1] is
incompatible with HTMLElement [2].

 [1] 11│   targetRef: HTMLElement | null,
       :
     17│   touchStart: number;
     18│
     19│   componentDidMount() {
     20│     this.startListening(this.props.targetRef);
     21│   }
     22│   componentDidUpdate(prevProps) {
     23│     if (prevProps.targetRef != this.props.targetRef) {
       :
 [2] 31│   startListening(el: HTMLElement) {
```

If I remove `null` from `targetRef`:

IBLczp

It eliminates all errors but one:

```
Missing type annotation for prevProps.

     19│   componentDidMount() {
     20│     this.startListening(this.props.targetRef);
     21│   }
     22│   componentDidUpdate(prevProps) {
     23│     if (prevProps.targetRef != this.props.targetRef) {
     24│       if (prevProps.targetRef) this.stopListening(prevProps.targetRef);
     25│       this.startListening(this.props.targetRef);
```

Which seems to be a missing type for the new `prevProps` argument you've added.

I may be completely off the mark here for all of the above, but I wanted to give it a go.

Let me know!

Rall3n:
Hey @bladey,

I was not sure how to type a legacy `ref`.
The prop `targetRef` could be `null` on component mount. That is why I added the `componentDidUpdate` lifecycle and the `null` typing to the prop.
If it is not necessary, then it can be removed.

The `prevProps` typing is simple. I can add a commit right away for that.

packages/react-select/src/internal/ScrollCaptor.js
```diff
@@ -2,27 +2,31 @@
...
 
-import NodeResolver from './NodeResolver';
-
 export type CaptorProps = {
   children: Element<*>,
   onBottomArrive?: (event: SyntheticEvent<HTMLElement>) => void,
   onBottomLeave?: (event: SyntheticEvent<HTMLElement>) => void,
   onTopArrive?: (event: SyntheticEvent<HTMLElement>) => void,
   onTopLeave?: (event: SyntheticEvent<HTMLElement>) => void,
+  targetRef: HTMLElement | null,
```
bladey, line 11: To be checked

packages/react-select/src/internal/ScrollBlock.js
```diff
@@ -2,30 +2,20 @@
...
-import NodeResolver from './NodeResolver';
 import ScrollLock from './ScrollLock/index';
 
 type Props = {
   children: Element<*>,
   isEnabled: boolean,
-};
-type State = {
-  touchScrollTarget: HTMLElement | null,
+  targetRef: HTMLElement | null,
```
bladey, line 10: To be checked

bladey:
Thanks @Rall3n! Looks good, I'll leave it to @JedWatson to check the two comments above for the tests to start passing or if we need further work.

Rall3n:
@bladey @JedWatson Any information regarding this pr / the failed checks for the reference typing?

packages/react-select/src/internal/ScrollCaptor.js
```diff
@@ -2,27 +2,31 @@
...
+  }
+  componentDidUpdate(prevProps: CaptorProps) {
+    if (prevProps.targetRef != this.props.targetRef) {
+      if (prevProps.targetRef) this.stopListening(prevProps.targetRef);
+      this.startListening(this.props.targetRef);
+    }
   }
   componentWillUnmount() {
-    this.stopListening(this.scrollTarget);
+    this.stopListening(this.props.targetRef);
```
Methuselah96, line 29: Can we be sure that `this.props.targetRef` is not `null` her? To be honest, I've never quite understood when a ref can or can't be `null`, but based on [the documentation](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs) I'm wondering if it could be `null` in this scenario:
> React will call the ref callback with the DOM element when the component mounts, and call it with null when it unmounts. Refs are guaranteed to be up-to-date before componentDidMount or componentDidUpdate fires.


I'd be happy to get the Flow types working with `| null` if that's helpful. I have it working on a local branch.

Methuselah96 (changes requested):

packages/react-select/src/internal/ScrollCaptor.js
Rall3n: Well, `componentWillUnmount` is neither `componentDidMount` or  `componentDidUpdate`.

> `componentWillUnmount()` is invoked immediately before a component is unmounted and destroyed. Perform any necessary cleanup in this method, such as invalidating timers, canceling network requests, or cleaning up any subscriptions that were created in `componentDidMount()`.

This lifecycle is made for such things, so I am pretty confident that the reference will never be null at this point.

Methuselah96: Yeah, I agree that `componentWillUnmount` is usually a good place to remove event listeners from refs, etc. However, this is not a normal ref situation since the grandparent has a ref to the grandchild and is passing it into the parent of the grandchild. Therefore, I'm more wary than usual.

I made [a CodeSandbox](https://codesandbox.io/s/frosty-snowflake-bdmd3?file=/src/App.js) to see how this situation would behave, and it is possible to have a `null` value in `componentWillUnmount` if the top-level parent doesn't re-render (if you press "Toggle mount status of ReactSelect" immediately after loading the page). However, if the component is re-rendered (by pressing "Set state on ReactSelect") the ref will be set in the props in `componentWillUnmount`.

It looks like the `Select` component in `react-select` is always re-rendered by something pretty quickly, but I don't think it's worth relying on that being the case. That's why I'd rather be safe and just check for `null` where possible.

Rall3n: The example seems very specific, but the reasoning is understandable. 

I will push a commit with a `null`-check.

Methuselah96: Thanks, I'm sorry if my concerns are coming off as pedantic, just trying to make sure we catch the edge cases.

After thinking about this some more, I realized that this means that if `ScrollBlock` and `ScrollCaptor` fail to re-render after the `menuListRef` in `Select` has been set, then the scroll blocking and capturing won't work correctly. I've been trying to figure out whether we know for sure if those re-renders will happen or not (they seem to re-render in every case I've tried), but I'm having trouble proving it one way or the other. Do you have any insight?

If we can't know for sure that those components will re-render, I was thinking that a pretty safe way to make sure they re-render would be to introduce a `ScrollManager` component where the `ref` is in `state`, so it will re-render on state change. Something like this:
```
export default function ScrollManger({
  captureMenuScroll,
  onMenuScrollToBottom,
  onMenuScrollToTop,
  menuShouldBlockScroll,
  children,
}: Props) {
  const childElement = Children.only(children);
  const { innerRef } = childElement.props;

  // must be in state to trigger a re-render, only runs once per instance
  const [scrollTarget, setScrollTarget] = useState();
  const refCallback = useCallback(element => {
    innerRef(element);
    if (element === scrollTarget) return;
    setScrollTarget(element);
  });

  return (
    <ScrollCaptor
      isEnabled={captureMenuScroll}
      onTopArrive={onMenuScrollToTop}
      onBottomArrive={onMenuScrollToBottom}
      scrollTarget={scrollTarget}
    >
      <ScrollBlock
        isEnabled={menuShouldBlockScroll}
        scrollTarget={scrollTarget}
      >
        {React.cloneElement(childElement, { innerRef: refCallback })}
      </ScrollBlock>
    </ScrollCaptor>
  );
}
```

Still looking into it, let me know if you have any ideas.

Methuselah96: Looks like the re-render is due to [a `setState` in `Menu`](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/components/Menu.js#L302) which is guaranteed to happen.

My concern is that it would be easy to overlook a change that doesn't cause `Menu` to re-render, which would inadvertently make it so that `ScrollCaptor` and `ScrollBlock` don't work correctly, so my preference would still be to explicitly make sure that we're updating the state after receiving the `ref` using something like the code snippet I posted above. Let me know what you think.

Rall3n: I think your code snippet would be too over engineered, just simple overkill. I believe a better way would be to let the components have their own reference to the DOM Element, and not through props.

But I also think this should be part of another discussion/PR. I already have something in mind and will get to it.

The sole focus of this PR is to remove deprecated methods without changing current functionality, and I think this has been accomplished. Everything else is just preventing this from being merged.

Methuselah96: I completely agree that my proposed solution is not ideal and that letting the components have their own reference to the DOM element would be much better (I would be interested to see how these components could get a ref to the DOM node without getting it through `innerRef`). And I also completely agree on what the focus of this PR should be (although I still think having a technical discussion on how to accomplish the goal is still appropriate).

My reasons for preferring a solution that explicitly causes a re-render are:

1. The current solution in this PR relies on implicit behavior of the `Menu` component working a certain way. This means if these components are either no longer a child of the `Menu` or the functionality of `Menu` changes that it would be very easy to miss the fact that that would cause the `ScrollBlock` and `ScrollCaptor` to no longer work (since it's not obvious that these components depend on the functionality of `Menu`). Making the re-render explicit would make it more likely that we won't break this functionality by mistake when working with parts of the code that aren't related to the scroll components.
2. It's hard to predict how long a certain piece of code is going to live for or in what context someone's going to try to use your code (unless there's an active PR that is about to get merged). Therefore, if there's an easy solution that is a better long-term solution, I usually prefer the better long-term solution.
3. This functionality is not tested, so it would be easy for it to break without anyone noticing.
4. My code snippet was completely based on logic that was already in `ScrollBlock` (putting the `ref` in `state` in order to force a re-render), I just put it in a higher component so that it would apply to both `ScrollBlock` and `ScrollCaptor`. (I tried having each component put the `ref` in state, but it would be hard to make sure that the `innerRef` passed into `MenuList` isn't called multiple times in that situation.)

As you noted, if both of us think our approach is the better solution, then this discussion is holding up the merging of this PR. Thanks for bearing with me while I figure out how some of these internal components work.

Methuselah96: I think I gave up too easily on making `ScrollBlock` and `ScrollCaptor` responsible for getting the DOM node themselves. I came with [a solution](https://github.com/Methuselah96/react-select/commit/14451b77afb5b6348a7b84cdfd64e69760a154e2) that doesn't rely on a `ScrollManager`, is a very minimal code change, and handles the `ref` internally inside of `ScrollBlock` and `ScrollCaptor` (which means that we're no longer relying on the `Menu` state change to trigger a re-render), Can you take a look and see what you think?

Methuselah96: It seems like there ought to be some HOC or React hooks way to do this generically, but I can't seem to figure it out.

Rall3n: I´m not a fan of using `cloneElement` to force extra props onto a component. And I´m guaranteed not a fan of storing refs in the state.

I agree that the components should handle the references by themselves, but not like this. Instead I would suggest a solution using `React.createRef` and using a function as a child (much like the `MenuPlacer` component) to forward the ref to the DOM element.

```jsx
<ScrollCaptor>{scrollTargetRef => (
    <ScrollBlock>{blockTargetRef => (
        <div ref={instance => {
            this.getMenuListRef(instance)
            scrollTargetRef.current = instance;
            blockTargetRef.current = instance;
        }}
        >
            { ... }
        </div>)}
    </ScrollBlock>)}
</ScrollCaptor>
```
This makes the components independent of the `menuListRef` reference and React guarantees that the references are set before the lifecycle methods are called (`componentDidMount`, `componentDidUpdate`).

------

Tested it without the `MenuPlacer` and `Menu` components to prevent possible re-renders. Both elements worked as expected.

------

But as I already said and will say again: **This is a matter for another PR. Get this merged to resolve `StrictMode` errors as soon as possible.**

Methuselah96:
Replaced by https://github.com/JedWatson/react-select/pull/4330.

PR #3666, opened 2019-07-05
add autoComplete as prop to the Select component

lvl99:
As mentioned  in issue https://github.com/JedWatson/react-select/issues/3500 and  https://github.com/JedWatson/react-select/pull/2395 and having my own need to set an explicit `autoComplete` value (namely to prevent Google Chrome from showing its autofill drop-down on top of the `react-select` drop-down) I've attempted a quick simple fix.

Issue #3500, opened 2019-04-03
Disable Chrome autofill allowing autoComplete="new-password" to be passed into the Input component properties


rdsedmundo:
Currently, the default Input component is forcefully setting `autoComplete="off"` which doesn't disable Chrome's autofill functionality on latest browser versions. For achieving that, we need to pass `autoComplete="new-password"`.

The problem is that the value is hardcoded and can't be configured unless we implement our own Input component and configure this property properly there.

It's hardcoded here: https://github.com/JedWatson/react-select/blob/ba76246a92fe9371b5d4f8795d30119b045dcaba/src/Select.js#L1401

And it gets passed to the default Input component here: https://github.com/JedWatson/react-select/blob/ba76246a92fe9371b5d4f8795d30119b045dcaba/src/components/Input.js#L53

As you can see, there's no way of configuring that property manually.

kylehurt-rkv:
I agree that this would be a welcome change.  I have been fighting with this for the last couple of days.  It is a dealbreaker for our app because the autocomplete control overlays the react-select menu.  We have resorted to forking the project, but it would be nice if this were supported natively.

Rall3n:
> As you can see, there's no way of configuring that property manually.

@rdsedmundo Yes, there is a possibility to configure it. It is hard coded as a prop to the `Input` component, that is true. But this library offers in its current version the possibility to overwrite its internal components using its [components framework](https://react-select.com/components).

You just have to overwrite the `Input` component and overwrite the prop being passed to the component.

```jsx
import Select, { components } from 'react-select';

const Input = ({ autoComplete, ...props }) => <components.Input {...props} autoComplete="new-password" />;

<Select
    { ... }
    components={{
        Input
    }}
/>
```

> We have resorted to forking the project, but it would be nice if this were supported natively.

@kylehurt-rkv Here is your native support, without forking the repository.

rdsedmundo:
That's a smarter solution than what I did. I knew I could just overwrite the `Input`, that's what I did for fixing it, but I just copied its source code originally and pasted on my codebase. I haven't thought about the possibility of just importing it from the package.

I still can see value of having it configurable though.

kylehurt-rkv:
@Rall3n We actually tried that exact solution, but whenever the page would try to load, Chrome would get of memory errors and crash.  We are supplying several other custom components into the Select component, so it may be possible that something we did on one of the other custom components was conflicting with custom input component.  Or maybe we are doing something in a non-standard way.
[...]

PR #2395, opened 2018-02-22
Support Chrome autofill

davidcoleman007:
This PR addresses #1397 .  

Originally a PR was raised by @sevenval and @mmintel. that pr is: https://github.com/JedWatson/react-select/pull/2174

The purpose of this PR is to resolve the merge conflicts present in the original PR, and provide a working proposal for merge with current master as of 2/22/18.

My work is available at: https://github.com/davidcoleman007/react-select

PR #4105, opened 2020-07-02
add autoComplete as prop to the Select component

bladey:
Based off PR #3666 by @lvl99

Updated tests and other misc changes.

> As mentioned in issue #3500 and #2395 and having my own need to set an explicit autoComplete value (namely to prevent Google Chrome from showing its autofill drop-down on top of the react-select drop-down) I've attempted a quick simple fix.

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -51,17 +51,17 @@ import { defaultTheme, type ThemeConfig } from './theme';
 
 import type {
   ActionMeta,
-    ActionTypes,
-    FocusDirection,
-    FocusEventHandler,
-    GroupType,
-    InputActionMeta,
-    KeyboardEventHandler,
-    MenuPlacement,
-    MenuPosition,
-    OptionsType,
-    OptionType,
-    ValueType,
+  ActionTypes,
+  FocusDirection,
+  FocusEventHandler,
+  GroupType,
+  InputActionMeta,
+  KeyboardEventHandler,
+  MenuPlacement,
+  MenuPosition,
+  OptionsType,
+  OptionType,
+  ValueType,
 } from './types';
 
 type MouseOrTouchEvent =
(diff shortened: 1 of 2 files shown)


bladey:
@JedWatson to be reviewed on commented on.

gotrecillo:
please this is super needed to avoid that awful address manager of chrome

CoOlDud3:
Any updates on this?

viniciuscdavila:
My project is also super needing this!

graham-saunders:
This would be a great addition to react-select. Any updates?

Methuselah96:

ebonow:
This seems somewhat unnecessary given the existing component api.

```JS
const Input = props => {
  const { autoComplete='off' } = props.selectProps;
  return <components.Input {...props} autoComplete={autoComplete} />
}

const MySelect => <Select autoComplete="new-password" components={{ Input }} {...otherProps} />
```

cullylarson:
@ebonow Your solution doesn't work for me using `autoComplete='address-level1'`. The field is filled, but is then immediately cleared and the select is focused. Not sure if this is an issue with react-select focusing the field after input, which somehow triggers the field to be cleared (e.g. race condition), or potentially an issue with controlled form state.

ebonow:
@cullylarson Adding autoComplete to the input prop will add the attribute to the input, but using the autoFill selection is a different conversation.

In regards to your findings, you are also correct. The issue is that browser support for autoFilling inputs is neither standard nor well supported. Here is an [older SO link](https://stackoverflow.com/questions/11708092/detecting-browser-autofill) highlighting the issue. The first posted answer gives some specific details, whereas the second answer gives a solution as to how to accomplish detecting an autoFill change by the animationStart/End events. 

This approach also appears the basis of another PR to support this exact kind of behavior https://github.com/JedWatson/react-select/pull/2395 

The issue with it, is that it only seems to be supported by Webkit browsers which may be good for most, but likely wouldn't be a good fit for integrating into the react-select library code without a more complete solution.

I am open to conversation but the present api should be enough to roll your own version. I truly believe that the goal of react-select is not to be everything for everyone but instead to provide the tools be anything for anyone, which this seems to fulfill.

PR #4109, opened 2020-07-04
Replace componentWillReceiveProps with componentDidUpdate in docs and async component

manvydasu:
partial fix for 4094 (Remove UNSAFE methods from docs).

Other UNSAFE methods are removed with  #4313


bladey:
Thanks for this PR @manvydasu!

TrejoCode:
<3

jleider:
Any timeline on getting this merged and releasing a new version? With strict mode on this reports as a console error which winds up polluting the console with library based messaging making it hard to notice actual errors in your app.

xuniaoji:
Do we have any plan to release a new version to fix this?

pervezalam83:
Please update when this PR will be merged. It seems we can not get rid of the "Warning: Using UNSAFE_componentWillReceiveProps in strict mode is not recommended and may indicate bugs in your code" until this is merged

brenobaptista:
This PR would be very useful, please merge it if possible 😄 thanks, guys!

Methuselah96:

docs/App/Header.js
```diff
@@ -117,10 +117,11 @@ class Header extends Component<HeaderProps, HeaderState> {
...
     this.getStarCount();
   }
-  UNSAFE_componentWillReceiveProps({ location }: HeaderProps) {
+  componentDidUpdate(prevProps: HeaderProps) {
     const valid = ['/', '/home'];
-    const shouldCollapse = !valid.includes(this.props.location.pathname);
-    if (location.pathname !== this.props.location.pathname && shouldCollapse) {
+    const shouldCollapse = !valid.includes(prevProps.location.pathname);
+
+    if (prevProps.location.pathname !== this.props.location.pathname && shouldCollapse) {
```
Methuselah96, line 124: Doing this on update means that the first time you click away from the home page, the transition won't work (which is actually true of the status quo as well).

Instead I would suggest that we call `this.setState({ contentHeight: this.content.scrollHeight });` immediately in `componentDidMount` and get rid of `toggleCollapse` altogether.

packages/react-select/src/Async.js
```diff
@@ -86,15 +86,17 @@ export const makeAsyncSelect = <C: {}>(
         });
       }
     }
-    UNSAFE_componentWillReceiveProps(nextProps: C & AsyncProps) {
+    componentDidUpdate(prevProps: C & AsyncProps) {
```
Methuselah96, line 89: This method is better replaced with `getDerivedStateFromProps`. The work to do that has already been done https://github.com/JedWatson/react-select/pull/4313, so we're likely to merge that one first.

Methuselah96 (changes requested):
Thanks for this PR! As noted below, the work for `Async` will probably become redundant soon and the `Header` modification needs to be changed. Really appreciate your help!

packages/react-select/src/Async.js
manvydasu: I agree. I removed my changes regarding `Async` component in this PR.

docs/App/Header.js
manvydasu: You're right. I removed `toggleCollapse` alltogether and removed `contentHeight` from the `state` itself. My suggested approach will get rid of one additional re-render after `componentDidMount` and the functionality remains the same.

ebonow:
It seems a bit sluggish when transitioning between pages. I ran a profiler and it seems that it is noticable. It could also likely be due to these change being applied against v3 as opposed to v4 (perhaps giving some credit to the performance gains from the lifecycle method replacements from @Methuselah96) , but I'll try these changes against v4 and see if the differences flatten out.
Screen Shot 2021-02-12 at 9 53 02 AM

manvydasu:
@ebonow 
Not sure how these changes could increase rendering time of the component/page. 
I removed additional re-render and didn't introduce any setStates or anything.
I can rebase this PR on 4.* if that's needed.

ebonow:
I just ran it against v4 and it seems negligible. Could just be an environmental difference between running local dev server against production server.

Once of the changes I noticed that would show up if you rebase is that jsx is now instead imported from 
```
import { jsx } from '@emotion/react';
```

Would you be opposed to helping us rid all of the lifecycle method warnings from the current docs? 

Looks like a few of these components could use a version bump specifically in react-router.

Screen Shot 2021-02-12 at 11 02 14 AM

ebonow:
TLDR: Don't worry about the other dependencies...

Putting this here just to document findings...

Did some more research. Other dependencies causing warnings:
```
    "react-helmet": "^6.1.0",
    "react-router-dom": "^5.2.0",
```

Outside of that are the atlas dependencies, but these also introduce a peer dependency on styled-components:
```
    "@atlaskit/button": "^15.1.4",
    "@atlaskit/icon": "^21.2.4",
    "@atlaskit/modal-dialog": "^11.2.6",
    "@atlaskit/spinner": "^15.0.6",
    "@atlaskit/tooltip": "^17.1.2",
    "styled-components": "^5.2.1"
 ```
 
Started getting react key errors in the console, and we're replacing the documentation in the near future, so not worth putting more time into it.

docs/App/Header.js
ebonow: @Methuselah96 I tested this locally and doesn't seem to impact anything negatively. Any other concerns or can this conversation be resolved?

ebonow (approved):

JedWatson (approved):

JedWatson:
Thanks for the PR @manvydasu and the help getting this done everyone, merging now 🙂

PR #4112, opened 2020-07-06
Update code example for CX and Custom Components documentation

bladey:
@ebonow and @Rall3n to review.

### DIFF
--- a/docs/examples/StyleCompositionExample.js
+++ b/docs/examples/StyleCompositionExample.js
@@ -1,4 +1,4 @@
-import React from 'react';
+/** @jsx jsx */
 import { jsx } from '@emotion/core';
 import Select from 'react-select';
 import { colourOptions } from '../data';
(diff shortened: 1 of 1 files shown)


docs/examples/StyleCompositionExample.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { css } from 'emotion';
+import { jsx } from '@emotion/core';
 import Select from 'react-select';
```
Rall3n, line 3: The `React` import **has** to be removed and replaced by following comment:
```jsx
/** @jsx jsx */
import { jsx } from '@emotion/core';

/* ... */
```

The `css` prop won't exist/work if the `@jsx` comment doesn't exist.

bladey: Thanks @Rall3n.

PR #3150, opened 2018-10-29
Docs: custom 'filterOption' function example

cutterbl:
Updating the 'Advanced' docs, making a clear example of overriding the `filterOption` with one's own custom filtering method.

PR #4113, opened 2020-07-06
Docs: custom 'filterOption' function example

bladey:
Based off PR #3150 by @cutterbl

Updated code, example and tests.

> Updating the 'Advanced' docs, making a clear example of overriding the filterOption with one's own custom filtering method.


ebonow:
Greetings again @cutterbl and thank you for the submission. Also apologies for the delay in a response to this PR.

The example file name is being changed in a reverse word order from the prop name which might make finding the example less intuitive. While showing regex as a possible filter condition might be helpful for some, I would expect those familiar enough with regex to already consider it when doing string matching. As mentioned in the previous PR, given the number of examples and how they are presented in the documentation, we'd like to focus on PR's that are filling in less trivial implementation gaps in regards to the react-select api.

Issue #3632, opened 2019-06-17
[Multi select] onChange returns null when no values instead of []


nicolasletoublon:
All is in the title. We implemented a version of the Multi select. and it seems that when you clear the last value (or the only one) of the field, the onChange returns you null. I would clearly expect an empty array at this stage :) 

Here is your sandbox updated with the onChange callback :) 
https://5jtqt.codesandbox.io/

Thanks for looking at it!

ragesoss:
This appears to be the intended behavor with v3: https://github.com/JedWatson/react-select/pull/3416

I just ran into a bug in my app because of this change as well. Curious as to how others handle it.

speir-wang:
And interestingly, if user uses the clear all button to remove all the values in Multi select, the value would be set to an empty array. 

I was wondering why removing the last value results in `null` but clicking clear all button results in `[]`? Shouldn't they be treated in the same manner?

someden:
Simple example:
```js
import React from 'react';
import Select from 'react-select';

export default () => (
  <Select
    isMulti
    options={[{ value: 'ocean', label: 'Ocean' }, { value: 'blue', label: 'Blue' }]}
    onChange={console.log}
  />
);
```
![ScreenRecording2019-06-18at1](https://user-images.githubusercontent.com/7009699/59666389-645a5980-91bd-11e9-8ef3-5eb05d4dc146.gif)

someden:
Temporary workaround for this bug:
```js
import React, { Component } from 'react';
import ReactSelect from 'react-select';

class Select extends Component {
  defaultProps = {
    onChange: () => {},
  };

  handleChange = selected =>
    this.props.onChange(!selected && this.props.isMulti ? [] : selected);

  render() {
    return <ReactSelect {...this.props} onChange={this.handleChange} />;
  }
};

export default Select;
```

stearruda:
I was having some problems with that as well and to fix it I've redefined the value on the onChange to an empty array.

```
const MultiSelect = () => {
  const handleOnChange = value => {
    if (value === null) {
      value = [];
    }
  };

  return (
    <Select
      isSearchable
      isMulti
      onChange={handleOnChange}
    />
  );
}

export default MultiSelect;
```

TravnikovRN:
I am still having the same Issue

luksari:
I was having the very same problem, needed to use `onChange` from `redux-form` in my component so, this is workaround for `redux-form` and `typescript`

[...]

PR #4117, opened 2020-07-09
Don't return null if the last item is manually removed

Mecit:
Fixes the `Invalid attempt to spread non-iterable instance` which occurs when items are removed from the input one by one.


bladey:
Thanks @Mecit!

ebonow (changes requested):
Great work. Perhaps you can add one small fix as well while this is open.

Can you add the following after line 34: (or something equivalent)

``` JS
if (value.find(val => val && val.value === inputValue)) return;
```

There is another bug present in this example if you try to add a duplicate word.

Methuselah96:
Thanks for the PR! I believe this is better fixed by actually fixing the code instead of the docs (see https://github.com/JedWatson/react-select/issues/3632), but it makes sense to merge this until that is fixed.

ebonow (approved):

Methuselah96:
Seems like the build is failing. Perhaps ESLint doesn't like the `??`?

Mecit:
@Methuselah96 Yes, it seems [ESLint 7.2.0](https://eslint.org/blog/2020/06/eslint-v7.2.0-released#highlights) is needed for the nullish coalescing feature. If upgrading the package would be too bothersome, we can consider using || instead. `{ value: value || [] }` would yield the same result.

Methuselah96:
@Mecit It's probably easier to go with `||` for now thanks! I'll work on getting ESLint upgraded in a separate PR.

Methuselah96:
@Mecit This should actually be no longer necessary with the merge and release of https://github.com/JedWatson/react-select/pull/4339. Apologies for the back-and-forth on this and it no longer being necessary in the end. We really appreciate the contribution!

Mecit:
@Methuselah96 no problem. I'm aware my commit was a quick fix to the problem and I'm glad the underlying issue was resolved in a more convenient way.

PR #4132, opened 2020-07-20
feat(a11y): add support for adding aria attributes to input element

kiranyedidi:
We are using react-select in our project and we are doing accessibility check and found out that it is not compliant to WCAG 2. Right now, there is no way to provide the required aria attributes to the input element. This PR is to add the ability to provide aria attributes which can be added to the input element.

https://www.digitala11y.com/aria-autocomplete-properties/

Above is the reference link to what we are trying to achieve. Below is the short story from the above link.

```
If an element has aria-autocomplete set to list or both, authors MUST ensure both of the following conditions are met:

The element has a value specified for aria-controls that refers to the element that contains the collection of suggested values.
Either the element or a containing element with role combobox has a value for aria-haspopup that matches the role of the element that contains the collection of suggested values.
```

Please let me know if I miss something.

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -75,6 +75,10 @@ type FormatOptionLabelMeta = {
 };
 
 export type Props = {
+  /* Aria attributes to support accessibility  */
+  aria?: {
+    [key: string]: any,
+  },
   /* Aria label (for assistive tech) */
   'aria-label'?: string,
   /* HTML ID of an element that should be used as the label (for assistive tech) */
@@ -376,10 +380,13 @@ export default class Select extends Component<Props, State> {
         const [newProps, newSelectValue] = (newArgs: [Props, OptionsType]);
         const [lastProps, lastSelectValue] = (lastArgs: [Props, OptionsType]);
 
-        return isEqual(newSelectValue, lastSelectValue)
-          && isEqual(newProps.inputValue, lastProps.inputValue)
-          && isEqual(newProps.options, lastProps.options);
-      }).bind(this);
+        return (
+          isEqual(newSelectValue, lastSelectValue) &&
+          isEqual(newProps.inputValue, lastProps.inputValue) &&
+          isEqual(newProps.options, lastProps.options)
+        );
+      }
+    ).bind(this);
     const menuOptions = props.menuIsOpen
       ? this.buildMenuOptions(props, selectValue)
       : { render: [], focusable: [] };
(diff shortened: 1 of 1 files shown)


packages/react-select/src/Select.js
```diff
@@ -1426,6 +1436,13 @@ export default class Select extends Component<Props, State> {
       'aria-labelledby': this.props['aria-labelledby'],
     };
 
+    // Helper function to form aria attributes from aria object prop
+    const attributesFromObject = (prefix, items) =>
+      Object.keys(items).reduce((acc, name) => {
+        acc[`${prefix}-${name}`] = items[name];
```
darkowic, line 1442: To be honest, this is not straightforward API. For me, I would like to pass the aria attributes to the component directly
```jsx
<Select aria-describedby="sth" />
```
not via separate property. Also, now there are 2 different ways to provide the aria props. And this is not tested nor documented too.

kiranyedidi: So the reason to go this route (aria prop) was to make it generic for adding aria props, so that even in future, if need to add few more aria props, we do not have to open another pull request. This is how `react-modal` is currently supporting for adding aria props. 

And I just did not want to remove the existing aria props because of backward compatibility.

LKF92:
We've run into the same problem regarding the aria-autocomplete attribute being not allowed on the input (tested with Continuum Explorer Community tool).
Would this PR allow us to remove the aria-attribute completely ?
Thanks!

packages/react-select/src/Select.js
kylemh: I don't think we would want to expose every possible aria attribute.

For example, look at the [list of aria attributes in @types/react](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1558). Not all of these make sense and can lead to even more broken keyboard accessibility for people who have best intentions.

Instead, we should indeed limit it to just the attributes that make sense as the value input for react-select.

ebonow:
Greetings @kiranyedidi and my apologies for such a long wait until receiving a response to this PR.

Myself, the team and the community appreciate any and all efforts to improve react-select.

I wanted to let you know that the collaboration team reviewed the code and decided that it would not be a good fit to be merged. We agree with the thoughts above that it does not do enough to limit the aria attributes only to what would be relevant and would prefer a strategy that allows users to enter the aria props directly.

There is currently another PR #4322 that is more thorough in regards to implementation and I think if we are going to address accessibility, we should consolidate our attention there. The challenges of applying the correct roles and attributes are not arbitrary so we do want to do so in a way that does not conflict with the aria-live implementation which I am currently reviewing.

In the time being, aria-attributes can be added via custom components. This is not to understate the importance of the issue, but simply to offer that the tools are available to provide accessibility until the proper strategy can be implemented without having to break users should we release something that isn't completely thought out.

Again thank you for your contribution and looking forward to any support you can provide us with the impending discussions on accessibility.

PR #4139, opened 2020-07-26
Add TypeScript infrastructure

Methuselah96:
See https://github.com/JedWatson/react-select/issues/4142 for overall plan.

This PR starts adding the TypeScript infrastructure necessary in order to incrementally start transitioning to TypeScript. The idea would be to convert some of the `docs` to TypeScript, add definition files from the `@types/react-select` package, and then convert the source code to TypeScript little by little and removing the definition files as we went.


.eslintrc.js
```diff
@@ -39,4 +39,38 @@ module.exports = {
...
   },
+  settings: {
+    react: {
+      version: 'detect',
+    },
+  },
+  overrides: [
+    {
+      files: ['*.ts', '*.tsx'],
+      extends: ['plugin:@typescript-eslint/base'],
```
kylemh, line 50: if we're making a breaking change for TS anyways, and are adding a linter. I wanted to recommend we had the following config for boolean naming of variables:

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md#enforce-that-boolean-variables-are-prefixed-with-an-allowed-verb

This library in particular seems to have splitting behavior (`isClearable` / `meniIsOpen`)

kylemh (approved):

.eslintrc.js
Methuselah96: This seems like mostly an issue of style. Not saying it would be a bad change, but that that we would need to discuss this as a team (as with any breaking change). I try to stay out of stylistic issues as much as possible (since I think the disagreement over them are often not worth the possible benefit of a better style). My personal opinion is that the pain of everyone having to change their code would be worse than the consistency benefits in this case.

kylemh: Certainly. To your point, just because it's a breaking change... doesn't mean we want every little bit of the API broken.

Methuselah96:
Closing in favor of a single PR.

PR #4140, opened 2020-07-26
Support for required attribute

natanaelsimoes:
Required prop added to base Select and also to Input component.

At Input, a regular HTML Input was created just to handle the required as the AutosizeInput does not handle it (don't know why). The selected value is sent to that hidden input, if any.

Now native HTML validation works!

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -245,6 +245,8 @@ export type Props = {
   value: ValueType,
   /* Sets the form attribute on the input */
   form?: string,
+  /** Checks if the select has some value */
+  required?: boolean,
 };
 
 export const defaultProps = {
@@ -1411,6 +1413,8 @@ export default class Select extends Component<Props, State> {
       isSearchable,
       inputId,
       inputValue,
+      value,
+      required,
       tabIndex,
       form,
     } = this.props;
@@ -1467,7 +1471,8 @@ export default class Select extends Component<Props, State> {
         form={form}
         theme={theme}
         type="text"
-        value={inputValue}
+        value={value ? value.value : inputValue}
+        required={required}
         {...ariaAttributes}
       />
     );
--- a/packages/react-select/src/components/Input.js
+++ b/packages/react-select/src/components/Input.js
@@ -16,7 +16,11 @@ export type InputProps = PropsWithStyles & {
   isDisabled?: boolean,
   className?: string,
   /** The ID of the form that the input belongs to */
-  form?: string,
+  form ?: string,
+  /** Checks if the select has some value */
+  required ?: boolean,
+  /* The value of the select; reflected by the selected option */
+  value?: string,
 };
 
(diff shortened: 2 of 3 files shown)


Rall3n:
Making the `Select` component not searchable (setting `isSearchable` to false) will remove the `Input` component and the underlying (new) `<input>` element handling the `required` attribute.

chocobuckle:
Any update on this folks? Support for the 'required' attribute is quite a dealbreaker!

irfanalef2060:
@chocobuckle  same here dude. For now, I am thinking to write a wrapper component to support the required functionality.

Methuselah96:

willymwai:
Actually there's a way to enable the required attribute

`import Select, {components} from "react-select";`

`const Input = props => {`
               ` if (props.isHidden) {`
                    `return <components.Input {...props} required={true}/>;`
                `}`
                `return (`
                        `<components.Input {...props} required={true}/>`
                `);`
           ` };`

<Select components={{Input}}
/>

dolie:
I need it !

packages/react-select/src/components/Input.js
```diff
@@ -16,7 +16,11 @@ export type InputProps = PropsWithStyles & {
   isDisabled?: boolean,
   className?: string,
   /** The ID of the form that the input belongs to */
-  form?: string,
+  form ?: string,
+  /** Checks if the select has some value */
+  required ?: boolean,
```
kylemh, line 21:
```suggestion
  form?: string,
  /** Checks if the input has some value */
  required?: boolean,
```

packages/react-select/src/Select.js
```diff
@@ -1467,7 +1471,8 @@ export default class Select extends Component<Props, State> {
         form={form}
         theme={theme}
         type="text"
-        value={inputValue}
+        value={value ? value.value : inputValue}
```
ebonow, line 1474: This would be a breaking change as existing behavior is to clear the input onBlur.

ebonow (changes requested):

ebonow:
Greetings @nathsimpson and my apologies for such a long wait until receiving a response to this PR.

Myself, the team and the community appreciate any and all efforts to improve react-select.

That said, I wanted to inform you that this PR will be closed for the following reasons:
1) Introduces breaking behavior in regards to how blurring the input is handled.
2) Does not cover the use-case for when isSearchable is false isMulti is true.

Regarding the first point, I agree that it is less than ideal to create a controlled input simply to add required validation and we  agree that validation is important. I have created issue https://github.com/JedWatson/react-select/issues/4327 to discuss the complexities and possible approaches to support it for all possible combinations that is allowed by the Select.

Should you be able to implement the behavior without breaking changes, a more complete implementation to cover all variations of Select, and add unit tests to ensure there are all covered, we can of course revisit this and re-open the PR.

Again thank you for the contribution.

PR #4148, opened 2020-07-29
Subtracting the size of the border when calculating the max height

martinadamec:
Menu border is done by `box-shadow`, but when you use `border`, the width is not subtract from max height calculation


Methuselah96:
@martinadamec Thanks for the PR! Is there an issue for this or can you can provide a CodeSandbox demonstrating the issue?

Issue #4137, opened 2020-07-24
Creatable Multi-select text input while  creating 2 or more elements with same content and when removing one of the tags, creating phantom tags that can not be removed


srinivasmerugu:
Issue with creatable multi select text input -- Try to create similar options along with other values and delete one of the similar tag that will create a phantom tags which cannot be removed and cannot proceed further.

I have replicated the same in react-select sandbox

Replication steps:

Goto Creatable Multi-select text input and try to create options as below for example

Example of list:
test and other test

After pressing enter, will generate 4 tags:
test - and - other - test

If click to remove 1 duplicated value/tag, it will remove the values from the component state and it keep 1 ghost tag that can not be removed

The same applies to two or more copies,
if there are 3 equal values, it will generated 2 ghost tags
if there are 4 equal values, it will generated 3 ghost tags
And so on...

![image](https://user-images.githubusercontent.com/29884577/88393841-00508e80-cddc-11ea-8746-509a5eb5dc4f.png)

srinivasmerugu:
Creatable Multi-select text input while  creating 2 or more elements with same content and when removing one of the tags, creating phantom tags that can not be removed.

Issue with creatable multi select text input -- Try to create similar options along with other values and delete one of the similar tag that will create a phantom tags which cannot be removed and cannot proceed further.

I have replicated the same in react-select sandbox and attached the video here.

Replication steps:

Goto Creatable Multi-select text input and try to create options as below for example 

Example of list:
test and other test

After pressing enter, will generate 4 tags:
test - and - other - test

If click to remove 1 duplicated value/tag, it will remove the values from the component state and it keep 1 ghost tag that can not be removed (video attached).

The same applies to two or more copies,
if there are 3 equal values, it will generated 2 ghost tags
if there are 4 equal values, it will generated 3 ghost tags
And so on...

creatable.webm

PR #4154, opened 2020-07-31
Fix removing MultiValues that have identical values. Issue #4137

brenshanny:
Fix for issue #4137 

A multi value Creatable with a value array containing any number of identical values will not properly delete all identical values, and will leave MultiValue components that cannot be interacted with.

This fix adds on the index to the key for the MultiValue as a way to identify unique MultiValue components even when they share the same value.

Considerations:
  - Using index on its own as a key is considered an anti-pattern for React, so introducing a uniqueId generator would be a better fix
   - This fix will have the remove-value action remove all identical values, not just the one clicked


JedWatson (approved):

Methuselah96:
Thanks!

NullPainter2:
Hi,

using item's value might not be unique at all for `key`.

Please, make key for an item customizable by user by some callback, like this.getReactKey(opt,i), because I as a user know how to specify unique id for my data.

PR #3550, opened 2019-05-10
Kashkovsky intl a11y

gwyneplaine:
Resolves a few bugs in #3358 
Thanks to @kashkovsky for doing most of the heavy lifting. 
This PR makes it so that the accessibility prop is merged with a default accessibility config, instead of being a hard override. 
```
{
 ...this.props.accessibility 
...defaultAccessibilityProps
}
```

As well as minor fixes to flow types around the newly exposed accessibility prop.

PR #4161, opened 2020-08-13
Kashkovsky intl a11y, attempt 2

radegran:
Fixed conflicts and added test for PR #3550 that has come to a halt.


packages/react-select/src/Select.js
```diff
@@ -73,8 +75,41 @@ type FormatOptionLabelMeta = {
   inputValue: string,
   selectValue: ValueType,
 };
+export type AccessibilityProp = {
```
TheHollidayInn, line 78: What do you think about moving these to a separate file?

packages/react-select/src/Select.js
```diff
@@ -1815,10 +1868,14 @@ export default class Select extends Component<Props, State> {
   renderLiveRegion() {
```
TheHollidayInn, line 1868: It might be good to abstract this to a separate component at this time. What do you think?

TheHollidayInn (commented):
Testing manually seems to work!! Good job :D

packages/react-select/src/Select.js
radegran: I think you are right! I'll try to find a good place for them.

packages/react-select/src/Select.js
radegran: Thanks for your feedback!
I am not sure. Looking at other render helpers, `renderLiveRegion` seems to line up pretty well? On the other hand one could argue for a decoupling exercise of the whole `Select` class, being ~1600 lines tall. But perhaps that work should be logged in a separate work package. What are you thoughts?

TheHollidayInn: I agree! You are following the convent. Also agree that this is 1600 lines. So maybe another PR would be good to refactor some of the render* function into separate components.

horrigan:
Hi guys, the example looks great! Btw Is anyone working on merging this PR?  Do you guys need any help? It seems @TheHollidayInn has done all the job.

ebonow (changes requested):
@TheHollidayInn Would you mind taking a look at the comments regarding the propName?  `accessibility` as a prop name simply would not be descriptive enough. Recommendation would be `ariaLiveConfig` or something of a similar nature.

ebonow:
### Overview - This proposed PR adds the ability for users to configure the ariaLive messaging configuration to allow custom messages.

#### Problem - This addresses two gaps:
1) The hardcoded aria live message templates are written in English and therefore do not support localization.
2) The templates are not configurable and therefore custom user messages are not possible especially when additional data can be presented in a custom Option component such as when formatOptionLabel returns a JSX component instead of a string

#### Solution - The internal `instructionsAriaMessage` class should be exposed so that a developer can override any of the default aria-live event messages.

ebonow:
@kylemh regarding your comments from [here](https://github.com/JedWatson/react-select/issues/4293#issuecomment-745102241), perhaps you were mistaking this with another PR because of the `accessibility` prop which would likely be better named as something like `ariaLiveConfig`.

kylemh:
💯  thank you for helping me read correctly

The name change would help for sure.

ebonow:
Note to future self, pull this PR with documentation PR and rename prop name

ebonow:
Please note that there has been no response to requests for updates. I have began creating a new PR inspired by this and hope to address several other gaps in the aria live messaging. I will close this once I have a submission ready.

ebonow:
Greetings @radegran ,

Thank you for your efforts resurrecting this PR. Your efforts are much appreciated. There were some issues with merging and code review suggestions that were not addressed so I have taken these efforts as a starting point to create #4390 . I will be closing this issue so we can continue on with attempt 3.

radegran:
> Greetings @radegran ,
> 
> Thank you for your efforts resurrecting this PR. Your efforts are much appreciated. There were some issues with merging and code review suggestions that were not addressed so I have taken these efforts as a starting point to create #4390 . I will be closing this issue so we can continue on with attempt 3.

@ebonow, I have not been keeping track of this PR for some time. Glad to see you moving forward with it, best of luck!

Issue #3189, opened 2018-11-08
Add option to keep incomplete input onBlur


agonsalves:
Currently, if I start typing into the field and have not yet selected anything, then I leave the field, my input gets cleared out.

It would be nice to have an option to keep the input there.

paigeflourin:
hi @agonsalves ,
not sure if this is helpful but i was trying to do this as well and i managed to keep the input on mouse blur with this:

```
 public handleBlur = (event) => {
      const { inputValue } = this.state;
      const { value } = this.props;

      if (!_.isEmpty(inputValue)) {
        this.setState({
          inputValue: ''
        });
        this.props.onChange([...value,createOption(inputValue)]);
        event.preventDefault();
      }

    }
```
and i call it:
```
 <CreatableSelect
            components={components}
            inputValue={inputValue}
            isClearable
            isMulti
            menuIsOpen={false}
            onChange={this.handleChange}
            onInputChange={this.handleInputChange}
            onBlur={this.handleBlur}
            onKeyDown={this.handleKeyDown}
            value={value}
            className={styles.questionInput}
            styles={questionStyles}
          />

```

agonsalves:
@paigeflourin Thanks. I am not sure what createOption() does, so I am not able to replicate what you've got. Any insight?

redheadedstep:
This option used to be in react-select v1 `onBlurResetsInput`, but is missing in v2.  The line in question is in `Select.js on line 1074`:

```
onInputBlur = (event: SyntheticFocusEvent<HTMLInputElement>) => {
    if(this.menuListRef && this.menuListRef.contains(document.activeElement)) {
      this.inputRef.focus();
      return;
    }
    if (this.props.onBlur) {
      this.props.onBlur(event);
    }
    this.onInputChange('', { action: 'input-blur' }); // <-- have an option to skip this line
    this.onMenuClose();
    this.setState({
      focusedValue: null,
      isFocused: false,
    });
  };
```

In react-select v1, `Select.js lines 416-418`, this is how it used to work:

```
if (this.props.onBlurResetsInput) {
    onBlurredState.inputValue = this.handleInputValueChange('');
}
```

I'm working on a PR to include this option.

Rall3n:
Another option would be to use the `onInputChange` prop and manage the value of the search input via state. You just have to prevent state updates for the actions `input-blur` and `menu-close`.

```jsx
class KeepSearchOnBlurSelect extends Component {
  state = {
    inputValue: ""
[...]

PR #4207, opened 2020-09-10
added clearInputOnBlur property to clear input value onBlur

marianorodriguez:
this PR is an up-to-date version of #3242, I just renamed the property for more clarity

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -201,6 +201,8 @@ export type Props = {
   noOptionsMessage: ({ inputValue: string }) => Node | null,
   /* Handle blur events on the control */
   onBlur?: FocusEventHandler,
+  /* Whether to clear partial input when the control is blurred */
+  clearInputOnBlur: boolean,
   /* Handle change events on the select */
   onChange: (ValueType, ActionMeta) => void,
   /* Handle focus events on the control */
@@ -278,6 +280,7 @@ export const defaultProps = {
   openMenuOnFocus: false,
   openMenuOnClick: true,
   options: [],
+  clearInputOnBlur: true,
   pageSize: 5,
   placeholder: 'Select...',
   screenReaderStatus: ({ count }: { count: number }) =>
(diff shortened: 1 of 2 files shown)


Rall3n:
> [...] but there is still no solution for this specific problem.

If you had read the comments on #3242 you would have seen a mention of a solution in https://github.com/JedWatson/react-select/issues/3189#issuecomment-442792535 (by yours truly).

But for the lazy:
```jsx
const KeepInputOnBlurSelect = ({...props}) => {
    const [inputValue, setInputValue] = React.useState('');

    const handleInputChange = React.useCallback((value, meta) => {
        if ([ 'input-blur', 'menu-close' ].indexOf(meta.action) === -1) {
            setInputValue(value);
            return value;
        }
      
        return inputValue;
    }, [inputValue]);

    return <Select {...props} inputValue={inputValue} onInputChange={handleInputChange} />;
}
```

kylemh:
I would not merge this. The solution Rall3n posted isn't a workaround - it's a solution. The idea is that we use the existing callbacks to adjust the value at the form level (as opposed to on a per input basis).

ebonow:
I agree with all points with one caveat.

This seems absolutely essential to consider adding `required` as a prop which appears to be the top request as clearly an uncontrolled input in react-select will always fail the required validation due to the input being wiped out on blur.

I agree that it can be straightforward enough to set up, but to be honest, it is a lot of boilerplate code to override what is already a somewhat controversial behavior, that I believe most people would prefer to easily opt out without the need of having to create controlled inputs simply to make an input required.

On its own, I agree, this is the encouraged solution. But looking at it as a necessary prerequisite for every user supporting validation, I think that it deserves consideration.

Then again, I could be missing an easier implementation for text validation, so open to feedback.

mercteil:
@Rall3n workaround/fix works flawlessly... and I am kind of ok with adding some local state to control/observe input actions BUT: if you have a searchable input ala [advanced/experimental usage in docs](https://react-select.com/advanced#experimental) mobile devices such as iOS automatically open the keyboard which in return cannot be hidden without bluring the input and thus clearing it (which is the default behaviour). The user experience becomes horrible if you want to search for a term and afterwards hide the keyboard and swipe too your desired option. I understand the argument of bloating the API, but from the end user's point of view having a prop for the mentioned problem is probably a necessity.

ebonow:
This was recently mentioned by Jed as something for consideration. 

To add context, the reason why the default behavior was set to clear the input was because the Menu will otherwise filter out all the options except for the currently selected (or defaulted) value. For instance, when viewing the first example on the documentation page, opening the menu would only show the option "Ocean" which is confusing for a user to open a dropdown and only see one option.

I'm not sure there is a right approach for this especially for Async users or users with an extremely large dataset.

Issue #4206, opened 2020-09-10
makeCreatableSelect(...) not working


DragonRaider5:
Hey,

I'm trying to create a creatable select using `makeCreatableSelect(...)`. The default export `<CreatableSelect />` component works as expected. However when I switch to using `makeCreatableSelect(...)`, I don't get the option to create a new entry anymore.

Replicated in this CodeSandbox: https://codesandbox.io/s/modest-chandrasekhar-u8yki?file=/src/App.js

Or is it me not properly using `makeCreatableSelect(...)`?

Best regards

ebonow:
Greetings @DragonRaider5 , 

`makeCreateableSelect` is intended to be used only as an internal method to access the entry point (or so it would seem from the TODO comment).

The reason it does not work in the manner you are using it is because this method needs to be wrapped in a `manageState` wrapper which more specifically is the StateManager as you can see [here](https://github.com/JedWatson/react-select/blob/5bee3f40e74c165117a6765588ae906f07ad76f9/packages/react-select/src/Creatable.js#L189).

As for why it is designed as such, I would suspect that it makes for better testing separation. 

That's all to say that importing `CreateableSelect` would be a better and more stable approach than trying to recreate the default export than to be using `makeCreatableSelect`.

DragonRaider5:
Hey @ebonow ,

I think this is very confusing from an API standpoint, as `makeAsyncSelect` f.e. can also be used without any issues. In addition `makeCreatableSelect` is mentioned in the docs, so one would expect it to work.

I am aware that it apparently doesn't work because the StateManager is missing. This is because the `makeCreatableSelect` HOC depends on the `inputValue` being a controlled property.
I provided a PR to fix this issue and clean up the code a little.

dakur:
I've come into this as well. My use-case is to make creatable from custom react-select component (basically just extending Select and rewriting one method). That's what `makeCreatableSelect()` is useful for but now I am unable to do it.

DragonRaider5:
Same for me, I have a custom styled `Select` and would like to use it with `makeCreatableSelect`.

andresmatasuarez:
any updates here? My use case is the same as the above ones: have a custom <Select> used in lots of places in my codebase and want to reuse it to support creatable options.

ebonow:
Good news, bad news.

Bad news, `makeCreatable` is being removed in version 5.

[...]

PR #4210, opened 2020-09-11
Fix makeCreatableSelect(...) for uncontrolled inputValue

marcesengel:
Hey,

currently it is not possible to use `makeCreatableSelect(...)` without controlling the `inputValue` property, as filed in #4206.
This is also why the default export was wrapped in `manageState(...)`, which I was able to remove after fixing the mentioned issue.

I also added a Jest test case for the described behavior and refactored `makeCreatableSelect(...)` to not use `UNSAFE_componentWillReceiveProps(...)` anymore and improved its memoization of the generated `options`.

@JedWatson this is a pure bugfix, I'd be very happy about a timely merge. In case anything is missing/faulty, you have questions or something else comes up, just ping me.

Thanks in advance.

### DIFF
--- a/.changeset/seven-mayflies-brush.md
+++ b/.changeset/seven-mayflies-brush.md
@@ -1,0 +1,5 @@
+---
+'react-select': patch
+---
+
+Fixed `makeCreatableSelect(...)` for uncontrolled `inputValue` prop
--- a/packages/react-select/src/Creatable.js
+++ b/packages/react-select/src/Creatable.js
@@ -93,42 +93,7 @@ export const makeCreatableSelect = <C: {}>(
   class Creatable extends Component<CreatableProps & C, State> {
     static defaultProps = defaultProps;
     select: ElementRef<*>;
-    constructor(props: CreatableProps & C) {
-      super(props);
-      const options = props.options || [];
-      this.state = {
-        newOption: undefined,
-        options: options,
-      };
-    }
-    UNSAFE_componentWillReceiveProps(nextProps: CreatableProps & C) {
-      const {
-        allowCreateWhileLoading,
-        createOptionPosition,
-        formatCreateLabel,
-        getNewOptionData,
-        inputValue,
-        isLoading,
-        isValidNewOption,
-        value,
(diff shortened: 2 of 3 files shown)


Rall3n:
This PR breaks common behavior and functionality of the base `Creatable` component, which were provided by the `manageState` HOC. E.g. clicking on a `Creatable` results in a React crash because of missing props.

The `manageState` HOC is essential to making the `Select` components (in React terms) uncontrollable. Removing that would mean removing basic API, which could be breaking for running projects if released.

A more suitable and easier change would be to make `manageState` exportable instead of replicating that functionality in the `makeCreatableSelect` HOC.

DragonRaider5:
Hey @Rall3n,

Thanks for your feedback. Are you sure about that? None of the integration nor unit tests failed after the change.

This is because the Select component itself already is wrapped in `manageState`, isn't it?

Rall3n:
> Are you sure about that? None of the integration nor unit tests failed after the change.

I'm positive. I tested it myself using the build from this PR. 

> This is because the Select component itself already is wrapped in manageState, isn't it?

It is not. The base `Select` component is only wrapped with the `manageState` HOC in the default export of this library (`index.js`).
```tsx
export default manageState<ElementConfig<typeof SelectBase>>(SelectBase);
```

DragonRaider5:
@Rall3n apologies - in this case yes, the default export should also be wrapped. We'd also need to add a test case for this, as I didn't encounter any failed tests.

DragonRaider5:
Regarding exporting `manageState`: In my opinion this should not be done and kept internal.

The requirement of wrapping `makeCreatableSelect(Select)` (where `Select` is the default export `import Select from 'react-select'`) with `managState` for a **second time** came from `makeCreatableSelect` depending on `inputValue` as a controlled property to trigger `UNSAFE_componentWillReceiveProps`.
I consider this to not be the cleanest approach + it makes use of a lifecycle which was marked as deprecated + the old implementation does unnecessary state updates. That's why I think we should keep going forward with this PR.

PR #4213, opened 2020-09-18
protect against reading removeEventListener from el if null

eythort:
Causes a crash in react app in some rare cases. Probably a race condition. Seems a fix, checking for el, was added for the mount part, but left out for the unmount


JedWatson (approved):

Methuselah96:
Thanks!

nifgraup:
The `if (!el) return;` check would not be needed if `el` was really of type `HTMLElement`.

dfiad77pro:
I think you have to release a minor version with this fix, I have the same problem when I upgrade from react 16--> 17 ( react-select is  included on child npm package),
with a fork with your fix, all work :)

Methuselah96:
@dfiad77pro It would be great if you could provide a reproducible sample of this. Can you make a CodeSandbox showing it not working?

dfiad77pro:
Its currently difficult for me to do that because my code is protected by my company, but i will try to reproduce this on more conventional way (ps : sorry for my english :) )

AdDfNet:
finally i will have not time to reproduce issue on CodeSandbox :( i will make a temp fork on react select for my company and when this commit will be published i will remove my fork (published on private repository).

ps : I can show you the problem with skype/teams private meeting. But i think you will waste time because this fix is good :)

AdDfNet:
I resolve my issue by setting captureMenuScroll to false (instead of true or undefined)

Issue #4232, opened 2020-10-02
Error in running docs website


cseas:
I just made a fresh clone of the react-select repository and tried to run the docs but I keep getting a webpack error.

Steps to reproduce:
```sh
git clone https://github.com/JedWatson/react-select
cd react-select/docs
npm install
npm start
```

Error message produced:
```sh
Child html-webpack-plugin for "index.html":
     1 asset
    Entrypoint undefined = index.html
    [./node_modules/html-webpack-plugin/lib/loader.js!./index.html] 1.75 KiB {0} [built]
    [./node_modules/lodash/lodash.js] 530 KiB {0} [built]
    [./node_modules/webpack/buildin/global.js] (webpack)/buildin/global.js 472 bytes {0} [built]
    [./node_modules/webpack/buildin/module.js] (webpack)/buildin/module.js 497 bytes {0} [built]
ℹ ｢wdm｣: Failed to compile.
```

Operating system: Ubuntu 20.04.1 LTS

Rall3n:
@cseas The docs come without a compiled `react-select` build. You have to run the `build` task in the projects root directory before running the `start` task. Alternatively you can run the `watch` task in parallel.

cseas:
@Rall3n I went to the root folder and ran the following commands:
```
npm install
npm run build
```
After the build finished, I went back to the `docs/` folder and ran `npm start`.

The error changed slightly but still the same thing:
![Selection_089](https://user-images.githubusercontent.com/29686866/94995463-15a41f00-05bc-11eb-8724-237951a224cc.png)

Rall3n:
My mistake. I mistook the docs with the test suite.

But I could reproduce your problem. Seems like the environment does not get correctly set up using `npm`. But it works using `yarn`.

I would suggest trying it again with `yarn`.

cseas:
@Rall3n I installed `yarn` and tried running the docs website with it. It seemed to work fine, although I did get a message about linux incompatibility. That's possibly a separate issue so I'll close this one.

PR #4238, opened 2020-10-04
Add instructions for running docs locally

cseas:
It took me a couple of days to figure out how to run the docs website locally. Hopefully, these changes will make it easier to understand for other people. 

The current CONTRIBUTING file makes no mention that the repository only works with `yarn` and not `npm`. There is also no mention that a local build task needs to be run before starting a local instance of the docs.

Please feel free to suggest any changes if I've gotten something wrong in the instructions but I really think these should be in the CONTRIBUTING file. Many people are used to `npm` and might run into the same errors I did.

Related: #4232

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,3 +23,21 @@ changes, feel free to ping @gwyneplaine
 * Please [follow our established coding conventions](https://github.com/keystonejs/keystone/wiki/Coding-Standards)
 (with regards to formatting, etc)
 * All new features and changes need documentation.
+
+## Running the docs website
+
+You can run the website in a local environment using the following steps:
+
+```sh
+git clone https://github.com/JedWatson/react-select
+cd react-select
+yarn
+yarn build
+cd docs
+yarn
+yarn start
+```
+**Note:**
+
+* We recommend using [yarn](https://classic.yarnpkg.com/en/) for setting up the local development environment for this project. Using `npm` might raise unexpected errors.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -23,3 +23,21 @@ changes, feel free to ping @gwyneplaine
...
+## Running the docs website
+
+You can run the website in a local environment using the following steps:
+
+```sh
+git clone https://github.com/JedWatson/react-select
+cd react-select
+yarn
+yarn build
+cd docs
```
manvydasu, line 36: This feels wrong to me, since the script itself is described as
```"start": "cd docs && yarn start",```
It does the `cd` part

Comment (manvydasu) [inline CONTRIBUTING.md:36]: This feels wrong to me since the script itself is describe as
`"start": "cd docs && yarn start",`

Comment (cseas) [reply CONTRIBUTING.md:36]: That bit actually feels a bit unnatural to me. As someone who visited the repository for the first time, I saw a `package.json` inside the docs folder and assumed the website's an isolated folder, which it is too. The "start" command at the root level's just an abstraction. 

I didn't even know going to the root folder's required for running docs until the discussion in #4232

But anything's fine here I guess. Should I change the commands to the below code?

```sh
cd react-select
yarn
yarn build
yarn start
```

bladey: This looks good @cseas, could you please change the commands and we'll look at getting this merged in.

Methuselah96: I don't think you even need to run `yarn build` since `preconstruct dev` is run as a `postinstall` script. I believe it can just be:

```
cd react-select
yarn install
yarn start
```

ebonow:
PR repo has been deleted so closing this PR

Issue #4255, opened 2020-10-16
Select should not be styled as having focus when disabled


arthurakay:
# Are you asking a question?

I asked a [question on SO](https://stackoverflow.com/questions/64391456/react-select-remove-focus-when-disabled)... though after digging into the source code, I believe this may be the solution.

# Are you reporting a bug or runtime error?

This is a bug in v3.1.0

https://codesandbox.io/s/react-select-v3-sandbox-forked-t8hol

Steps to reproduce in the linked example:

1. Choose a color in the first `Select` box
2. Click into the second `Select` box; notice the options correctly appear and the `Select` receives "focus"
3. Do not select an option in the second `Select`; with the options menu still open, click the "X" to clear the value in the first `Select` box
4. Notice that the second `Select` is correctly disabled (you cannot interact with it) but there are two issues:

  * First, the menu options are still visible (despite being disabled and you cannot interact with them)
  * Second, the disabled `Select` has styling applied to it indicating the element has focus (despite the fact it does not)

# Proposed solution

1. The menu options should automatically close when the `Select` component is disabled
2. The `Select` box should not have styling applied to it indicating focus when disabled

I have been able to create a workaround for the first issue using the following configuration... though it would be great if this were simply the default behavior:

```
    // handles edge-case where menu is open but MODEL is cleared
    menuIsOpen={state.model === null ? false : undefined}
```

The second of these two items should be solved by changing the following line (45?) in [`Control.js`](https://github.com/JedWatson/react-select/blob/b7868d23ccd0b9e623b277f971e89020d426fbdb/packages/react-select/src/components/Control.js#L45)

Change from:
```
boxShadow: isFocused ? `0 0 0 1px ${colors.primary}` : null,
```
to:
```
boxShadow: !isDisabled && isFocused ? `0 0 0 1px ${colors.primary}` : null,
```

ebonow:
@arthurakay I've taken a little bit deeper look into this on my flight yesterday, and I agree it's an edge-case, but feels like the proposed solutions are a band-aid for the fact that a disabled Select is not properly doing the following things:
- blurring focus
- not closing the menu (assuming menuIsOption prop is undefined)
- not disabling menu options (assuming menuIsOpen is true)

I have begun to put together a PR that addresses these concerns...

ebonow:
I checked in a PR for this. 

If `menuIsOpen=true`, then the menu will stay open, but otherwise it will close the menu and blur the Select

ebonow:
[...]

PR #4256, opened 2020-10-16
Fix for #4255 -- disabled component should not be styled as focused

arthurakay:
See issue #4255


ebonow:
Greetings @arthurakay ,

I reviewed the code changes but feel that these changes only fix the cosmetic issues when in fact the input should never be both focusable and disabled.

I appreciate the contribution and bringing this to our attention and am looking at creating a PR that will resolve the underlying issues so that these styles are not necessary.

Issue #3006, opened 2018-09-04
Improve documentation


gwyneplaine:
This is an Issue for us to keep track of ways we need to improve the documentation, feel free to add points in the comments below. 
- Add example for getOptionValue method 
- Add documentation around the expected default properties of options (value: [any], label: string).

wejrox:
The [source links](https://jedwatson.github.io/react-select/) are pointing to the wrong URL for the examples. Presumably they have changed and not been updated.

Rall3n:
@wejrox https://jedwatson.github.io/react-select/ has been replaced with https://v1.react-select.com. There all example links are pointing to the correct source.

ebonow:
- [ ] Add Accessibility section
- [ ] Add examples for debounce with async

PR #4283, opened 2020-11-15
Upgrades Emotion to v11

majgaard:
Hi 👋 

With [Emotion v11](https://emotion.sh/docs/emotion-11) being released and many other packages (including [Chakra-UI](https://github.com/chakra-ui/chakra-ui)) upgrading their dependencies, `react-select` should follow suit.

The predominant reason for this is that having mismatched versions of Emotion can cause lots of issues. See [this issue](https://github.com/chakra-ui/chakra-ui/issues/2414) for one of many examples.


babel.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   plugins: [
-    'emotion',
+    '@emotion',
```
belgattitude, line 3: quick fix: "@emotion/babel-plugin"

.changeset/olive-toys-try.md
```diff
@@ -0,0 +1,6 @@
+---
+'@react-select/docs': patch
+'react-select': patch
```
belgattitude, line 3: Desserves a minor at least. Not everyone has updated to 11

package.json
```diff
@@ -25,9 +25,9 @@
     "@babel/runtime": "^7.4.4",
     "@changesets/cli": "^2.0.3",
     "@changesets/get-github-info": "^0.2.1",
-    "@emotion/cache": "^10.0.9",
-    "@emotion/core": "^10.0.9",
-    "@emotion/css": "^10.0.9",
+    "@emotion/cache": "^11.0.0",
+    "@emotion/react": "^11.0.0",
```
belgattitude, line 29: Would force a minimum for "@emotion/react" to "^11.1.1" (others are still 11.0.0). Emotion 11 is new, I wouldn't care.

If done, must be done in all package.json

package.json
```diff
@@ -37,7 +37,7 @@
     "babel-eslint": "^9.0.0",
     "babel-jest": "^23.6.0",
     "babel-loader": "^8.0.0",
-    "babel-plugin-emotion": "^10.0.9",
+    "@emotion/babel-plugin": "^11.0.0",
```
belgattitude, line 40: Would be nice to move this into depsDependencies. I mean it's not the only one that should move but that would be a start.

(not to do in root package.json, there it's fine)

belgattitude: Forget about this one, mistakenly took the package.json for a package one ;)

giovannigiordano:
@belgattitude @majgaard any update on this?

majgaard:
This is largely good-to-go from my end. There are 2 Flow-typing issues. One can be fixed somewhat easily, but I'm not sure how the maintainers would prefer to fix it. The other issue is because Flow can't find the Stylis module. I'm not particularly familiar with Flow and I have been unable to fix the issue.

sneridagh:
@majgaard @JedWatson This is starting to be painful, any update? Sorry, but I can't do Flow, so I'm not able to jump in.

Methuselah96:
@sneridagh We're planning to include this in a 4.0 release (which we're hoping will be soon). I'm looking into the Flow errors now.

with-heart:
Hi @Methuselah96, Chakra UI maintainer here. Glad to hear this is scheduled for v4 🎉 Let me know if there's anything I can do to assist with this!

yarn.lock
```diff
@@ -12543,6 +12584,11 @@ stylis@^3.5.0:
   resolved "https://registry.yarnpkg.com/stylis/-/stylis-3.5.4.tgz#f665f25f5e299cf3d64654ab949a57c768b73fbe"
   integrity sha512-8/3pSmthWM7lsPBKv7NXkzn2Uc9W7NotcwGNpJaa3k7WMM1XDCA4MgT5k/8BIexd5ydZdboXtU90XH9Ec4Bv/Q==
 
+stylis@^4.0.3:
+  version "4.0.3"
+  resolved "https://registry.yarnpkg.com/stylis/-/stylis-4.0.3.tgz#0d714765f3f694a685550f0c45411ebf90a9bded"
+  integrity sha512-iAxdFyR9cHKp4H5M2dJlDnvcb/3TvPprzlKjvYVbH7Sh+y8hjY/mUu/ssdcvVz6Z4lKI3vsoS0jAkMYmX7ozfA==
+
```
Methuselah96, line 12591: If you remove these lines from the `yarn.lock` and re-run `yarn install`, the flow error regarding `stylis` should be resolved (the [fix](https://github.com/thysultan/stylis.js/pull/241) for the issue causing this was released in 4.0.4).

Methuselah96 (changes requested):

packages/react-select/src/NonceProvider.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import React, { Component, type Node } from 'react';
-import { CacheProvider } from '@emotion/core';
+import { CacheProvider } from '@emotion/react';
 import createCache from '@emotion/cache';
```
Methuselah96, line 4: As for the new required `key` option in `createCache`, I would propose we just add it to the `props` of `NonceProvider` and make the user provide it. I'm not thinking of any good ways to deal with it on our end. We're going to be releasing this PR in a major release anyway.

Methuselah96 (dismissed):

packages/react-select/src/NonceProvider.js
Methuselah96: I suppose we could do something similar to an auto-incrementing `instanceId` that we have for `Select`, but I think it'd be preferable to have the user handle the `key`.

packages/react-select/src/NonceProvider.js
```diff
@@ -1,24 +1,28 @@
...
 import memoizeOne from 'memoize-one';
 
 type NonceProviderProps = {
   nonce: string,
   children: Node,
+  key: string,
```
JedWatson, line 10: this should be something like `cacheKey` so it's not confused with the standard React `key` prop

```suggestion
  cacheKey: string,
```

.changeset/olive-toys-try.md
```diff
@@ -0,0 +1,6 @@
+---
+'@react-select/docs': minor
+'react-select': minor
```
JedWatson, line 3: This would be a major release, because a major update to emotion effectively changes our public API for react-select (even though it should be pretty painless for users to upgrade)

```suggestion
'react-select': major
```

JedWatson (changes requested):
Looks good, two small changes and we can get this merged

majgaard:
We should be all good! 🎉

kpunith8:
@majgaard looks like `@emotion/jest` is not updated to the latest version, it is still pointing to `jest-emotion`, and is used here, https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/__tests__/Select.test.js#L15 I found this while I was forking and updating to emotion-11 using the codemod rule `"@emotion/pkg-renaming": "error"`

Not sure this is a blocker for this PR to go.

majgaard:
@kpunith8  b573dca should take care of that. Had to regenerate snapshots because of the updated serializer.

JedWatson (approved):

jdelStrother:
Can anyone suggest what a good cacheKey for NonceProvider is? I'm not familiar with emotion, so I'm not sure whether it's supposed to be a different cacheKey for each component, or each nonce, or each page ..?

Methuselah96:
I believe it should be a unique `cacheKey` per `NonceProvider` (i.e., a unique `cacheKey` per Emotion cache since each `NonceProvider` has its own cache).

ebonow:
I added a note to #3006 to add examples to the documentation for NonceProvider. I struggled finding adequate information on this trying to search through Emotions's documentation.

jdelStrother:
If it’s per-NonceProvider, wouldn’t it make more sense for NonceProvider to generate its own key rather than making other developers figure this out?

Methuselah96:
@Andarist @mitchellhamilton Can you provide some guidance here? We provide a [`NonceProvider`](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/NonceProvider.js) component that can be used to wrap the `Select` component in order to add nonce support. What should be the scope of the `key` being passed into `createEmotionCache` (i.e., per app, per component, per component instance)?

Andarist:
The scope of the `key` should be, well, per the cache **instance** and ideally, no 2 caches should use the same key. If you share a single cache through different CacheProviders then this should be OK.

jdelStrother:
> The scope of the key should be, well, per the cache instance

Forgive me, I think I must be missing something important in how emotion works or how it implements caching.  Doesn't that mean that emotion ought to be responsible for scoping its own cache keys? What would be the benefit in a client supplying the same cache key to different cache instances?

Andarist:
> What would be the benefit in a client supplying the same cache key to different cache instances

I'm saying that this should not be done - each new cache should have a new key. Caches integrate with the DOM and it's a shared, global resource - hence the need for keys.

Methuselah96:
> I'm saying that this should not be done - each new cache should have a new key.

If this is the case then why can't Emotion generate the key themselves, one per cache instance?

Andarist:
Mainly because of SSR

jdelStrother:
Say if I have two `<Select>` components on the page, each wrapped in their own `<NonceProvider>`. The two NonceProviders share the same nonce. Can both of them use the same cacheKey? If the two select components have different `style`  props, does that change the answer?

Just using a static cacheKey seems like it works fine, but I'm not really sure what I'm breaking by doing so.

ebonow:
> Say if I have two `<Select>` components on the page, each wrapped in their own . The two NonceProviders share the same nonce. Can both of them use the same cacheKey? If the two select components have different `style` props, does that change the answer?
> 
> Just using a static cacheKey seems like it works fine, but I'm not really sure what I'm breaking by doing so.

My suspicion would be that you may run into issues if you try to apply two different styles to the same component if the using the same `cacheKey`. 

For instance if you want to set the border red in one of your Selects, 
``` js
const styles = { control: css => ({ ...css, borderColor: 'red' }) }
```

It could be merging all of the styles together and likely would result in all Selects have a red border. I could be wrong though, and would be interested in seeing the impact or having better familiarity with the impact.

Andarist:
> Say if I have two <Select> components on the page, each wrapped in their own . The two NonceProviders share the same nonce. Can both of them use the same cacheKey?

No - they couldn't use the same `cacheKey` **but** you could share the very same cache instance for both, resulting in the same cacheKey being used without any problems. The difference is that you suggest using 2 caches while I propose using a single cache. You just can't create caches per NonceProvider in this scenario - you can still create them lazily there but you'd have to use memoization to actually reuse the same cache for the second NonceProvider.

> It could be merging all of the styles together and likely would result in all Selects have a red border. I could be wrong though, and would be interested in seeing the impact or having better familiarity with the impact.

I'm not aware of how React Select works but I would suspect that this is not true - styles should be still scoped per component and Emotion caches don't have much to do with that. They are mainly responsible for injecting a given ruleset once to the CSSOM (and have some global-like options).

jdelStrother:
Thanks @Andarist, that clears things up.

For the react-select people - I don't think NonceProvider is ideal for our use case, where we have a multi-root react app (boring old static html pages, with multiple react components mounted at different locations on the page). I ended up re-implementing our own version of NonceProvider to share a single cache with a fixed key. A tiny snippet in case it helps anyone else:

```tsx
import { CacheProvider } from "@emotion/react";
import createCache from "@emotion/cache";
import memoizeOne from "memoize-one";
import cspNonce from "lib/csp_nonce";

const sharedCache = memoizeOne((nonce) =>
  createCache({ nonce, key: "ab-react-select" })
);
function withNonceProvider<Config extends {}>(
  Component: React.ComponentType<Config>
): React.ComponentType<Config> {
  return (props: any) => (
    <CacheProvider value={sharedCache(cspNonce())}>
      <Component {...props} />
    </CacheProvider>
  );
}

```

Issue #4265, opened 2020-10-29
innerProps for menuList not applied.


supersheen:
I am trying to remove the top and bottom white margin between menu and menuList, but seems the innerProp is not applied, here is example. Can you have a look and let me know why? thanks. 
https://codesandbox.io/s/react-select-v3-sandbox-forked-xe8ms?file=/example.js

Rall3n:
@supersheen Unfortunately the example does not show what you are trying to accomplish / proof. Could you perhaps update the example?

supersheen:
Thanks your reponse @Rall3n yes the example doesn't work, I am trying to figure it out. Here `<components.MenuList  {...props} innerProps={Object.assign({}, props.innerProps, {'data-role': 'menuList'})} />` I am hoping to add an attribute `data-role: menuList` in output to get rid of the top and bottom white margin using `padding-bottom: 0; padding-top: 0;` but it's not working. I checked source code, seems `innerProps` is not assigned for `menuList`? `menu` is working fine for me.

supersheen:
![Screenshot 2020-10-29 at 18 54 51](https://user-images.githubusercontent.com/12728272/97619691-6f133880-1a18-11eb-9024-22fa262dff83.png)

Rall3n:
@supersheen You can override the style using the styling framework.

Removing the styles for `paddingTop` and `paddingBottom` from the `menuList` styles should result in what you want to achieve.

```jsx
<Select
  styles={{
    menuList: ({ paddingTop, paddingBottom, ...baseStyles }, props) => baseStyles
  }}
/>
```

supersheen:
thanks @Rall3n, I thought about that. Does it mean `menuList` doesn't support `innerProp`? I also need to add other none style props like `data-testId`.

ebonow:
Greetings @supersheen ,

The code above is simply to address the styling of the component. You would still need to create your own custom menuList component if you need to modify the component in any way such as adding a data attribute.

If you need a look at how this would be done, you can refer to this example on how to get started. If you have an additional questions after taking a look, just let us know.

ebonow:
@supersheen I wanted to revisit your question as I wasn't sure it was completely resolved. @Rall3n seems to have already answered your concern on styling the component. 

[...]

PR #4289, opened 2020-11-20
add innerProps to MenuList

slimklim:


JedWatson:
Context for this PR: see issue #4265

Without this change, you can add custom props to the MenuList component using custom components, but you have to pick between two bad options:

### 1. Add another DOM element

```js
const MenuList = ({ children, ...props }) => <components.MenuList {{...props}}><div data-role="menuList">{children}</div></components.MenuList>
```

### 2. Replicate the entire implementation of the internal MenuList component

Full example here: https://github.com/JedWatson/react-select/issues/4265#issuecomment-738341678

This is bad because it's a lot of duplication of core logic for what should be a trivial use case.

### The solution

With this PR, we can enable the following, which is a meaningful improvement on the two options above:

```js
const MenuList1 = ({ children, ...props }) => <components.MenuList {...props} innerProps={{ 'data-role': 'menuList' }}>{children}</components.MenuList>
```

Benefits:

* No additional DOM wrapping required
* Minimises unnecessary forking of internal code in userland

JedWatson (approved):

PR #4297, opened 2020-11-25
fix null value error

Olaide-EO:
fix null value error. This example breaks when a user clears all the input and tries to type in a new value. Value is null when input is cleared.
Screenshot 2020-11-25 at 7 13 36 PM


ebonow:
Duplicate of existing PR https://github.com/JedWatson/react-select/pull/4117

PR #4313, opened 2020-12-06
Remove usages of UNSAFE React methods

Methuselah96:
Fixes part of https://github.com/JedWatson/react-select/issues/4094.


packages/react-select/src/Select.js
```diff
@@ -608,7 +690,7 @@ export default class Select extends Component<Props, State> {
     });
     this.announceAriaLiveContext({
       event: 'menu',
-      context: { isDisabled: isOptionDisabled(options[nextFocus]) },
+      context: { isDisabled: this.isOptionDisabled(options[nextFocus], selectValue) },
```
Methuselah96, line 693: I assumed that this was a bug. It was calling the builtin `isOptionDisabled`, instead of the `this.isOptionDisabled` which calls `this.props.isOptionDisabled`.

packages/react-select/src/Select.js
```diff
@@ -363,29 +487,49 @@ export default class Select extends Component<Props, State> {
...
-    this.buildMenuOptions = memoizeOne(
-      this.buildMenuOptions,
-      (newArgs: any, lastArgs: any) => {
-        const [newProps, newSelectValue] = (newArgs: [Props, OptionsType]);
-        const [lastProps, lastSelectValue] = (lastArgs: [Props, OptionsType]);
-
-        return isEqual(newSelectValue, lastSelectValue)
-          && isEqual(newProps.inputValue, lastProps.inputValue)
-          && isEqual(newProps.options, lastProps.options);
-      }).bind(this);
```
Methuselah96, line 382: I removed the memoization here since `buildMenuOptions` is refactored in this PR in a way that makes it hard to memoize it the same way it was before. The memoization was originally added in https://github.com/JedWatson/react-select/pull/3868, but it doesn't look the memoization was an essential part of that PR.

packages/react-select/src/Select.js
```diff
@@ -608,7 +723,9 @@ export default class Select extends Component<Props, State> {
     });
     this.announceAriaLiveContext({
       event: 'menu',
-      context: { isDisabled: isOptionDisabled(options[nextFocus]) },
+      context: {
+        isDisabled: this.isOptionDisabled(options[nextFocus], selectValue),
+      },
```
Methuselah96, line 728: I'm assuming this is a bug (introduced [here](https://github.com/JedWatson/react-select/pull/3445/files#diff-7064e1eb0f7889ae735e705d019ebfa160fbb668d7733f73ee6f107b8cf16fadR591)). Instead of calling the builtin `isOptionDisabled`, we should be calling `this.isOptionDisabled` (that everything else calls) which calls `this.props.isOptionDisabled`.

packages/react-select/src/Select.js
```diff
@@ -808,17 +898,12 @@ export default class Select extends Component<Props, State> {
   getElementId = (element: 'group' | 'input' | 'listbox' | 'option') => {
     return `${this.instancePrefix}-${element}`;
   };
-  getActiveDescendentId = () => {
-    const { menuIsOpen } = this.props;
-    const { menuOptions, focusedOption } = this.state;
```
Methuselah96, line 813: I removed this method because it was not being used any more (it's last usage was removed [here](https://github.com/JedWatson/react-select/pull/2581/files#diff-7064e1eb0f7889ae735e705d019ebfa160fbb668d7733f73ee6f107b8cf16fadL1128)) and it wasn't trivial to figure out how it should work with some of the changes I made.

packages/react-select/src/Select.js
```diff
@@ -363,29 +487,49 @@ export default class Select extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
     const { value } = props;
-    this.cacheComponents = memoizeOne(this.cacheComponents, isEqual).bind(this);
```
Methuselah96, line 366: I removed the memoization here because the memoization seems unnecessary (`cacheComponents` just ends up doing an object spread). It is now a `getComponents` method. The memoization was of the components was originally introduced [here](https://github.com/JedWatson/react-select/pull/2814).

packages/react-select/src/Select.js
```diff
@@ -808,17 +942,35 @@ export default class Select extends Component<Props, State> {
...
+    return defaultComponents(this.props);
+  };
 
-    const index = menuOptions.focusable.indexOf(focusedOption);
-    const option = menuOptions.render[index];
+  buildCategorizedOptionsFromPropsAndSelectValue = memoizeOne(
+    buildCategorizedOptions,
+    (newArgs: any, lastArgs: any) => {
+      const [newProps, newSelectValue] = (newArgs: [Props, OptionsType]);
+      const [lastProps, lastSelectValue] = (lastArgs: [Props, OptionsType]);
```
Methuselah96, line 954: This method is memoized per instance of this component, however since `getDerivedStateFromProps` is `static`, there is no way to memoize the call that `getDerivedStateFromProps` makes to `buildCatgeroizedOptions` and have it behave the same with multiple instances as it does with one instance.

packages/react-select/src/Select.js
```diff
@@ -316,22 +482,20 @@ export default class Select extends Component<Props, State> {
...
 
   // Misc. Instance Properties
   // ------------------------------
 
   blockOptionHover: boolean = false;
   isComposing: boolean = false;
-  clearFocusValueOnUpdate: boolean = false;
   commonProps: any; // TODO
-  components: SelectComponents;
-  hasGroups: boolean = false;
```
Methuselah96, line 331: I removed this field because it was not being used any more (it's last usage was removed [here](https://github.com/JedWatson/react-select/pull/2295/files#diff-7064e1eb0f7889ae735e705d019ebfa160fbb668d7733f73ee6f107b8cf16fadL785)) and it wasn't trivial to figure out how to preserve its functionality with some of the changes I made.

packages/react-select/src/Select.js
```diff
@@ -826,6 +818,10 @@ export default class Select extends Component<Props, State> {
     return option && option.key;
   };
 
+  getComponents = () => {
+    return defaultComponents(this.props);
+  };
```
Methuselah96, line 948: Add back memoization.

packages/react-select/src/Select.js
Methuselah96: I actually ended up adding back the memoization in a latter commit, but still moved it be a getter instead of a class instance field.

corysimmons:
Amazing PR. Love the comments accompanying it.

If this gets merged it'll solve many issues.

ebonow (approved):

corysimmons:
:shipit:

JedWatson (approved):

Issue #4293, opened 2020-11-22
React-select v3.1.1 and beyond...


ebonow:
Greetings everyone,

It's been quite a while since we've had an update so I wanted to talk a bit about how to proceed without spamming subscribers on the previous pinned post.

In full disclosure, I have reached out directly to @JedWatson  about a month ago to gauge his interest in having a conversation about the future of the project and identifying ways that I could become a more effective contributor. He replied positively and am still awaiting a date/time we can discuss over Zoom. He likely has hands full with his family and other projects and I am optimistic that we will have a productive discussion when the timing is right for him.

That said, I think what keeps this project alive is the support of the community. Too much has gone into this project to let it die, let alone all the number of existing projects and application that rely on this package. I am also aware that there are still some ["open source gardeners"](https://github.com/JedWatson/react-select/issues/3910#issuecomment-635499277) who are actively participating to help quell the creeping tide of "issues" with generally one of two outcomes: 1) Help me fix my code or 2) React-select code/documentation needs update

Number 2 is especially frustrating as the only response available is "Yes it's broken, no ETA on a fix, and even if you fix it, no ETA on a PR". @Methuselah96 has started a fork for this reason, but I think the community is of course hesitant to splinter off onto yet another one of 3.6k forks. I would much rather exhaust all other options.

This leads me to the purpose of this post. A plan forward is better than no plan at all, and I am interested in addressing existing concerns with active community members. The hope is that we can identify a plan forward and to start by having a meaningful conversation about what the future of react-select looks like as a roadmap:
- What PR's are still relevant?
- What are the bug priorities?
- What are the feature priorities?
- What features are missing?
- What introduces breaking changes?
- Missing/broken documentation?
- Issue management
- What does a regular release cycle look like?

Perhaps @Methuselah96, @Rall3n, @bladey or others would be interested in virtually meeting up to help compile and track what's left to be done or merged.  I'm open to whatever communication platform makes sense.

Also curious if anyone else had any ideas for keeping react-select moving forward. Thanks!

bladey:
Hey @ebonow,

Thanks for reaching out, I agree with all of the above and would like to help put together a plan going forward.

Let's meet up virtually as you suggested - and we'll have that call over Zoom with @JedWatson as well once he's ready.

Rall3n:
Interest is definitely there.

Methuselah96:
[...]

Issue #2456, opened 2018-03-21
A possible plan to improve v2 accessibility and make it ARIA compliant


afercia:
Sorry to open a so broader issue, it's meant more for initial discussion and can be closed at any time, no problem :)

Pretty excited to see al the new features coming in V2, especially the ability to use [custom components](https://deploy-preview-2289--react-select.netlify.com/components#components). Very interesting architectural direction.

On the other hand, accessibility is very important to me and for the team I work with, where we're evaluating the adoption of react-select. FWIW I'm also collaborating with the WordPress accessibility team, where in the past we've done extensive research and testing about similar (non-react) tools, namely Select2, in the hope to find one with a high level of accessibility and to try to help improving it.

After some initial testing and digging into V2, I've noticed there's room for accessibility improvements and I'd be happy to help. I'd tend to think the best option would be sticking to the latest recommended practices and trying to implement an ARIA "Combobox" as defined in the ARIA authoring practices. Please notice the implementation in ARIA 1.1 has changed and it's a bit different from the one used in ARIA 1.0, Some references:

WAI-ARIA Authoring Practices 1.1
https://www.w3.org/TR/wai-aria-practices/

Combobox:
https://www.w3.org/TR/wai-aria-practices/#combobox

> NOTE:
> The options for a combobox to popup a grid, tree, or dialog were introduced in ARIA 1.1. **Changes made in the ARIA 1.1 specification also add support for a code pattern that enables assistive technologies to present the textbox and popup as separately perceivable elements.** both ARIA 1.0 and 1.1 patterns are described in the following sections. While using the ARIA 1.1 pattern is recommended as soon as assistive technology support is sufficient, there are no plans to deprecate the ARIA 1.0 pattern.

See also one of their example, please notice the differences in the markup:

**Examples of ARIA 1.1 Combobox with Listbox Popup:**
https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

I understand this would require some good refactoring, so I'd kindly ask if there's interest and availability in collaborating to improve accessibility before opening specific issues and submitting PRs :) 

Worth noting that even fully implementing the pattern described in the ARIA Authoring Practices wouldn't solve all the accessibility issues. Thinking for example at the selected values and the "tags" as you've also pointed out in the readme [accessibility notes](https://github.com/JedWatson/react-select#accessibility-note).

Thanks very much for reading this, any feedback would be greatly appreciated :)

[...]

PR #4322, opened 2020-12-09
[ARIA] Add more Accessible Tags in Input and MenuList

Progyan1997:
Added combo box role for Input item, listbox role to MenuList and option role to MenuItems. Added relationship between elements by ID.

Signed-off-by: Progyan Bhattacharya <bprogyan@gmail.com>


packages/react-select/src/Select.js
```diff
@@ -1420,9 +1420,12 @@ export default class Select extends Component<Props, State> {
     const { inputIsHidden } = this.state;
 
     const id = inputId || this.getElementId('input');
+    const listBoxId = this.getElementId('listbox');
 
     // aria attributes makes the JSX "noisy", separated for clarity
     const ariaAttributes = {
+      role: 'combobox',
```
kylemh, line 1427: I believe this should be `'combobox'` if it's searchable, but `'listbox'` otherwise; however, I'm not certain what that means for other attributes (re: `'aria-owns'`).

Progyan1997: Aria owns is for searchbox, the list that it owns

Progyan1997:
@kylemh Can you review again?

packages/react-select/src/Select.js
```diff
@@ -1420,9 +1420,12 @@ export default class Select extends Component<Props, State> {
     const { inputIsHidden } = this.state;
 
     const id = inputId || this.getElementId('input');
+    const listBoxId = this.getElementId('listbox');
```
kylemh, line 1423:
```suggestion
    const menuId = isSearchable ? this.getElementById('combobox') : this.getElementId('listbox');
```

packages/react-select/src/Select.js
```diff
@@ -1420,9 +1420,12 @@ export default class Select extends Component<Props, State> {
...
 
     // aria attributes makes the JSX "noisy", separated for clarity
     const ariaAttributes = {
+      role: isSearchable ? 'combobox' : 'listbox',
+      'aria-owns': listBoxId,
```
kylemh, line 1428:
```suggestion
      'aria-owns': menuId,
```

packages/react-select/src/Select.js
```diff
@@ -1727,6 +1730,8 @@ export default class Select extends Component<Props, State> {
       menuShouldScrollIntoView,
     };
 
+    const listBoxId = this.getElementId('listbox');
```
kylemh, line 1733:
```suggestion
    const menuId = isSearchable ? this.getElementById('combobox') : this.getElementId('listbox');
```

packages/react-select/src/Select.js
```diff
@@ -1749,6 +1754,7 @@ export default class Select extends Component<Props, State> {
               <ScrollBlock isEnabled={menuShouldBlockScroll}>
                 <MenuList
                   {...commonProps}
+                  id={listBoxId}
```
kylemh, line 1757:
```suggestion
                  id={menuId}
```

packages/react-select/src/components/Menu.js
```diff
@@ -380,6 +384,7 @@ export const MenuList = (props: MenuListComponentProps) => {
         className
       )}
       ref={innerRef}
+      aria-expanded="true"
```
kylemh, line 387: I'm feeling a bit unsure here, but I think `role="listbox"` should go higher up the DOM (the listbox "container" if you will), but since MenuList is conditionally rendered, `aria-expanded="true"` is correct.

kylemh:
Great work on this btw. This is a complex component for sure!

packages/react-select/src/components/Menu.js
Progyan1997: Listbox should contain options, hence I am putting it with `MenuList`

packages/react-select/src/Select.js
```diff
@@ -1421,9 +1421,13 @@ export default class Select extends Component<Props, State> {
     const { inputIsHidden } = this.state;
 
     const id = inputId || this.getElementId('input');
+    const menuRole = isSearchable ? 'combobox' : 'listbox';
+    const menuId = isSearchable ? this.getElementId('combobox') : this.getElementId('listbox');
```
kylemh, line 1425:
```suggestion
    const menuRole = isSearchable ? 'combobox' : 'listbox';
    const menuId = this.getElementId(menuRole);
```

packages/react-select/src/Select.js
```diff
@@ -1728,6 +1733,8 @@ export default class Select extends Component<Props, State> {
       menuShouldScrollIntoView,
     };
 
+    const menuId = isSearchable ? this.getElementId('combobox') : this.getElementId('listbox');
```
kylemh, line 1736:
```suggestion
    const menuId = this.getElementId(menuRole);
```

packages/react-select/src/components/Menu.js
```diff
@@ -231,7 +231,7 @@ export type MenuProps = MenuAndPlacerCommon & {
 };
 export type MenuPlacerProps = MenuAndPlacerCommon & {
   /** The children to be rendered. */
-  children: ({}) => Node,
+  children: ({ }) => Node,
```
kylemh, line 234:
```suggestion
  children: ({}) => Node,
```

kylemh (commented):
The last thing that would go far is if you added snapshot tests that actually output `role="listbox"` it seems like all the snapshot tests are for searchable selects.

Progyan1997:
@kylemh Added a snapshot in `Select.test`, let me know if you need me to add in other tests as well!

kylemh (approved):
Looks good to me! May wanna @ one of the contributors.

Progyan1997:
@JedWatson @gwyneplaine Either of you can look?

kylemh:
Maybe @ebonow

ad1992:
@kylemh we are waiting for this to get merged. When can we expect the next release to happen?

ebonow:
@ad1992 Sounds like Jed will be putting some time aside early next week to continue to groom through the remaining PRs. He is very anxious to move forward with [a new release/releases](https://github.com/JedWatson/react-select/issues/4293#issuecomment-743781662) so could be as early as then. We got through about 1/4 of the existing PR's the last time we met up, so it all depends on how far we get. 

That said, there was a pre-existing implementation of aria tags throughout the library, but it sounds like there were some [inconsistencies](https://github.com/JedWatson/react-select/issues/2456#issuecomment-414197563) so it's hard to say what was tried and how this is different from that previous approach vs perhaps advances in the aria standards.

ad1992:
@ebonow sure if the release can happen sometime this week or early next week would be great. In case it gets delayed we would like to release a forked version and start using it, would be great if you can guide us on how we can do that.

kylemh:
You should simply fork and keep eyes on this PR!

ad1992:
> You should simply fork and keep eyes on this PR!

@kylemh I meant the steps needed to build and publish this branch so we can start using it in production in case the release gets delayed

kylemh:
There's no guarantee this PR even gets in the next release.

ebonow:
What @kylemh said. 

As mentioned above, there is some uncertainty about why react-select removed their previous implementation of aria tags to instead providing aria live messaging instead especially with all of the various configurations of react-select.

This all said, most all of the changes can already be done using the components api. The only difference appears to be automatically generating an id for the menuList, which should be straight forward enough to pass that in too to the Select.

```jsx
const Option = props => { 
  const innerProps = { ...props.innerProps, role: 'option', 'aria-disabled': props.isDisabled };
  return <components.Option {...props} innerProps={innerProps} />
}

const MenuList = props => {
  const innerProps = { ...props.innerProps, id: props.selectProps.menuId, role: 'listbox', 'aria-expanded': true };
  return <components.MenuList {...props} innerProps={innerProps} />
}

const Input = props => {
    // aria-label and aria-labelledby are already spread by the existing props if provided
  const innerProps = { 
    ...props.innerProps, 
    role: props.selectProps.isSearchable ? 'combobox' : 'listbox';,
    'aria-owns': props.selectProps.menuId,
    'aria-autocomplete': 'list',
  };

  return <components.Input  {...props} innerProps={innerProps} />
}

const ariaSelect = props => <Select menuId="unique_menu_id" components={{ Option, MenuList, Input }} {...props} />
```

@kylemh please correct me if Im wrong, but does this not accomplish everything the PR is doing and without changing any internal methods outside of providing a unique id for the menu?

packages/react-select/src/components/Menu.js
```diff
@@ -370,9 +372,17 @@ export const menuListCSS = ({
   WebkitOverflowScrolling: 'touch',
 });
 export const MenuList = (props: MenuListComponentProps) => {
-  const { children, className, cx, getStyles, isMulti, innerRef, innerProps } = props;
+  const { children, className, cx, getStyles, isMulti, innerRef, innerProps, id } = props;
+
+  // aria attributes makes the JSX "noisy", separated for clarity
+  const ariaAttributes = {
+    role: 'listbox',
```
kylemh, line 379: Oh I missed this.

How does this work in conjunction with the changes made [here](https://github.com/JedWatson/react-select/pull/4322/files#diff-28278ba3fce9a624f09452aa1d9ee109d9e183bce08a032eff72a83a03a9252a) is role defined twice?

kylemh:
Indeed @ebonow 

To highlight the problem and why it's not as simple as always providing these defaults, imagine if a developer were to make an instance of `react-select` where the menu is already open and also where an option moves the app's focus to the next form element.

This would essentially make the `react-select` behave like a radio input and require different considerations for aria attributes. Finding a system to that provides accessible default for majority use cases is important, which is why I reviewed this PR; however, we don't have a system of delivering those defaults without fucking over the consumers using this component in a more unique manner.

ebonow:
@kylemh I'm confused with what "defaults" you are referring to. I simply moved the internal "ariaProps" component logic into these custom components. I was hoping you could review to see if I missed anything as it seems you are the most active/familiar with this PR with the intention was to help @ad1992 (and any others) to achieve the exact same functionality today without unnecessarily forking. 

I appreciate your passion for accessibility and want to reiterate that I agree that accessibility is both important and should be integrated, but that said, there was a decision to move from aria tags to aria-live messaging instead. I think this merits a better understanding of why that decision was reached and I am hoping to accomplish that tonight to better convey any short-comings that were experienced. 

As is, it's possible that this could conflict with the `assertive` live-messaging but I haven't tested this. If it were the case that this conflicts with the messaging, it likely means we would need some way to provide the flexibility to let the dev choose which implementation which could be less straightforward than proposed in the PR.

I hope this seems like a reasonable consideration. In the meanwhile, given the PR changes, is there anything the code I supplied doesn't do that that this PR does from a functionality perspective?

kylemh:
I wasn't contradicting you or debating with you! I was trying to agree with you with an additional perspective for the others to see. I brought up ideal defaults because I believe that was the goal in creating this PR; however, as you've pointed out - it may not be correct.

packages/react-select/src/components/Menu.js
ebonow: The innerProps of a custom component always override the the existing props including aria attributes which is the basis of this solution.

packages/react-select/src/Select.js
```diff
@@ -806,7 +806,7 @@ export default class Select extends Component<Props, State> {
     const custom = this.props.styles[key];
     return custom ? custom(base, props) : base;
   };
-  getElementId = (element: 'group' | 'input' | 'listbox' | 'option') => {
+  getElementId = (element: 'group' | 'input' | 'listbox' | 'combobox' | 'option') => {
```
ebonow, line 809: I'm not a fan of this approach and I understand you are adapting to existing code.

This should likely just be `menu` so we don't have to know the role of the element to know its id.

packages/react-select/src/components/Option.js
```diff
@@ -84,6 +84,7 @@ const Option = (props: OptionProps) => {
   } = props;
   return (
     <div
+      role="option"
```
ebonow, line 87: You should also add the following:

`aria-disabled: isDisabled,`

packages/react-select/src/components/Menu.js
```diff
@@ -383,6 +393,7 @@ export const MenuList = (props: MenuListComponentProps) => {
       )}
       ref={innerRef}
       {...innerProps}
+      {...ariaAttributes}
```
ebonow, line 396: `ariaAttributes` should come before `innerProps`

ebonow (changes requested):

Progyan1997:
@ebonow Updated as you asked.

JedWatson:
As discussed above, and to @kylemh's points, this is a pretty complex and opinionated thing to default that we can't predict will be accurate for all use cases so I've removed it from 3.2 for the time being and we can revisit it in the next major version

ebonow:
@Progyan1997 I also wanted to add good job. I appreciate the speedy replies and want to be responsible in turn to keep you up-to-date.

Accessibility is something we want to prioritize and the existing implementation supports aria live messaging as the primary means of accessibility as it provides a more consistent and customizable experience than what is supported by various screen readers on various OS's especially as react-select continues to grow in complexity.

One of the blockers for this will be testing against changes being made to the aria live messaging. It could likely be counter-productive to include two competing standards and so the ability to apply these default values will likely be contingent on being able to disable aria live messaging.

Please continue to be patient as there is a bit of a backlog on issues and PR's that we are working through. I and the other collaborators are committed to ensuring that contributors are being acknowledged for their work and we are working through the list and will do our best to keep you in communication as I believe this PR will likely be a good starting point to address the ability to better support WAI-ARIA should users decide that the existing aria live messaging is not adequate for their purposes.

Progyan1997:
Thanks for the support @ebonow and thanks @JedWatson for reviewing it.

gauravahir28:
Good work @Progyan1997  by adding the roles and attributes on the dropdown elements!
    
    We wanted to use this PR in hour project, but I have found that only adding roles and attribute creates problems with MAC and IPhone.
    On IPhone, The voiceover isn't allowing me to select the  options as I am focusing on the listbox, the listbox gets collapsed.
    And on MAC, As I navigate within the dropdown using arrow keys, the voiceover isn't reading anything!
    
    therefore, it seems that adding roles and attributes isn't enough to achieve expected accessibility as per the W3C's combobox design pattern.

Thanks.

Progyan1997:
Hey @gauravahir28, can't say about IPhone, but I developed and thoroughly tested this feature on Mac. I have also tested this on Linux system. I believe the problem you're facing is not related to this changes.

Annshuk:
> @ad1992 +1 to what @kylemh said.
> 
> As mentioned above, there is some uncertainty about why react-select removed their previous implementation of aria tags to instead providing aria live messaging instead especially with all of the various configurations of react-select.
> 
> This all said, most all of the changes can already be done using the components api. The only difference appears to be automatically generating an id for the menuList, which should be straight forward enough to pass that in too to the Select.
> 
> ```js
> const Option = props => { 
>   const innerProps = { ...props.innerProps, role: 'option', 'aria-disabled': props.isDisabled };
>   return <components.Option {...props} innerProps={innerProps} />
> }
> 
> const MenuList = props => {
>   const innerProps = { ...props.innerProps, id: props.selectProps.menuId, role: 'listbox', 'aria-expanded': true };
>   return <components.MenuList {...props} innerProps={innerProps} />
> }
> 
> const Input = props => {
>     // aria-label and aria-labelledby are already spread by the existing props if provided
>   const innerProps = { 
>     ...props.innerProps, 
>     role: props.selectProps.isSearchable ? 'combobox' : 'listbox',
>     'aria-owns': props.selectProps.menuId,
>     'aria-autocomplete': 'list',
>   };
> 
>   return <components.Input  {...props} innerProps={innerProps} />
> }
> 
> const ariaSelect = props => <Select menuId="unique_menu_id" components={{ Option, MenuList, Input }} {...props} />
> ```
> 
> @kylemh please correct me if Im wrong, but does this not accomplish everything the PR is doing and without changing any internal methods outside of providing a unique id for the menu?
> 
> Oh which reminds me that this PR is missing aria-disabled for the disabled options which the above code accounts for.

this is what we are also looking for.. react select is very nice library but in terms of accessibility it not much good. also the html elements used not the standard one, it should be wrapped in UL LI so it would be more readable as well when render

pleunv:
> this is what we are also looking for.. react select is very nice library but in terms of accessibility it not much good. also the html elements used not the standard one, it should be wrapped in UL LI so it would be more readable as well when render i.e can refer below

Wrapping things in `<ul>` and `<li>` serves no purpose here. `ul` has an implicit "list" role while `li` has an implicit "listitem" role, yet you are overriding both roles with "listbox" and "option" respectively. `<div>` is just fine, not to say better.

PR #4329, opened 2020-12-12
4255: Check props to ensure Select not focused when isDisabled changed

ebonow:
### Minor Bug Fix for #4255 

**Expected:** When a Select is changed to disabled, it should no longer be focused and the menu should closed (unless specified otherwise by prop `menuIsOpen`).

**Actual:** When a Select is changed to disabled, the Control remains styled as if still focused and the state still indicates that the Select is focused. Additionally, the Menu remains open but in a state that is not Selectable.

Example sandbox of behavior:
https://codesandbox.io/s/react-select-4255-focus-while-disabled-7jr14

**Solution:** Check props on componentDidUpdate for any changes to prop isDisable and resolve discrepancies there.

**Notes:**
- This is indeed an edge case as it would be rare to programmatically set a Select to disabled while maintaining focus on the input. 
- This does not resolve issue of phantom options and menu if menuIsOpen is true. While attempting to modify isOptionDisabled by checking props.isDisabled, this method is not called during all state changes and thus does not successfully modify all options as disabled. I left this portion of code out until it would be effective but could be added as such:

```js
  isOptionDisabled(option: OptionType, selectValue: OptionsType): boolean {
    if (props.isDisabled) return true;
    return typeof this.props.isOptionDisabled === 'function'
      ? this.props.isOptionDisabled(option, selectValue)
      : false;
  }
```
- Credit for reporting this issue and creating an initial PR to resolve this goes to @arthurakay


lijovijayan:
### Here is a temporary solution for this issue
https://github.com/JedWatson/react-select/issues/762#issuecomment-750250172

JedWatson (approved):

PR #4330, opened 2020-12-13
Resolve `findDOMNode` deprecation by merging `ScrollBlock` and `ScrollCaptor` into `ScrollManager`

Rall3n:
This is an alternative PR to #4102  to resolve the `findDOMNode` deprection of #4094.

As #4102 tried to solve the issue through adding the reference as a prop, this PR aims to provide a solution by merging the components `ScrollBlock` and `ScrollCaptor` into a single `ScrollManager` component. This component will have **its own reference** to the needed DOM element (instead of one through props), which resulted in the `children` prop being a function.

In its conception both `ScrollBlock` and `ScrollCaptor` were planned to have their own references with `children` also being a function. But to minimize possible performance issues they resulted in being merged into the `ScrollManager` component.


packages/react-select/src/internal/ScrollManager.js
```diff
@@ -0,0 +1,189 @@
...
+
+  isBottom: boolean = false;
+  isTop: boolean = false;
+  touchStart: number;
+
+  state = {
+    enableLock: false
+  }
+
+  targetRef = React.createRef<HTMLElement>();
```
Methuselah96, line 38: Is there an advantage to using `React.createRef()` here if we're just going to be setting its `current` value manually? I was under the impression that `React.createRef` was just a way to reduce boilerplate, but that since we are already using a callback, it doesn't actually do that.

Methuselah96 (changes requested):
Looks good! I'm assuming we can get rid of the `ScrollBlock` and `ScrollCaptor` components in this PR as well?

packages/react-select/src/internal/ScrollManager.js
```diff
@@ -0,0 +1,189 @@
...
+    this.touchStart = event.changedTouches[0].clientY;
+  };
+  onTouchMove = (event: SyntheticTouchEvent<HTMLElement>) => {
+    const deltaY = this.touchStart - event.changedTouches[0].clientY;
+    this.handleEventDelta(event, deltaY);
+  };
+
+  setTargetRef = (instance: HTMLElement | null) => {
+    this.targetRef.current = instance;
+    this.setState({ enableLock: !!instance });
```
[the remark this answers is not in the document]
Rall3n: > If the intent is to trigger a re-render, can we store the element in state, similar to what ScrollBlock does?

>  I know you've said you're against storing an element in state, I'd be curious to know why.

I learned and made it my ideology, that `state` is **explicitly** for data, while references to DOM elements and possibly React components should be kept in `refs`. That is why I`m against it. Period.

> Another option would be to make ScrollLock not a React component anymore since all it does is attach event listeners to the element and it doesn't actually render anything.

The React mindset is all about components. And I think either would not be an improvement over the other.

packages/react-select/src/internal/ScrollManager.js
Rall3n: Using `createRef` is a preference. And it is more prominently used in React documentation.

The callback is to ensure that the `ScrollLock` will get the current reference through state update. My prefernce would be  not to use it, but that would not ensure that `ScrollLock` will get the current reference value.

packages/react-select/src/internal/ScrollManager.js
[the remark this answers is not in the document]
Methuselah96: That's fine, I would prefer not putting DOM elements in `state` as well (although I think that's sometimes the best solution available, just not in this case).

While the React mindset is all about components, that doesn't mean literally every piece of logic has to live in a component. The design of `react-select` is clearly not against taking logic that makes most sense in a `componentDidMount`, etc. and abstracting it into a function based on `utils.js`.

I would also argue that the mindset of React is also against taking DOM elements received from refs and passing them down to child components through props, because it makes it difficult to make sure that those elements are properly received without forcing what should be an unnecessary re-render on the parent component. The reason why `componentDidMount`, etc. exist are to perform actions on DOM elements that are received via refs after the `render` has taken place, but that breaks down as soon as you start passing elements received from refs as props to other components.

packages/react-select/src/internal/ScrollManager.js
Methuselah96: Sounds good, I agree it is a matter of preference and doesn't have much practical difference, although obviously my preference would be to use a normal callback.

packages/react-select/src/internal/ScrollManager.js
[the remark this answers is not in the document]
Methuselah96: Also, components that return `null` (like `ScrollLock`) don't really fit into the React mindset either.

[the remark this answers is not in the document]
Rall3n: Back to topic: I don't think it's worth the effort to rewrite `ScrollLock` (for this PR). Maybe add it to the backlog?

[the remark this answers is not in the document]
Rall3n: Maybe we could rewrite `ScrollLock` to be a wrapping component that also has its own reference to the DOM element.

Or we could use a library for that: https://github.com/jossmac/react-scrolllock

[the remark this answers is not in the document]
Methuselah96: Sounds fine, I'll make a follow-up PR to avoid the re-render in `ScrollManager` and hopefully we can get them merged in back-to-back for 4.0, or if there is a problem with the follow-up PR, we can just wait to merge it in order to get this released ASAP.

---

Would we have to use the same pattern of having a `(ref) => { }` child function if it were a wrapping component? That seems like a necessary pattern in these use cases, but I personally think it adds unnecessary complexity when you're doing it more than once for the same DOM element. I understand the intention to abstract the functionality it into a reusable component, but I think there's a more simple way to achieve this behavior that follows good coding practices and is still in the React mindset (imo).

Yeah, using a library sounds good too, although I'm not sure I understand how it works if the child is a component and not a function because it seems like it's just getting a ref to the child component and not a ref to the DOM element (based on [this line](https://github.com/jossmac/react-scrolllock/blob/master/src/TouchScrollable.js#L52))?

packages/react-select/src/internal/index.js
```diff
@@ -5,3 +5,4 @@ export { default as DummyInput } from './DummyInput';
 export { default as NodeResolver } from './NodeResolver';
 export { default as ScrollBlock } from './ScrollBlock';
 export { default as ScrollCaptor } from './ScrollCaptor';
```
Methuselah96, line 7: Can we remove these components in this PR since they've been replaced by `ScrollManager`?

Methuselah96 (changes requested):

packages/react-select/src/internal/ScrollManager.js
```diff
@@ -1,30 +1,65 @@
...
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    if (prevProps.captureEnabled !== this.props.captureEnabled) {
+      if (this.props.captureEnabled) {
+        this.startListening(this.targetRef.current);
+      } else {
+        this.stopListening(this.targetRef.current);
+      }
+    }
```
Methuselah96, line 55: Will this work if the DOM element changes?

Methuselah96 (changes requested):

packages/react-select/src/internal/ScrollManager.js
```diff
@@ -1,30 +1,65 @@
...
   }
```
Methuselah96, line 56: Will this work if the DOM element changes?

Methuselah96 (changes requested):

packages/react-select/src/internal/ScrollManager.js
Rall3n: The problem has not occurred so far with `NodeResolver` and I think it should not be considered now. If it ever comes up, then we should come up with a solution.

JedWatson (approved):

Issue #2909, opened 2018-08-08
Unable to apply custom-styles for top menu-portal element via CSS


ixrock:
I'm using `classNamePrefix="Select"` prop to style inner elements from external css-file.
Everything works fine, except `menuPortal` element. It does not receive any specific className and when used in combination with `menuPortalTarget={document.body}` it's impossible to style that top element from portal via `.Select__someElemName {}`.

Currently only possible solution to add styles to `menuPortal` element is to use inline styles:
```
  <Select
    menuPortalTarget={document.body}
    styles={{
      menuPortal: styles => ({ ...styles, zIndex: 100 }) //  >= dialog's z-index
    }}
  />
```

The portal rendered  in `document.body` has closest human-readable className `Select__menu`:
<img width="561" alt="screenshot 2018-08-08 18 28 08" src="https://user-images.githubusercontent.com/6377066/43847596-a516c0b0-9b39-11e8-8948-e5ee46bd430e.png">

It would be good to have ability to add custom class-names to this `menuPortal` element, since it might be rendered outside of `<Select/>` component.

*Please fix it and thanks for great job!* 👍

igoriok1994:
I have the same problem.
I'm handling z-index dynamically in scss files.
It would be great, to set classNamePrefix at first div
![image](https://user-images.githubusercontent.com/37294916/49445915-adc3f800-f7db-11e8-9f47-ed73a9892be5.png)

andypearson:
@ixrock @igoriok1994 I don't know if you ever found a work around for this issue, but I am using `menuPortalTarget={document.body}` so can target the portal `<div>` with the following (probably evil) CSS selector:

```
body > [class^="css-"] {
  border: 3px solid yellow;
}
```

crathgeb:
A workaround is to add a placeholder in the dom for the react-select-portals and then to apply classes you want to it:

![image](https://user-images.githubusercontent.com/4104620/71489372-42d4cd00-27f3-11ea-9cc9-ddf10c1a33f6.png)

Then specify the portal to the portal placeholder: 
![image](https://user-images.githubusercontent.com/4104620/71489401-744d9880-27f3-11ea-9c01-066cc8ae3dee.png)

igoriok1994:
> A workaround is to add a placeholder in the dom for the react-select-portals and then to apply classes you want to it:

Thats nice :) Btw, what syntax highlighting theme and editor you using?  Looks good.

bladey:
Hello -

[...]

PR #4333, opened 2020-12-16
Refactor ScrollCaptor and ScrollLock into hooks

Methuselah96:
This is a follow-up PR to https://github.com/JedWatson/react-select/pull/4330, and should not be merged until that one gets merged (since it includes the changes from https://github.com/JedWatson/react-select/pull/4330). This PR is not necessary in order to remove `findDOMNode`, so it should not block a 4.0 publish.

This PR removes the need force a re-render in `ScrollManager` since all of the logic is run as an effect.


packages/react-select/src/internal/useScrollLock.js
```diff
@@ -0,0 +1,191 @@
...
+          preventInertiaScroll,
+          listenerOptions
+        );
+        touchScrollTarget.addEventListener(
+          'touchmove',
+          allowTouchMove,
+          listenerOptions
+        );
+      }
+    }
```
ebonow, line 126: Maybe this is minor, but `useScrollCapture` adds and removes event listeners in a single line. 
Is there a reason for the differentiation in line breaks between these two files?

My preference is to consolidate them to one line like in `useScrollCapture`  for the sake of consistency as I believe this is done in other places such as `select.js` (ie: L395)

packages/react-select/src/internal/useScrollLock.js
```diff
@@ -0,0 +1,191 @@
...
+            'touchstart',
+            preventInertiaScroll,
+            listenerOptions
+          );
+          touchScrollTarget.removeEventListener(
+            'touchmove',
+            allowTouchMove,
+            listenerOptions
+          );
+        }
```
ebonow, line 171: Same comment as above regarding line breaks...

ebonow (commented):
Looks good. I like the changes though I would like to clear up some of the line spacing to make it more consistent, but that is minor compared to how much better this is to have the functionality better composed into hooks for separation of concerns without having to wrap in 2 HOC components

ebonow:
@Methuselah96 I was responding to a user about some of the menuPortal styling and props. 
https://github.com/JedWatson/react-select/issues/2909#issuecomment-748693376

I wanted to get your feedback based on what the documentation states and the fact that you happen to be working on this portion of code at the moment.

packages/react-select/src/internal/useScrollLock.js
Methuselah96: Yeah, I understand wanting them to be consistent. This is just the way that Prettier decided to format them. I assume that these ones are split-up because they exceed 120 characters, but the other ones are below 120 characters.

Methuselah96:
@ebonow Thanks, I'll take a look when I get a chance.

packages/react-select/src/internal/useScrollLock.js
ebonow: I believe that the line character limit is 80 and not 120

Methuselah96: You're right, I meant 80.

JedWatson (approved):

PR #4342, opened 2020-12-21
Standardize innerProps and className props on customizable components

Methuselah96:
Resolves https://github.com/JedWatson/react-select/issues/2909.

See https://github.com/JedWatson/react-select/pull/4289#issuecomment-742936297 for the explanation of why we spread `innerProps` to all components.

This PR:

- Adds `innerProps` to components that were missing them (`Group`, `MenuPortal`, `ValueContainer`, `IndicatorsContainer`).
- Adds `classname` prop to `MenuPortal`.
- Moves all of the `innerProps` spreads to be the last prop for consistency so that all props are overridable if desired.


ebonow (approved):
Looks good! Thanks for the quick turnaround on this.

JedWatson (approved):

JedWatson:
Moving this to 3.3 because it needs more review, and 3.2 is about to be released

ebonow:
@Methuselah96 I was hoping to get your thoughts on one more thing. What are your thoughts on adding Emotion labels to the Indicator components like many of the other components?

This would be add consistency and I think achieves a bit more visibility especially for automated testing as there is no reliable method to identify the specific indicators to reliably trigger menu down or trigger a clear event.

Here is what I have in mind for indicators.js
```jsx
const indicatorCSS = (label: string) => ({ ...baseCSS, label });

export const dropdownIndicatorCSS = indicatorCSS('dropdownIndicator');
export const loadingIndicatorCSS = indicatorCSS('loadingIndicator');
```

The indicatorSeparator and loadingIndicator can have their labels applied directly to their CSS objects, but I think the `dropdown` and `clear` indicators are probably the most important (as well as `multiValueClear`) to create some kind of more reliable DOM selection attributes for these 3 interactive components which each reliably trigger some action.

Thoughts?

Methuselah96:
@ebonow Yeah, I'm not opposed. Looks like both `dropdownIndicator` and `clearIndicator` share [the `indicatorContainer` label](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/components/indicators.js#L63), so it would be a breaking change if we wanted to distinguish between the two. (You mentioned `loadingIndicator` in your comment, but it looks like `loadingIndicator` [already has a unique label](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/components/indicators.js#L63).)

Either way, it probably makes to keep the changes in a separate PR so the current changes don't get held up.

ebonow:
Totally fair. I think I mistook `indicatorsContainer` and `indicatorContainer`. It makes sense that they have the same label given they share the same css. Also a bit confusing since there is no specified `indicatorContainer`.

JedWatson (approved):
👍

PR #3046, opened 2018-09-17
Pass down extra props to group header as item.data

akre54:
I need the ability to add a click handler to the group header. This allows arbitrary props to get passed down.

PR #4391, opened 2021-01-22
Pass and sanitize commonProps passed to GroupHeader and Input

ebonow:
### Purpose: 
When styling components, it would be beneficial to have `commonProps` accessible as the documentation states that they should be. This also resolves a bug introduced in v3.2

### Proposal:
Currently these two components do not use `commonProps`. It is likely because they pass extra props directly to the components wrapper as DOM attributes. It's worth noting that they do not follow the same `innerProps` paradigm as the other components. 

Since all the cleanedProps are passed directly to the DOM as attributes, it was just a matter of removing the `commonProp` keys from the props as a util function.

### Resolves:
**Issue**: _"input" object of the styles object should receive the same state as the other components #3982_ 

**Issue**: Unreported side effect from 
_Pass down extra props to group header as item_.data #3046

Result of the PR is that GroupHeader now renders with data attribute in the DOM represented as `data=[Object Object]`
Screen Shot 2021-01-21 at 4 12 55 PM

The included code resolves this by destructuring `data` from the sanitized props prior to passing them in as "innerProps"


Methuselah96:
I'm probably missing something, but I don't see how the `commonProps` are getting passed to `GroupHeading`.

ebonow:
Common props weren't being passed and this aims to add them for the sake of consistency. 

The previous implementation was...
```jsx
  const { className, cx, getStyles, theme, selectProps, ...cleanProps } = props;
   return (
    <div
      css={getStyles('groupHeading', { theme, ...cleanProps })}
      className={cx({ 'group-heading': true }, className)}
      {...cleanProps}
    />
  );
```

Both GroupHeading and Input lack `innerProps` and instead all the extra props are passed to the wrapper DOM element as if they were innerProps. My guess is this is why there are no commonProps added.

The intention with this was to create a function that would cleanse the common props in a similar way, so that all of the commonProps are available to the styles api, but then sanitized for the DOM in a consistent way.

Methuselah96:
Okay, got it, should we go ahead and pass `...commonProps` to `GroupHeading` then?

JedWatson (approved):

Issue #4398, opened 2021-01-22
Unable to navigate grouped options via up/down arrow keys


ebonow:
**keyboard navigation (up/down arrow key) not working** - we can open menu but not move focus over the options

the same example (just change react-select version to 4.0.1)
https://codesandbox.io/s/7utj1?module=/example

_Originally posted by @mrudowski in https://github.com/JedWatson/react-select/issues/4395#issuecomment-765692120_

PR #4399, opened 2021-01-22
Fix building focusable options from groups

Methuselah96:
Resolves https://github.com/JedWatson/react-select/issues/4398.


ebonow (approved):
Awesome and thank you for adding a test especially given some discrepancies we are likely to run into handling groups vs options. 💯

PR #4404, opened 2021-01-26
Remove unnecessary dependency on @emotion/css

Methuselah96:
Missed in https://github.com/JedWatson/react-select/pull/4283. We don't need to depend on `@emotion/css`, only `@emotion/react`.


JedWatson (approved):
👍

PR #4406, opened 2021-01-27
Fixed typo in StateManagerProps>Option docs

Reflex-Gravity:
This PR consists of a fix for a typo for **Type** under docs for **StateManagerProps > Option**.
Below is the screenshot of the typo.

react-select-issue-Screenshot_2021-01-27_15-24-18


ebonow (approved):

packages/react-select/src/components/Option.js
```diff
@@ -31,8 +31,8 @@ export type OptionProps = PropsWithStyles &
     innerProps: InnerProps,
     /* Text to be displayed representing the option. */
     label: string,
-    /* Type is used by the menu to determine whether this is an option or a group.
-    In the case of option this is always `option`. */
+    /** Type is used by the menu to determine whether this is an option or a group.
+    In the case of option this is always `option`. **/
```
Methuselah96, line 35: We should make this conform with [official JSDoc](https://jsdoc.app/about-getting-started.html).
```suggestion
    /**
     * Type is used by the menu to determine whether this is an option or a group.
     * In the case of option this is always `option`.
     */
```

Methuselah96 (dismissed):

Issue #4411, opened 2021-01-29
Screenreader status reads back "0 results" when menuIsClosed after option selected


ebonow:
Reported in this discussion: https://github.com/JedWatson/react-select/discussions/4363

Tested version: 4.0.2

<img width="757" alt="Screen Shot 2021-01-29 at 10 05 08 AM" src="https://user-images.githubusercontent.com/1007579/106311585-3bfc6300-621a-11eb-8b4b-de6e082b2502.png">

Proposed solution: Screenreader status should only be read when the menuIsOpen

PR #4390, opened 2021-01-21
Aria Live Configuration

ebonow:
## WIP: recreation of Kashkovsky support intl a11y
Original PR: #3550 
Revised PR: #4161 
Documentation PR: https://github.com/JedWatson/react-select/pull/4203

### Purpose
The purpose of this PR is to allow the aria live messages to be customized to support internationalization or serve to provide more context to assistive technologies https://github.com/JedWatson/react-select/issues/3352

### Proposal
prop: `AriaLiveMessages` is an object supporting several formatting functions for various on the aria-live messages types
_note: prop and methods are tentatively named_
- hoverOption
- hoverValue
- selectOption
- filterResults
- instructions

### Aria-live TODOs
[ ] revise naming to fit better with events
[ ] support for focusing Select that has value https://github.com/JedWatson/react-select/issues/4343
[ ] support for loading https://github.com/JedWatson/react-select/issues/3351
[ ] support for no results message 
[ ] support to show accurate number of results when filtering? (ie: option 3 of 10. 6 results available)
[ ] support for groups (group option index/total and header) https://github.com/JedWatson/react-select/issues/4277
[ ] verify if https://github.com/JedWatson/react-select/issues/3356 is still an issue
[ ] add tests and snapshots
[ ] add assertive and polite aria-live regions on render
[ ] encapsulate aria-live logic in component
[ ] do not read back results count if menu is not open https://github.com/JedWatson/react-select/issues/3006

Issue #4343, opened 2020-12-24
Announce selected or not while option is focused in the list menu of select box


mohit181191:
While navigating the options through the list menu using keyboard up/down arrow key. The screen reader announces the focused option and its position. Example below
If the option is X its position in 2 out 6
`option X focused, 2 of 6. 6 results available. Use Up and Down to choose options, press Enter to select the currently focused option, press Escape to exit the menu, press Tab to select the option and exit the menu.`

It does not announce if the option is already selected and exists in the selected list of options. The users who use screen readers would only be notified when it is actually selected or deselected using keyboard enter key.

Expected Behaviour(something similar to the below example)
`option X focused, 2 of 6. selected. 6 results available. Use Up and Down to choose options, press Enter to select the currently focused option, press Escape to exit the menu, press Tab to select the option and exit the menu.`

It would only be applicable when `hideSelectedOptions` prop is set to false

ebonow:
Greetings @mohit181191,

I have addressed the work required for this as a part of PR #4414 which is being reviewed. This issue will be closed once it is merged.

Issue #4277, opened 2020-11-11
[BUG] "Grouped" options breaks screen readers


carsonpowers:
Using the "grouped" feature breaks screen readers.

Because `group` sub-component elements exist as the first item underneath `menuList`; aria readers will think that the dropdown list contains however many group elements there are, NOT however many option items there are, like it should.

![image](https://user-images.githubusercontent.com/11263423/98876693-6ec96300-2444-11eb-9dbd-fef98934da8d.png)

To reproduce 

1. Navigate to the "Grouped" example on the main react-select homepage: https://react-select.com/home
![image](https://user-images.githubusercontent.com/11263423/98877371-f368b100-2445-11eb-8fd6-4cdbcf44dd73.png)
2. Focus the grouped example element with NVDA (or JAWS) enabled. 
3. Click spacebar to open the menu.
4. Click down the the keyboard arrow keys until the option "Purple" is selected.
5. Screen-reader will call out that the element is in position "0 of 2" despite the fact that there are 14 items in the list and purple is the third member of that list; it should read "3 of 14".

It's saying 0 of 2 because it's counting the number of `group` elements under `menuList` (`Colors` and `Flavors`), not the `Options` nested further down inside `Group`.

Note> Adding `role="presentation"` to `group` may fix this issue.

ebonow:
There has been some discussion also about supporting nested options so I am not sure if there is a recommended approach for having screen readers handle both this current implementation as well as the proposed solution correctly. 

I'm hoping to address many of the a11y issues with this into a near future release and would certainly appreciate any support from those who have deeper experience in this area to contribute whatever they know about appropriate syntax and best practices.

ebonow:
@carsonpowers I have been attempting to read a bit more about how screenreaders handle `optgroup`. It appears that OSX VoiceOver does indeed list all of the options under the total amount, but additionally, optgroup tags are focusable but disabled. I have also read that WAI-ARIA has poor support for `optgroup` so it's also unclear to me what could/should be done for that approach.

From an accessibility perspective, I am curious if @kylemh has any thoughts regarding best practices for supporting groups.

kylemh:
Coverage without using `optgroup` exists [here](https://w3c.github.io/aria-practices/examples/listbox/listbox-grouped.html) FWIW; however, semantic markup is always preferable...

Where did you read about there being poor support for `optgroup`? Obviously, [this](https://a11ysupport.io/tech/html/optgroup_element) doesn't look ideal...

[...]

PR #4414, opened 2021-02-01
Aria Live Configuration

ebonow:
## WIP: recreation of Kashkovsky support intl a11y
Original PR: #3550 
Revised PR: #4161
First pass: #4390 
Documentation PR: https://github.com/JedWatson/react-select/pull/4203

### Purpose
The purpose of this PR is to allow the aria live messages to be customized to support internationalization or serve to provide more context to assistive technologies https://github.com/JedWatson/react-select/issues/3352

### Proposal
prop: `AriaLiveMessages` is an object supporting several formatting functions for various aria-live messages and events
- onChange
- onFocus
- onFilter
- guidance

Additionally: a new prop `aria-live` is introduced to allow users to change the voice of the aria-live component. This should satisfy users unhappy with passive voice (due to listbox guidance interfering with focus events) and users who want to implement a solution without any aria-live.

This PR also introduces a new `LiveRegion` component thought it is not yet shared publicly. As relevant state changes are made in Select.js, they are passed as props to LiveRegion and then use `useEffect` hooks to determine what changes should be made to the rendered aria live messaging, This also has the benefit of removing 100+ lines of code from Select.js

While several other aria-live issues exist, this one is comprehensive enough to test first and then isolate other issues in a follow up. I will keep the to-dos below to keep visibility.

#### Allow aria-live and role to be configurable
- Aria roles appear to have a blocking affect on the aria-live implementation once it begins to speak on the directions for the listbox. Keyboard interaction is blocked and in some cases queued up forcing several "Focused option" messages to be read off consecutively.
- Setting role to alert seems to improve functionality in OSX VoiceOver, but unsure of impact on other SR's (especially since aria-live is still set to polite)
- Setting aria-live to assertive creates the most predictable results, but apparently blocks some other expected functionality so it may be worth considering one or the other to be disabled.
- Setting aria-atomic to true was suggested in another thread, but this causes issues with the selection text then being read on every update as all children nodes are read back instead of the just the changing context.

#### Concerns
- The guidance instructions read back on every input and focus option are a bit tedious. Perhaps this is something we only want to read when the context changes.
- It seems both redundant and inaccurate to state the index of the focused option as it applies to all the options as not all options are showing especially when the results as returned from the `screenreaderStatus` prop return the number of focusableOptions
- It seems like a serious missing that the selected options are not read back when a user focuses on the select.

#### Groups
- Group support is completely broken...
- Group Headers are not available to be read with the option. There is no read-only function to identify the string mapping for the group headers. Would recommend using `formatGroupHeader` as a formatting function and creating a new bulitin called `getGroupHeaderLabel`. 
- Need to introduce a way to identify group heading from the option
- Index of results always reads back 0 of # of groups, because groups are the child component of the menuList
- Should headers be focusable? Current select behavior is such that they are focusable but disabled.

### Aria-live TODOs
- [x] revise naming to fit better with events
- [x] render aria-live region on render
- [x] encapsulate aria-live logic in component
- [x] add tests and snapshots
- [x] Screenreader status reads back "0 results" when menuIsClosed after option selected #4411
- [x] Announce selected or not while option is focused in the list menu of select box #4343
- [x] customize aria-live attribute as optional prop #4447 
- [ ] add more to newly created accessibility section in documentation about a11y and message format

and...
- [ ] support for loading https://github.com/JedWatson/react-select/issues/3351
- [ ] support for no results message 
- [ ] support to show accurate number of results when filtering? (ie: option 3 of 10. 6 results available)
- [ ] support for groups (group option index/total and header) https://github.com/JedWatson/react-select/issues/4277
- [ ] verify if https://github.com/JedWatson/react-select/issues/3356 is still an issue
- [ ] Screen reader read's selected value as 'blank' (when there is a selectedValue) #1585


docs/pages/advanced/index.js
```diff
@@ -33,6 +34,20 @@ export default function Advanced() {
       {md`
       # Advanced
 
+      ## Accessibility
```
ebonow, line 37: TODO: More feedback from maintainers about how accessibility and the new api should be addressed.

packages/react-select/src/__tests__/Select.test.js
```diff
@@ -197,6 +197,7 @@ cases(
     'single select > should match accented char': {
       props: {
         ...BASIC_PROPS,
+        inputValue: '',
```
ebonow, line 200: Unit tests were throwing warnings because `inputValue` was being changed on re-render without being originally set.

packages/react-select/src/__tests__/Select.test.js
```diff
@@ -1862,6 +1872,29 @@ test('accessibility > interacting with disabled options shows correct A11yText',
   );
 });
 
+test('accessibility > interacting with multi values options shows correct A11yText', () => {
```
ebonow, line 1875: Added unit test to add code coverage of use cases of focusing multiValues.

packages/react-select/src/accessibility/index.js
```diff
@@ -1,92 +1,133 @@
...
-      return `option ${value}, deselected.`;
-    case 'select-option':
-      return isDisabled
-        ? `option ${value} is disabled. Select another option.`
-        : `option ${value}, selected.`;
-  }
+export type AriaLiveMessagesProps = {
+  onChange?: (value: ValueType, context: SelectionContext) => string,
+  onFocus?: (focused: OptionType, context: FocusedContext) => string,
+  onFilter?: (args: {
```
ebonow, line 49: Perhaps considering renaming to `onInput`. Thoughts?

packages/react-select/src/components/LiveRegion.js
```diff
@@ -0,0 +1,186 @@
...
+      inputValue,
+      menuIsOpen,
+      messages,
+      options,
+      screenReaderStatus,
+    ]
+  );
+
+  useEffect(
+    function onContext() {
```
ebonow, line 141: rename function to relate to guidance messaging

packages/react-select/src/Select.js
```diff
@@ -583,6 +578,7 @@ export default class Select extends Component<Props, State> {
   componentDidUpdate(prevProps: Props) {
     const { isDisabled, menuIsOpen } = this.props;
     const { isFocused } = this.state;
+    //this.ariaLiveMessages = this.getAriaMessages();
```
ebonow, line 581: Delete this

packages/react-select/src/Select.js
```diff
@@ -1917,14 +1788,32 @@ export default class Select extends Component<Props, State> {
...
+      ariaSelection,
+      focusedOption,
+      focusedValue,
+      isFocused,
+      selectValue,
+    } = this.state;
+
+    const { inputValue, menuIsOpen, options } = this.props;
+
+    const focusableOptions = menuIsOpen ? this.getFocusableOptions() : [];
```
Methuselah96, line 1802: `getFocusableOptions` already checks `menuIsOpen`, so you should be able to just call `this.getFocusableOptions()`.

packages/react-select/src/Select.js
```diff
@@ -1917,14 +1788,32 @@ export default class Select extends Component<Props, State> {
...
+    const { commonProps } = this;
+    const {
+      ariaSelection,
+      focusedOption,
+      focusedValue,
+      isFocused,
+      selectValue,
+    } = this.state;
+
+    const { inputValue, menuIsOpen, options } = this.props;
```
Methuselah96, line 1800: Should we just have `LiveRegion` pull these off of `selectProps` inside the `LiveRegion` component like it does for a lot of the other props?

Methuselah96 (changes requested):
I like the direction this is headed! I haven't done a thorough review of `Select.js` yet, but I thought I should provide some initial comments. Sorry it took so long for me to take a look!

packages/react-select/src/accessibility/index.js
```diff
@@ -1,92 +1,133 @@
...
-  selectValue,
-}: {
-  focusedValue: OptionType,
-  getOptionLabel: (option: OptionType) => string,
-  selectValue: OptionsType,
-}) =>
-  `value ${getOptionLabel(focusedValue)} focused, ${selectValue.indexOf(
-    focusedValue
-  ) + 1} of ${selectValue.length}.`;
+export function getAriaLiveMessages() {
```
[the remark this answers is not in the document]
ebonow: Certainly. The singular objective of getAriaLiveMessages is simply to return an immutable object.

packages/react-select/src/components/LiveRegion.js
```diff
@@ -0,0 +1,184 @@
...
+    isDisabled,
+    isMulti,
+    isOptionDisabled,
+    isSearchable,
+    screenReaderStatus,
+    tabSelectsValue,
+  } = selectProps;
+  const ariaLabel = selectProps['aria-label'];
+
+  const defaultMessages = useRef(getAriaLiveMessages()).current;
```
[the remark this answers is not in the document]
ebonow: The result of getAriaLiveMessages() doesn't change at all so I am open to however we store the default object should we need to remerge with any new props the user passes in (eg: change languages, user passes in new config). Whether we access the default as an object, or a function which returns the object, I am open to whatever is best practice.

packages/react-select/src/Select.js
ebonow: I meant to also ask here if this is causing any additional overhead by calling `getFocusableOptions` which in turn calls `buildFocusableOptions`? When I originally was working on this, focusableOptions were in state.

Methuselah96: Not overhead that doesn't already exist.

packages/react-select/src/components/LiveRegion.js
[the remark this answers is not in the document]
ebonow: Refactored getAriaLiveMessages into an object and used as a direct import instead to resolve this and related conversation.

packages/react-select/src/accessibility/index.js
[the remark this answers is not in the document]
ebonow: Refactored this asRefactored getAriaLiveMessages into an object and used as a direct import instead to resolve this and related conversation.

packages/react-select/src/accessibility/index.js
```diff
@@ -1,92 +1,136 @@
...
-      return isDisabled
-        ? `option ${value} is disabled. Select another option.`
-        : `option ${value}, selected.`;
-  }
+export type AriaLiveMessagesProps = {
+  onChange?: (value: ValueType, context: SelectionContext) => string,
+  onFocus?: (focused: OptionType, context: FocusedContext) => string,
+  onFilter?: (args: {
+    inputValue: string,
+    screenReaderMessage: string,
```
bozdoz, line 55: `screenReaderMessage` should be `resultsMessage`

packages/react-select/src/accessibility/index.js
```diff
@@ -1,92 +1,136 @@
...
-  inputValue,
-  screenReaderMessage,
-}: {
-  inputValue: string,
-  screenReaderMessage: string,
-}) =>
-  `${screenReaderMessage}${
-    inputValue ? ' for search term ' + inputValue : ''
-  }.`;
+export const defaultAriaLiveMessages = {
```
bozdoz, line 60: Make this type-safe with:

```ts
export const defaultAriaLiveMessages: AriaLiveMessagesProps = {
```

bozdoz: Actually this wouldn't work as I expected.  Sorry, I'm from the TypeScript world, and I can't find a Flow-equivalent to TypeScript's `Required<T>` utility type.

bozdoz: I may have spoken too soon (Forgive my wordiness); just found a use with `$Shape`:

What you could do is make `AriaLiveMessagesProps` all non-optional, and then `export $Shape<T>`:

```typescript
type RequiredAriaLiveMessagesProps = {
  onChange: (value: ValueType, context: SelectionContext) => string,
  onFocus: (focused: OptionType, context: FocusedContext) => string,
  onFilter: (args: {
    inputValue: string,
    resultsMessage: string,
  }) => string,
  guidance: (type: GuidanceType, context?: GuidanceContext) => string,
};

export type AriaLiveMessagesProps = $Shape<RequiredAriaLiveMessagesProps>

export const defaultAriaLiveMessages: $Exact<RequiredAriaLiveMessagesProps> = {
```

ebonow: I appreciate you taking a look at this PR.

It's worth mentioning that the @Methuselah96 has put a lot of time into converting the project from Flow to TypeScript and the plan is to make the move shortly after this PR is released.

bozdoz: That sounds great.  Let me know if you need some help, or some :eyes: @Methuselah96 

Also, yes, I am actively testing this PR. 😄

packages/react-select/src/components/LiveRegion.js
```diff
@@ -0,0 +1,197 @@
...
+    [ariaLiveMessages]
+  );
+
+  // Update aria live selected option when prop changes
+  useEffect(
+    function onChange() {
+      let message = '';
+      if (ariaSelection && messages.onChange) {
+        const { value, ...context } = ariaSelection;
+        message = messages.onChange(value, { ...context });
```
bozdoz, line 84: nit: unnecessary spread of context: `{...context}` => `context`

packages/react-select/src/components/LiveRegion.js
```diff
@@ -0,0 +1,197 @@
...
+      messages,
+      options,
+      screenReaderStatus,
+    ]
+  );
+
+  useEffect(
+    function guidance() {
+      let guidanceMsg = '';
+      if (messages.guidance) {
```
bozdoz, line 152: do we need to check `messages.guidance` if `defaultAriaLiveMessages` is spread into `messages`?

packages/react-select/src/components/LiveRegion.js
```diff
@@ -0,0 +1,197 @@
...
+      aria-live={ariaLive}
+      aria-atomic="false"
+      aria-relevant="additions text"
+    >
+      {isFocused && (
+        <React.Fragment>
+          <span id="aria-selection">{ariaSelected}</span>
+          <span id="aria-context">{ariaContext}</span>
+        </React.Fragment>
+      )}
```
bozdoz, line 192: I've tested this, and find it troublesome after selection, blur, and regaining focus (Screenreader never again reveals the selected value because it hasn't changed).  Here's how I changed it locally to make it work, as a suggestion (let me know what you think):

(1) For these highlighted lines of code (removed condition and fragment):

```suggestion
<span id="aria-selection">{ariaSelected}</span>
<span id="aria-context">{ariaContext}</span>
```

(2) Made `ariaContext` conditional on `isFocused`:

```tsx
const ariaContext = isFocused ? `${ariaFocused} ${ariaResults} ${ariaGuidance}` : '';
```

(3) Made `ariaSelected` conditional on `isFocused`; in line 82, also adds `isFocused` to that `useEffect` dependencies:

```tsx
...
if (isFocused && ariaSelection && messages.onChange) {
...
```

And (4) added `removals` to `aria-relevant`

So now when the component loses focus the aria-live message becomes empty, sees that as relevant, and will announce the selected option again after announcing the label of the combobox.

Anyway, curious your thoughts on this.

packages/react-select/src/accessibility/index.js
bozdoz: I like `onFilter`, personally

packages/react-select/src/components/LiveRegion.js
ebonow: Which screenreader are you using for your testing out of curiosity? I am using VoiceOver on OSX and hadn't run into this though I may have to re-test after removing the "role=alert"

packages/react-select/src/components/LiveRegion.js
ebonow: @bozdoz Good question with a few thoughts here:

The thought here was if a user wanted to remove any of the messages, they could pass in 
```jsx
ariaLiveMessages={ guidance: null }
```

Guidance in particular can get a bit wordy and the default behavior is such that screen readers will read both this and the listbox or textbox instructions.

I also envisioned some advanced level of customization where a function can return false or null if some criteria is met. In this case, if the user is navigating a listbox, they may not want to hear the directions after some amount of time, interaction, etc... so the thinking was to make this as dynamic as possible.

packages/react-select/src/components/LiveRegion.js
ebonow: @bozdoz the intention is to only read the "selected" option onChange. 

The issue you are referring to is something I thought about, but think it's a bit more complicated. It could be possible that the Select has a value or a defaultValue without any selection by the user. The other use case to be considered is if the select `isMulti`, then only the last selected value is read back but not all of values.

I think this needs more consideration overall, but my thoughts on this are to clear the ariaSelected value from the Select state when the select loses focus. This way we can derive and announce all selected values instead from the selectValue instead, though there is no existing support for announcing multiple options at present.

bozdoz: > Which screenreader are you using for your testing out of curiosity? I am using VoiceOver on OSX and hadn't run into this though I may have to re-test after removing the "role=alert"

I am using the same. Though I did notice a change since updating to Big Sur.

bozdoz: > The issue you are referring to is something I thought about, but think it's a bit more complicated. It could be possible that the Select has a value or a defaultValue without any selection by the user. The other use case to be considered is if the select `isMulti`, then only the last selected value is read back but not all of values.

For a defaultValue I would think you would still want the screen reader to announce that value; the user should know there's an option chosen and can move to another form input for example.  I haven't tested multiple choices yet, but thanks for the heads up.  That's something I wasn't thinking about.  Actually, I just tested a native <select multiple> element, and voiceover is truly awful with it.

packages/react-select/src/components/LiveRegion.js
bozdoz: `guidance: null` looks good to me; the types did not make that API apparent to me.  Perhaps you could change the functions to be Maybe types (as opposed to optional):

```diff
- guidance?: (type: GuidanceType, context?: GuidanceContext) => string,
+ guidance?: ?(type: GuidanceType, context?: GuidanceContext) => string,
```

That way `messages.guidance` can be null

packages/react-select/src/components/LiveRegion.js
ebonow: The selection event is currently only called by the `onChange` event so the ability to get the select's value would have to be derived differently. Will be addressed but wanted to maintain some parity to ensure nothing was affected especially sinceiIt's already a fairly big change stripping all of these state changes out of Select.js into its own component. 

By the way, one of the things I noticed while I was testing is that the screenreader's instructions about keyboard interactions in a listbox can interfere with the aria-live arrowing over different focused options because it's in the passive voice so it doesn't interrupt no matter how many different options the user interacts with. This is part of the reason I wanted to add aria-live as a prop to allow users to choose between off/passive/assertive to determine what's best for their own needs.

bozdoz: I thought it best to remove the guidance message in my implementation.  Do you think assertive works better than passive?

ebonow: I personally think so, but apparently not everyone https://github.com/JedWatson/react-select/issues/3353

The described behavior above only happened when I had passive voice. Assertive or even a non-standard string would allow the narrator to be interrupted with new events, whereas passive did not.

bozdoz: Wow. Yeah, assertive is way better than passive!  I'll do some tests for default value and multiple values.

packages/react-select/src/Select.js
```diff
@@ -1261,16 +1177,11 @@ export default class Select extends Component<Props, State> {
...
     }
-    this.setState({ inputIsHiddenAfterUpdate: false });
-    this.announceAriaLiveContext({
-      event: 'input',
-      context: { isSearchable, isMulti },
-    });
     this.setState({
+      inputIsHiddenAfterUpdate: false,
       isFocused: true,
     });
```
bozdoz, line 1186: I added this after line 1186 to test screenreader with default value, and it seems to be working quite well:

```ts
this.ariaOnChange(this.props.value, {
  action: 'select-option',
});
```

For your consideration

packages/react-select/src/components/LiveRegion.js
bozdoz: So after seeing how great `assertive` is (and noting it does have a downfall of cutting off the screen reader from reading the combobox label), I realized we *could* move the aria-live attributes to the child tags instead.

For your consideration:

```tsx
<A11yText>
  <span
    aria-live="polite"
    aria-atomic="false"
    aria-relevant="all"
    id="aria-selection"
  >
    {ariaSelected}
  </span>
  <span
    aria-live="assertive"
    aria-atomic="false"
    aria-relevant="additions text"
    id="aria-context"
  >
    {ariaContext}
  </span>
</A11yText>
```

Here the screenreader will read the combobox label, and not be cut off by repeating the selected value (because it's set to polite).  Once the user has engaged with the component and is selecting options an assertive message pops up, which I believe is ideal since it corresponds with the user interrupting the label announcement (and subsequent option choices) with their own actions.  Anyway, this may be opinionated, but I think that covers a lot of the bases I've been trying to cover.  Also note the difference in `aria-relevant`.  

Let me know what you think! 😄 

Let me ekno

ebonow: I considered this and played around with the different aria relevant and atomic settings. The one thing that would have to be considered in your example (that is fixed in this PR) is that the aria-live nodes are present onMount as this apparently creates some issues for some screen readers otherwise. 

I'm not sure which parts should be made assertive or passive. There are arguments to make selected and focused assertive, but perhaps onResults too? I was playing around with adding support for loading, but the timing of the loading seemed to vary the messaging (especially when in passive mode) and there's further conversations to be had about how persistent the navigation guidance instructions should be (probably not needed on every new addition)

For a first pass, I think exposing string formatting functions and a universal aria-live prop will perhaps give us a better idea how to improve or perhaps even expose this as a public component so each of the components of the aria context and selection can be customized.

bozdoz: I have not yet looked into async loading.  I can't imagine needing that announced at all, especially if the total results count changes.  But I will do some research into it.

bozdoz: Also found that the original ticket has something similar with two elements, one polite, and one assertive: https://github.com/JedWatson/react-select/pull/3550/files#diff-7064e1eb0f7889ae735e705d019ebfa160fbb668d7733f73ee6f107b8cf16fadR1844-R1849

ebonow: Yes, and I think this needs more thought to better understand how we want to structure all the components of the ariaContext and would be worth doing that when we have more resources available to know how this all works with JAWS and NVDA.

I'd like to wait until we get more testing available for all of the screen readers before experimenting and testing these and possibly other DOM changes for other aspects of ariaContext.

packages/react-select/src/components/LiveRegion.js
```diff
@@ -0,0 +1,167 @@
...
+    return message;
+  }, [ariaSelection, messages]);
+
+  const ariaFocused = useMemo(() => {
+    let focusMsg = '';
+    const focused = focusedOption || focusedValue;
+    const isSelected = !!(
+      focusedOption &&
+      selectValue &&
+      selectValue.includes(focusedOption)
```
bozdoz, line 80: this fails to determine `isSelected` if `isMulti` and multiple options are chosen.  I tried out react-fast-compare (which doesn't appear to be used anywhere) and made it work; for your consideration:

```suggestion
      !!selectValue.find((selected) => exportedEqual(selected, focusedOption))
```

that function imported at the top: 

```js
import exportedEqual from '../internal/react-fast-compare';
```

ebonow:
@bozdoz , so I reworked a few things about this component over the last week.

- I condensed all of the formatting functions to take a single prop.
  - The rationale was that it's not clear to me whether it makes sense to have different things in different places and just makes it harder to deconstruct. 
  - At first I wanted synchronicity with the `onChange` function, but found that to be confusing since they are essentially sued for two different things. 
    - onChange as a select prop function is expected to pass value back to state for controlled input. 
    - onChange as a formatting function is expected to pass back information regarding the selected/removed option to vocalize it to the user and the biggest difference between the two is that the formatting function fires when a user selects a disabled option. 
  - It's also worth noting that the `option` prop is not populated in the actionMeta of onChange when `select-option` is called on a single select, but is when it isMulti, so for consistency, I return a `selected` prop and the derived label of that, because I find the onChange actionMeta overly complicated when option could simply be present on every action.

packages/react-select/src/components/LiveRegion.js
ebonow: Thanks, let me take a look at this and add some tests to make sure this doesn't get missed.

ebonow: Do you have latest and can you provide an example of it not working correctly? I do not seem to have the same problem of the select identifying menu options that are selected.
Screen Shot 2021-03-01 at 3 51 14 PM

packages/react-select/src/accessibility/index.js
```diff
@@ -7,66 +7,85 @@ import {
...
 export type AriaLiveMessagesProps = {
-  onChange?: (value: ValueType, context: SelectionContext) => string,
-  onFocus?: (focused: OptionType, context: FocusedContext) => string,
-  onFilter?: (args: {
-    inputValue: string,
-    screenReaderMessage: string,
-  }) => string,
-  guidance?: (type: GuidanceType, context?: GuidanceContext) => string,
+  // Guidance message used to convey component state and specific keyboard interactivity
+  guidance?: (tprops: AriaGuidanceProps) => string,
```
bozdoz, line 70: typo `tprops`

packages/react-select/src/accessibility/index.js
ebonow: Replaced this with single prop so should be resolved now.

packages/react-select/src/components/LiveRegion.js
bozdoz: I do have latest.  I don't know how to provide a good  example, but I'll try.  I'm using hooks:

```tsx
const FRUITS_AND_VEGGIES = [
    'Apple',
    'Artichoke',
    'Asparagus',
    'Banana',
    'Beets',
    'Bell pepper',
    'Broccoli',
    'Brussels sprout',
]
export const MultipleValues = () => {
    const options = FRUITS_AND_VEGGIES.map((x, i) => ({
        value: x,
        label: x,
        isDisabled: i % 5 === 1,
    }));

    const [value, setValue] = useState<any>(options[2]);

    return (
        <div>
            <label id="ex3-label">Fruits and Vegetables</label>
            <Select
                isMulti
                value={value}
                onChange={(v) => setValue(v)}
                options={options}
                closeMenuOnSelect={false}
                aria-labelledby="ex3-label"
                isSearchable
                isClearable={false}
            />
       </div>
    );
};
```

Here's what I'm getting:

Screen Shot 2021-03-01 at 6 31 45 PM

But it's a different message initially:

Screen Shot 2021-03-01 at 6 32 30 PM

bozdoz: Oh, you know what's happening is that the function is being rerendered, and the array map is generating new objects (in my example).  So your test is checking for identical objects, which fails in my case.  If it were defined outside of the function component, then it should work with the simple equality check. Makes sense.  For your consideration.

ebonow: There is some internal discussion regarding the conversation of comparing objects by value or reference. I don't disagree with a value based approach over one that is referential, but opted to remain consistent with the existing approach for the sake of consistency.

packages/react-select/src/Select.js
```diff
@@ -1917,14 +1786,27 @@ export default class Select extends Component<Props, State> {
...
+    const focusableOptions = this.getFocusableOptions();
+
     return (
-      <A11yText aria-live="polite">
-        <span id="aria-selection-event">
-          &nbsp;{this.state.ariaLiveSelection}
-        </span>
-        <span id="aria-context">&nbsp;{this.constructAriaLiveMessage()}</span>
-      </A11yText>
+      <LiveRegion
```
bozdoz, line 1801: Low effort improvement would be to include this in `getComponents`:

In Select.js
```tsx
const { LiveRegion } = this.getComponents()
```

And then for components/index.js:
```tsx
LiveRegion: ComponentType<LiveRegionProps>,
```

ebonow: I think exposing it in the public facing components is a near future goal but not for this release.

JedWatson (approved):

jc00973:
Hey! I read all this thread and I couldn't find if it exists a component where I can directly define the AriaLiveMessages so I can translate it to another language other than English. 

This was already reported in  #3352 and #3691, but they were closed claiming that issue would be treated here, but I'm guessing the internationalization features are still lacking. 

Can someone confirm if there is already a feature to translate the AriaLiveMessages to other languages? If not, I would be interested in working on it.

jc00973:
Hey! I read all this thread and I couldn't find if it exists a component where I can directly define the AriaLiveMessages so I can translate it to another language other than English. 

This was already reported in  #3352 and #3691, but they were closed claiming that issue would be treated here, but I'm guessing the internationalization features are still lacking. 

Can someone confirm if there is already a feature to translate the the default AriaLiveMessages to other languages? If not, I would be interested in working on it.

Issue #3802, opened 2019-10-04
ScrollLock dist missing window undefined check


swensorm:
While attempting to upgrade react-select from version 2.4.4 to the latest 3.0.8 I'm running into issues with server side rendering and the following error:
```
ErrorDetails : ReferenceError: window is not defined
    at ./node_modules/react-select/dist/base/dist/react-select-cac0a5ae.browser.esm.js (server.min.js:117809:17) -> var canUseDOM = !!(window.document && window.document.createElement);
```

We use this within a .Net CMS by way of ReactJS.Net so the `window` object is not available while it is being rendered on the server.  The strange thing is that looking at the dist output from the npm package, the check for window being undefined isn't there:
![image](https://user-images.githubusercontent.com/1808452/66236311-c8932180-e6b7-11e9-947b-2bc3cff6e8bf.png)
However, looking at the src it is:
![image](https://user-images.githubusercontent.com/1808452/66236359-e9f40d80-e6b7-11e9-8d7a-456fa0b73a9e.png)

The check did exist in the dist output of the 2.4.4 version so something with the build process may have been changed with version 3.0.0 that could have caused this?  Thank you for looking at the issue!

krzysztof4M:
@swensorm  Have you found any solution?

mgrzyb:
The bundle you use on the server should be built with webpacks target: 'node' to resolve modules from 'main' or 'module' fields and not using client-side specific aliases.

Since you're not running it under nodejs (ReactJS.NET does not provide full nodejs environment) you will get "require is not defined" when trying to use bundle built like this but there is away.

In the server build, keep target: 'web' (default when unspecified) but change resolve.aliasFields to an empty array, like this:
```
module.export = {
 ...
 resolve: {
    aliasFields: []
  }
}
```
This way webpack will not use browser specific aliases when resolving react-select module... Ufff... :)

alverdal:
I was able to solve a similar issue with "window is not defined" when using SSR with reactjs.net by using your fix @mgrzyb . 
Issue was described in ticket https://github.com/marcelltoth/react-select-reborn/issues/8.

swensorm:
While I can confirm that adding `aliasFields: []` to the webpack config did indeed fix the drop-down rendering server side, it had the unfortunate side effect of causing over 1 MB of locale data from the `Intl` package to be included as well.

rafalschmidt97:
[...]

PR #4423, opened 2021-02-05
Remove browser alias fields

Methuselah96:
Resolves https://github.com/JedWatson/react-select/issues/3802.

The `browser` alias outputs aren't doing much anyway besides causing the bug in SSR (they just remove the `typeof window !== undefined` check in one place), so I think it practically makes sense to remove them even though we're technically doing everything correctly.

### DIFF
--- a/.changeset/perfect-pandas-argue.md
+++ b/.changeset/perfect-pandas-argue.md
@@ -1,0 +1,5 @@
+---
+"react-select": patch
+---
+
+Remove browser alias fields in order to fix SSR apps
--- a/packages/react-select/animated/package.json
+++ b/packages/react-select/animated/package.json
@@ -1,10 +1,6 @@
 {
   "main": "dist/react-select.cjs.js",
   "module": "dist/react-select.esm.js",
-  "browser": {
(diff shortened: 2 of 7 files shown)


Methuselah96:
Need to look into this some more.

JedWatson:
@mitchellhamilton do you know anything about this that would make this PR not the best way forward?

mitchellhamilton:
There's nothing really wrong with merging this PR since in this case, the browser builds are pointless. react-select depends on emotion though and emotion has browser builds which have a non-trivial bundle size saving so this change won't really solve the problem for people.

Methuselah96:
True. I've added a [concluding comment](https://github.com/JedWatson/react-select/issues/3802#issuecomment-776333493) to the linked issue.

@JedWatson Up to you whether this is worth merging. The benefit of merging it would be that it would then definitely not be our problem and we could tell people to bug Emotion, their SSR library, and/or Webpack instead of bugging us. The browser builds (as Mitchell pointed) are currently pointless.

JedWatson (approved):

PR #4436, opened 2021-02-08
Allow the formatGroupLabel be dependent to inputValue

montaseri:
The `formatGroupLabel` should have the same feature like `formatOptionLabel`.

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -65,6 +65,9 @@ type FormatOptionLabelMeta = {
   inputValue: string,
   selectValue: ValueType,
 };
+type FormatGroupLabelMeta = {
+  inputValue: string
+};
 
 export type Props = {
   /* Aria label (for assistive tech) */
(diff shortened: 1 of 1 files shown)


Rall3n:
> The `formatGroupLabel` should have the same feature like `formatOptionLabel`.

@montaseri Could you elaborate on why it **should** have the same features (in regards to `inputValue`)?

ebonow:
>  Could you elaborate on why it **should** have the same features (in regards to `inputValue`)?

One possible use-case is if you want to filter groups by input or apply syntax highlighting to the group labels depending on the input.

I believe there may be a larger discussion perhaps to be had about accessibility and whether group labels should be focusable and creating a built-in to access the group heading as a string such as `getGoupLabel`...

JedWatson:
I'm generally 👍 to this change, but it would be good to link to an issue that describes the reasoning behind the change. I can think of some but not sure if they're actually valid real-world things.

Other than that, this needs to pass CI / have a changeset / be up to date with master... @montaseri there's an option to let people with write access to the target repo update the PR, if you tick that we can help you out

montaseri:
I would like to close the PR for now.

Issue #2729, opened 2018-06-19
non-passive event listener in scroll blocking events


gugurete:
chrome warnings popping up about non-passive event listeners

react-select.es.js:571 [Violation] Added non-passive event listener to a scroll-blocking 'wheel' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952

[Violation] Added non-passive event listener to a scroll-blocking 'touchstart' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952

gwyneplaine:
hello @gugurete thanks for raising this, which version of react-select are you using? 
Could you also supply us with a [codesandbox](https://codesandbox.io/s/new) with your select configuration, just so we can debug/ reason about this a bit easier.

gugurete:
It is version: 2.0.0-beta.6, and the offending code is in react-select.es.js starting with line 564
I believe it should be something like  **el.addEventListener('wheel', this.onWheel, {passive: true});**
to avoid chrome warning messages.

    key: 'startListening',
    value: function startListening(el) {
      // bail early if no scroll available
      if (el.scrollHeight <= el.clientHeight) return;

      // all the if statements are to appease Flow 😢
      if (typeof el.addEventListener === 'function') {
        el.addEventListener('wheel', this.onWheel, false);
      }
      if (typeof el.addEventListener === 'function') {
        el.addEventListener('touchstart', this.onTouchStart, false);
      }
      if (typeof el.addEventListener === 'function') {
        el.addEventListener('touchmove', this.onTouchMove, false);
      }
    }

gugurete:
https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md

zike-Q:
I am getting the same warnings on Chrome with "react-select 2.0.0-beta.7":

> Added non-passive event listener to a scroll-blocking 'wheel' event.
> Added non-passive event listener to a scroll-blocking 'touchstart' event.
> Added non-passive event listener to a scroll-blocking 'touchmove' event.

xakep139:
Same problem with `react-select` 2.0.0, Chrome version is 68.0.3440.106 (64-bit)

jossmac:
@gugurete looks good, would you mind submitting a PR?

xakep139:
@jossmac #2933

jossmac:
Re-opened due to #2940

[...]

PR #4437, opened 2021-02-08
Declare event listeners as non-passive to remove browser warnings

ebonow:
### Problem:
Issue: non-passive event listener in scroll blocking events #2729

Chrome issues console warnings when adding some event listeners without declaring them as passive. Three of these warnings present themselves in a non-mobile view when opening the menu.

Screen Shot 2021-02-08 at 8 13 03 AM

### Resolution:
By explicitly setting passive is false, Chrome will no longer show this warning in the console.


JedWatson (changes requested):
I took a look and there is a backwards compatibility concern here with older browsers -- https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md#feature-detection

Off the top of my head I'm **not confident** that this wouldn't cause a problem for browsers that react-select supports **in the wild**, so unless we wait for a major release, I'd be more comfortable if we added the compatibility check in here.

(note: with the next major version we should really clarify our browser support so stuff like this doesn't come up again! I think evergreen is fair these days, and IE is dead, but when we rewrote for v2 IE11 was still actually a target for Atlassian so unless someone knows other reasons it wouldn't still work on IE, it may, and if so people using it may also be relying on that 🤷‍♂️)

ebonow:
@JedWatson Thanks for the heads up on that and I did miss that consideration, though I did see this was something that had been explored before.

This may be a case where we'd like to include the polyfill and consider removing it for some major release. I understand the warnings are fairly innocuous, but ultimately it reflects poorly on on the devs using react-select to have Chrome console spamming 3 warning messages every time a menu is opened.

ebonow:
@JedWatson , I added a polyfill to support legacy browsers. 

Also in regards to this line...
https://github.com/JedWatson/react-select/blob/9aaf35b0c41b8c89da42430365c8c58cc3b41815/packages/react-select/src/internal/useScrollCapture.js#L106

I know I am fairly new to Flow, but the events are being attached directly to the DOM, so wouldn't they be actual Wheel/Touch events and not synthetic events?

JedWatson (approved):

PR #4444, opened 2021-02-10
Use accessor props to get value and label in `compareOption`

Rall3n:
By default the `compareOption` internal function for the `Creatable` component defaults to using the `value` and `label` properties of options to resolve and find options matching a search string. Besides `getOptionLabel` and `getOptionValue` people would have to also use the `isValidNewOption` prop to accommodate for their custom options.

This PR forwards the values of the accessor props `getOptionLabel` and `getOptionValue` to the `compareOption` function to allow for the usage of custom options without the need to overwrite the default `isValidNewOption` function.

There was PR #2956 which tried to do the same but failed in using the proper functions (defaulted to hardcoded built-ins).

### DIFF
--- a/.changeset/bright-tigers-tie.md
+++ b/.changeset/bright-tigers-tie.md
@@ -1,0 +1,5 @@
+---
+'react-select': minor
+---
+
+Use accessor props to get value and label in `compareOption`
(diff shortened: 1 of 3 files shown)


JedWatson (approved):

JedWatson:
Thanks @Rall3n!

PR #4465, opened 2021-02-24
fix: updates Select types

bozdoz:
Adds CommonProps to fix an outstanding TODO related to Select['commonProps'] being typed as `any`.  Updates other related types to adjust for this change.

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -330,7 +330,7 @@ type CommonProps = {
   ): void,
   selectProps: Props,
   theme: typeof defaultTheme,
-}
+};
 
 function toCategorizedOption(
   props: Props,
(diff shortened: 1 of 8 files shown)


ebonow:
Hey @bozdoz 

Jed is a big fan of this PR. The biggest issue at the moment though is that the TypeScript rewrite is nearing completion soon and want to avoid and potential merge conflicts since it's almost complete at this point. We may still follow up with a quality of life release for v4 for those who want to continue Flow development, but thought to at least thank you for this and that it was well received.

Same in regards to your findings about aria input values being announced. We have a bit of a pause on larger initiatives until we have the TypeScript version released or at least in alpha.

Methuselah96:
Thanks for the work on this, we've recently converted the code over to TypeScript (https://github.com/JedWatson/react-select/pull/4489) in preparation for v5 so the Flow types will no longer be relevant. The `CommonProps` should be updated with more accurate types in the TypeScript types.

PR #4480, opened 2021-03-08
Fix skipped unit tests for autoFocus and onChange

ebonow:
When running the test suite, several tests were marked as skip. The purpose of this PR is to fix these tests so that they work as expected.

- autoFocus calls onFocus test for both single and multi
  - This test had a comment displaying a codesandbox showing that it wouldn't be called, but instead the wrong prop name was being called (`autofocus` instead of `autoFocus`)
- spacebar does not trigger select onChange method
  - Test said to revalidate this, so removed call setting focusedOption (as first menu option is focused anyway) and the onChange method was not called as expected.

### DIFF
--- a/.changeset/little-ants-sort.md
+++ b/.changeset/little-ants-sort.md
@@ -1,0 +1,5 @@
+---
+'react-select': patch
+---
+
+Fixed 3 unit test test cases instead of skipping them
--- a/.changeset/little-ants-sort.md
+++ b/.changeset/little-ants-sort.md
@@ -1,5 +1,0 @@
----
-'react-select': patch
----
-
-Fixed 3 unit test test cases instead of skipping them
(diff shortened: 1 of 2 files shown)


.changeset/little-ants-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-select': patch
+---
+
+Fixed 3 unit test test cases instead of skipping them
```
Methuselah96, line 5: Maybe we don't need a changeset for this one because it's not modifying anything that a user of `react-select` would use?

ebonow: I wasn't sure since it lives in the source, but also doesn't contribute or change any functionality so id be in favor of removing it.

JedWatson (approved):

PR #4484, opened 2021-03-10
Fix scrolling issues

vadimka123:
https://github.com/JedWatson/react-select/issues/3232

### DIFF
--- a/packages/react-select/src/internal/useScrollCapture.js
+++ b/packages/react-select/src/internal/useScrollCapture.js
@@ -96,6 +96,20 @@ export default function useScrollCapture({
     },
     [handleEventDelta]
   );
+  const onMouseDown = useCallback(
+    (event: SyntheticMouseEvent<HTMLElement>) => {
+      // set mouse start so we can calculate mousemove delta
+      touchStart.current = event.clientY;
+    },
+    []
+  );
+  const onMouseUp = useCallback(
+    (event: SyntheticMouseEvent<HTMLElement>) => {
+      const deltaY = event.clientY - touchStart.current;
+      handleEventDelta(event, deltaY);
+    },
+    [handleEventDelta]
+  );
 
   const startListening = useCallback(
     el => {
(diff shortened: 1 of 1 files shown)


vadimka123:
Hi, @ebonow 

I tested only in desktop this behaviour
It's separate events: only touch events fired when I make scroll use mouse wheel, only this mouse events fired when I make mouse drag

ebonow:
No worries, I think I answered my own question anyway as I confused touchmove with touchend. I'll do a quick smoke test and verify that it works and resolves both of the issues that are mentioned.

ebonow:
@vadimka123 I had an opportunity to test this. My concern is that this seems like a partial solution. It is only called when the mouse button is let go so if the user drags to the bottom, doesn't see what they want, drags back up, and then lets go of the mouse button, the onScrollToBottom handler is never called.

It's also unclear to me how this resolves #2862 as this doesn't appear to address the scrollTarget ref being null when first called.

I appreciate the contribution and simply want to make sure we are adding code that resolves the core issues instead of continuing to patch this up with work arounds. I think long term there are some other solutions we can look at, but I'm not sure that adding 2 more event listeners does that, though it also likely doesn't hurt either.

vadimka123:
@ebonow , sorry, #2862 added by mistake 

I agree that a partial solution
Looks like full working solution will be onScroll event listener instead of all others, but it's removed in version 2
Do you know why?

ebonow:
> I agree that a partial solution
> Looks like full working solution will be onScroll event listener instead of all others, but it's removed in version 2
> Do you know why?

I do know why. It is related to how iOS used to only call the onScroll event when the scrolling has stopped which was further complicated by the built in inertia.
https://www.tjvantoll.com/2012/08/19/onscroll-event-issues-on-mobile-browsers/

I believe this has since been addressed in iOS8 or later, and put together this codesandbox to test if it can be worked back in.
https://codesandbox.io/s/fancy-cherry-l3vuh

I tested this against an iPhone simulator as well as someone else tested against an iPhone and it looks like scrolling is being updated in real-time now so I think that would be the better way to go. Not sure if you would be open to having a stab at this, though it may be worth waiting until after TypeScript refactoring is completed.

ebonow:
Greetings @vadimka123 ,

We appreciate you putting this together but I think we would like to put together a more complete scrolling solution that would likely overwrite the already existing and this proposed code so we will be closing this. I will be sure to include you on any new developments. 

Thank you again for putting this together and let us know if you would be instead in putting together a PR that uses the onScroll event handler.

Issue #4142, opened 2020-07-27
Convert to TypeScript


Methuselah96:
This is an issue to track the progress of converting this project to TypeScript. The plan is to do this incrementally in order to make the transition as smooth as possible and not make the review process too overwhelming. The overall plan is:

1. Incrementally convert `docs` to TypeScript.
  - TypeScript files can import Flow files if they have a corresponding TypeScript declaration file, so we'll just do a breadth-first conversion.
  - Import `react-select` types from `@types/react-select` in this step.
  - Fix types in `@types/react-select` to make it work with `docs`.
  - PRs:
    - Add TypeScript infrastructure (https://github.com/JedWatson/react-select/pull/4139)
    - Convert Webpack config to TypeScript (https://github.com/JedWatson/react-select/pull/4143)
    - Incrementally convert docs to TypeScript (part 1) (https://github.com/JedWatson/react-select/pull/4144)
    - Incrementally convert docs to TypeScript (part 2) (TODO)
    - etc.
2. Move Typescript declaration files from `@types/react-select` to this repository.
3. Incrementally convert main source files to TypeScript.

bladey:
Thank you for all your efforts @Methuselah96, this looks great.

Methuselah96:
@kylemh Thanks for all your help so far on PRs! I saw on https://github.com/JedWatson/react-select/issues/4293 that you're looking to help out with the TS conversion. Most of the work is actually already done (I have the entire code-base converted to TypeScript on [a branch](https://github.com/Methuselah96/react-select/tree/typescript-final), it's just a matter of getting it merged in.

The PRs I've made so far have been to incrementally convert the docs to TypeScript using the types that currently exist on DefinitelyTyped. When I run across a problem in the types, I make a PR to DefinitelyTyped. This works, however it's slow because there are no active reviewers for react-select on DefinitelyTyped. Would be willing to update the types on DefinitelyTyped to 3.1 (by adding [the `isLoading` prop to `AsyncSelect`](https://github.com/JedWatson/react-select/releases/tag/react-select%403.1.0)) and also add yourself as a definition author. That way we can review each others PRs and this process can be sped up quite a bit.

Thanks for all your hard work!

kylemh:
Down; however, [that looks to already be done, no?](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-select/src/Async.d.ts#L28)

Methuselah96:
[...]

PR #4489, opened 2021-03-11
Convert to TypeScript

Methuselah96:
Resolves https://github.com/JedWatson/react-select/issues/4142.


Methuselah96:
As a note to anyone looking at this: my plan is to add defaults to the exported generic types to reduce the need for users to specify generics where it shouldn't be necessary.

Methuselah96:
Note that there's no changelog yet. I'll leave that to you @JedWatson.

JedWatson (approved):
🎉 

Thanks so much for your work on this @Methuselah96 

Let's get this merged, and out for testing as a beta while we pick up the rest of the tasks for the 5.0 release!

PR #4495, opened 2021-03-16
Using Index as a key is an anti-pattern in React

vadimka123:
Fixed issue when on D&D new MultiValue component rendered instead of update exists components
Also in general it's anti-pattern

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -1463,7 +1463,7 @@ export default class Select extends Component<Props, State> {
             }}
             isFocused={isOptionFocused}
             isDisabled={isDisabled}
-            key={`${this.getOptionValue(opt)}${index}`}
+            key={this.getOptionValue(opt)}
             index={index}
             removeProps={{
               onClick: () => this.removeValue(opt),
@@ -1767,14 +1767,12 @@ export default class Select extends Component<Props, State> {
       } else {
         const input =
           selectValue.length > 0 ? (
-            selectValue.map((opt, i) => (
-              <input
-                key={`i-${i}`}
-                name={name}
-                type="hidden"
-                value={this.getOptionValue(opt)}
-              />
-            ))
+            selectValue.map((opt, i) => {
+              const value = this.getOptionValue(opt);
+              return (
+                <input key={value} name={name} type="hidden" value={value} />
+              );
+            })
           ) : (
             <input name={name} type="hidden" />
           );
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -1767,7 +1767,7 @@ export default class Select extends Component<Props, State> {
       } else {
         const input =
           selectValue.length > 0 ? (
-            selectValue.map((opt, i) => {
+            selectValue.map(opt => {
               const value = this.getOptionValue(opt);
               return (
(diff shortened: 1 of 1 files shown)


ebonow:
Looks good and makes sense to me though I know this exists in a few other places in Select.js

packages/react-select/src/Select.js
```diff
@@ -1463,7 +1463,7 @@ export default class Select extends Component<Props, State> {
             }}
             isFocused={isOptionFocused}
             isDisabled={isDisabled}
-            key={`${this.getOptionValue(opt)}${index}`}
+            key={this.getOptionValue(opt)}
```
razh, line 1466: For reference, this was added to fix https://github.com/JedWatson/react-select/pull/4154.

Rall3n (changes requested):
This would/could introduce a breaking change resulting in `Duplicate key` warnings.
I believe a better and more robust solution should be considered instead.

vadimka123:
@Rall3n , maybe additional prop ? duplicateValues or maybe pass func for build key ?

packages/react-select/src/Select.js
```diff
@@ -1767,14 +1767,12 @@ export default class Select extends Component<Props, State> {
...
-            selectValue.map((opt, i) => (
-              <input
-                key={`i-${i}`}
-                name={name}
-                type="hidden"
-                value={this.getOptionValue(opt)}
-              />
-            ))
+            selectValue.map(opt => {
+              const value = this.getOptionValue(opt);
```
NullPainter2, line 1771: This value might not be unique at all for `key`. 


Make this customizable by user instead by some callback, like `this.getReactKey(opt,i)`, because I as a user know how to specify unique id for my data.

NullPainter2 (changes requested):
React-select can't specify `key` because it does not know what is unique in user's data. Please add callback so user can specify unique ID herself.

packages/react-select/src/Select.js
NullPainter2: @Rall3n Is "value" of option required to be unique or not  ? I assume it should be ... but I found no information about that in docs.

ebonow: @NullPainter2 it's likely assumed, but not sure that it's a stated requirement. Perhaps something to consider. I wonder if something like `${label}_${value}` would be acceptable as I can't think of a compelling reason why or how a user would have duplicate derived value and label values. Perhaps in different option groups, but they would have a different parent node so the keys I believe wouldn't need to be unique.

ebonow:
Greetings @vadimka123 ,

Thank you for putting this together. We discussed this with Jed and the rest of the team, and this is something we would certainly like to address, but given that there is no requirement for values to be unique, this could potentially be a breaking change. 

We would likely would want a more complete solution perhaps as @NullPainter2 had suggested with a new built-in prop for more advanced use cases like this. We want to address this, but are currently in the middle of a total rewrite to TypeScript. This will be something that we will revisit once it's complete,

pgmanutd:
[One issue](https://github.com/JedWatson/react-select/issues/4602) was raised today around this I think.

SalahAdDin:
@vadimka123 Can you solve it, please?

SalahAdDin:
@vadimka123 what?

PR #4513, opened 2021-03-22
Apply reset for box shadow

kluvin:
Hello!

I ran into an issue with Tailwind's forms plugin, where a box shadow is applied on focus. This might also be a general issue wherever a box shadow is used for the input box. 

This patch resolves the issue for me. Let me know if I should change anything.

### DIFF
--- a/packages/react-select/src/components/Input.js
+++ b/packages/react-select/src/components/Input.js
@@ -25,6 +25,7 @@ export const inputCSS = ({
   margin: spacing.baseUnit / 2,
   paddingBottom: spacing.baseUnit / 2,
   paddingTop: spacing.baseUnit / 2,
+  boxShadow: 'none',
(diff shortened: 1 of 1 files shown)


ebonow:
Greetings @kluvin ,

I appreciate the contribution, but this could be a breaking change for some users.

To answer the root problem you are running into, can you try the following and see if this resolves your issue without impacting the source code?

```jsx
const styles = {
  input = css => ({ 
    ...css, 
    input: { boxShadow: 'none' }   
})

return <Select styles={styles} {...otherProps} />
```

kluvin:
More or less that is my current solution. Although one needs to nest down to actual input:

```js
const styles = {
  input = css => ({
    ...css,
    '& div': {
      '& input': {
        boxShadow: 'none !important',
      }
    }
  })
}
```

Issue #4504, opened 2021-03-19
Select does not show opi


AliNjie:

h3rmanj:
I've also noticed this, as I used the workaround for limiting displayed options in large sets, which no longer works.  
https://github.com/JedWatson/react-select/issues/126#issuecomment-478603553

AliNjie:
Interesting side note;
Setting the `limit` variable to `colourOptions.length + 1` displays one option in the menu, `colourOptions.length + 2` displays two and so on.

h3rmanj:
After some debugging, I've found that it's this change that broke it. https://github.com/JedWatson/react-select/pull/3990
Specifically the equality check on `selectValue`. I'll try to throw in a PR.

ebonow:
Greetings,

Thanks for the find @h3rmanj ! 

Some next steps I think to put around this:

1) Memo the buildCategorizedOptions is a good first step
2) There should be a feature request added so that `maxOptions` can be specified in the filterOptions and/or the filterOptions function exposes an index to allow the user this ability.

Thoughts?

h3rmanj:
I agree with 1, and have made a PR in #4514.  
As for 2, I still think using `filterOption` to limit the amount of rendered options is somewhat clunky, and tried to just directly add a prop that limits the rendered options in #4515. 

I did them in seperate PRs so they could be addressed individually. Either would solve this issue, but both would be nice to have.

PR #4514, opened 2021-03-22
Memoize buildCategorizedOptions

h3rmanj:
Memoize buildCategorizedOptions, as it is called multiple times per render, on option lists possibly containing thousands of options.
Since I mainly tried to solve #4504, I only tweaked it to a point where you could limit the number of rendered options using `filterOption`.

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -1,6 +1,7 @@
 // @flow
 
 import React, { Component, type ElementRef, type Node } from 'react';
+import memoize from 'memoize-one';
 import { MenuPlacer } from './components/Menu';
 import LiveRegion from './components/LiveRegion';
 
@@ -341,7 +342,7 @@ function toCategorizedOption(
   };
 }
 
-function buildCategorizedOptions(
+let buildCategorizedOptions = memoize(function(
   props: Props,
   selectValue: OptionsType
 ): CategorizedGroupOrOption[] {
@@ -374,7 +375,18 @@ function buildCategorizedOptions(
     })
     // Flow limitation (see https://github.com/facebook/flow/issues/1414)
     .filter(categorizedOption => !!categorizedOption): any[]);
-}
+}, function([newProps, newSelectValue], [lastProps, lastSelectValue]) {
+  // Shallow compare props
+  // ignore menuIsOpen, as it's not used by sub-functions
+  // use find instead of filter(!"menuIsOpen").map(equal).includes(false)
+  // to traverse the properties as few times as possible
+  let propsEqual = !Object.entries(newProps).find(([key, value]) => key !== 'menuIsOpen' && value !== lastProps[key]);
+
+  // Sometimes we recieve a new instance of an empty array, check for that
+  let selectValueEqual = newSelectValue === lastSelectValue || (newSelectValue.length === 0 && lastSelectValue.length === 0);
+
+  return propsEqual && selectValueEqual;
+});
 
 function buildFocusableOptionsFromCategorizedOptions(
   categorizedOptions: CategorizedGroupOrOption[]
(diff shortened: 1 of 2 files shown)


packages/react-select/src/Select.js
```diff
@@ -374,7 +375,18 @@ function buildCategorizedOptions(
     })
     // Flow limitation (see https://github.com/facebook/flow/issues/1414)
     .filter(categorizedOption => !!categorizedOption): any[]);
-}
+}, function([newProps, newSelectValue], [lastProps, lastSelectValue]) {
+  // Shallow compare props
+  // ignore menuIsOpen, as it's not used by sub-functions
+  // use find instead of filter(!"menuIsOpen").map(equal).includes(false)
+  // to traverse the properties as few times as possible
+  let propsEqual = !Object.entries(newProps).find(([key, value]) => key !== 'menuIsOpen' && value !== lastProps[key]);
```
h3rmanj, line 383: Optimally I would only compare the props that are needed by this function, but the whole prop object is passed to sub-functions, so it would be to hard to keep track of.

I don't check `menuIsOpen` because it's not used here and is changing when I need memoization of this function.

packages/react-select/src/Select.js
```diff
@@ -341,40 +342,59 @@ function toCategorizedOption(
...
+      .filter(categorizedOption => !!categorizedOption): any[]);
+  },
+  function(newArgs: any, lastArgs: any) {
+    const [newProps, newSelectValue] = (newArgs: [Props, OptionsType]);
+    const [lastProps, lastSelectValue] = (lastArgs: [Props, OptionsType]);
+
+    // Shallow compare props
+    // ignore menuIsOpen, as it's not used by sub-functions
+    for (let key in newProps) {
+      if (key === 'menuIsOpen') continue;
```
h3rmanj, line 384: Optimally I would only compare the props that are needed by this function, but the whole prop object is passed to sub-functions, so it would be to hard to keep track of.

I don't check `menuIsOpen` because it's not used here and is changing when I need memoization of this function.

packages/react-select/src/__tests__/Select.test.js
```diff
@@ -2765,3 +2765,17 @@ test('renders with custom theme', () => {
     window.getComputedStyle(firstOption).getPropertyValue('background-color')
   ).toEqual(primary);
 });
+
+test('filterOption only called once per option when opening menu', () => {
```
h3rmanj, line 2769: I'm not sure how I would go about testing the memoization of buildCategorizedOptions. So I test how many times filterOption gets called when opening menu. Don't know if this is necessary though.

Methuselah96:
I'm not sure that we can memoize this function reliably. Specifically in the case where there are multiple `Select` components on the page, they would all share the same memoized function so there could be cases where it would have the props from a previous `Select` cached and call `filterOption` more times than necessary.

h3rmanj:
The goal of this PR is to memoize it reliably per render, not between. As you only interact with one select at a time, that shouldn't be an issue;

https://codesandbox.io/s/react-codesandboxer-example-forked-tr1dk

h3rmanj:
Closing because of the major rewrite in progress.

joshhunt:
@Methuselah96 @h3rmanj How do you feel about me picking this back up soon? We're trying to [improve performance of react-select](https://github.com/grafana/grafana/issues/57675) in Grafana, and this seems like a low-hanging fruit to target?

Methuselah96:
@joshhunt Thanks for the ping, I'm working on a PR to memoize `buildCategorizedOptions`. Should be ready soon.

Methuselah96:
Hopefully https://github.com/JedWatson/react-select/pull/5450 will help a bit.

PR #4515, opened 2021-03-22
Add resultLimit prop limiting the amount of rendered options

h3rmanj:
Add a `resultLimit` prop that prevents rendering too many results. Useful if the number of options are in the thousands.

Indirectly addresses #4504  
Addresses #4516

### DIFF
--- a/packages/react-select/src/Select.js
+++ b/packages/react-select/src/Select.js
@@ -1653,34 +1653,36 @@ export default class Select extends Component<Props, State> {
     let menuUI;
 
     if (this.hasOptions()) {
-      menuUI = this.getCategorizedOptions().slice(0, this.props.resultLimit || undefined).map(item => {
-        if (item.type === 'group') {
-          const { data, options, index: groupIndex } = item;
-          const groupId = `${this.getElementId('group')}-${groupIndex}`;
-          const headingId = `${groupId}-heading`;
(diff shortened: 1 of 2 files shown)


Methuselah96:
Could you filter the options before passing them into the Select component?

h3rmanj:
> 
> 
> Could you filter the options before passing them into the Select component?

I could, yeah. But it's incredibly frustrating that I'd have to hoist and manage input state and search logic, which Select already has, just because i want to limit the amount of options that are rendered. This feature would be really helpful to have available, when dealing with thousands of options, which is often the case.

ThomDevine:
I cannot filter options before passing them into react-select, because some of our lists have big numbers (1200+ is slow in IE11) and I want to make use of the typeahead filter.

I tried using a custom MenuList component to render a single "Too many results - keep typing" option until the list is under the limit.  But in MenuItem, props.children already exists and altering the props.options list does nothing.

If there is a higher-level component that has "the list of filtered options" and does NOT already have "the rendered list of options as props.children", that will work.  I have not found this component in the docs though.  If someone can tell me here if that component exists and what it is, it will solve the "max options" issue without an API change.

If it exists, it's probably worth adding to the Examples page.

brotherbarz (approved):

h3rmanj:
Closing because of the major rewrite in progress.

PR #4528, opened 2021-03-27
Anshuk/accessbility and Performance optimized

Annshuk:
added ul as menu container and Li as list options, also specify the role, so it can be accessible with Jaws, Component Input is memorization now, it it rendering whenever doing the focus in, please review and this is my first time to contribute in react-select so any review feedback would be appreciated


docs/examples/CustomSingleValue.js
```diff
@@ -15,16 +15,6 @@ export default class CustomControl extends Component<*, State> {
...
-        styles={{
-          singleValue: base => ({
-            ...base,
-            padding: 5,
-            borderRadius: 5,
-            background: colourOptions[2].color,
-            color: 'white',
-            display: 'flex',
-          }),
-        }}
```
Rall3n, line 27: Why were these styles removed? They visually convey what was changed.

packages/react-select/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-select",
-  "version": "4.3.0",
```
Rall3n, line 3: Version should not be changed by hand. This is handled through the release process.

packages/react-select/src/components/Input.js
```diff
@@ -46,16 +46,16 @@ const Input = (props: InputProps) => {
   );
 
   return (
-    <div css={getStyles('input', props)}>
+    <label css={getStyles('input', props)}>
```
Rall3n, line 49: What purpose does the use of the `label` tag serve here?

packages/react-select/src/components/Menu.js
```diff
@@ -383,9 +386,13 @@ export const MenuList = (props: MenuListComponentProps) => {
     innerProps,
     innerRef,
     isMulti,
+    selectProps:{ menuIsOpen }
   } = props;
   return (
-    <div
+    <ul
```
Rall3n, line 392: I don't think we should change the tags to `<ul>` and `<li>`. They are no requirement for a list of selectable items. `role` should be sufficient enough for this.

packages/react-select/src/components/Option.js
```diff
@@ -80,10 +80,15 @@ const Option = (props: OptionProps) => {
...
+    type,
     innerProps,
   } = props;
+
   return (
-    <div
+    <li
+      role={type}
+      aria-disabled = {isDisabled}
+      aria-selected={isFocused}
```
Rall3n, line 91: `isFocused` does not equal a selected option. This should be `isSelected`.

Furthermore, if the option is disabled, this attribute should somehow be omitted:

> If an option is not selectable, omit the aria-selected.
> _[MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role#associated_aria_roles_states_and_properties)_

Rall3n (changes requested):
Some questions, some changes.

-----

Other than this revision, I believe that the "accessibility" and "performance optimization" topics should be split into separate PRs.

packages/react-select/src/components/Option.js
ebonow: @Rall3n regarding the `aria-disabled` attribute, I believe we want this because the options are not "truly" disabled and we allow the user to click on disabled options so that the information is relayed back to the aria-live region. Let me know if I am missing something.

dcousens:
Marking as stale, @Annshuk if you want to update your pull request from the feedback provided, please re-open :yellow_heart:

PR #4542, opened 2021-04-16
#4146 Add menu position recalculation after async data is loaded

### LINKED ISSUE #4146: Async menu list height remains unchanged after options loaded
I've set a `maxMenuHeight` on an `Async` component, but after the options are loaded, the max height is not calculated and instead is set to the `maxMenuHeight` that is provided. This causes the menu to be cut off by the bottom of the screen and not flip to the top of the select when enough options are returned. I would expect the menu to not flow beyond the window after the options are loaded and still not flip to the top if `minMenuHeight` has not been reached either.

I was able to write a hack around this by adding a custom component for `Menu` and forcing `innerRef` to be called _(so that `getPlacement` on `MenuPlacer` will be triggered and the max height recalculated)_ whenever the supplied options have changed _(using an MD5 hash for this to ensure that the options truly have changed)_.

**Version:** 3.1.0

**Test Case:**
```jsx
import ReactSelectAsync from "react-select/async";

export default function TestCase() {
  return (
    <ReactSelectAsync
      maxMenuHeight={800}
      loadOptions={() => new Promise((resolve) => {
        setTimeout(function () {
          const ret = [];

          for (let i = 0; i < 50; i++) {
            ret.push({
              value: i,
              label: `Item ${i}`
            });
          }

          resolve(ret);
        }, 1000);
      }}
    />
  );
}
```

**My Workaround:**
```jsx
import { components } from "react-select";
import { MD5 } from "object-hash";

export class CustomMenu extends React.Component {
  _isFirstRender = true;
  _menuRef = React.createRef();

  render() {
    const { children, ...props } = this.props;

    return (
      <components.Menu {...props} innerRef={this._innerRef}>
        {children}
      </components.Menu>
    );
  }

  componentDidUpdate() {
    const optionsHash = MD5(this.props.options);

    if (this._prevOptionsHash !== optionsHash) {
      this._prevOptionsHash = optionsHash;

      if (!this._isFirstRender && this._menuRef.current != null) {
        this.props.innerRef(this._menuRef.current);
      }
    }

    this._isFirstRender = false;
  }

  _innerRef = (...args) => {
    processDomRef(this._menuRef, ...args);
    this.props.innerRef(...args);
  };
}
```

Wunsz:
Fixes #4146 by calling `getPlacement` function in `MenuPlacer` once async call is finished (both promise and callback approach).

### DIFF
--- a/packages/react-select/src/Async.js
+++ b/packages/react-select/src/Async.js
@@ -32,6 +32,9 @@ export type AsyncProps = {
   /* Will cause the select to be displayed in the loading state, even if the
      Async select is not currently waiting for loadOptions to resolve */
   isLoading: boolean,
+  /* Will prevent menu position and size recalculation on async data
+     load finish */
+  preventMenuPositionRecalculation: boolean,
 };
 
 export type Props = SelectProps & AsyncProps;
@@ -41,6 +44,7 @@ export const defaultProps = {
   defaultOptions: false,
   filterOption: null,
   isLoading: false,
+  preventMenuPositionRecalculation: false,
 };
 
 type State = {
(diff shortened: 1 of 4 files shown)


ebonow:
Greetings @Wunsz ,

This looks interesting and thank you for contributing. We are in the process of replacing Flow with TypeScript so to avoid complexities, we are pausing on merging PR's. Once the TypeScript conversion and any other potential breaking changes have been pushed in the major release, we will take a look at this PR along with any other work associated with Menu positioning, sizing, and event handling.

PR #4569, opened 2021-05-06
remove ert from docs and replace with magical-types

gwyneplaine:
This PR removes ERT in favor of magical-types, due to the former being better designed for typescript type definitions. 

In doing so we've introduced a pre-build step that does the following:
* generates a directed graph of typescript types from the root entrypoint of the react-select workspace, as well as additional re-exported types from the docs PropTypes folder (this generates an empty manifest in test environments)
* serializes and chunks the generated graph for consumption at runtime 

We've also introduced a useMagicalNodes hook, which deserializes and lazily loads the chunks at runtime, such that we only pull in the relevant chunks as we need them (reach further into each type).


docs/PropTypes/Async.ts
```diff
@@ -1,10 +1 @@
...
-//
-// import { AsyncProps } from 'react-select/src/Async';
-// import { GroupBase, OptionBase } from 'react-select';
-//
-// export default class Select<
-//   Option extends OptionBase,
-//   IsMulti extends boolean,
-//   Group extends GroupBase<Option>
-// > extends Component<AsyncProps<Option, IsMulti, Group>> {}
+export { AsyncAdditionalProps } from 'react-select/src/useAsync';
```
gwyneplaine, line 1: magical-types resolves the type signature down through interface extensions. 
We only pull in AsyncAdditionalProps here to make the docs experience nicer.

docs/PropTypes/Creatable.ts
```diff
@@ -1,10 +1 @@
...
-//
-// import { CreatableProps } from 'react-select/src/Creatable';
-// import { GroupBase, OptionBase } from 'react-select';
-//
-// export default class Select<
-//   Option extends OptionBase,
-//   IsMulti extends boolean,
-//   Group extends GroupBase<Option>
-// > extends Component<CreatableProps<Option, IsMulti, Group>> {}
+export { CreatableAdditionalProps } from 'react-select/src/useCreatable';
```
gwyneplaine, line 1: magical-types resolves the type signature down through interface extensions. 
We only pull in CreatableAdditionalProps here to make the docs experience nicer.

docs/pages/props/index.tsx
```diff
@@ -93,17 +88,9 @@ export default function Api() {
...
-            <Prop
-              {...props}
-              shapeComponent={() => null}
-              type="All Components Object"
-            />
-          ),
-        }}
+      <ShowTypes
+        getNode={getNode}
+        index={selectTypes['SelectInstance'].index}
```
mitchellhamilton, line 93: This should be the select props, right? it looks like it's displaying the select instance though? (and same for the one above)

docs/generate-magical-types/src/generate.ts
```diff
@@ -0,0 +1,111 @@
...
+  item,
+  pkgExports,
+}: {
+  sourceFile: any;
+  item: string;
+  pkgExports: MagicalNodesForPackage;
+}) => {
+  let exportedDeclarations = sourceFile.getExportedDeclarations();
+  for (const [exportName, declaration] of exportedDeclarations) {
+    if (item === 'icon' && exportName !== 'IconProps') continue;
```
mitchellhamilton, line 48:
```suggestion
```

docs/pages/props/index.tsx
gwyneplaine: yep, good catch, changed both

docs/pages/props/index.tsx
```diff
@@ -83,51 +81,37 @@ export default function Api() {
     please see the [controlled props](/advanced#controlled-props) section of the advanced page.
 
     ${(
-      <Props
-        props={require('!!extract-react-types-loader!../../PropTypes/stateManager')}
+      <ShowTypes
+        getNode={getNode}
+        type={stateManagerTypes?.StateManagerProps}
```
mitchellhamilton, line 86: When I look at this locally, I can't see the state manager props here

JedWatson (approved):
Awesome, thanks @gwyneplaine 👏

Issue #4559, opened 2021-04-29
Releasing v5 [TypeScript] Umbrella


JedWatson:
As originally discussed in #4142 we've rewritten react-select in TypeScript (🙌  **huge** props to @Methuselah96 for the heavy lifting here!)

This issue is to track the work involved in merging the TypeScript conversion (#4489), and getting a new major version (v5) released.

## Steps

- [x] Disable the website auto-publish on updates to master @bladey 
- [x] Merge #4489 @JedWatson 
  - [x] Fix `__isNew__` handling
  - [x] Merge https://github.com/JedWatson/react-select/pull/4702
- [x] Cleanup tasks (because we'll be bumping a major version)
  - [x] Switch to CSS Grid over `react-autosize-input` @ebonow PR https://github.com/JedWatson/react-select/pull/4625
  - [x] Update some styling and CSS decisions (including closing out #4720) @seancurtis
- [x] Update the website to use magical types @gwyneplaine in a subsequent PR
  - [x] look at what we can do to reduce the anticipated bundle size increase @mitchellhamilton 
- [x] Write the changelog and upgrade notes #4762 @Methuselah96 
  - [ ] Add a banner to the website
  - [ ] Call out updated Browser Support (Evergreen, officially not IE11)
- [ ] Release version 5 🎉

Bosch-Eli-Black:
Any chance of publishing a beta build to NPM? 🙂

ebonow:
Everyone, please note that this major release will be a complete rewrite so we officially halted PR's until release so we can focus on addressing on breaking changes and documentation. It is absolutely inexcusable that so many PR's have not been addressed but also know that quite a bit focus around Menu positioning, scrolling and sizing behavior are being looked at holistically and may be resolved in batch as we continue to experiment with Popper as a solution.

Methuselah96:
We've released a `5.0.0-beta.0`.

nicoburns:
This new update looks great (especially the TypeScript support). However dropping IE11 support unfortunately makes it a non-starter for us. I was wondering whether that decision could be reconsidered. I see that the main reason for dropping IE11 support seems to be lack of support for CSS grid, but IE11 does partially support CSS grid so I am wondering if it would be possible to get the auto-sizing grid hack to work with it.

@ebonow Did you attempt to get the grid hack to work with IE11? Is there a specific reason why it won't work?

yournatalita:
[...]

PR #4573, opened 2021-05-07
Add meaningful CSS class to AutosizeInput wrapper div

trumbitta:
This little div:

```html
<div
  class="select__value-container select__value-container--has-value css-g1d714-ValueContainer"
>
  <div class="select__single-value css-1uccc91-singleValue">Ocean</div>

  <!-- This one right here below -->
  <div class="css-b8ldur-Input">

    <div class="select__input" style="display: inline-block">
      <input [...] />
      [...]
    </div>
  </div>
</div>
```

bears some CSS rules that would be much easier to customize (via class name / styled-components) if it had a proper, meaningful, fixed class name.

This little PR adds that class name:

```html
<div
  class="select__value-container select__value-container--has-value css-g1d714-ValueContainer"
>
  <div class="select__single-value css-1uccc91-singleValue">Ocean</div>

  <!-- This one right here below -->
  <div class="css-b8ldur-Input select__input-wrapper">

    <div class="select__input" style="display: inline-block">
      <input [...] />
      [...]
    </div>
  </div>
</div>
```

Without it, I'm having to do this:

```css
    /*
      This is the parent of the div that hides the real HTML input element.
      The problem is that this element causes unwanted margin, height, padding but
      it has no proper name / class while its direct child has a proper name / class but at the time
      of writing we can't target a parent by its child in CSS (https://caniuse.com/css-has)

      So, I guess I'll just 🤞
    */
    div:nth-child(2) {
      margin: 0;
      padding: 0;
    }
```

but I would rather not :D

### DIFF
--- a/packages/react-select/src/__tests__/__snapshots__/Async.test.tsx.snap
+++ b/packages/react-select/src/__tests__/__snapshots__/Async.test.tsx.snap
@@ -172,7 +172,7 @@ exports[`defaults - snapshot 1`] = `
           Select...
         </div>
         <div
-          class="emotion-5"
+          class=" emotion-5"
         >
           <div
             class=""
--- a/packages/react-select/src/__tests__/__snapshots__/AsyncCreatable.test.tsx.snap
+++ b/packages/react-select/src/__tests__/__snapshots__/AsyncCreatable.test.tsx.snap
@@ -172,7 +172,7 @@ exports[`defaults - snapshot 1`] = `
           Select...
         </div>
         <div
-          class="emotion-5"
+          class=" emotion-5"
         >
           <div
             class=""
--- a/packages/react-select/src/__tests__/__snapshots__/Creatable.test.tsx.snap
+++ b/packages/react-select/src/__tests__/__snapshots__/Creatable.test.tsx.snap
@@ -172,7 +172,7 @@ exports[`defaults - snapshot 1`] = `
           Select...
         </div>
         <div
-          class="emotion-5"
+          class=" emotion-5"
         >
           <div
             class=""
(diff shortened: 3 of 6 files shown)


JedWatson:
Hey @trumbitta thanks for this PR!

We're currently working through the updates we're making for 5.0 (see #4559) and have two thoughts on this:

1. We are looking into replacing `react-input-autosize` with a css-only solution now that we have dropped IE11 support, which would clean this whole part of the component up considerably. If we do that, we **may** be able to remove one of these two wrapping divs entirely 
2. If we do keep both wrapping divs, I'd be very happy to merge this PR, but for consistency with other wrapper class names we should change it to `input-container` and not `input-wrapper`

Will update you shortly with how this is going

trumbitta:
Hey @JedWatson thanks for the heads-up!

While you decide about the best course of action, I thought I'd better help by doing what I could and  so I updated the name to `input-container` and rebased the branch to get the latest updates from master in.

ebonow:
@trumbitta just wanted to say thanks for this PR. 

One of the goals moving into a new major version was to identify any breaking changes and autosizeinput was one of those long overdue changes. Much of the Input component was thus overhauled as a part of this PR. https://github.com/JedWatson/react-select/pull/4625

The impact of the referenced PR is that it duplicates this functionality.

Some other things to note:
- The DOM structure was simplified so the Input component is only one div wrapping an input
- inputClassName was added as a prop to allow more control in adding styles to the input
- `.prefix__input-container` refers to the div wrapper whereas `.prefix__input` will now refer to the input element. This will likely cause some initial confusion, but feel it's much more consistent with the styles.

That all said, I agree with the intention of this PR, but will close this as this functionality will is included as a part of another PR.

trumbitta:
Hey, thanks for the heads-up!

I'm excited for the upcoming release, but at the same time I don't know when I will be able to use it since a sizable slice of our users unfortunately still uses IE11.

Is there any chance you'll release a patch version with this PR before going forward with the next major? 🤞🤞🤞

PR #4583, opened 2021-05-13
Prevent ArrowLeft & ArrowRight key action when controlShouldRenderValue=false

kidokidozh:
**Problem**
We are using the select component with `controlShouldRenderValue = false`. And we have defined event handler for `ArrowLeft & ArrowRight` key on the container element. But the event handler here does not return, and go to `event.preventDefault()`, which caused our event handler not work.

**Solution**
Just return when `controlShouldRenderValue = false`, since it does not make sense to `focusValue` when no value rendered.

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -1367,6 +1367,7 @@ export default class Select<
       onKeyDown,
       tabSelectsValue,
       openMenuOnFocus,
+      controlShouldRenderValue,
     } = this.props;
     const { focusedOption, focusedValue, selectValue } = this.state;
 
@@ -1383,11 +1384,11 @@ export default class Select<
     this.blockOptionHover = true;
     switch (event.key) {
       case 'ArrowLeft':
-        if (!isMulti || inputValue) return;
+        if (!isMulti || inputValue || !controlShouldRenderValue) return;
         this.focusValue('previous');
         break;
       case 'ArrowRight':
-        if (!isMulti || inputValue) return;
+        if (!isMulti || inputValue || !controlShouldRenderValue) return;
         this.focusValue('next');
         break;
       case 'Delete':


kidokidozh:
@JedWatson Could you review this? Thanks

Rall3n:
Hello @kidokidozh,

thank you for your contribution. We are currently in the middle of our TypeScript/v5 beta release. Therefore we are currently not merging any PR until this is finished.

But I had a quick look at your proposed changes and I think that these changes are not necessary.
I don't know how you applied your event handler, but the `Select` component provides a prop named `onKeyDown` with which you can overwrite the handling of certain or all key events which are handled by the component.

```jsx
<Select
    onKeyDown={e => {
        // Skip if not keys we want
        if (!['ArrowLeft', 'ArrowRight'].includes(e.key)) return;
        
        // Need to prevent or default handling runs after this.
        e.preventDefault();
        
        // Do your code
    }}
/>
```
Based on what you wrote in your first comment, this should be sufficient. If not, then please explain.

kidokidozh:
Hi @JedWatson , thanks for your reply, sorry I didn't explain the problem clearly.

We are using Select inside a table like this. 
image

For each row, the expect behavior is pressing `ArrowLeft/ArrowRight` will move the cursor to next column's input, e.g. from "Name" input to "Type" select. So we add an event handler on the row element.
```js
<tr onKeyDown={e => {
    if (['ArrowLeft', 'ArrowRight'].includes(e.key)) { //focus to next column's input }
}}>
    <td><input /></td>
    <td><Select /></td>
    <td><Select /></td>
</tr>
```

The problem is Select will catch the Arrow key event, and prevent default in this line. This causes our row event handler not working.
https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.tsx#L1489

My understanding is that, since the `ArrowLeft/ArrowRight` event defined here is used to focus to previous/next value under multi value mode, it's reasonable to add a return condition when `controlShouldRenderValue=false`. Because when it's false the default multi value component is not rendered, it makes no sense to focus previous or next value.
https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.tsx#L1385
```js
case 'ArrowLeft':
    if (!isMulti || inputValue || !controlShouldRenderValue) return;
    this.focusValue('previous');
    break;
```

BGDSHKA (approved):

bbsimonbb:
preventDefault and stopping propagation are very aggressive manoeuvres. In my code, I add a `history` string array to all events, Then any function that handles an event leaves its trace, and downstream handlers can examine what action has already been taken on the event. It's such a good idea, I'm happy to share it.

```
export interface IHasHistory {
    history: string[];
}

export function withHistory<TargetShape>(target: any): TargetShape & IHasHistory {
    if (!target.history) {
        (target as IHasHistory).history = [];
    }
    return target;
}
```

then, any time you call a handler function...

`onclick={(e)=>myHandler(withHistory(e))}`

and

```
function myHandler(e:ClickEvent & IHasHistory){
   if(e.history.length === 0){ // or whatever test you like
      ...
      e.history.push("myHandler")
   }
}
```

bbsimonbb:
@ebonow how can I modify your sandbox such that the right arrow key saves the currently focussed option in the react-select please ?

Issue #3111, opened 2018-10-10
Feature Request: Filter From the MenuList


cutterbl:
We're migrating an application from Angular to React. This is a gradual, incremental replacement, so it's critical that the user experience does not change. We decided to go with react-select as our replacement for UI-Select. So I'm tasked with creating the implementation.

I've been able to replicate nearly all functionality so far, except some final styling and function on the control.

I am attempting to separate the inputValue filter from the root Control and move it into the MenuList. I realize that I can override both of these Components to customize this functionality. The first hurdle I've encountered is that I cannot add an Input control (any input control) to the MenuList and give it focus. Any 'click' action just bounces off. 

I'm looking to create something sort of similar to what is seen in the Experimental popup example, while keeping my ValueContainer in the core control. Anyone have direction on how to go about this?

cutterbl:
![screen shot 2018-10-11 at 11 44 18 am](https://user-images.githubusercontent.com/666987/46870193-8c561980-cdf3-11e8-86b3-f8aec6bc5dfb.png)

This is what it is I'm trying to accomplish. As you can see in the shot, I've been able to manage most everything, with the exception of the 'search'. Since the focus is retained on the Control, all key presses are retained by the Control itself.

chiplay:
I'm interested in something similar, any luck on this one?

dan-kez:
Would definitely love to have this feature!

There is also a stack overflow thread about this that I made about a year ago.

https://stackoverflow.com/questions/48930622/react-select-show-search-bar-in-dropdown?noredirect=1#comment96510392_48930622

Rall3n:
Here is a [working example](https://codesandbox.io/s/m75wlyx3oy) using the [Components Framework](https://react-select.com/components).

I had to handle `isFocused` and `menuIsOpen` on my own by using a wrapper and checking, if a click in the document was inside the wrapper.

Also I had to prevent `onMouseDown` and `onTouchDown` to keep the events from bubbling to keep the menu open when clicking the input.

cutterbl:
@Rall3n Great stuff! I'll try to integrate some of what you've done with my own example, and see how far I can get...

bladey:
Hello -

In an effort to sustain the `react-select` project going forward, we're closing old issues.

We understand this might be inconvenient but in the best interest of supporting the broader community we have to direct our efforts towards the current major version.

If you aren't using the latest version of `react-select` please consider upgrading to see if it resolves any issues you're having.

[...]

PR #4590, opened 2021-05-18
Add `searchInMenu` property

moelders:
Hello, I open this PR because I needed to display the search input as part of the menu instead of as part of the control.

You can see a sample in the samples file:

image

Sample code:

```html
<Select
  defaultValue={[colourOptions[2]]}
  isMulti
  searchInMenu
  name="colors"
  options={colourOptions}
/>
```

I was searching for workarounds to do this without modifying the package source code before doing the changes, but, unfortunately, the available workarounds did not work well.

I hope you can analyze and validate the changes.

Thank you, regards.

### DIFF
--- a/docs/examples/SearchInMenu.tsx
+++ b/docs/examples/SearchInMenu.tsx
@@ -1,0 +1,14 @@
+import React from 'react';
+
+import Select from 'react-select';
+import { colourOptions } from '../data';
+
+export default () => (
+  <Select
+    defaultValue={[colourOptions[2]]}
+    isMulti
+    searchInMenu
+    name="colors"
+    options={colourOptions}
+  />
+);
--- a/docs/examples/index.tsx
+++ b/docs/examples/index.tsx
@@ -42,6 +42,7 @@ export { default as CustomIsOptionDisabled } from './CustomIsOptionDisabled';
 export { default as DefaultOptions } from './DefaultOptions';
 export { default as Experimental } from './Experimental';
 export { default as FixedOptions } from './FixedOptions';
+export { default as SearchInMenu } from './SearchInMenu';
 export { default as MultiSelectSort } from './MultiSelectSort';
 export { default as Popout } from './Popout';
 export { default as StyledMulti } from './StyledMulti';
--- a/docs/pages/home/index.tsx
+++ b/docs/pages/home/index.tsx
@@ -10,6 +10,7 @@ import {
   BasicMulti,
   CreatableSingle,
   FixedOptions,
+  SearchInMenu,
   StyledMulti,
   StyledSingle,
 } from '../../examples';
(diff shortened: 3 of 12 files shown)


Rall3n:
Hello @moelders,

thank you for your contribution.

Unfortunately this has already been done with the configuration options provided by this library.
You should check out #3111 (in particular [this comment](https://github.com/JedWatson/react-select/issues/3111#issuecomment-470911304) in particular) for a solution.

This does not mean that your PR is invalid, but that the same is achievable with the tools provided.

moelders:
Thank you for the answer @Rall3n.

I already knew that sample and I tried to work with it. It has some issues with the focus, but it mostly works well.

However, that solution should be implemented whenever you want to have that feature, so, my proposal is to make it standard for everyone.

moelders:
I tested the solution that you proposed one more time, and it has some issues:

- The search input in the menu is not focused when you open the dropdown. The original does it by default.
- When the search input in the menu is focused the dropdown looses the focus.

moelders:
Hello again @Rall3n,

Can you tell me if the search-in-menu feature will be considered in the future?

I am not saying that my implementation will be accepted, but I would like to know it so I can adapt my application accordingly.

Thank you.

ebonow:
Greeting @moelders we are behind on PR's (as always) but working towards releasing version 5.0 which is almost an entire rewrite based on converting to TypeScript. As such, we are putting all PR's are on hold until we can get that released (barring some urgent fix).

Regarding this particular PR, while we haven't met as a group to discuss, I'd likely be in favor of leaving this kind of functionality up to the user as opposed to implementing this as a built-in solution. My rationale is that while some users may find this helpful, it's possible that other users may want additional customizations on top of this (such as indicators in the input). What starts simply as a "doThis" boolean prop then becomes a list of new "doThisWithThis" and "doThisWithThat" all for what should be achievable with custom components.

Therefore, I would be more interested in looking into the root causes that prevent this from being easily implemented with custom components.

ebonow:
@moelders 

I looked at @Rall3n 's example and added the ability to autoFocus on the input when the menu is opened.
Code example [here](https://codesandbox.io/s/react-select-input-in-menulist-forked-29f4e?file=/example.js)

Regarding this...
> When the search input in the menu is focused the dropdown looses the focus.

I am unsure of what type of behavior you are expecting. Are you saying that both the input and the Select should be focused at the same time?

moelders:
Thank you for the reply @ebonow. I understand your point of view, and I respect it.

If in the next version you can provide a good way to have this, it will be very welcome.

Regarding the focus, what I said, is that right now, you click on the dropdown and it gains the focus on it. With my changes, when you click the dropdown, the focus is set in the text input in the menu, just one input focused.

Thank you again, regards.

PR #4598, opened 2021-05-21
Fix for #3945 - Problem when you hit Enter, it submits the forms

ankurmittal092:


ankurmittal092:
@ebonow Please review this..

PR #4609, opened 2021-05-28
Update README to clarify need for @types for v4 and not for v5

ngault:
Confusion about Typescript arises because the HEAD of the current master branch, although still numbered v4.x for the npm-published module, still shoes the Typescript code to be releases as v5.

Unfamiliar users might assume that, variously, v4.x is written in Typescript (in fact, it's in JavaScript) or that the typings in this code work for v4.x (they don't, as they're not compatible with the required `@types` version.

These 2 statements should clear up confusion.


JedWatson (approved):
Thanks @ngault!

Issue #4088, opened 2020-06-15
Dropdown remains fixed in place and doesn't scroll with parent container


bladey:
_This issue is a round-up of multiple past issues documenting the same bug. Feel free to check out the linked issues below for more information. This issue is the source of truth going forward to investigate the issue, report findings, and implement a bug fix._

**Issue:** 

![Screen Capture on 2020-06-15 at 14-43-03](https://user-images.githubusercontent.com/737821/84618854-c061d600-af16-11ea-94a0-2448f5470821.gif)

Dropdown remains fixed in place and doesn't scroll with parent container.

The commit responsible for this issue is located here - https://github.com/JedWatson/react-select/commit/691a01171969da89e20097f9d603651bb18a8510

Version `v3.0.4` does not have this bug however the latest release `v3.1.0` does.

**Issues:**

https://github.com/JedWatson/react-select/issues/4554
https://github.com/JedWatson/react-select/issues/4020
https://github.com/JedWatson/react-select/issues/4052
https://github.com/JedWatson/react-select/issues/3734
https://github.com/JedWatson/react-select/issues/4085
https://github.com/JedWatson/react-select/issues/3929
https://github.com/JedWatson/react-select/issues/3349
https://github.com/JedWatson/react-select/issues/3646

**Examples:**

https://codesandbox.io/s/react-select-v3-sandbox-forked-lgkms
https://codesandbox.io/s/compassionate-cookies-85j2n
https://codesandbox.io/s/react-select-v3-sandbox-kr5wf
https://codesandbox.io/s/festive-dirac-izf5y
https://codesandbox.io/s/react-codesandboxer-example-xy47y

If anyone would like to share any further information, please do so.

mjb95:
Hi - for info, I have explicitly tested this by installing v3.0.4 and v3.0.3, and I experience the same issue in both versions.  Is it possible that the bug existed even before the commit that you have highlighted above?

palashkaria:
I think @mjb95 is right, as this is the behaviour of React Select in v2 also; menuPosition fixed & portal, both do not update the positioning of the Menu on scroll - so the menu is stuck. All the sanboxes here have menuPosition fixed, and portal, and nothing related to it was changed in that commit.

The bug mentioned by @bladey is related to a _workaround_ (`menuShouldBlockScroll`) to handle the issue of the menu not scrolling - as mentioned in this comment https://github.com/JedWatson/react-select/issues/3929#issuecomment-604866966.


[...]

PR #4616, opened 2021-06-01
[WIP] Use popper for menu positioning

Methuselah96:
Looks into resolving https://github.com/JedWatson/react-select/issues/4088 and many other issues related to menu positioning. To test this out you can go to the `cypress-tests` route on the website where there are a couple examples of portalling.

Things that currently work:
- Menu position updating when scrolling while using portalling
- Menu position updating when resizing window
- Bottom/top/auto placement works as before
- You can now close the menu when scroll is blocked (but you can scroll the smaller divs for some reason)

Things that don't currently work:
- Automatically scrolling the menu into view
- Using a constrained height for the menu if necessary and taking the min height into account
- Scroll blocking doesn't seem to be correctly blocking scrolling in the smaller div (in the portalling case)


ebonow:
It seems there are possibly some things that can be looked at regarding modifiers.
https://popper.js.org/docs/v2/modifiers/community-modifiers/

There is a link to a `maxSize` modifier. Playing with the codepen there, it is still possible to add a min-height and max-height to the #popper selector. 

Here's a link to my fork of the codepen, where you can remove the max and min heights and have it behave as expected.
https://codesandbox.io/s/great-frost-ucqgm?file=/src/styles.css:205-226

Methuselah96:
Yeah, for sure, I was thinking the same thing regarding max and min heights, just hadn't gotten around to implementing it. I can guess the question is whether scrolling can be accomplished using a modifier. I haven't looked much into our scrolling code, but that would be the last thing that would have to be figured out for us to switch in my opinion.

PR #4625, opened 2021-06-04
Remove AutoSizeInput

ebonow:
⚠️ **Issue**: 
`react-autosize-input` is a mono-repo that is maintained specifically for react-select. Not only does this add extra complexity and code weight, but it is also at the root of several issues. Additionally, it adds one extra layer of DOM complexity for users to access the DOM input.

✅  **Solution**:
This proposal is to completely remove the AutoSizeInput component and replace it with a 100% pure CSS solution. Heavily inspired by this [codepen example by Shaw](https://codepen.io/shshaw/pen/c72208b279d750f1ab46b6a45bbd04d7) found via this [CSS Tricks article](https://css-tricks.com/auto-growing-inputs-textareas/) 

💪 **How it works**:
The wrapper component now is styled as an inline-grid component with a pseudo :after element. This pseudo element has the content of the `data-value` attribute of the wrapper and occupies the same cell as the input causing it to expand as needed without javascript needing to copy and calculate new widths as the value changes. An early codesandbox example of this can be seen [here](https://codesandbox.io/s/react-select-autosizinginput-dgyy1?file=/src/style.css).

👍🏽  **Features**:
- Less package dependencies and code weight.
- Removed unnecessary nested div element.
- Allows CSS to naturally resize input without relying on 3rd party component ref callback methods
- Fixes CSSTransitions of React Select
- Input will now flex the remaining width of the row (but still wrap to next line when necessary)
  - This creates a more consistent UX for users attempting to right click or long press to paste
- When used with classNamePrefix, classnames are now generated for both the container (`prefix__input-container`) and input (`prefix__input`) 
- `inputClassName` was added as a new prop to the Input component to make it easy to style components via classNames (hello Tailwind)

👎🏽  **Breaking Changes**:
- IE11 no longer works as it does not fully support CSS grid
- Renaming the `.prefix__input` now targets the input and NOT the container. Unfortunate but overdue and perhaps opens the door to completely decoupling the input from needing a container if autosizing is not important.


ebonow:
May consider addition suggested in https://github.com/JedWatson/react-select/pull/4513 to remove any boxShadow

ebonow:
Please note that this also addresses another closed issue https://github.com/JedWatson/react-select/pull/4435 by using the shorthand css style `font` to inherit size, family, weight, and line-height.

JedWatson (approved):
🎉  thanks @ebonow!

PR #4634, opened 2021-06-11
Refactor how the DummyInput works when isSearchable={false}

### LINKED ISSUE #3291: iOS Safari "tabbing" doesn't work for non-searchable Selects
iOS Safari displays a nifty little toolbar whenever an input element is focused, which lets you skip directly to the next (or previous) input on the page.

This works as expected for the plain Select:

<details>
<summary>
Working demo
</summary>

demo of working navigation to searchable select

</details>

<br />

Setting `isSearchable={false}` has the side effect of making the `<input>` rendered by `<DummyInput>` `readonly`, which [removes it from the tab order in iOS Safari](https://stackoverflow.com/a/9811664):

<details>
<summary>
Broken demo
</summary>

demo of broken navigation to non-searchable select

</details>

<br />

It's still in the tab order in desktop Safari, though, so… that's a thing that happens.

I tried working around this by avoiding the soft keyboard without the `readonly` attribute, but I couldn't do it, hence this issue.

seancurtis:
Fixes #3291

Now that we've dropped IE11 we can use some newer platform features to achieve the same effect, but with a more accessible result.

The `readonly` attribute was added only when `isSearchable={false}` and achieved 2 things from what I've seen:

* hides the flashing cursor
* prevents devices from showing a virtual keyboard

We can solve these by:

* using the newer `caret-color` CSS prop and setting it to `transparent`
* set the `inputMode` attribute on the `<input>` to `none`

Tested on Chrome/Firefox for MacOS, and iOS 14 via the Simulator.

-


ebonow:
[](url)Unfortunately setting inputmode="none" does not appear to have any influence on iOS per my testing on a Simulator running iOS v12.4. Can you confirm otherwise?

Screen Shot 2021-06-11 at 7 52 43 AM

Methuselah96:
Yeah, looks like this is confirmed by [CSS Tricks](https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/#none) as well.

seancurtis:
Updating the PR with my comments from Slack:

I've tested this on my physical device (iOS 14) and also the upcoming iOS 15 version of Safari. It works perfectly in both of these browsers. Both caret-color (https://caniuse.com/css-caret-color) and inputMode (https://caniuse.com/input-inputmode) are also supported in the mobile browsers. However `inputMode` is not currently supported in Firefox or Safari on desktop — but I don't think virtual keyboards are used with those desktop browsers.

ebonow (approved):

JedWatson (approved):
Thanks @seancurtis

PR #4676, opened 2021-07-05
Improve a11y announcements

MatthewCharlton:
There are currently a few issues with the aria-live announcements that this PR aims to solve.

- NVDA will now announce the context text when initially focused
- VoiceOver now announces the context text when re-focusing
- Placeholder text is now announced
- The selected option/s (single and multi) are now announced when being focused
- The deselected option/s (single and multi) are now announced when cleared


packages/react-select/src/components/LiveRegion.tsx
```diff
@@ -77,11 +77,43 @@ const LiveRegion = <
...
+          return opt.reduce(
+            (acc, optionItem: Option) => {
+              const { value: optionVal, label } = optionItem;
+              const sep = acc.label ? ', ' : '';
+              acc.label = acc.label + sep + (label || optionVal);
+              return acc;
+            },
+            // we only need the label for the returned object
+            { label: '' }
+          ) as Option;
```
Methuselah96, line 106: I'm not sure we can create a fake object because it ends up getting passed into `isOptionDisabled` and `getOptionLabel` which is are props that the users can provide.

Methuselah96 (changes requested):

packages/react-select/src/components/LiveRegion.tsx
MatthewCharlton: @Methuselah96 Yes, this is a really good call out that I had not considered. I have refactored the implementation to allow using getOptionLabel function. As a consequence I removed this hacky approach which is much nicer.

packages/react-select/src/types.ts
```diff
@@ -160,14 +161,20 @@ export interface CreateOptionActionMeta<Option extends OptionBase>
...
 
 export type ActionMeta<Option extends OptionBase> =
   | SelectOptionActionMeta<Option>
   | DeselectOptionActionMeta<Option>
   | RemoveValueActionMeta<Option>
   | PopValueActionMeta<Option>
   | ClearActionMeta<Option>
-  | CreateOptionActionMeta<Option>;
+  | CreateOptionActionMeta<Option>
+  | InitialInputFocusedActionMeta<Option>;
```
Methuselah96, line 177: I don't think we should be adding `InitialInputFocusedActionMeta` to `ActionMeta`, because that implies that the `onChange` prop can be called with `InitialInputFocusedActionMeta` for its action meta and that the user needs to handle it. However at the moment `InitialInputFocusedActionMeta` is only used for aria selection.

Methuselah96 (changes requested):

packages/react-select/src/Select.tsx
```diff
@@ -628,13 +630,28 @@ export default class Select<
     this.instancePrefix =
       'react-select-' + (this.props.instanceId || ++instanceId);
     this.state.selectValue = cleanValue(props.value);
+
+    // If `value` or `defaultValue` props are not empty then announce them
+    // when the Select is focused
+    if (this.state.selectValue.length) {
+      this.state.ariaSelection = {
+        value: this.state.selectValue as OnChangeValue<Option, IsMulti>,
```
Methuselah96, line 638: I'm not sure we should be okay with casting this. `selectValue` is an array, but the `value` property in this context is referring to either an array for multi-select or a single value (or `null`) for single-select. I think we probably want to use what we use elsewhere:

```
value: valueTernary(
  this.props.isMulti,
  this.state.selectValue,
  this.state.selectValue[0] || null
);
```

packages/react-select/src/Select.tsx
```diff
@@ -714,6 +739,47 @@ export default class Select<
       scrollIntoView(this.menuListRef, this.focusedOptionRef);
       this.scrollToFocusedOptionOnUpdate = false;
     }
+
+    const setAriaSelectionToNull = () => this.setState({ ariaSelection: null });
+    const setAriaSelectionToSelectedValues = () =>
+      this.setState({
+        ariaSelection: {
+          value: selectValue as OnChangeValue<Option, IsMulti>,
```
Methuselah96, line 747: Also applies here.

Methuselah96 (changes requested):

packages/react-select/src/Select.tsx
```diff
@@ -628,13 +630,28 @@ export default class Select<
...
     props: Props<OptionBase, boolean, GroupBase<OptionBase>>,
     state: State<OptionBase, boolean, GroupBase<OptionBase>>
   ) {
-    const { prevProps, clearFocusValueOnUpdate, inputIsHiddenAfterUpdate } =
-      state;
+    const {
+      prevProps,
+      clearFocusValueOnUpdate,
+      inputIsHiddenAfterUpdate,
+    } = state;
```
Methuselah96, line 654: Different version of Prettier?

Methuselah96 (dismissed):

packages/react-select/src/Select.tsx
MatthewCharlton: Haha, I yeh. I thought I removed the re-formatting changes before pushing.

packages/react-select/src/Select.tsx
MatthewCharlton: I have refactored a bit and am using this pattern now

packages/react-select/src/types.ts
MatthewCharlton: I removed this from the ActionMeta type and used the ```InitialInputFocusedActionMeta``` interface directly in the ```AriaSelection``` type

seancurtis (approved):
Tested OK in VO with Chrome and Safari, NVDA with Edge, and JAWS 21 with Edge

PR #4679, opened 2021-07-06
Add support for aria-invalid & aria-errormessage

dgalarza:


JedWatson (approved):
Thanks @dgalarza! we're focusing on improving accessibility at the moment and this is really helpful.

PR #4699, opened 2021-07-20
feature/always-focus-first-menu-option

andersonmfjr:
Adds a new property called "alwaysFocusFirstMenuOption" that makes it possible to keep focus on the first item when the list of options is changed. An example can be seen below:

- Without prop:
   ezgif com-gif-maker

- With prop:
   ezgif com-gif-maker (1)

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -74,6 +74,8 @@ export interface Props<
   IsMulti extends boolean,
   Group extends GroupBase<Option>
 > {
+  /** Always focus first option on menu list */
+  alwaysFocusFirstMenuOption: boolean;
   /** HTML ID of an element containing an error message related to the input**/
   'aria-errormessage'?: string;
   /** Indicate if the value entered in the field is invalid **/
@@ -268,6 +270,7 @@ export interface Props<
 }
 
 export const defaultProps = {
+  alwaysFocusFirstMenuOption: false,
   'aria-live': 'polite',
   backspaceRemovesValue: true,
   blurInputOnSelect: isTouchCapable(),
@@ -484,8 +487,21 @@ function getNextFocusedOption<
   Option extends OptionBase,
   IsMulti extends boolean,
   Group extends GroupBase<Option>
->(state: State<Option, IsMulti, Group>, options: Options<Option>) {
+>(
+  state: State<Option, IsMulti, Group>,
+  options: Options<Option>,
+  props?: Props<Option, IsMulti, Group>
+) {
   const { focusedOption: lastFocusedOption } = state;
+
+  if (props) {
+    const { alwaysFocusFirstMenuOption } = props;
+
+    if (alwaysFocusFirstMenuOption) {
+      return options[0];
+    }
+  }
+
   return lastFocusedOption && options.indexOf(lastFocusedOption) > -1
     ? lastFocusedOption
     : options[0];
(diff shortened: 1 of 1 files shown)


Annie-Huang:
Is this one similar to the request of https://github.com/JedWatson/react-select/pull/4080?

Is this one going to be merged into master soon?

stnwk:
This PR (open for 1 year) and #4080 (open for 2 years) fix the same thing, that currently cannot be worked around. It would be extremely cool to finally merge this one. The solution is simple and elegant but nobody is merging anything.

Pinging @andrevitalb because I saw recent commit/merge activity from you. Do you think you can help? Thanks! :)

packages/react-select/src/Select.tsx
```diff
@@ -478,8 +481,21 @@ function getNextFocusedOption<
...
+    const { alwaysFocusFirstMenuOption } = props;
+
+    if (alwaysFocusFirstMenuOption) {
+      return options[0];
+    }
+  }
+
   return lastFocusedOption && options.indexOf(lastFocusedOption) > -1
     ? lastFocusedOption
     : options[0];
```
andrevitalb, line 501: Why pass the whole props object? Maybe passing only the relevant prop here would be a wiser choice.

```suggestion
>(
  state: State<Option, IsMulti, Group>,
  options: Options<Option>,
  alwaysFocusFirstMenuOption: boolean
) {
  const { focusedOption: lastFocusedOption } = state;
  return !alwaysFocusFirstMenuOption &&
    lastFocusedOption &&
    options.indexOf(lastFocusedOption) > -1
    ? lastFocusedOption
    : options[0];
```

Propagating suggestion to next code change.

packages/react-select/src/Select.tsx
```diff
@@ -653,7 +669,11 @@ export default class Select<
       const focusedValue = clearFocusValueOnUpdate
         ? getNextFocusedValue(state, selectValue)
         : null;
-      const focusedOption = getNextFocusedOption(state, focusableOptions);
+      const focusedOption = getNextFocusedOption(
+        state,
+        focusableOptions,
+        prevProps
```
andrevitalb, line 675: Ensuring we always get a value, even when the prop doesn't exist

```suggestion
        !!prevProps?.alwaysFocusFirstMenuOption
```

andrevitalb (changes requested):
Functionality-wise, LGTM

andrevitalb:
> This PR (open for 1 year) and #4080 (open for 2 years) fix the same thing, that currently cannot be worked around. It would be extremely cool to finally merge this one. The solution is simple and elegant but nobody is merging anything.
> 
> Pinging @andrevitalb because I saw recent commit/merge activity from you. Do you think you can help? Thanks! :)

I'm just able to add reviews, don't have write access, We still need @JedWatson for that.

stnwk:
Oh, thanks for your time and work anyways! Hopefully we can move forward here

andersonmfjr:
@andrevitalb can you review again? any chance of the functionality being merged?

andrevitalb (approved):
Again, LGTM

andrevitalb:
> @andrevitalb can you review again? any chance of the functionality being merged? thanks for your time

I don't have any write permissions for the repo. We'd have to wait for @JedWatson's approval

dcousens:
@andersonmfjr thanks for the pull request, but I wonder, maybe the more sustainable approach here (for the library) is to have 1 behaviour, and then encourage developers to control the component themselves.

andersonmfjr:
> @andersonmfjr thanks for the pull request, but I wonder, maybe the more sustainable approach here (for the library and maintainers) is to only support 1 behaviour, and then encourage consumers to control the component themselves if they need something else.
> 
> Related
> 
> * [Add focusDefaultOption prop to prevent focusing first option #4080](https://github.com/JedWatson/react-select/pull/4080)

Do you think it is possible to achieve this same behavior with the components control?

In the related PR it only shows that workarounds were done. And apparently they didn't work out so well.

PratikAwaik:
@JedWatson can we please merge this, if possible? Lot of people are requesting for this feature and we don't want to do unnecessarily complex workarounds

Issue #4696, opened 2021-07-19
types too strict OptionBaseType v5


shulcsm:
```tsx
interface Entity {
  id: number;
  name: string;
}

const s = <Select<Entity> getOptionValue={(e) => e.id.toString()} getOptionLabel={(e) => e.name} />;

>>> Type 'Entity' has no properties in common with type 'OptionBase'
```

This used to work with v3/4 https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-select/src/types.d.ts#L4
but i'm not sure that is correct.

This does work but is not very ergonomic and causes different issues:
```tsx
interface Entity extends OptionBase {
  id: number;
  name: string;
}
```

Methuselah96:
Thanks for the issue. That's definitely a fair criticism, let me think about the best solution.

PR #4702, opened 2021-07-23
Remove OptionBase type

Methuselah96:
Resolves https://github.com/JedWatson/react-select/issues/4696.

Unfortunately due to [weak type detection](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-4.html#weak-type-detection), it doesn't work to have an `Option` generic type that doesn't have any properties in common with `OptionBase` since all of the properties on `OptionBase` are optional. See the attached issue for a more concrete example that causes issues.

```
export interface OptionBase {
  readonly label?: string;
  readonly value?: unknown;
  readonly isDisabled?: boolean;
  readonly __isNew__?: true;
}
```

### DIFF
--- a/.changeset/tame-suits-lick.md
+++ b/.changeset/tame-suits-lick.md
@@ -1,0 +1,6 @@
+---
+"react-select": patch
+"@react-select/docs": patch
+---
+
+The Option generic is no longer required to extend the OptionBase type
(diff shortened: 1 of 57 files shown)


JedWatson (approved):
Thanks @Methuselah96

PR #4720, opened 2021-08-13
Single value layout fix

seancurtis:
The absolute positioning pulls both the value and the placeholder out of the flow, and thus the component itself collapses down when used as a flex child. This can be solved by wrapping the component in another div, but we can solve it better by using CSS Grid instead.

before-after

### DIFF
--- a/packages/react-select/src/components/Placeholder.tsx
+++ b/packages/react-select/src/components/Placeholder.tsx
@@ -29,11 +29,9 @@ export const placeholderCSS = <
 }: PlaceholderProps<Option, IsMulti, Group>): CSSObjectWithLabel => ({
   label: 'placeholder',
   color: colors.neutral50,
+  gridArea: '1 / 2',
   marginLeft: spacing.baseUnit / 2,
   marginRight: spacing.baseUnit / 2,
-  position: 'absolute',
-  top: '50%',
-  transform: 'translateY(-50%)',
 });
 
 const Placeholder = <
--- a/packages/react-select/src/components/Placeholder.tsx
+++ b/packages/react-select/src/components/Placeholder.tsx
@@ -29,7 +29,7 @@ export const placeholderCSS = <
 }: PlaceholderProps<Option, IsMulti, Group>): CSSObjectWithLabel => ({
   label: 'placeholder',
   color: colors.neutral50,
-  gridArea: '1 / 2',
+  gridArea: '1 / 1 / 2 / 3',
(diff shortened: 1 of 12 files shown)


packages/react-select/src/components/containers.tsx
```diff
@@ -85,9 +85,10 @@ export const valueContainerCSS = <
   Group extends GroupBase<Option>
 >({
   theme: { spacing },
+  isMulti
 }: ValueContainerProps<Option, IsMulti, Group>): CSSObjectWithLabel => ({
   alignItems: 'center',
-  display: 'flex',
+  display: `${isMulti ? 'flex' : 'grid'}`,
```
Rall3n, line 91: Question of style, not critical, but why literals when you can do it without them?

Rall3n (changes requested):
image

Value, placeholder and input are centered by default instead of left aligned.

seancurtis:
@Rall3n thanks for the catch in the layout when expanded. I was working on it using a fixture to test the flex content and as such it was constantly squashed. I just reverted it before pushing. I did some more fixes to the layout/alignment as I noticed another issue too. It's much more solid now.

packages/react-select/src/components/Placeholder.tsx
```diff
@@ -29,11 +29,9 @@ export const placeholderCSS = <
 }: PlaceholderProps<Option, IsMulti, Group>): CSSObjectWithLabel => ({
   label: 'placeholder',
   color: colors.neutral50,
+  gridArea: '1 / 2',
```
Rall3n, line 32: image

Forgot to provide the `Placehholder` component with the latest fix :)

Rall3n (changes requested):
One last thing, but everything else looks ok.

seancurtis:
D'oh! Not sure how I missed that one tbh. Should be good now.

Rall3n (approved):
Looks good now 👍

packages/react-select/src/components/containers.tsx
```diff
@@ -84,9 +84,10 @@ export const valueContainerCSS = <
   Group extends GroupBase<Option>
 >({
   theme: { spacing },
+  isMulti
```
Methuselah96, line 87: To make Prettier happy:
```suggestion
  isMulti,
```

JedWatson (approved):
Nice! Ha, love css grids, this is so neat now. Thanks @seancurtis

PR #4731, opened 2021-08-18
MultiValue key - made more unique by adding a hyphen

sgadman:
The key for MultiValue in Select.tsx is not unique enough and can have some edge cases where there are duplicated keys.

For example, if the value is 32 and the index is 6 and another value is 3 and the index is 26.  In that case both keys will be 326.

I propose we just add a hyphen between the value and the index.  That way the keys will be `32-6` and `3-26` in this instance.

### DIFF
--- a/.changeset/gold-tomatoes-smoke.md
+++ b/.changeset/gold-tomatoes-smoke.md
@@ -1,5 +1,5 @@
 ---
-"react-select": patch
+'react-select': patch
 ---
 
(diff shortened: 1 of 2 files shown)


githubwebexcellisofficial (approved):

githubwebexcellisofficial (approved):

stemail23 (approved):
We have exactly this issue with a select box with values "-12" … "+12". "-11" and "-1" collide with keys "-111".

JedWatson (approved):
This looks good - thanks @sgadman

Issue #4743, opened 2021-08-27
Feature Request: merge styles with provided by default


brandonchinn178:
# My use case

I'm adding a multiselect field to a form, that's basically intended to be a nicer version of a `<textarea>` with comma separated values. As such, I want the cursor to still show the `text` cursor when hovering over the input, but not over the `multiValue` elements. Specifically, I have the following:
```ts
styles={{
  valueContainer: (provided: object) => ({
    ...provided,
    cursor: 'text',
  }),
  multiValueLabel: (provided: object) => ({
    ...provided,
    cursor: 'default',
  }),
  multiValueRemove: (provided: object) => ({
    ...provided,
    cursor: 'pointer',
  }),
}}
```
I find it rather annoying to have to repeat `...provided` every time; in fact, I can't personally think of an instance where I would want to throw away `provided` entirely and start completely from scratch. Most if not all of the time, I want to add styles on top of `react-select`'s built-in styles.

# Proposal

1. By default, merge styles with `provided`. `() => ({ foo: bar })` should be equivalent to `(provided) => ({ ...provided, foo: bar })`
2. If a special `override` key is set to `true`, don't merge. `() => ({ override: true, foo: bar })` is equivalent to today's `() => ({ foo: bar })`
3. As a shorthand, a plain object may be specified. `valueContainer: { foo: bar }` should be equivalent to `valueContainer: () => ({ foo: bar })` which merges `{ foo: bar }` with provided, per (1)

ebonow:
Greetings @brandonchinn178 ,

I had some concerns about making this part of the common library. 

I would agree that in most use-cases, the provided properties are spread into the style, but unfortunately this isn't always the case. The are times users may simply want to remove some css property and default to something either in their stylesheet or browser default. While there are `unset`, `initial`, `inherit`, etc... I think these add more complexity.

Additionally, there have been other requests to disable all injected styles. The resolution to this was using a `Proxy` to return an empty object. https://github.com/JedWatson/react-select/issues/2872#issuecomment-654251799

Your proposal would likely be a shortcut for most users, but it would ultimately remove this flexibility required for some.

The other critical thing here is that this change would likely require its own major release as it would break every user of the library using the styling api as the first variable would no longer be the provided props, but instead the `state` object.

[...]

PR #4744, opened 2021-08-27
[RFC] Merge styles by default

brandonchinn178:
Resolves https://github.com/JedWatson/react-select/issues/4743

### DIFF
--- a/.changeset/sour-adults-attend.md
+++ b/.changeset/sour-adults-attend.md
@@ -1,0 +1,28 @@
+---
+'react-select': major
+---
+
+Styles now merge with React-Select's styles by default. To avoid merging, add `override: true` to the object. You may also specify an object directly, which is equivalent to a function returning the object.
+
+```js
+styles={{
+  // merges with React-Select's default styling
+  container: () => ({
+    padding: '10px',
+  }),
+  // same as above
+  control: {
+    padding: '10px',
+  },
+  // does not merge
+  input: () => ({
+    padding: '10px',
+    override: true,
+  }),
+  // same as above
+  menu: {
+    padding: '10px',
+    override: true,
+  },
+}}
+```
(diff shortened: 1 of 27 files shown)


brandonchinn178:
@Rall3n any update on this?

ebonow:
Greetings @brandonchinn178 ,

Here are some of my concerns about making this part of the common library. 

The critical thing here is that this change would likely require its own major release as it would break every user of the library using the styling api as the first variable would no longer be the provided props, but instead the `state` object.

Regarding the issue that is being solved here, I would agree that in most use-cases, the provided properties are spread into the style, but this isn't always the case. The are times users may simply want to remove some css property and default to something either in their stylesheet or browser default. While there are `unset`, `initial`, `inherit`, etc... I think these add more complexity.

Additionally, there have been other requests to disable all injected styles. The resolution to this was using a `Proxy` to return an empty object. https://github.com/JedWatson/react-select/issues/2872#issuecomment-654251799. In regards to that specific solution, it seems that it would be possible to replicate this by returning `{ override: true }` instead of `{}`, but it feels a bit odd to me to include `override` in a styles object. I appreciate that there is an "escape hatch" if it was needed, and am open to getting thoughts from others regarding this.

The team has discussed some preliminary thoughts about providing alternative means of styling that could remove the dependency on Emotion. We are still very much in preliminary discussion about what that would look like, specifically with getting 5.0 and its documentation working properly, so any kind of change would likely be a bit more down the road.

I am open to thoughts from others, but this certainly is a breaking change and would require a new major version if it were to be considered a MR candidate.

brandonchinn178:
Thanks for getting back, and looking into it so deeply! I definitely agree it's a breaking change; as you mentioned, it does break users who are *not* merging styles that would now be unexpectedly merging styles. Happy to hear from other people on this; I would find this useful, but it's certainly workaroundable.

Another option is to provide a separate attribute like `addStyles` that merges.

Issue #4776, opened 2021-09-14
noOptionsMessage prop is confusingly typed


luxalpa:
The typings for the noOptionsMessage only allow a string return (as described here: https://github.com/JedWatson/react-select/issues/3251 although since it's quite old I don't know if this is still the current state of the matter). Since this is apparently intentional then I don't understand why in the official docs it says you should return a react.Node: https://react-select.com/props#select-props
```
({
inputValue string required
}) => One of <
react.Node,
null
>
```

Excuse me if this is already resolved in the upcoming release. But this really confused me. Am I supposed to return a react.Node or a string?

Methuselah96:
The issue you linked to is discussing `getOptionLabel` (which should return a `string`) and `formatOptionLabel` (which should return a `ReactNode`) so that discussion is irrelevant to the `noOptionsMessage` prop.

The official docs are correct. `noOptionsMessage` should return a `ReactNode` (and `string`s are valid `ReactNode`s). Does that answer your question?

luxalpa:
I thought it was related due to this issue: https://github.com/JedWatson/react-select/issues/3511

So if that is actually intentional, then I am guessing that the Typescript typings on DT need to be corrected? (See: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-select/src/Select.d.ts#L150)

Methuselah96:
Got it, no it's fine for the `loadingMessage` and `noOptionsMessage` to be any React node since they aren't used in the ARIA messages.

Yeah, feel free to make a PR on DT, although we're working on releasing [a TypeScript rewrite](https://github.com/JedWatson/react-select/issues/4559) so the type for `noOptionsMessage` will be correct in 5.0. However it looks like the `loadingMessage` prop still needs to be updated on this repo.

PR #4782, opened 2021-09-16
Allow loadingMessage prop to return any ReactNode

Methuselah96:
Resolves https://github.com/JedWatson/react-select/issues/4776.

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -187,7 +187,7 @@ export interface Props<
   /** Whether to enable search functionality */
   isSearchable: boolean;
   /** Async: Text to display when loading options */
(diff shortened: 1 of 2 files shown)


ebonow (approved):

packages/react-select/src/Select.tsx
```diff
@@ -187,7 +187,7 @@ export interface Props<
   /** Whether to enable search functionality */
   isSearchable: boolean;
   /** Async: Text to display when loading options */
-  loadingMessage: (obj: { inputValue: string }) => string | null;
+  loadingMessage: (obj: { inputValue: string }) => ReactNode;
```
ebonow, line 190: My initial concern was accessibility but I don't believe that this is something being implemented at present, so this is likely a non-issue.

JedWatson (approved):

PR #4803, opened 2021-09-25
Import CSSObject from @emotion/react instead of @emotion/serialize

Methuselah96:
This is helpful for people using a more strict package manager like Yarn PnP since we don't have a direct dependency on @emotion/serialize.


package.json
```diff
@@ -174,5 +174,8 @@
   "browser": {
     "./dist/react-select.cjs.js": "./dist/react-select.browser.cjs.js",
     "./dist/react-select.esm.js": "./dist/react-select.browser.esm.js"
+  },
+  "resolutions": {
+    "csstype": "^3.0.2"
```
Methuselah96, line 179: This is necessary because we still have some dependencies on the website that use Emotion 10 which depends on csstype 2 which clashes with the csstype 3 types and produces build errors.

Rall3n (approved):

ebonow (approved):

PR #4810, opened 2021-09-26
Fix "Props" documentation page not loading types

### LINKED ISSUE #4796: https://react-select.com/props loading issue
I'm not sure if this is the correct place to report this, as it is not part of the actual library, but I went to look at the prop documentation today and noticed that the page is failing to load after "internal types", note the "loading" in the sections:

Screen Shot 2021-09-23 at 3 17 06 PM

Tested on Chrome:

Screen Shot 2021-09-23 at 3 16 23 PM

and Safari:

Screen Shot 2021-09-23 at 3 16 18 PM

### LINKED ISSUE #4808: The page `https://react-select.com/props` is broken
Browse to [https://react-select.com/props](https://react-select.com/props), and you'll see all examples are "loading". 

The console shows the following error:

> Uncaught (in promise) SyntaxError: Unexpected token < in JSON at position 0

The error is thrown from `useMagicalNodes` on this line:

> `.then((x) => x.json())`

The "json" URL is:

> https://react-select.com/magical-types/magical-types-0-0.pl5a2gtoah.json

But this file is actually an HTML.

Please fix :pray:

Rall3n:
`magical-types` created files where not added to the `dist` directory when building the documentation.
This causes the props on the "Props" page not to load.

This PR includes those files in the build process.

Closes #4796

### DIFF
--- a/docs/webpack.config.ts
+++ b/docs/webpack.config.ts
@@ -4,6 +4,7 @@ import CopyWebpackPlugin from 'copy-webpack-plugin';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
 import ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin';
 import { config } from 'dotenv';
+import { CleanWebpackPlugin } from 'clean-webpack-plugin';
 
 config();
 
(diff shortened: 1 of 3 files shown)


docs/webpack.config.ts
```diff
@@ -51,12 +52,18 @@ const webpackConfig: webpack.Configuration = {
     //   'process.env.CLIENT_ID': `'${process.env.CLIENT_ID}'`,
     //   'process.env.CLIENT_SECRET': `'${process.env.CLIENT_SECRET}'`,
     // }),
+    new CleanWebpackPlugin(),
```
Methuselah96, line 55: Can we use [`output.clean`](https://webpack.js.org/guides/output-management/#cleaning-up-the-dist-folder) instead?

Methuselah96 (changes requested):

docs/webpack.config.ts
Rall3n: `output.clean` is Webpack v5+ only.

Methuselah96: Ah, I missed we were still on Webpack 4.

Methuselah96 (approved):

Methuselah96:
Merging this since the website is broken.

PR #4833, opened 2021-10-03
Set ValueContainer display grid if isMulti and has no value

### LINKED ISSUE #4797: Insertion point on multi valued select v5 moved from position in v4

I hope my pointing at the examples on https://react-select.com/home are sufficient.

# Details

In v4, when the select is set up to show multi-values, when there is no value, the insertion point is placed before the placeholder text.  In v5, however, it appears after the placeholder text.  This only happens with the multi-value case.  Single value works unchanged.

## Example

| V4 | V5 |
|--|--|
|  image |  image |

This turns out to be pretty easy to reproduce on the react-select web site.  Notice that if you delete the pills in the "Multi" case, the insertion point appears after "Select..."

image

If you scroll down to (say) Async > callbacks on that page, the single select there always shows the insertion point before the select

image

Thanks!  We love `react-select`!

ebonow:
closes #4797 

This PR is a fix for the ValueContainer styling so that the display property is correctly set to grid if the Select isMulti and has no value. This will ensure that the Input will be positioned over the Placeholder.


seancurtis (approved):
LGTM

ebonow:
@newanalog The issue is that we have two distinct use-cases for the ValueContainer.
1. When it is a single select, the display should be `grid` so that the Placeholder and the Input are overlapping in the same cell 
2. When it is a multiSelect, then the displayed should be `flex` so that the selected values can remain inline with each other and so that the input can fill out the rest of the row or expand to a new row when the input expand.

The issue this PR solves is the empty case of when the Select isMulti and there are no selected values. The ValueContainer should in this case mimic that of the Single Select and have its display set to `grid` until there has been a selected value so that the Input and Placeholder can share the same grid cell and appear as though "the cursor was moved to the beginning of the placeholder text"

If you are still seeing the input and placeholder misaligned, then there might be a separate issue. Can you confirm what version of react-select your screenshot was taken from?

PR #4860, opened 2021-10-13
Fixes for animated value container and unique keys for options

### LINKED ISSUE #4602: The exiting animation on MultiValues doesn't look correct
Hi there 👋 
The current exiting animation on the multi values looks a bit janky, the demo below was recording on https://react-select.com/home
home

image
Click on "Perth" to remove it
↓
image
The 2nd "Perth" was removed, but the last 3 items ("Perth", "Darwin", "Melbourne") on the first pic were added to the end
↓
image
3. The last 3 items were removed 

My assumption is the `index` was added to the `key` field in the `<MultiValue>` [here](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.tsx#L1604). Every time a value get removed from the list, the index would change so the `key` of the MultiValue get changed as well, which causes the rerender 🤔 

The index was add on v3.2.0, also made a codesandbox to verify that: https://codesandbox.io/s/sad-northcutt-3s5ce?file=/src/App.tsx
the animation looks good when using react-select v3.1.1 but not on v3.2.0

### LINKED ISSUE #4844: 'animatedComponents' is not behave as before in version 5
Clearing animation in version 5, producing a gap in the Input field.
Same result with disabled custom styles.

React 17.0.2
Microsoft Edge 96.0.1032.0 / Windows 11
Mozila Firefox 93.0 / WSL Ubuntu

ebonow:
closes #4844 , closes #4602 , 

Animated components were broken by the display styling change on the ValueContainer as the transition effect occurs after the display switches from flex to grid causing the last value when removed to appear on a new line below the input until the component collapses and is removed.

https://user-images.githubusercontent.com/1007579/137392104-90c19bee-7a63-4b9d-93cb-d357e03e29b4.mov

This PR addresses this by passing in an onExited prop to the AnimatedMultiValue as a callback to set the display to grid only after the last value has been removed. Additionally, it accounts for the Placeholder component hiding it until the container has been full set to grid position.

Additionally, previous MultiValue animations were broken as the index of each Option was changed when options were removed from the beginning or middle of the list, because the options index was being used to generate the key. This PR addresses this by generated the key based on the existing methods used to stringify key and value and concatenate them.

https://user-images.githubusercontent.com/1007579/137392248-a14a707a-2ea5-42c2-b2e0-6aa76c6d2c12.mov

Here is a video screen share with the above methods being used demonstrating the differences.

https://user-images.githubusercontent.com/1007579/137392332-f68e261c-38a3-4470-a789-3b78e900dd7c.mov


Rall3n (commented):
Animated seems "jumpy" when selecting options in an empty `Select` (`Placeholder` stays behind the options for less then a second).

Also the `Placeholder`s fade animation doesn't seem to trigger. Doubt it is a cause of this PR, but it should be noted.

Saga4:
Q: Can anyone suggest on how to make an animated expandable search bar with AsyncSelect? And shall we file an issue for this feature here?

PR #4882, opened 2021-10-28
Add `required` prop

Rall3n:
This PR adds the new `required` prop to allow for required validation in forms.

This also adds a new prop `requiredMessage` to customize the content of the HTML5 validation message. I would have wanted to use a default browser message, but there is no API to get them, and the default for `<input type='text'>` is imo not applicable for a dropdown component.

@ebonow I don't know if and how this should be implemented into our `LiveRegion` component. Maybe you can find a solution.

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -1588,6 +1588,7 @@ export default class Select<
       tabIndex,
       form,
       menuIsOpen,
+      required
     } = this.props;
     const { Input } = this.getComponents();
     const { inputIsHidden, ariaSelection } = this.state;
@@ -1604,6 +1605,7 @@ export default class Select<
       'aria-invalid': this.props['aria-invalid'],
       'aria-label': this.props['aria-label'],
       'aria-labelledby': this.props['aria-labelledby'],
+      'aria-required': required,
       role: 'combobox',
       ...(menuIsOpen && {
         'aria-controls': this.getElementId('listbox'),
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -262,8 +262,6 @@ export interface Props<
   value: PropsValue<Option>;
   /** Sets the form attribute on the input */
   form?: string;
-  /** Text to display if the component fails the `required` validation  */
-  requiredMessage: () => string;
   /** Marks the value-holding input as required for form validation */
   required?: boolean;
 }
@@ -1588,7 +1586,7 @@ export default class Select<
       tabIndex,
       form,
       menuIsOpen,
-      required
+      required,
     } = this.props;
     const { Input } = this.getComponents();
     const { inputIsHidden, ariaSelection } = this.state;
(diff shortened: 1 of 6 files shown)


packages/react-select/src/__tests__/Select.test.tsx
```diff
@@ -3112,3 +3112,21 @@ test('renders with custom theme', () => {
     window.getComputedStyle(firstOption!).getPropertyValue('background-color')
   ).toEqual(primary);
 });
+
+test('form validation with `required` prop', () => {
+  const components = (value?: Option) => (
+    <form id="formTest">
+      <Select {...BASIC_PROPS} required value={value || null} />
```
ebonow, line 3119: For testing, it might be better to test that the value is explicitly passed as null instead of coalescing in case the behavior differs between undefined and null. 

Not that it would here, but it's possible some future use-case changes the default behavior and it gets missed.

packages/react-select/src/__tests__/Select.test.tsx
```diff
@@ -3112,3 +3112,21 @@ test('renders with custom theme', () => {
...
+  const { container, rerender } = render(components());
+
+  expect(
+    container.querySelector<HTMLFormElement>('#formTest')?.checkValidity()
+  ).toEqual(false);
+  rerender(components(BASIC_PROPS.options[0]));
+  expect(
+    container.querySelector<HTMLFormElement>('#formTest')?.checkValidity()
+  ).toEqual(true);
+});
```
ebonow, line 3132: Can we also add a test to cover use-cases for `isMulti` and `isSearchable=false`?

Rall3n: Extended test cases have been added.

Rall3n:
Removed the `requiredMessage` prop. Can be achieved using `onInvalid` on a `<form>` element.

lukebennett88 (approved):

packages/react-select/src/Select.tsx
```diff
@@ -1992,7 +2009,7 @@ export default class Select<
               />
             ))
           ) : (
-            <input name={name} type="hidden" />
+            <input name={name} type="hidden" value="" />
```
nderkim, line 2012: Just wondering, does setting value="" mean that it can't be used as an uncontrolled component anymore?

Rall3n: It would prevent React from throwing a "Switching from uncontrolled to controlled" error because React reuses the `<input>` element for `RequiredInput` component and if a value is set.

I don't know if it was really applicable here anymore, but better safe than sorry.

srinivasmerugu:
@Rall3n @ebonow With this requiredInput component changes we see an accessibility issue on this new <input/> tag .

Screenshot 2022-12-10 at 10 09 36 PM

I think on the below  element we may need to pass the type="hidden" similar to other <input /> elements in the Select.tsx file  since its has tabindex="-1".

 <input required="" name="licenseOrCertification" tabindex="-1" class="css-5kkxb2-requiredInput-RequiredInput" value="">

could you please add this in the next minor release as axe dev tools (tool which is used to test the accessibility) is throwing a critical error.

osvaldokalvaitir:
@Rall3n @ebonow
I tested it and the required property didn't work

PR #4883, opened 2021-10-28
Consolidate dependencies and remove duplicates

jgeurts:
Hi! Hope this PR finds you well. I'm not sure how useful this is since it seems that preconstruct does similar at build time. 

* Move most dependencies to be devDependencies
* Reduce duplication of dependencies

I ran through all of the scripts in package.json and things appear to work as they did prior to this PR.


Methuselah96 (changes requested):
Thanks for the PR! Can you keep all the dependencies for the root package and `docs` in `dependencies` instead of splitting it into `devDependencies` and `dependencies`? For those packages, it doesn't really matter whether it's a dependency or dev dependency since it will not ever get installed by another package.

jgeurts:
I'm curious how come you prefer to put everything in `dependencies` and not lean on `devDependencies`?

Methuselah96:
@jgeurts It's purely a stylistic choice (imo) that Jed said he preferred in our maintainers meeting when we looked at this PR. Since those packages are never used outside of this repository there is no functional difference between them being dependencies or devDependencies so might as well just stick them all in dependencies.

(Personally I prefer the semantic difference between dependencies and devDependencies even for apps where it doesn't make a functional difference since it still designates which dependencies will used at runtime, but to each his own and it's Jed's project.)

docs/package.json
```diff
@@ -69,11 +48,7 @@
     "style-loader": "^0.23.1",
     "styled-components": "^3.4.10",
     "ts-node": "^9.1.1",
-    "typescript": "^4.1.3",
-    "unfetch": "^3.0.0",
-    "webpack": "^4.30.0",
-    "webpack-cli": "^3.3.1",
-    "webpack-dev-server": "^3.3.1"
```
Methuselah96, line 76: I think since these dependencies are only used by the `docs` package they should remain in the `docs` package and be removed from the root package.

Methuselah96 (changes requested):

Methuselah96:
Or I guess the other way of thinking about it is that the docs and root packages are primarily Node packages and the "dev dependencies" are in fact a runtime dependency since the main function of those packages is to start a Webpack build.

jgeurts:
> @jgeurts It's purely a stylistic choice (imo) that Jed said he preferred in our maintainers meeting when we looked at this PR. Since those packages are never used outside of this repository there is no functional difference between them being dependencies or devDependencies so might as well just stick them all in dependencies.

Cool, thanks for the insight!

dcousens (approved):

dcousens:
This is always going to be an arbitrary balance, but I think this is helpful enough as is that we can refine anything else as needed in future pull requests.

PR #4903, opened 2021-11-08
Add renderClearableIndicatorForEmptyValue prop

thepocp:
Adds a new property called "renderClearableIndicatorForEmptyValue".
Usage example:
```tsx
const hasDifferentValues = props.value === DIFFERENT_VALUES;

return (
    <MultiSelect
        value={hasDifferentValues ? [] : props.value}
        onChange={props.onChange}
        className={hasDifferentValues ? styles.has_different_values : undefined}
        placeholder={hasDifferentValues ? 'There are different values' : ''}
        renderClearableIndicatorForEmptyValue={hasDifferentValues}
    />
);
```

### DIFF
--- a/docs/pages/components/index.tsx
+++ b/docs/pages/components/index.tsx
@@ -196,7 +196,7 @@ export default function Components() {
 
         * When \`isClearable\` is false, or when \`isMulti\` is false, and \`isClearable\` is undefined
         * When the select is disabled
-        * When the select has no value
+        * When the select has no value and \`renderClearableIndicatorForEmptyValue\` is not true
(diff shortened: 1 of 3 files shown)


thepocp:
I didn't write what this change is for. Here is the example from my project: several entities are modified in the form, and if the values of entities'attributes are not equal then in the Select the following placeholder is shown "There are different values". And the user has an opportunity to clear the field and/or choose other options.

xiaogwu:
Can we get a maintainer to review this PR?  I think this is a good feature to add.

Issue #4940, opened 2021-12-06
TypeScript: Incompatible `'aria-invalid'` types for `Select`


prichey:
I'm having some issues satisfying TypeScript when upgrading from `react-select` v4 (4.3.1) to v5 (5.2.1).

The crux of it is this:

```
Type '{ styles: StylesConfig<SelectInputOption<T>, false, GroupBase<SelectInputOption<T>>>; options: SelectInputOption<T>[]; ... 356 more ...; className: string; }' is not assignable to type 'Omit<Pick<Props<SelectInputOption<T>, false, GroupBase<SelectInputOption<T>>>, "aria-errormessage" | "aria-invalid" | ... 28 more ... | "form"> & Partial<...> & Partial<...>, StateManagedPropKeys>'.
  Types of property ''aria-invalid'' are incompatible.
    Type 'boolean | "false" | "true" | "grammar" | "spelling" | undefined' is not assignable to type 'boolean | undefined'.
      Type '"false"' is not assignable to type 'boolean | undefined'.
```

I can provide a relatively complete reproduction of my types if useful but TL;DR: up the tree I'm using `AllHTMLAttributes<HTMLSelectElement>` which pulls from React's own [AriaAttributes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1567) type (deep link to `'aria-invalid'` [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1654)). This type differs from what's declared as `Props` in [Select.tsx](https://github.com/JedWatson/react-select/blob/20ea0308199ce27309ff89df30c21ecd1c9c5166/packages/react-select/src/Select.tsx#L79), so when I spread props of one type onto the `Select` component, `tsc` yells at me.

IMO the types in `Select` should be modified to match those defined in `AriaAttributes`, since presumably the attribute should follow the same spec. Let me know if I'm missing something on my end, though!

prichey:
I have a naive PR up for this: https://github.com/JedWatson/react-select/pull/4941. This fix is preventing me from upgrading to v5 so please let me know if there's anything else I can do here!

prichey:
@Rall3n tagging you since I see you've been doing triage on recent issues, please let me know what you think whenever you have a chance!

Methuselah96:
As far as preventing you from upgrading to v5, can you just cast until it gets fixed here?

prichey:
[...]

PR #4941, opened 2021-12-06
Use React's `AriaAttributes` type directly rather than recreating

prichey:
Fixes https://github.com/JedWatson/react-select/issues/4940

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -1,4 +1,5 @@
 import React, {
+  AriaAttributes,
   Component,
   FocusEventHandler,
   FormEventHandler,
@@ -13,12 +14,7 @@ import LiveRegion from './components/LiveRegion';
 
 import { createFilter, FilterOptionOption } from './filters';
 import { DummyInput, ScrollManager } from './internal/index';
-import {
-  AriaInvalid,
-  AriaLive,
-  AriaLiveMessages,
-  AriaSelection,
-} from './accessibility/index';
+import { AriaLiveMessages, AriaSelection } from './accessibility/index';
 
 import {
   classNames,
@@ -75,15 +71,15 @@ export interface Props<
   Group extends GroupBase<Option>
 > {
   /** HTML ID of an element containing an error message related to the input**/
-  'aria-errormessage'?: string;
+  'aria-errormessage'?: AriaAttributes['aria-errormessage'];
   /** Indicate if the value entered in the field is invalid **/
-  'aria-invalid'?: AriaInvalid;
+  'aria-invalid'?: AriaAttributes['aria-invalid'];
   /** Aria label (for assistive tech) */
-  'aria-label'?: string;
+  'aria-label'?: AriaAttributes['aria-label'];
   /** HTML ID of an element that should be used as the label (for assistive tech) */
-  'aria-labelledby'?: string;
+  'aria-labelledby'?: AriaAttributes['aria-labelledby'];
   /** Used to set the priority with which screen reader should treat updates to live regions. The possible settings are: off, polite (default) or assertive */
-  'aria-live'?: AriaLive;
+  'aria-live'?: AriaAttributes['aria-live'];
   /** Customize the messages used by the aria-live component */
   ariaLiveMessages?: AriaLiveMessages<Option, IsMulti, Group>;
   /** Focus the control when it is mounted */
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -14,6 +14,7 @@ import LiveRegion from './components/LiveRegion';
 import { createFilter, FilterOptionOption } from './filters';
 import { DummyInput, ScrollManager } from './internal/index';
 import {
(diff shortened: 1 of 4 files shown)


packages/react-select/src/accessibility/index.ts
```diff
@@ -11,6 +11,8 @@ export type OptionContext = 'menu' | 'value';
 
 export type GuidanceContext = 'menu' | 'input' | 'value';
 
+export type AriaInvalid = boolean | 'false' | 'true' | 'grammar' | 'spelling';
```
prichey, line 14: Pulled from here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1654

packages/react-select/src/Select.tsx
```diff
@@ -76,7 +77,7 @@ export interface Props<
   /** HTML ID of an element containing an error message related to the input**/
   'aria-errormessage'?: string;
   /** Indicate if the value entered in the field is invalid **/
-  'aria-invalid'?: boolean;
+  'aria-invalid'?: AriaInvalid;
```
Methuselah96, line 80: Instead of redefining `AriaInvalid`, can we do `AriaAttributes['aria-invalid']` here?

Methuselah96 (changes requested):

packages/react-select/src/Select.tsx
prichey: Yeah, definitely. I'll make that change now

prichey: Do you think it's worth updating the other aria attributes to pull from that type in that case?

Methuselah96: If you're up for that, that would be great! Thanks.

packages/react-select/src/Select.tsx
```diff
@@ -74,15 +71,15 @@ export interface Props<
   Group extends GroupBase<Option>
 > {
   /** HTML ID of an element containing an error message related to the input**/
-  'aria-errormessage'?: string;
+  'aria-errormessage'?: AriaAttributes['aria-errormessage'];
```
prichey, line 74: `AriaAttributes['aria-errormessage']` includes `| undefined` so this is technically equivalent to

```
'aria-errormessage': AriaAttributes['aria-errormessage'];
```

but I'll defer to your preferences here @Methuselah96

packages/react-select/src/Select.tsx
prichey: Done ✅

packages/react-select/src/Select.tsx
Methuselah96: I'm not sure it's always equivalent ([TS playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4Aoc4AOxiSk3STgHkwZgJqUAbABRxgAznADe5OJLh0hMAPwAuOLKg0A5nAA+cAK7UAJkkw0k+igF9KaHiiEi2HLrzhIAHnQMjk6GADoAwriQ1Ei0ADwOnNz8gkIAfGISUkQG9AAUAJSJUjlwRDA6UNRw1Do8PBQ5lpbkaFyyJVyRTjwAqkIoaswAvHAR7FHOAPRxFFS09IxozAByTQMtAhDC2VIyMEoq6lq6qcYhZuQ11rYic9TN0S7uofpexH6B4FyhMGHnl7xLwgniOSmGKCZVa5PJIApFEplCpJSTVKz1eAwAAW6gATHBeu95o4riMKEA)). Either way, I like the semantic nuance of declaring it as optional.

Methuselah96 (approved):
Thanks!

prichey:
@Methuselah96 took the liberty to update the only other place we were recreating types from `AriaAttributes`: https://github.com/JedWatson/react-select/pull/4941/commits/9ace36c0b5a83625e589357beec70c71ebdac99f. Going to stop poking at this now! Looking forward to 5.3 😄

Methuselah96:
Thanks! Forgot to ask earlier: can you add changeset?

prichey:
@Methuselah96 Just added one - let me know if I did it correctly, this was my first time using that package

Methuselah96:
Yeah, looks good.

prichey:
@Methuselah96 How do you determine when to cut a new release? It'd be nice to upgrade without having to do my own manual patching. Thanks!

Methuselah96:
@prichey Thanks for the reminder, as you can see we haven't been doing a whole lot of merging/releasing recently. I'll try to get a release out soon. I noticed you made this a minor instead of a patch. Any particular reason?

prichey:
@Methuselah96 No problem! Tbh it could totally be a patch, I think I did that after you had tagged it with 5.3 so I assumed you wanted it to be a minor release. Patch would be awesome if that helps it get released!

Methuselah96:
@prichey Published in [v5.2.2](https://github.com/JedWatson/react-select/releases/tag/react-select%405.2.2).

PR #4955, opened 2021-12-13
refactor docs index page

Gyekye:


Rall3n:
I don't see a valid reason to accept these changes and would opt to reject this PR.

Any thoughts @ebenow @Methuselah96 @JedWatson @bladey?

Issue #4962, opened 2021-12-18
@types/react interferes with preact types


weedz:
I noticed this package uses `@types/react-transition-group` as a regular dependency which in turn depends on `@types/react`. `@types/react` interferes with the JSX types from preact so we get warnings like:
```
'MenuLink' cannot be used as a JSX component.
  Its instance type 'Component<IMenuLink, {}>' is not a valid JSX element.ts(2786)
```

Are there any problems with moving `@types/react-transition-group` into devDependencies? This seems to fix our issue.

PR #4972, opened 2021-12-29
Dependancies update

uzikilon:
This PR is a suggestion to move `emotion` to `"peerDependencies"`   avoid conflict with apps running other version of `emotion`


weedz:
This will also close https://github.com/JedWatson/react-select/issues/4962 as it moves `@types/react-transition-group` to devDependencies :+1:

kauffmanes:
There's a bug in emotion that caused `useInsertionEffect` to be referenced directly in the specifiers list of the import statement. I was going to open a bug to update the version of emotion but I think this issue would take care of that problem as well?

https://github.com/emotion-js/emotion/issues/2649
https://github.com/emotion-js/emotion/pull/2651
https://stackoverflow.com/questions/71185635/material-ui-export-useinsertioneffect-imported-as-useinsertioneffect1-was

lukebennett88 (dismissed):

.changeset/fair-vans-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-select': patch
+---
+
+Move emotion packages and types for react-transition-group to peerDependencies
```
nderkim, line 5: Is it still `patch` if a dependency has been moved to peerDependencies?

lukebennett88: Code that ships to the user has been changed, but it doesn't add any new functionality (minor) or change the existing API (major), so I'd consider it a patch (unless I'm missing something?)

nderkim: They might not already have `emotion` explicitly added as a dependency in their application, so wouldn't it be a potentially breaking change?

lukebennett88: Thats a good point. 
I think it varies by package manager, so it _might_ technically be a breaking change. I'll try to find someone who can give us a definitive answer.

lukebennett88:
Thanks so much for your PR @uzikilon.
Unfortunately this _is_ a breaking change so we can't merge it until we're ready to do another major release.

sorenhoyer:
In my opinion runtime css-in-js is the real problem here. Many have already started using zero-runtime css-in-js such as Linaria, and tbh I don't think the consumer who opted out of these messy libraries (emotion, styled-components etc.) should pay the price by having such a library installed and included in their code neither implicit nor explicitly,

So instead of asking the question whether it should be a peer- or normal dependency, perhaps there should be a PR for migrating away from using runtime css-in-js in the first place?

Methuselah96:
See https://github.com/JedWatson/react-select/issues/5451.

PR #4992, opened 2022-01-13
Added aria-autocomplete prop to select input

abhijeetranmale7:
### How it was:
There is no prop available for react select to change aria-autocomplete attribute like we can change aria-label.
The value for aria-autocomplete is set to "list" and is unable to change through props.
Refer the following url to the line of code.
https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.tsx#L1560
Because of aria-autocomplete="list", the screen reader reads out things that the user shouldn't really need to hear. something like "edit text autocompletion lists" is read out.
to prevent that we need to set aria-autocomplete to none. But since it has constant value "list", we are not able to do so.

### What I did:
I have added aria-autocomplete prop to the select input with default value as 'list'

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -79,6 +79,8 @@ export interface Props<
   'aria-invalid'?: boolean;
   /** Aria label (for assistive tech) */
   'aria-label'?: string;
+  /** Aria autocomplete */
+  'aria-autocomplete'?: 'none' | 'inline' | 'list' | 'both';
   /** HTML ID of an element that should be used as the label (for assistive tech) */
   'aria-labelledby'?: string;
   /** Used to set the priority with which screen reader should treat updates to live regions. The possible settings are: off, polite (default) or assertive */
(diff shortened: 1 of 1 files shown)


chiragshahklc (approved):

chiragshahklc:
Can we review and approve this please? It's somewhat required for us. Thank you.
@Rall3n @Methuselah96

packages/react-select/src/Select.tsx
```diff
@@ -77,6 +77,8 @@ export interface Props<
   'aria-invalid'?: AriaAttributes['aria-invalid'];
   /** Aria label (for assistive tech) */
   'aria-label'?: AriaAttributes['aria-label'];
+  /** Aria autocomplete */
+  'aria-autocomplete'?: 'none' | 'inline' | 'list' | 'both';
```
kethomas, line 81: inline is not a valid option for this component - https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#wai-aria-roles-states-and-properties-6

abhijeetranmale7: Hi @kethomas, Thanks for the review. I've removed the "inline" option. Please check.

chiragshahklc:
Hey @Methuselah96,

Can we approve these changes and merge this PR? 
It's somewhat required.

Thank you.

halain:
I would like to know why a year later this PR is still open and it did not merge, it is something very important for a11y

Thank you

halain (approved):

Issue #5055, opened 2022-02-14
loadOptions not properly typed anymore


josias-r:
- The argument `inputValue` provided to `loadOptions` is inferred with type `any`, instead of `string` (unless the second argument, callback, is also extracted) -> type overload seems to be broken.
- The return value of `loadOptions` can be anything, no type error is ever reported.

See example:
https://codesandbox.io/s/react-select-v5-ts-9sxmq

This worked prior to v5

Rall3n:
It's curios that TypeScript can't coalesce the second variant provided in its current typing. Maybe because the function always gets the callback as its second parameter?

josias-r:
Could the problem be that one of the two types for `loadOptions` is
```
((inputValue: string, callback: (options: OptionsOrGroups<Option, Group>) => void) => void)
```
instead of
```
((inputValue: string, callback: (options: OptionsOrGroups<Option, Group>) => void) => Promise<OptionsOrGroups<Option, Group>>) 
```

I don't know why the return type would be `void`.
Maybe the two different return types + two different argument types cause typescrip to report type `any` because it has no clue which one it is of the two.


Also, wouldn't a single type like this (notice the optional callback argument):
```
((inputValue: string, callback?: (options: OptionsOrGroups<Option, Group>) => void) => Promise<OptionsOrGroups<Option, Group>>) 
```
...be sufficient? Instead of having two types for with and without the callback argument

Methuselah96:
The return type is `void` because you should be able to pass in a function that doesn't return a promise. We have an example of that [in the docs](https://github.com/JedWatson/react-select/blob/master/docs/examples/AsyncCallbacks.tsx#L16-L23). If we updated the `loadOptions` type to have the return type always be a `Promise` that example wouldn't compile.

josias-r:
@Methuselah96 Ah yeah of course. That's the point of having a callback. I have no clue where types breaks here 😕

Rall3n:
> Then I have no clue where the types breaks here

I don't know exactly how TypeScript determines which types are applicable, but I believe it checks the source to see where something would apply.

With that in mind I also believe the issue is that we only call the prop with both arguments (`inputValue`, `callback`), never with only the first argument (see [source](https://github.com/JedWatson/react-select/tree/master/packages/react-select/src/useAsync.ts#L120)).

@Methuselah96 Would a change to the old type `(...) => Promise<...> | void` sufficient and acceptable?

Methuselah96:
[...]

PR #5081, opened 2022-02-28
Fix type inference for Async's loadOptions prop

Methuselah96:
Fixes https://github.com/JedWatson/react-select/issues/5055.

Basically, TypeScript handles function overloading better than taking the union of functions.


Methuselah96:
Hmm... it isn't quite that easy.

Issue #5050, opened 2022-02-08
Opening dropdown via dropdown icon hijacks events (react-datepicker example)


rkulinski:
When using `react-select` with `react-datepicker` it seems that some architectural issue of `react-select` shows up.

Repro:
https://stackblitz.com/edit/react-o6gkep?file=src/App.js

Correct behaviour:
1. Open datepicker.
2. Click on select (not the part with icon on right) and observe datepicker is closed correctly.

Issue:
1. Open datepicker.
2. Click on the part with icon on right of select and observe datepicker is not closed.

I think there's a problem with how this div with icon is designed. Seems to me like it's highjacking events. That doesn't happen for other part of the select and IMO it shouldn't happen here as well.

It was reported here:
https://github.com/JedWatson/react-select/issues/1539
but closed without any resolution.

Versions:
```
    "react": "17.0.2",
    "react-datepicker": "4.6.0",
    "react-dom": "17.0.2",
    "react-select": "^5.2.2"
```

rkulinski:
I think the issue is that dropdown indicator has it's own method for click (onDropdownIndicatorMouseDown). It calls `stopPropagation` which causes e.g. `react-datepicker` to not receive an event. I honestly don't understand why there are two handlers (with different logic inside) for something that seems should be working exactly the same.

hushee69:
For information, the bug isn't reproductible in mobile view for some reason (used this [https://stackblitz.com/edit/react-o6gkep?file=src/App.js](https://stackblitz.com/edit/react-o6gkep?file=src/App.js) while on mobile view and DatePicker does close correctly when select is clicked using the icon on right)

rkulinski:
@hushee69 what do you mean by "mobile view" - mobile device or narrow window? Note that this issue occurs only when you click "icon area" as described in issue.

rkulinski:
Will be fixed in https://github.com/JedWatson/react-select/pull/5138

PR #5085, opened 2022-02-28
Remove duplicate handler for indicator

rkulinski:
I have no idea why two different handlers should be applied for single action. The approach with two handlers was forcing to call `stopPropagation` which was causing this issue to occur
https://github.com/JedWatson/react-select/issues/5050

### DIFF
--- a/.changeset/allow-event-bubbling-for-all-controls.md
+++ b/.changeset/allow-event-bubbling-for-all-controls.md
@@ -1,0 +1,5 @@
+---
+'react-select': patch
+---
+
+Replace stopPropagation with target check.
--- a/.changeset/share-click-handler-for-dropdown-and-select.md
+++ b/.changeset/share-click-handler-for-dropdown-and-select.md
@@ -1,0 +1,5 @@
+---
+'react-select': patch
+---
+
+Use single click handler for opening select options. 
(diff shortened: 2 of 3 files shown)


Rall3n (changes requested):
Removing the dedicated event handler for the `DropdownIndicator` component introduces breaking changes and also breaks interaction with the component when using the prop `openMenuOnClick`:
* setting `openMenuOnClick={false}` does not open the menu when clicking on the indicator
* additionally setting `isSearchable={false}` shows no indication of interaction with the component upon clicking the indicator (resembles an `<select readonly />` element)

Even though both handlers do the same actions, they react in different ways to the interaction. One should not be removed on the basis that they both end in the same behavior. Instead, a way should be found to replace `stopPropogation` without changing current behaviour.

rkulinski:
@Rall3n got it, I wasn't able to find any difference in how it was working, but I'm not that familiar with library code so I missed some edge case scenarios. I'll try to somehow get rid of `stopPropagation`, but I'm not sure if this will be easy.

rkulinski:
@Rall3n I wasn't able to find any better and easier workaround for this with how components are structured.

rkulinski:
Closing in favor of https://github.com/JedWatson/react-select/pull/5089

PR #5167, opened 2022-04-27
Make "MenuPlacer" component customizable and switchable.

kkamik:
I need to redefine the component "MenuPlacer" because I have issue with viewHeight
https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/components/Menu.tsx#L78

In my case, the Select component renders inside scrollable container that have 50vh, and `const viewHeight = window.innerHeight` is a problem for my case.

I think make customizable MenuPlacer is the easiest way to fix it.


eugene-platov:
LGTM

needpower:
Good for me also 👍

kkamik:
@Rall3n @Methuselah96 Could you review PR?

kkamik:
> To be clear, it seems like the issue you're describing is actually a bug, is that correct?

Correct.

> Would using the height of the scroll parent when the menu is not using fixed positioning ([like this](https://github.com/JedWatson/react-select/pull/4913/files#diff-0b70b19d504021d4f2acd88c29ed2b3447116a02bf1679653779da6206db85dcR79-R81)) work for you?

I guess it can help.

>  If so, I wonder if it would be better to merge specifically that change since it seems like an improvement to the status quo while we still work on more fully fixing menu positioning by switching to Popper.

@Methuselah96 So, let's merge it :)

Methuselah96:
https://github.com/JedWatson/react-select/pull/5177

PR #5173, opened 2022-05-01
Fix : color issue with Dropdown indicator on Menu open

rohanvachheta:
- There was issue with one of the examples from https://react-select.com/components page where the dropdown indicator was getting grey color and due to which it wasn't seem good. Just fix that issue here with adding the valid color.

Screenshot (36)


rohanvachheta:
@Rall3n Could you review/merge this PR?

Issue #5159, opened 2022-04-20
docs(Sortable MultiSelect): replace react-sortable-hoc example by dnd-kit or another


SFromentarius:
In docs (Advanced) in **Sortable MultiSelect**, your advising to use `react-sortable-hoc`:
https://github.com/JedWatson/react-select/blob/fe7dab371290b87c425d9036b66febcf4d1beead/docs/pages/advanced/index.tsx#L51

But it's no longer recommended, from the lib itself:
https://github.com/clauderic/react-sortable-hoc#%EF%B8%8F-important-notice-for-new-consumers

They suggest to look on [`dnd-kit`](https://github.com/clauderic/dnd-kit) side.

Methuselah96:
Thanks for the issue, we should indeed update that example. Feel free to make a pull request if you have time.

SFromentarius:
Well noted, I'll see if I have time :)

codercodingthecode:
Opened a pr for this issue
https://github.com/JedWatson/react-select/pull/5212

PR #5212, opened 2022-06-02
Update the Sortable Multiselect component documentation

codercodingthecode:
This PR address the documentation for the Sortable Multiselect Component example. 
Due to the deprecation of the old Library [react-sortable-hoc](https://www.npmjs.com/package/react-sortable-hoc), recommended the use of the DnD-Kit library.

## Included in this PR.
- Add and use DnD-Kit Library
- Update documentation for Sortable Multiselect
- Update the sample array to add an id property
- Remove the old library.

## Address 
this issue https://github.com/JedWatson/react-select/issues/5159

## Sample using DnD
Kapture 2022-06-01 at 23 47 25


docs/data.ts
```diff
@@ -1,22 +1,23 @@
...
+  { id: 1, value: 'ocean', label: 'Ocean', color: '#00B8D9', isFixed: true },
+  { id: 2, value: 'blue', label: 'Blue', color: '#0052CC', isDisabled: true },
+  { id: 3, value: 'purple', label: 'Purple', color: '#5243AA' },
+  { id: 4, value: 'red', label: 'Red', color: '#FF5630', isFixed: true },
+  { id: 5, value: 'orange', label: 'Orange', color: '#FF8B00' },
+  { id: 6, value: 'yellow', label: 'Yellow', color: '#FFC400' },
+  { id: 7, value: 'green', label: 'Green', color: '#36B37E' },
+  { id: 8, value: 'forest', label: 'Forest', color: '#00875A' },
+  { id: 9, value: 'slate', label: 'Slate', color: '#253858' },
+  { id: 10, value: 'silver', label: 'Silver', color: '#666666' },
```
Rall3n, line 20: Using the already unique `value` as the `id` would be better IMO.

Rall3n (changes requested):
If `dnd-kit` is a replacement for `react-sortable-hoc`, then I would like the example to behave the same as with the latter.

It is currently very restrictive in comparison and cancels the drag if moving too far right.

immasandwich:
Got this working to match the existing documentation with the following changes:

1. Use `ValueContainer` instead of `MultiValueContainer` as the droppable area, and ensure that you apply `flex-grow: 1` styling to it so it consumes the maximum width of the React Select component.
2. Use `closestCorners` as the collision detection algorithm (on the `DndContext` component)

Thanks for doing most of the leg work!

dacarmor:
Hello, how is this matter? Is there any forecast to integrate the changes?

docs/examples/MultiSelectSort.tsx
```diff
@@ -1,92 +1,119 @@
...
 
-export default function MultiSelectSort() {
-  const [selected, setSelected] = React.useState<readonly ColourOption[]>([
+const MultiSelectSort = () => {
+  const [selected, setSelected] = React.useState<ColourOption[]>([
     colourOptions[4],
     colourOptions[5],
   ]);
 
   const onChange = (selectedOptions: OnChangeValue<ColourOption, true>) =>
```
jacobsfletch, line 79: `onChange` should be a `useCallback` function, like this: 

```js
const onChange = useCallback((selectedOptions: OnChangeValue<ColourOption, true>) => {}, [])
```

docs/examples/MultiSelectSort.tsx
```diff
@@ -1,92 +1,119 @@
...
+    setSelected([...selectedOptions]);
 
-  const onSortEnd: SortEndHandler = ({ oldIndex, newIndex }) => {
-    const newValue = arrayMove(selected, oldIndex, newIndex);
-    setSelected(newValue);
-    console.log(
-      'Values sorted:',
-      newValue.map((i) => i.value)
-    );
+  const onSortEnd = (event: DragEndEvent) => {
```
jacobsfletch, line 82: `onSortEnd` should be a `useCallback` function, like this (and maybe rename it `onDragEnd`: 

```js
const onDragEnd = useCallback((event: DragEndEvent)=> {}, [])
```

carlesgomalb: Is there any way I can do this? I tried to get it done in @codercodingthecode repository but i don't have permissions, neither in this repository.

codercodingthecode: Open a pr against that repo and we’ll update here.

codercodingthecode:
I’m no longer using this feature at work. So, I’d like to know if there’s is any interest in moving this forward before I put more of my time into it.
To anyone wanting to contribute, open a pr against current branch and will update here. 
I will resolve the current conflicts for now.

docs/examples/MultiSelectSort.tsx
carlesgomalb: done! #https://github.com/codercodingthecode/react-select/pull/1

oscarjg (approved):

codercodingthecode:
resolving conflicts after work today.

codercodingthecode:
can someone look into the CI Pipeline?

carlesgomalb (approved):

chrissysemens:
Hi, I have a few issues trying to use this example:

MultiValue is returning JSX and not components.MultiValue directly and so I receieve a type error:
image

Also, there's a distance prop being passed to Select which is not expected?

image

NotDemonix:
I suggest to remove `strategy={horizontalListSortingStrategy}`, works the same and is better if the component goes into 2 lines.
Next thing to change is `transform: CSS.Transform.toString(transform),` into `transform: CSS.Translate.toString(transform),`, this fixes the issue of element width if we have 2 diffrent widths for options. Would open a pr but it doesn't let me.

NotDemonix:
`distance={4}` doesn't seem to be doing anything

KubaBir:
For anyone looking to use this. There is a problem with using it on mobile which can be resolved by adding 
`touchAction: 'none'` to `MultiValue` styles

```
const MultiValue = (props: MultiValueProps<Item>) => {
  const onMouseDown: MouseEventHandler<HTMLDivElement> = (e) => {
    e.preventDefault()
    e.stopPropagation()
  }
  const innerProps = { ...props.innerProps, onMouseDown }
  const { attributes, listeners, setNodeRef, transform, transition } = useSortable({
    id: props.data.value,
  })
  const style = {
    transform: CSS.Translate.toString(transform),
    transition,
    touchAction: 'none',
  }

  return (
    <div style={style} ref={setNodeRef} {...attributes} {...listeners}>
      <components.MultiValue {...props} innerProps={innerProps} />
    </div>
  )
}
```

https://docs.dndkit.com/api-documentation/sensors/pointer#touch-action

orbtl:
This isn't exactly an issue that was raised here, but for future visitors that may have had the issue I ran into:

When using dnd-kit with react-select (specifically in my case with `AsyncCreateableSelect`), after dragging selected options to re-order, the focus behaved very strangely.  The active element remained correct (the input element), but the caret was removed so no further typing was possible until a user blurred and then re-focused the element.

It seems like dnd-kit's sensors attach a `selectionchange` listener that calls `document.getSelection().removeAllRanges()` on a 50ms delay after drags end, which was removing the selection/clearing the input caret.

The solution was to override the `removeTextSelection` implementation on the dnd-kit sensors.

So I added something like:
```
export class CaretPreservingMouseSensor extends MouseSensor {
  removeTextSelection() {}
}

export class CaretPreservingTouchSensor extends TouchSensor {
  removeTextSelection() {}
}
```

And then used those sensors instead of the base sensors in my `useSensors` call:
```
  const sensors = useSensors(
    useSensor(CaretPreservingMouseSensor, {activationConstraint: {distance: 4}}),
    useSensor(CaretPreservingTouchSensor, {activationConstraint: {distance: 4}}),
  );

  {...etc}
  <DndContext
          sensors={sensors}
  {...etc}
  />
```

This prevented the 50ms delayed text selection removal being applied and fixed the issue

PR #5246, opened 2022-06-27
Fix re-focus of component in Firefox if being disabled while focused

### LINKED ISSUE #4906: ValueContainer loose click event after isDisabled state
## Are you reporting a bug or runtime error?

A bug.

## Explanation
An `isDisabled` property is dynamically set then deactivated on a bare select.   
After it being deactivated, the value container (`ValueContainer`) can't receive any click events, or at least does not have any effect anymore (opening menu, etc), while the dropdown indicator still can...   
Focusing out of the select, then clicking again works as expected.   

## Reproduction
See https://codesandbox.io/s/smoosh-cloud-3mqis?file=/src/App.js

## Versions
`react-select@latest` as well as `4.x`

Rall3n:
Fixes #4906.

As explained in https://github.com/JedWatson/react-select/issues/4906#issuecomment-1068329821, Firefox does not fire blur if component becomes disabled while focused.

When re-enabling, Firefox by default re-focuses the component without firing the focus event, causing in the focused state not being properly set.

This PR fixes this behavior by setting `isFocused` state if component was previously disabled and if `inputRef` is currently active element.

### DIFF
--- a/.changeset/neat-hounds-repeat.md
+++ b/.changeset/neat-hounds-repeat.md
@@ -1,0 +1,5 @@
+---
+"react-select": patch
+---
+
+Fix re-focus of component in Firefox if being disabled while focused
--- a/.changeset/neat-hounds-repeat.md
+++ b/.changeset/neat-hounds-repeat.md
@@ -1,5 +1,5 @@
 ---
-"react-select": patch
+'react-select': patch
 ---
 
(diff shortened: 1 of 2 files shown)


aesy (approved):
LGTM, I can confirm that this fixes the issue in Firefox :+1:

lukebennett88 (approved):

Issue #4336, opened 2020-12-17
Misaligned menu when using menuPortalTarget and menuPlacement = auto


ArnaudWeyts:
# Are you reporting a bug or runtime error?
A Bug

For some reason the menu is misaligned slightly to the left when using the body as a `menuPortalTarget` and setting the `menuPlacement` to auto, I'm not sure why this is the case and it only seems to happen in a specific layout (select on the right bottom), it seems like it could be adjusting for a potential scrollbar or something. When you hover over the items, the menu suddenly jumps to the correct position.

https://codesandbox.io/s/react-select-v3-sandbox-forked-d5txb?file=/example.js

![Dec-17-2020 16-39-03](https://user-images.githubusercontent.com/10011712/102509392-c9c72900-4086-11eb-9380-9ea163e2bbae.gif)

Thanks for your work on the library!

ebonow:
Greetings @ArnaudWeyts ,

I am unable to replicate the described behavior. 

- What browser and version and what OS are you using? 
- Are you seeing this behavior on any other browsers? 
- Is it possible that the viewport was zoomed in/out at all?
  - It doesn't appear to be the case, but in my experience that can cause strange visual artifacting

ArnaudWeyts:
@ebonow Hey, thanks for getting back to me

* Happens for me on the latest version of chrome on macos (Version 87.0.4280.88), but we've also internally had reports of this on windows machines
* The codesandbox reproduces the bug on every browser (Firefox, Safari, Edge)
* The viewport is at a standard 100%, It happens both on my macbook pro 2015's retina screen (2880 × 1800p) and a regular 1920 x 1080p screen

ebonow:
I'm running Chrome Version 87.0.4280.88 on Mac OSX and I am unable to replicate. Can someone else perhaps take a look and verify what they see? 

https://user-images.githubusercontent.com/1007579/103964584-a553ef00-5110-11eb-88ec-17725db2c95d.mov

ArnaudWeyts:
@ebonow I've just discovered that this only seems to be the case when a mouse is connected to my macbook, and causes the scrollbars to be displayed. (As you can see in my gif, the scrollbar is displayed, but when using the trackpad, macbooks don't display scrollbars by default)

ebonow:
[...]

Issue #4159, opened 2020-08-05
Menu is re-rendering after scrolling the page and trying to click on an option


Wojtasik:
When using `react-select` with `menuPostion="fixed"` I've noticed a strange behaviour when trying to select an option after scrolling down.

The issue can be reproduced in this sandbox: https://codesandbox.io/s/react-codesandboxer-example-mxslt?file=/example.js

Steps:
- click on the select to open the options menu;
- scroll down the page (not the options menu) to the bottom (would be best if the select is no longer visible in the viewport);
- try to select any option that is not highlighted;
- the menu will re-render at different postion.

Is this an expected behaviour?

Here is what I found after digging into the code:
- when user hovers on an option in the menu, `onOptionHover` gets called https://github.com/JedWatson/react-select/blob/32ad5c040bdd96cd1ca71010c2558842d684629c/packages/react-select/src/Select.js#L1142-L1147 which calls `setState`;
- because the `setState` was called the menu is rerendered and freshly calculated offset is applied: https://github.com/JedWatson/react-select/blob/32ad5c040bdd96cd1ca71010c2558842d684629c/packages/react-select/src/components/Menu.js#L519-L521

Thanks,
Patryk.

P.S. Thanks for this awesome library!

ebonow:
Greetings @Wojtasik and Confirmed. 

It appears that scrolling the page while a fixed menu causes the menu position to get lost after an option is hovered. However, I am not certain what the expected/desired behavior is as the menu position is being specified as fixed and the page is being scrolled down. 

I will mark this is needs-review as its unclear to me what the expected solution should be.

Methuselah96:
Hey @Wojtasik! I know it's been a long time, but we're finally looking into old menu positioning issues. Can you help me understand what your use case for using `menuPosition="fixed"` is? I'm trying to determine in what scenarios someone would use `menuPosition="fixed"` and expect the position to change.

Wojtasik:
Hi @Methuselah96 - thanks for getting back to me with this one! 

> Can you help me understand what your use case for using menuPosition="fixed" is?
I cannot remember the use case exactly, but I think it was related to iPad (tablet) usage.

> I'm trying to determine in what scenarios someone would use menuPosition="fixed" and expect the position to change.
Agreed. The bug was about the menu jumping around after user has scrolled down. Just checked and still can see this bug happening in the code sandbox example in the ticket description.

Please let me know if it helps or if I can do anything else to help with this one. Thanks!

[...]

Issue #3533, opened 2019-05-01
MenuPortal - horizontal position of menu is incorrect when browser is zoomed


davidlucas:
A portaled menu when viewed on iOS Safari in a zoomed state on an iPhone/iPad has the wrong horizontal position.

This can be reproduced using the example on the website - https://react-select.com/advanced#portaling

bladey:
<img width="463" alt="IgFFKB" src="https://user-images.githubusercontent.com/737821/84865136-11123400-b0bb-11ea-8396-0144403cf75b.png">

AmrineA:
This can also be replicated on Chrome using the portal example. If you open the portal drop down and then resize the window horizontally, you will see it move.

![image](https://user-images.githubusercontent.com/6352189/88814346-f33ef100-d187-11ea-8df9-be2ef01a293a.png)

We've noticed this issue also when a header resizes with scrolling. For example, the header of a window might be 200px until the user scrolls down and then it resizes to 100px. At that point, the menu items "detach" from the menu control. It appears they will "reattach" when the mouse pointer hovers over the menu items.

andrzejwalk:
Any workaround for this?

ebonow:
Confirmed that this behavior exists but not sure what if anything can be done. It appears that there is no standard in place to identify if a viewport has been "zoomed"
https://css-tricks.com/can-javascript-detect-the-browsers-zoom-level/

It should also be noted that the menu "snaps" back into its expected position when the menu/option is hovered. This leads me to believe that it may be related to some other behavior that prevents the menu from staying positioned as expected:
#4336
#4375
#4159

They all share similar behavior in that the scroll position is offset from what is expected but then snaps into place once the menu is hovered. I think it may be worth consolidating these into a single use-case to better identify the root.

In this particular use-case though, I'm not sure how to detect a change in zoom or if there is some other means of positioning that should be considered.

ebonow:
Similar and perhaps related behavior
#3499

odusseys:
getting the same issue with no zoom (on desktop, chrome, mac)

marius-ionescu:
Not sure if this helps but, one workaround you can try is to set a **negative marginThreshold** to offset the difference between the zoomed (height or width) and actual size.
[...]

PR #5256, opened 2022-07-04
Auto-update menu position when using menu portalling

Methuselah96:
Fixes https://github.com/JedWatson/react-select/issues/3533.
Fixes https://github.com/JedWatson/react-select/issues/4088.
Fixes https://github.com/JedWatson/react-select/issues/4159.
Fixes https://github.com/JedWatson/react-select/issues/4336.

This PR makes it so that when using menu portalling, event listeners are attached to the scrolling parents of the Select component in order to update the position after scroll or resize of one of its parents. This is accomplished using [Floating UI's `autoUpdate` function](https://floating-ui.com/docs/autoUpdate) and is heavily inspired by [their `useFloating` hook](https://floating-ui.com/docs/react-dom#usage) (but doesn't directly use it in order to make this more minor of a change).

Previously I was trying to rewrite all of menu positioning, but I decided to go with this more modest approach to get this resolved more quickly and make it less of a breaking change. However a complete rewrite of menu positioning might still happen in the future if proven necessary.

I attempted to make this change as backwards compatible as possible. The only change should be that there are event listeners attached to the Select component's scroll parents if using menu portalling. It could be argued that this is either a bug fix, a new feature, or a breaking change. My leaning would be to release this in a minor version since I can't think of a use case where one would not want the menu position to update when the Control element moves.

### DIFF
--- a/.changeset/soft-bags-shave.md
+++ b/.changeset/soft-bags-shave.md
@@ -1,0 +1,5 @@
+---
+'react-select': minor
+---
+
+Auto-update menu position when using menu portalling
--- a/package.json
+++ b/package.json
@@ -176,6 +176,6 @@
     "./dist/react-select.esm.js": "./dist/react-select.browser.esm.js"
   },
   "resolutions": {
-    "csstype": "^3.0.2"
+    "csstype": "^3.1.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -176,6 +176,6 @@
     "./dist/react-select.esm.js": "./dist/react-select.browser.esm.js"
   },
   "resolutions": {
-    "csstype": "^3.1.0"
+    "csstype": "^3.0.2"
   }
 }
--- a/packages/react-select/package.json
+++ b/packages/react-select/package.json
@@ -13,7 +13,7 @@
     "@babel/runtime": "^7.12.0",
     "@emotion/cache": "^11.4.0",
     "@emotion/react": "^11.8.1",
-    "@floating-ui/dom": "^0.5.4",
+    "@floating-ui/dom": "^1.0.1",
     "@types/react-transition-group": "^4.4.0",
     "memoize-one": "^5.0.0",
     "prop-types": "^15.6.0",
--- a/packages/react-select/package.json
+++ b/packages/react-select/package.json
@@ -13,6 +13,7 @@
     "@babel/runtime": "^7.12.0",
     "@emotion/cache": "^11.4.0",
     "@emotion/react": "^11.8.1",
+    "@floating-ui/react-dom": "^0.7.2",
     "@types/react-transition-group": "^4.4.0",
     "memoize-one": "^5.0.0",
     "prop-types": "^15.6.0",
--- a/packages/react-select/package.json
+++ b/packages/react-select/package.json
@@ -13,11 +13,12 @@
     "@babel/runtime": "^7.12.0",
     "@emotion/cache": "^11.4.0",
     "@emotion/react": "^11.8.1",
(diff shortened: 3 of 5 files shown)


ryan-castner:
@Methuselah96 I tried out this branch and while I am no longer having the menu render off screen, I am noticing with `menuPosition="auto"` that it is always positioning above when there is space to position it below:

image

Methuselah96:
@ryan-castner Thanks for trying out the branch. I am not able to reproduce the issue you're experiencing. I've pushed a test of `menuPlacement="auto"` to https://github.com/JedWatson/react-select/tree/use-floating-with-test and get this result:

image

You can try it out by checking out that branch, starting the docs with `yarn start`, and navigating to `localhost:8000/cypress-tests`. Let me know if you can provide any more info regarding the issue you're experiencing.

lukebennett88 (approved):
This looks great @Methuselah96!
I couldn't replicate the issue that @ryan-castner mentioned.
I agree that a minor makes sense here 👍

jossmac (approved):

Methuselah96:
This has been released in [`react-select@5.5.0`](https://github.com/JedWatson/react-select/releases/tag/react-select%405.5.0). We anticipate this change to be non-breaking which is why we're releasing it in a minor release, but if it turns out that this change is (non-trivially) breaking, it's possible we might revert the change and re-release it in a major version.

alk-sdavid:
Hi @Methuselah96, thank you for this change, but I can confirm it's a major release which introduces some breaking change since it relies on `@floating-ui/dom` which relies itself on `ResizeObserver`. It breaks old browsers which does not support `ResizeObserver`. So I guess you should release it as version 6.0.0, and indicate this dependency in the changelog and the documentation.

Thanks again for your work and I hope it helps.

Methuselah96:
@alk-sdavid Which browsers are you trying to support? As far as I know we don't have any browser support policy at the moment besides not IE 11 (mentioned in [the v5 upgrade guide](https://react-select.com/upgrade#from-v4-to-v5)).

Methuselah96:
Looks like Safari on iOS 12.5 and Android Browser 4.4.4 might be the two browsers that don't support it if we're aiming supporting browsers with >.25% usage.

I've created https://github.com/JedWatson/react-select/pull/5381 to disable the `ResizeObserver` for now, and we can (potentially) re-enable it and make our browser support policy clear in the next major version.

alk-sdavid:
Hi, yes it was not IE11, I just tested react-select on browserstack with some browser which does not support ResizeObserver (https://caniuse.com/resizeobserver). I do not have any issue about it, but I was just thinking it would be better to make a major release and to make it clear in the changelog and the documentation.
FYI, I noticed the ResizeObserver because some of my company unit tests failed because of it on a Dependabot pull-request, and then I dug a bit…

kimi-truong:
Hi @Methuselah96, I've just upgraded to the latest version `v5.5.0` and all unit tests failed because of `ResizeObserver`
```
ReferenceError: ResizeObserver is not defined
```
Do you have any workaround on it?

Methuselah96:
It should be resolved by https://github.com/JedWatson/react-select/pull/5381 which will hopefully be released in the next couple days.

Methuselah96:
@alk-sdavid @kimi-truong This should be resolved in [`react-select@5.5.1`](https://github.com/JedWatson/react-select/releases/tag/react-select%405.5.1), let me know if you have any more issues.

kimi-truong:
@Methuselah96 thanks so much 🙏 , it works now.

Issue #5251, opened 2022-06-30
UI issue when using the makeAnimated()


Alejandro-Marco:
The select item container is longer that its content when using the makeAnimation() in components

Expected Behavior:
<img width="599" alt="1" src="https://user-images.githubusercontent.com/78371908/176674617-405202ec-dc93-4ed4-a615-d8ae40605621.png">

Actual Behavior
<img width="581" alt="5" src="https://user-images.githubusercontent.com/78371908/176674848-c362c133-1c47-443b-9105-ca130a2ad733.png">

Steps to reproduce:
1) Access https://react-select.com/home#animated-components
2) Remove all added items and type "O" ( any other character would also do as long you can select any item )
<img width="641" alt="2" src="https://user-images.githubusercontent.com/78371908/176675112-241fe15f-c18c-4c21-aa3d-a689d8cae7ac.png">
3) Select any item, the container of the item is longer than the content 
<img width="607" alt="3" src="https://user-images.githubusercontent.com/78371908/176675205-8ca667f9-b5e7-4907-9f00-8f694429898e.png">

This issue only occurs on the first item, any other item added after does not experience this.

Adding another item:
1) <img width="589" alt="4" src="https://user-images.githubusercontent.com/78371908/176675320-28c8419b-eac5-4d9e-a44a-e8d2e138a663.png">
2) <img width="581" alt="5" src="https://user-images.githubusercontent.com/78371908/176675342-838c5cb4-b3ba-4589-933b-fb0c5da8b828.png">

The issue only occurs when using makeAnimated(), so not that big of an issue. Really just wanted to make use of the animation.

vicnguyenqt:
Same issue here
![image](https://user-images.githubusercontent.com/93513978/189339541-3d9ae5c5-293c-481e-b6cc-89c824e61794.png)

jorge30fm:
I'm having this same issue. How can it be fixed?

Alireza17224:
Same issue !

lukebennett88:
Hi @Alireza17224, can you confirm that you're still seeing this issue? What version of react-select are you using? It _should_ be fixed in 5.5.2.

Alireza17224:
Hi yes ! I am using this version
react-select: "^5.4.0",
I will update it to version 5.5.2
Thank you very much

Alireza17224:
> Hi yes ! I am using this version react-select: "^5.4.0", I will update it to version 5.5.2 Thank you very much

FIXED BY UPDATING !

PR #5339, opened 2022-09-11
fixed selected option width issue

shreekeshmurkar:
Fixed following issue of width for selected option
https://github.com/JedWatson/react-select/issues/5251

Before fix:
image

After fix:
image

### DIFF
--- a/.changeset/heavy-plums-shake.md
+++ b/.changeset/heavy-plums-shake.md
@@ -1,0 +1,5 @@
+---
+'react-select': patch
+---
+
+Corrected width of selected option in multi-select options.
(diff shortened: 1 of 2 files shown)


shreekeshmurkar:
@lukebennett88 Please review

lukebennett88:
Thanks for your PR @shreekeshmurkar. 
Unfortunately this method no longer causes the element to be animated.

lukebennett88 (changes requested):
Thanks for your PR @shreekeshmurkar.
Unfortunately this method no longer causes the element to be animated, so we will have to think of another way to fix this issue.

lukebennett88:
I've taken a shot at fixing this bug in another PR
https://github.com/JedWatson/react-select/pull/5376

lukebennett88:
This has been fixed in [`react-select@5.5.2`](https://github.com/JedWatson/react-select/releases/tag/react-select%405.5.2).
Closing as a duplicate.

PR #5347, opened 2022-09-16
Make scroll lock div work on a document context it belongs to

### LINKED ISSUE #5346: menuShouldBlockScroll not working properly when the select is mounted in a separate window but execution context is in main window
## Are you reporting a bug or runtime error?

A **bug**.
react-select version `5.4.0`

I have an app that uses React Portals to mount components in separate windows/popups (we use [react-new-window
](https://www.npmjs.com/package/react-new-window) package to achieve that.

The thing is that in such a use case, the markup and components are rendered in the external HTML page (popup) but the javascript execution context is still in the main window so referring objects like `window` or `document` could produce errors because the document on the external page (where the component is rendered) is different than the one where the javascript executes.

One example of such an issue is when we use `menuShouldBlockScroll` config. When it's used, the library adds a [`fixed` positioned element](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/internal/ScrollManager.tsx#L48) to the page and attaches `onClick` event on it to blur the select. The issue is that the event listener looks for `document.activeElement` which is wrong in my use case as it refers to the main document, not the popup one where the component is mounted.

I found a simple solution for that one and will push a pull request with it soon.

aszmyd:
Fixes #5346


lukebennett88:
Thanks for your PR @aszmyd.
Could you please create a CodeSandbox example showing the use case you're trying to solve. The code looks pretty straight forward, but I'd like to test it a little more 🙂
Also this PR will require a changeset in order to get merged. If you haven't used changesets before, instructions on how to do that are available [here](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).

aszmyd:
I've updated the branch, included the changeset, and prepared two CodeSandbox instances to show the issue:
- [Sandbox running on react-select 5.7.4](https://codesandbox.io/s/react-codesandboxer-example-forked-3f5mjd?file=/index.js)
- [Sandbox running on commits from this PR - df685f08](https://codesandbox.io/s/react-codesandboxer-example-forked-9x5c28?file=/index.js) - it doesn't include the commit with the changeset but its irrelevant to the code logic

I also recorded a small demo of the bug ([link](https://youtu.be/ZjcLPnzHLTs)) - first part shows the behavior on the 5.7.4 version and second (on the second tab of the browser) shows the fix. The bug is that it's not possible to close the opened select by clicking outside of it when the `menuShouldBlockScroll` option is set to `true`.

I've used [react-new-window](https://www.npmjs.com/package/react-new-window) package to make a new document context case (it simply opens a new popup window and uses React Portals to render component on it) as I'm using this package also in my current production codebase.

As mentioned in the issue description, the problem only occurs when the `menuShouldBlockScroll` is set tu true.

Please let me know if there is anything else I do to make this merged. Our application has no workaround for this bug and it's quite annoying for the our users currently.

Thanks

lukebennett88 (approved):
Thanks so much for the PR, and for the thorough documentation and examples @aszmyd.
Everything looks good to me, so I'm going to approve and merge it.

PR #5358, opened 2022-10-03
docs: update links in readme

SebastiaanNijland:
- Update the link to point directly to the reactjs website, instead of a redirected url.
- Add https to the keystonejs link, so it doesn't need to redirect to https.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 # React-Select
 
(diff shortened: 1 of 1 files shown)


lukebennett88 (approved):
Thanks @SebastiaanNijland

PR #5381, opened 2022-10-14
Disable use of ResizeObserver for menu auto-updating

Methuselah96:
Resolves https://github.com/JedWatson/react-select/pull/5256#issuecomment-1278916683.

We do not have an official browser support policy, but it would be nice to delay relying on `ResizeObserver` for a major bump and make our official browser support policy clear in the next major.

I do not believe we anticipate the `ResizeObserver` getting triggered in many situations anyway since the menu itself shouldn't be resizing. The scroll events are much more critical to auto-updating working properly.

### DIFF
--- a/.changeset/new-sloths-wink.md
+++ b/.changeset/new-sloths-wink.md
@@ -1,0 +1,5 @@
+---
+'react-select': patch
+---
+
+Disable use of ResizeObserver for menu position auto-updating to avoid potential breaking changes.
--- a/packages/react-select/src/components/Menu.tsx
+++ b/packages/react-select/src/components/Menu.tsx
@@ -647,7 +647,8 @@ export const MenuPortal = <
       cleanupRef.current = autoUpdate(
         controlElement,
         menuPortalRef.current,
-        updateComputedPosition
+        updateComputedPosition,
+        { elementResize: false }
       );
     }
   }, [controlElement, updateComputedPosition]);


lukebennett88 (approved):

desbrowne:
Hi @Methuselah96 

For the major bump release, I wonder if you could use the following so that older browsers could still be supported without requiring a polyfill?
`{ elementResize: 'ResizeObserver' in window }`

Methuselah96:
Thanks for the suggestion, I'm not even sure we have any reason to re-enable it at this point. I'm having trouble trying to think of a situation where the ResizeObserver would be triggered but the other event listeners (scroll and resize on parent) would not, so we might be able to just do without it.

desbrowne:
Thanks. I'm not super familiar with MenuPlacer, but would async loading menu items or changing menu-item heights cause this?

Methuselah96:
There are two parts of menu positioning:
1. If `menuPlacement="auto"`, determine whether it should go on the top or bottom. Also scroll the menu into view if possible.
2. If using portalling (or fixed positioning), update the menu's absolute position based on the position of the control element.

This PR only addresses auto-updating the second part of menu positioning. I don't think we have determined yet whether we want to be auto-updating the first part.

The second part does not rely on the menu's height at the moment, so async loading menu items or changing menu-item heights would not affect the auto-updating of the menu position the way that it's currently implemented.

PR #5393, opened 2022-10-19
Change examples from `class` to `functional` components

nderkim:
This PR updates the examples to use `functional` components

### DIFF
--- a/docs/examples/AsyncCallbacks.tsx
+++ b/docs/examples/AsyncCallbacks.tsx
@@ -1,11 +1,7 @@
-import React, { Component } from 'react';
+import React from 'react';
 
 import AsyncSelect from 'react-select/async';
 import { ColourOption, colourOptions } from '../data';
-
-interface State {
-  readonly inputValue: string;
-}
 
 const filterColors = (inputValue: string) => {
   return colourOptions.filter((i) =>
@@ -22,24 +18,6 @@ const loadOptions = (
   }, 1000);
 };
 
-export default class WithCallbacks extends Component<{}, State> {
-  state: State = { inputValue: '' };
-  handleInputChange = (newValue: string) => {
-    const inputValue = newValue.replace(/\W/g, '');
-    this.setState({ inputValue });
-    return inputValue;
-  };
-  render() {
-    return (
-      <div>
-        <pre>inputValue: "{this.state.inputValue}"</pre>
-        <AsyncSelect
-          cacheOptions
-          loadOptions={loadOptions}
-          defaultOptions
-          onInputChange={this.handleInputChange}
-        />
-      </div>
-    );
-  }
-}
+export default () => (
+  <AsyncSelect cacheOptions loadOptions={loadOptions} defaultOptions />
+);
--- a/docs/examples/AsyncMulti.tsx
+++ b/docs/examples/AsyncMulti.tsx
@@ -1,11 +1,7 @@
-import React, { Component } from 'react';
+import React from 'react';
 
 import AsyncSelect from 'react-select/async';
 import { ColourOption, colourOptions } from '../data';
-
-interface State {
-  readonly inputValue: string;
-}
 
 const filterColors = (inputValue: string) => {
   return colourOptions.filter((i) =>
@@ -20,21 +16,11 @@ const promiseOptions = (inputValue: string) =>
     }, 1000);
   });
 
-export default class AsyncMulti extends Component<{}, State> {
-  state: State = { inputValue: '' };
-  handleInputChange = (newValue: string) => {
-    const inputValue = newValue.replace(/\W/g, '');
-    this.setState({ inputValue });
-    return inputValue;
-  };
-  render() {
-    return (
-      <AsyncSelect
-        isMulti
-        cacheOptions
-        defaultOptions
-        loadOptions={promiseOptions}
-      />
-    );
-  }
-}
+export default () => (
+  <AsyncSelect
+    isMulti
+    cacheOptions
+    defaultOptions
+    loadOptions={promiseOptions}
+  />
+);
--- a/docs/examples/BasicSingle.tsx
+++ b/docs/examples/BasicSingle.tsx
@@ -1,4 +1,4 @@
-import React, { Component, Fragment } from 'react';
+import React, { useState } from 'react';
 
 import Select from 'react-select';
 import { colourOptions } from '../data';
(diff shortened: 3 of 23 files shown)


docs/examples/AsyncCallbacks.tsx
```diff
@@ -22,24 +18,6 @@ const loadOptions = (
   }, 1000);
 };
 
-export default class WithCallbacks extends Component<{}, State> {
-  state: State = { inputValue: '' };
```
nderkim, line 26: I've removed the state logic to simplify the example. I don't think it was necessary to demonstrate how to use the async callbacks api.

docs/examples/CreatableAdvanced.tsx
```diff
@@ -25,50 +18,30 @@ const defaultOptions = [
...
-    value: undefined,
-  };
-  handleChange = (
-    newValue: OnChangeValue<Option, false>,
-    actionMeta: ActionMeta<Option>
-  ) => {
-    console.group('Value Changed');
-    console.log(newValue);
-    console.log(`action: ${actionMeta.action}`);
-    console.groupEnd();
```
nderkim, line 41: I've removed the console logs to simplify the example. I don't think they were necessary for the purpose of this example.

docs/examples/CreatableInputOnly.tsx
```diff
@@ -17,60 +16,33 @@ const createOption = (label: string) => ({
...
-    value: [],
-  };
-  handleChange = (
-    value: OnChangeValue<Option, true>,
-    actionMeta: ActionMeta<Option>
-  ) => {
-    console.group('Value Changed');
-    console.log(value);
-    console.log(`action: ${actionMeta.action}`);
-    console.groupEnd();
```
nderkim, line 37: I've removed the console logs to simplify the example. I don't think they were necessary for the purpose of this example.

docs/examples/CreatableMulti.tsx
```diff
@@ -1,26 +1,6 @@
...
 
-export default class CreatableMulti extends Component<{}> {
-  handleChange = (
-    newValue: OnChangeValue<ColourOption, true>,
-    actionMeta: ActionMeta<ColourOption>
-  ) => {
-    console.group('Value Changed');
-    console.log(newValue);
-    console.log(`action: ${actionMeta.action}`);
-    console.groupEnd();
```
nderkim, line 15: I've removed the console logs to simplify the example. I don't think they were necessary for the purpose of this example.

docs/examples/CreatableSingle.tsx
```diff
@@ -1,33 +1,6 @@
...
 
-export default class CreatableSingle extends Component {
-  handleChange = (
-    newValue: OnChangeValue<ColourOption, false>,
-    actionMeta: ActionMeta<ColourOption>
-  ) => {
-    console.group('Value Changed');
-    console.log(newValue);
-    console.log(`action: ${actionMeta.action}`);
-    console.groupEnd();
```
nderkim, line 15: I've removed the console logs to simplify the example. I don't think they were necessary for the purpose of this example.

docs/examples/DefaultOptions.tsx
```diff
@@ -20,20 +16,10 @@ const promiseOptions = (inputValue: string) =>
     }, 1000);
   });
 
-export default class WithPromises extends Component<{}, State> {
-  state: State = { inputValue: '' };
```
nderkim, line 24: I've removed the state logic from this example because it wasn't necessary for the purpose of this example.

docs/examples/Experimental.tsx
```diff
@@ -183,73 +183,47 @@ const Option = (props: OptionProps<DateOption, false>) => {
...
-export default class Experimental extends Component<{}, State> {
-  state: State = {
-    value: defaultOptions[0] as DateOption,
-  };
-  handleChange = (value: DateOption | null) => {
-    this.setState({ value });
-  };
-  render() {
-    const { value } = this.state;
-    const displayValue =
```
nderkim, line 246: I've removed this to simplify the example.

docs/examples/AsyncMulti.tsx
```diff
@@ -20,21 +16,11 @@ const promiseOptions = (inputValue: string) =>
     }, 1000);
   });
 
-export default class AsyncMulti extends Component<{}, State> {
-  state: State = { inputValue: '' };
-  handleInputChange = (newValue: string) => {
-    const inputValue = newValue.replace(/\W/g, '');
-    this.setState({ inputValue });
-    return inputValue;
```
nderkim, line 28: I've removed this because it wasn't being used

docs/examples/BasicSingle.tsx
```diff
@@ -10,86 +10,65 @@ const Checkbox = ({ children, ...props }: JSX.IntrinsicElements['input']) => (
...
 
-interface State {
-  readonly isClearable: boolean;
-  readonly isDisabled: boolean;
-  readonly isLoading: boolean;
-  readonly isRtl: boolean;
-  readonly isSearchable: boolean;
-}
+export default () => {
+  const [isClearable, setIsClearable] = useState(false);
```
lukebennett88, line 14: `isClearable` and `isSearchable` should start off as `true`

docs/examples/Experimental.tsx
```diff
@@ -183,73 +183,47 @@ const Option = (props: OptionProps<DateOption, false>) => {
...
-}
+  return (
+    <Select<DateOption, false>
+      {...props}
+      components={{ Group, Option }}
+      filterOption={null}
+      isMulti={false}
+      isOptionSelected={(o, v) => v.some((i) => i.date.isSame(o.date, 'day'))}
+      maxMenuHeight={380}
+      onChange={props.onChange}
```
lukebennett88, line 213: This is spread on the props that are spread in above

docs/examples/Experimental.tsx
```diff
@@ -183,73 +183,47 @@ const Option = (props: OptionProps<DateOption, false>) => {
...
+      onInputChange={handleInputChange}
+      options={options}
+      value={props.value}
```
lukebennett88, line 216: This is spread on the props that are spread in above

docs/examples/Experimental.tsx
lukebennett88: I kinda like it for this example ¯\_(ツ)_/¯
`value.value.toString()` is a bit grim though — maybe `const [dateOption, setDateOption] = useState<DateOption | null>(defaultOptions[0] as DateOption)` would be better?

lukebennett88 (approved):

docs/examples/BasicSingle.tsx
nderkim: good catch 👍

docs/examples/Experimental.tsx
nderkim: 🤔 Hm. Isn't that what i'm already doing in [line 222](https://github.com/JedWatson/react-select/pull/5393/files#diff-b0383a87b0483e38ff709daf238f585c7c7d5cac514b2bd81d1a6ae4c9b08ce4R222) (except i've just left it named as `value` instead of renaming it to `dateOption`)?

docs/examples/Experimental.tsx
nderkim: This is how it was in the original example. I'm just gonna leave it they way it was. Personally, i think the props spread should go and the `onChange` should stay (it's more explicit), but im too scared to delete the props spread in case i break something 😰

PR #5408, opened 2022-10-20
Set up Storybook with Vite builder

lukebennett88:
Test to see if Vite Storybook builder interferes with the repo less than the webpack one.


lukebennett88:
I got the webpack builder working, so there is no need for this anymore.

PR #5409, opened 2022-10-21
Update @preconstruct/cli

lukebennett88:
Updates to the latest version of Preconstruct as the version we were using is quite old (version two came out two years ago).
This requires moving some important files around, so we want to be extra careful that this outputs the same entrypoints etc so that it's not a breaking change.


.changeset/tall-dancers-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-select': minor
```
nderkim, line 2: should this be `patch` since it's not changing the api?

lukebennett88: I'm not 100%, I was trying to play it safe call it a minor as some things that aren't necessarily part of the API contract _might_ break after this gets merged.
e.g. if someone imports something directly from the `dist` folder.
This might even end up being a _major_ change (in which case, I guess we can't merge it) as I just found this section in the docs:
https://react-select.com/typescript#custom-select-props

lukebennett88: This PR would change this snippet from the docs:
```diff
- declare module 'react-select/dist/declarations/src/Select' {
+ declare module 'react-select/dist/declarations/src/base' {
  export interface Props<
    Option,
    IsMulti extends boolean,
    Group extends GroupBase<Option>
  > {
    myCustomProp: string;
  }
}
```

It would also change:

```diff
-import AnimatedSelect from 'react-select/animated/dist/react-select.cjs.js.cjs'
+import AnimatedSelect from 'react-select/animated/dist/react-select-animated.cjs'
```

But people shouldn't be importing directly from the dist folder for this sort of thing anyway.

nderkim: I think we need a decision on whether `react-select/dist/*` is considered part of the api, and if it is, then we should bump `major` version, otherwise if it isn't, it's only `patch`.

Alternatively, you could just leave the file structure the way it is (ie. don't move `Select.tsx` -> `base/Select.tsx`).
Although, i'm not sure how `preconstruct` would feel about the casing when it comes to `Creatable` and `Async` (because their entrypoints need to be the same name just lowercased)

lukebennett88: I updated the PR to _not_ move any files (and just add a bunch of folders and index files) so the PR is much smaller now.

The first example in my comment above should still work.

The second one I don't think I can fix, but I had a look and I don't think that would be a major problem. 
The only files built in the extra entrypoints are different build targets for the same file, and since none of the files have moved, the contents of `react-select/declarations/src` will be the same (apart from the new index files).

As such, I've switched the changeset to be a patch.

nderkim (approved):
Nice work! 🍉

PR #5414, opened 2022-10-21
Tidy up storybook and add more examples

lukebennett88:


storybook/.storybook/main.ts
```diff
@@ -6,6 +6,7 @@ const config: StorybookConfig = {
     builder: 'webpack4',
   },
   features: {
+    postcss: false,
```
lukebennett88, line 9: Hides a warning in the console about postcss.
I might turn it back on in a future PR to add a Tailwind styled story.

package.json
```diff
@@ -42,7 +42,7 @@
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^10.1.0",
     "babel-jest": "^23.6.0",
-    "babel-loader": "^8.0.0",
+    "babel-loader": "^8.2.5",
```
nderkim, line 45: Just curious, what was the need behind the version upgrades?

nderkim (approved):

package.json
lukebennett88: In the end I didn't actually need to.

Prior to https://github.com/JedWatson/react-select/pull/5418 I couldn't use yarn to add dependencies to Storybook (I had to manually edit the `package.json` file or I'd get an error).

Once that PR got merged, I could add them with yarn — which updated them to the latest version, and manypkg bumped them for me.

I could have rolled them back, but I figured react-select doesn't depend on these so they were safe ¯\_(ツ)_/¯

PR #5422, opened 2022-10-24
Add Tailwind example to Storybook

lukebennett88:
Add example styled with Tailwind to Storybook:

PixelSnap 2022-10-24 at 15 19 30@2x

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,3 +1,6 @@
 [build]
   command = "yarn build:docs"
   publish = "docs/dist"
+
+[build.environment]
+  YARN_VERSION = "1.22.19"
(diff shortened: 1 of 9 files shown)


netlify.toml
```diff
@@ -1,3 +1,6 @@
 [build]
   command = "yarn build:docs"
   publish = "docs/dist"
+
+[build.environment]
```
lukebennett88, line 5: `@storybook/addon-postcss` has an engines key in the `package.json` file that was causing the build to fail for the docs unless I updated this.

nderkim (approved):
🍥

Issue #5367, opened 2022-10-06
Dynamically updating the option list rests the focus to the first element.


MoritzKn:
When dynamically updating the options list rests the focus to the first element. 
Even though the options are the same, just a new object reference.

In my use case, this happens because the select controls a filter and the results change when selecting options. The filter options have the count (facet filtering) of the results associated with the filter option in their label. So when updating the results we also update the options to update the counts. This means that selecting an option with the keyboard triggers an update of the results, which leads to the recomputation of the options and shortly after the focus is set to the first option in the results list.

This makes keyboard usage very annoying.

This is a v5 bug: https://codesandbox.io/s/react-select-v5-sandbox-forked-s58hcn?file=/example.js

jcorevnue2022:
Hi @MoritzKn, I am currently encountering the same issue.

Did you have any workaround?

MoritzKn:
Unfortunately not. Can’t really think of any workaround for this either.

MoritzKn:
@jcorevnue2022 actually, I found a (very hacky) workaround:

```tsx
/**
 * Workaround for React Select issue "Dynamically updating the option list rests the focus to the first element"
 *
 * See: https://github.com/JedWatson/react-select/issues/5367
 */
function useFixReactSelectFocus<Option extends { value: string }>(
  selectRef: RefObject<SelectClass<Option, true>>,
  options: Option[]
) {
  // HACK: This accesses the select class via a ref
  const select = selectRef.current;
  if (select && select.state.focusedOption) {
    const focusedOptionValue = select.state.focusedOption.value;
    const focusedOption =
      options.find((o) => o.value === focusedOptionValue) || null;

    // HACK: manipulating the state without setState is evil (but needed here as we are mid-transition)
    select.state.focusedOption = focusedOption;
  }
}
```

Then in your component:
```tsx
import { useRef } from "react";
import type SelectClass from "react-select/dist/declarations/src/Select";

function MyComponent() {
  const selectRef = useRef<SelectClass<FilterOption, true>>(null);

  // [...]

  useFixReactSelectFocus(selectRef, options);

  return (
    <Component
      selectRef={selectRef}
      options={options}
    />
  );
}
```

linaia:
@MoritzKn would it be rather `ref={selectRef}` instead of `selectRef={selectRef}` ?

jukkahuuskonen:
[...]

PR #5425, opened 2022-10-24
Fix isSelected in accessibility LiveRegion

### LINKED ISSUE #5424: Selected options are not announced as selected if the object reference changed AriaLive
See:
https://codesandbox.io/s/react-select-v5-sandbox-forked-gzdeq6?file=/example.js

As far as I can tell this is due to this code:
https://github.com/JedWatson/react-select/blob/da617c51b810c67d809dfde9cb9f3b5815313d91/packages/react-select/src/components/LiveRegion.tsx#L112-L116

In most places, the object relevance is not relevant. Following this pattern, I think this should look like this:
```rsx
    const isSelected = !!(
      focusedOption &&
      selectValue &&
      selectValue.some(option => option.value === focusedOption.value)
    );
```

I'll send a PR.

MoritzKn:
Fixes #5424


packages/react-select/src/components/LiveRegion.tsx
```diff
@@ -112,7 +112,7 @@ const LiveRegion = <
     const isSelected = !!(
       focusedOption &&
       selectValue &&
-      selectValue.includes(focusedOption)
+      selectValue.some(option => option.value === focusedOption.value)
```
Rall3n, line 115: Should use `getOptionValue` from `selectProps` to ensure custom option format compatibility.

Rall3n (changes requested):

packages/react-select/src/components/LiveRegion.tsx
```diff
@@ -112,7 +113,7 @@ const LiveRegion = <
     const isSelected = !!(
       focusedOption &&
       selectValue &&
-      selectValue.includes(focusedOption)
+      selectValue.some(option => getOptionValue(option.value) === getOptionValue(focusedOption))
```
Rall3n, line 116: `... getOptionValue(option) ...`

Rall3n (changes requested):

packages/react-select/src/components/LiveRegion.tsx
MoritzKn: Arg, yes sorry, shouldn't make these quick changes without thinking about it

MoritzKn:
Firstly sorry for the half-harded PR. I thought this was a trivial change that I could do in the GitHub UI. And as it is most times, it was not. 🙃 

But there is another point I'm only now realizing:
Maybe the `focusedOption` should simply not be stored as an object reference but as the value of the focused option.
I.e. `focusedOptionValue`.

#5367 is another issue with the same root cause.

https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.tsx#L1397-L1402
https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.tsx#L1833-L1835

Or alternatively a isOptionFocused analogous to isOptionSelected and isOptionDisabled?

Rall3n (approved):

SetupCoding:
Any chance we can get this merged?

PR #5430, opened 2022-10-25
Fix focused option if `defaultMenuIsOpen` is set

### LINKED ISSUE #4013: 3.1.0: Selected value not focused/visible when using defaultMenuIsOpen or menuIsOpen
Example: https://codesandbox.io/s/zealous-leaf-58ett?file=/src/index.js

@Tirzono Here you go, thanks for looking at this

Rall3n:
Fix #4013

This PR fixes the behaviour around `defaultMenuIsOpen` so that the selected option is visible on mount.

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -623,6 +623,13 @@ export default class Select<
     this.instancePrefix =
       'react-select-' + (this.props.instanceId || ++instanceId);
     this.state.selectValue = cleanValue(props.value);
+
+    // Set focusedOption if menuIsOpen is set on init (e.g. defaultMenuIsOpen)
+    if (props.menuIsOpen && this.state.selectValue.length) {
+      const focusableOptions = this.buildFocusableOptions();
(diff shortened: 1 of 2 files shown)


packages/react-select/src/Select.tsx
```diff
@@ -623,6 +623,13 @@ export default class Select<
     this.instancePrefix =
       'react-select-' + (this.props.instanceId || ++instanceId);
     this.state.selectValue = cleanValue(props.value);
+
+    // Set focusedOption if menuIsOpen is set on init (e.g. defaultMenuIsOpen)
+    if (props.menuIsOpen && this.state.selectValue.length) {
+      const focusableOptions = this.buildFocusableOptions();
+      const optionIndex = focusableOptions.indexOf(this.state.selectValue[0]);
```
Rall3n, line 630: Object equality should not be a problem at this point as this happens before mount.
But could still be changed to compare values instead.

lukebennett88 (approved):

PR #5431, opened 2022-10-26
Change `class` components to `functional` components

nderkim:


packages/react-select/src/components/Menu.tsx
```diff
@@ -594,23 +599,12 @@ export const MenuPortal = <
...
     coercePlacement(menuPlacement)
   );
   const [computedPosition, setComputedPosition] =
     useState<ComputedPosition | null>(null);
 
-  // callback for occasions where the menu must "flip"
-  const getPortalPlacement = useCallback(
-    ({ placement: updatedPlacement }: CalculatedMenuPlacementAndHeight) => {
-      // avoid re-renders if the placement has not changed
-      if (updatedPlacement !== placement) {
```
nderkim, line 607: I don't think we need this check because React knows to bail out if the value did not change.

packages/react-select/src/animated/transitions.tsx
```diff
@@ -72,93 +66,62 @@ interface CollapseProps {
...
-export class Collapse extends Component<CollapseProps, CollapseState> {
-  duration = collapseDuration;
-  rafID?: number | null;
-  state: CollapseState = { width: 'auto' };
-  transition: { [K in TransitionStatus]?: CSSProperties } = {
-    exiting: { width: 0, transition: `width ${this.duration}ms ease-out` },
-    exited: { width: 0 },
-  };
-  nodeRef = createRef<HTMLDivElement>();
+export const Collapse = ({ children, in: _in, onExited }: CollapseProps) => {
```
jossmac, line 72: I assume that `in` is from "react-transition-group"—reserved word for a prop was such a bad idea...

nderkim: agreed 🥲

packages/react-select/src/animated/transitions.tsx
```diff
@@ -72,93 +66,62 @@ interface CollapseProps {
...
+        if (!current) return;
+        onExited?.(current);
+      }}
+      timeout={collapseDuration}
+      nodeRef={ref}
+    >
+      {(state) => (
+        <div
+          ref={ref}
+          style={{
```
jossmac, line 112: Nit: Could the nested ternaries be against an object with state keys, or some `getStyleFromState` fn with a switch?

packages/react-select/src/components/Menu.tsx
```diff
@@ -55,10 +55,10 @@ interface PlacementArgs {
 }
 
 export function getMenuPlacement({
-  maxHeight,
+  maxHeight: desiredMaxHeight,
```
jossmac, line 58: Glad these are now differentiated! I think "desired" isn't quite right, though. Maybe `preferredX` or `consumerX`?

packages/react-select/src/components/Menu.tsx
```diff
@@ -294,76 +294,81 @@ export const menuCSS = <
...
-  getPortalPlacement:
-    | ((menuState: CalculatedMenuPlacementAndHeight) => void)
-    | null;
-}>({ getPortalPlacement: null });
-
-interface MenuState {
-  placement: CoercedMenuPlacement | null;
-  maxHeight: number;
-}
+const PortalPlacementContext =
```
jossmac, line 297: Declaring `undefined` feels weird. Could this be:

```tsx
createContext<CtxType | undefined>()

// OR

createContext<CtxType | null>(null)
```

packages/react-select/src/components/Menu.tsx
```diff
@@ -294,76 +294,81 @@ export const menuCSS = <
...
+  } = props;
 
-    if (!ref) return;
+  const { setPortalPlacement } = useContext(PortalPlacementContext) || {};
+  const ref = useRef<HTMLDivElement | null>(null);
+  const [maxHeight, setMaxHeight] = useState(maxMenuHeight);
+  const [placement, setPlacement] = useState<CoercedMenuPlacement | null>(null);
+
+  useLayoutEffect(() => {
+    const { current } = ref;
```
jossmac, line 329: Nit: destructuring doesn't offer a lot of value here, prefer meaningful variable name e.g.

```jsx
useLayoutEffect(() => {
  const menuEl = ref.current
  if (!menuEl) return

  const state = getMenuPlacement({
    menuEl,
    ...
  })
})
```

packages/react-select/src/components/Menu.tsx
```diff
@@ -294,76 +294,81 @@ export const menuCSS = <
...
+  ]);
 
-    return children({
-      ref: this.getPlacement,
-      placerProps: this.getUpdatedProps(),
-    });
-  }
-}
+  return (
+    <Fragment>
```
jossmac, line 360: I get why this is here, but I think it's a mistake in our types 🤔 For another day:

```diff
export interface MenuPlacerProps {
-  children: (childrenProps: ChildrenProps) => ReactNode
+  children: (childrenProps: ChildrenProps) => ReactElement
}
```

packages/react-select/src/components/Menu.tsx
```diff
@@ -690,7 +684,7 @@ export const MenuPortal = <
   );
 
   return (
-    <PortalPlacementContext.Provider value={{ getPortalPlacement }}>
+    <PortalPlacementContext.Provider value={{ setPortalPlacement }}>
```
jossmac, line 687: State setters are equal between renders, but the object will be created each time. Might be worth memoizing to reduce consumer re-renders?

```tsx
const context = useMemo(() => ({ setPortalPlacement }), [setPortalPlacement])

return <PortalPlacementContext.Provider value={context} />
```

nderkim: nice catch! 🎉

jossmac (approved):
Couple of nit comments, but LGTM 🚀

Issue #5434, opened 2022-10-26
Documentation: Unacceptably low contrast on view sample source button


judfs:
For https://react-select.com/home:

Even drawing a giant red circle around it, its easy to look at this screenshot and not realize this button exists.
![image](https://user-images.githubusercontent.com/39133604/198116215-28b28cf5-a001-41ab-af53-26e7e989aa17.png)

I just spent a bunch of time trying to understand things in devtools because I had no clue this button existed.

There is no need de-emphasis the button blocks when not hovered, but `opacity: 0.2;` is absolutely not ok.

lukebennett88:
Thanks for raising this @judfs. I fully agree with you. 
We're planning to improve the docs soon, but in the meantime as a quick fix, I've opened a PR to improve the contrast: 
- https://github.com/JedWatson/react-select/pull/5436

PR #5436, opened 2022-10-27
Improve contrast of example buttons in docs

lukebennett88:
Closes https://github.com/JedWatson/react-select/issues/5434


nderkim (approved):
🍕

PR #5450, opened 2022-10-31
Memoize `buildCategorizedOptions`

Methuselah96:
Recreation of https://github.com/JedWatson/react-select/pull/4514. Reduces the render time when opening the menu with 10,000 options from ~300ms to ~200ms. Still not perfect, but it's a first step at reducing unnecessary work.

Before:
before

After:
after

As for [my concerns](https://github.com/JedWatson/react-select/pull/4514#issuecomment-804314463) in the initial PR, it is correct that memoization will still be helpful for performance within a single Select's render cycle, even if there are multiple Selects on the page.

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -19,6 +19,7 @@ import { DummyInput, ScrollManager } from './internal/index';
 import { AriaLiveMessages, AriaSelection } from './accessibility';
 
 import {
+  areArraysEqual,
   classNames,
   cleanValue,
   isTouchCapable,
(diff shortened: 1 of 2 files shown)


h3rmanj:
I'm not sure on the syntax, but couldn't the memoized function be bound to the class as well? That way it would even be memoized between renders with multiple selects. Something like:

```js
class Select {
  // ...
  memoizedBuildCategorizedOptions = memoizeOne(buildCategorizedOptionsFromProps)
  buildCategorizedOptions = () =>
    memoizedBuildCategorizedOptions(this.props, this.state.selectValue);
  // ...
}
```

Methuselah96:
The problem is that `getDerivedStateFromProps` is static and needs to call `buildCategorizedOptions`, so it can't call a class method.

h3rmanj:
The memoized function could technically be part of the state as well, although it might feel a little weird. This is still an improvement though, so it might be good enough until Select is refactored to use hooks.

Jdyn:
Any update on this? The menu open delay with large data is a huge drawback (already using virtualization etc). I think this can help a lot.

PR #5455, opened 2022-11-01
Tidy up examples from docs and move into Storybook

lukebennett88:


docs/package.json
```diff
@@ -29,7 +29,7 @@
     "@types/webpack": "^4.41.26",
     "@types/webpack-dev-server": "^3.11.1",
     "babel-plugin-macros": "^3.0.1",
-    "chroma-js": "^1.3.6",
+    "chroma-js": "^2.4.2",
```
lukebennett88, line 32: I added a Story that was using this package. Figured it doesn't hurt to bump the version to latest.

storybook/components/field.tsx
```diff
@@ -18,7 +18,6 @@ export function Field({
       style={{
         display: 'flex',
         flexDirection: 'column',
-        fontFamily: 'system-ui',
```
lukebennett88, line 21: `withSystemFont` in `preview.tsx` handles this now.

nderkim (approved):
🥞

Issue #5451, opened 2022-10-31
New classNames API for Tailwind (and other CSS library) support


JedWatson:
We're working on a new feature/release that I thought was worth documenting the plan and rationale here for ahead of the PRs and releases coming for it. If you've got any feedback, please share below.

## Goal 1: Tailwind Support

Tailwind (and other utility class frameworks) have gained a lot of momentum recently, and with increasing diversity of new front-end frameworks and build systems, it's less easy than it used to be to assume that CSS-in-JS / Emotion will "Just Work" for everyone. In particular, if you're building your own component library / design system with something like Tailwind and using React Select it's a fair bit of work to map design tokens and theme properly from CSS classes to React Select's theme context.

Wouldn't it be better if you could just style React Select with your own class names? **that's the plan** 😁 

### The new `classNames` API

Step one is adding a new `classNames` API. A small nod to a popular library, this will sit alongside the current `styles` prop, and let you provide (as the name suggests) class names instead of css property values for each Component in React Select, based on the same state argument provided to the `styles` functions.

So if you wanted a red border around your Select component on focus, where currently you could do this:

```
<Select styles={{
  control: (styles, { isFocused }) => ({ ...styles, borderColor: isFocused ? 'grey' : 'red' })
  }} />
```

Now you can apply class names to get the same effect:

```
<Select classNames={{
  control: ({ isFocused }) => isFocused ? 'border-red-600' : 'border-grey-300'
  }} />
```

ℹ️ Note: you won't have a good time using tailwind's native support for pseudo-states, because the browser doesn't think the Control component is focused when the input or menu are "focused" (internally we call this "pseudo-focus") so classes like `focus:border-grey-300` won't work. This is why we're sticking with the `(state) => string` signature for the new prop

## Goal 2: The new `unstyled` prop

Not necessarily in the same release (depends on how the implementation plays out) but we're also planning to introduce an `unstyled` prop which will remove all the presentation styles from React Select (leaving some important functional styles, like those for menu positioning and input width in multi select)

This will make it easier to completely specify your own styles **or** classNames to control the look of React Select, without having to specifically override the default theme we apply.

ℹ️ Note: removing the dependency on `emotion` for styles isn't specifically part of this same goal, although the two are related; the new `unstyled` prop is specifically only to make theming-from-scratch easier. We'll (probably) still be using emotion under the hood for applying functional styles.

## Goal 3: Decouple from Emotion

[...]

PR #5457, opened 2022-11-02
Add `classNames` API and `unstyled` prop

nderkim:
This PR adds the `classNames` API and the `unstyled` prop as defined in https://github.com/JedWatson/react-select/issues/5451

I've added `storybook` stories as examples, but no docs (yet)


.changeset/weak-roses-decide.md
```diff
@@ -0,0 +1,6 @@
+---
+'react-select': minor
+'@react-select/docs': patch
```
nderkim, line 3: How do we version docs? Should this be `minor` too?

packages/react-select/src/Select.tsx
```diff
@@ -1078,11 +1093,16 @@ export default class Select<
     key: Key,
     props: StylesProps<Option, IsMulti, Group>[Key]
   ) => {
-    const base = defaultStyles[key](props as any);
+    const { unstyled } = this.props;
```
nderkim, line 1096: I've made `unstyled` only available on the parent component (for now) because:

- It minimises the scope of this PR
- What happens if `unstyled` on the parent component is set to `true`, but on the inner component it's set to `false`?
- We can easily add it to the inner components later. If we added it now, and decided to change it later, it would be a breaking change.

packages/react-select/src/Select.tsx
```diff
@@ -1078,11 +1093,16 @@ export default class Select<
...
+    const base = defaultStyles[key](props as any, unstyled);
     base.boxSizing = 'border-box';
     const custom = this.props.styles[key];
     return custom ? custom(base, props as any) : base;
   };
+  getClassName = <Key extends keyof StylesProps<Option, IsMulti, Group>>(
+    key: Key,
+    props: StylesProps<Option, IsMulti, Group>[Key]
+  ) => this.props.classNames[key]?.(props as any);
```
nderkim, line 1105: There is a typing issue here (pre-existing) that I'm ignoring for now to keep the scope of this PR down

packages/react-select/src/components/Group.tsx
```diff
@@ -115,19 +118,24 @@ export const groupHeadingCSS = <
...
+  { theme: { spacing } }: GroupHeadingProps<Option, IsMulti, Group>,
+  unstyled: boolean
+): CSSObjectWithLabel => ({
   label: 'group',
   color: '#999',
   cursor: 'default',
   display: 'block',
   fontSize: '75%',
   fontWeight: 500,
   marginBottom: '0.25em',
```
nderkim, line 131: I haven't (yet) looked to deeply into why some of these values are hardcoded, so i'm keeping them in the "permanent" section of the styles

packages/react-select/src/components/Group.tsx
```diff
@@ -101,6 +103,7 @@ interface GroupHeadingPropsDefinedProps<
   selectProps: Props<Option, IsMulti, Group>;
   theme: Theme;
   getStyles: GetStyles<Option, IsMulti, Group>;
+  getClassName: CommonProps<Option, IsMulti, Group>['getClassName'];
```
nderkim, line 106: Not sure why these types are re-defined here, but just following the existing pattern to simplify this PR

.changeset/weak-roses-decide.md
lukebennett88: No one is consuming the docs as a package, so I probably wouldn't even worry about versioning it.

nderkim: I've noticed it's at version `3.1.2` though, and no idea how it got there 🤷

docs/examples/Experimental.tsx
```diff
@@ -123,6 +123,7 @@ const Group = (props: GroupProps<DateOption, false>) => {
   const {
     Heading,
     getStyles,
+    getClassName,
```
JedWatson, line 126: I would call this `getClassNames`

packages/react-select/src/components/Group.tsx
JedWatson: I think this was a mistake from when groups got added, they should be using theme variables. If you can line them up with existing theme values, please do 🙏

JedWatson (changes requested):

docs/index.css
```diff
@@ -15,7 +15,10 @@ body {
 }
 p > a,
 p > a:hover,
-p > a:visited {
+p > a:visited,
+li > a,
+li > a:hover,
+li > a:visited {
```
lukebennett88, line 21: This was a quick and dirty fix for some links being styled differently in the docs.
Fixing it properly would be more work and not sure how long for this world the existing docs site is.

packages/react-select/src/styles.ts
```diff
@@ -132,6 +108,27 @@ export const defaultStyles: StylesFunctions<
...
+  IsMulti extends boolean = boolean,
+  Group extends GroupBase<Option> = GroupBase<Option>
+> = {
+  [K in keyof StylesProps<Option, IsMulti, Group>]?: (
+    base: CSSObjectWithLabel,
+    props: StylesProps<Option, IsMulti, Group>[K]
+  ) => CSSObjectWithLabel;
+};
+
+export type ClassNamesConfig<
```
lukebennett88, line 122: This should also be exported from `packages/react-select/src/index.ts`

lukebennett88: I fixed this

packages/react-select/src/styles.ts
```diff
@@ -132,6 +108,27 @@ export const defaultStyles: StylesFunctions<
...
+  Option,
+  IsMulti extends boolean,
+  Group extends GroupBase<Option>
```
lukebennett88, line 125: Should these be given default values like StylesConfig above?
```suggestion
	Option = unknown,
  IsMulti extends boolean = boolean,
  Group extends GroupBase<Option> = GroupBase<Option>
```

nderkim: Since we are exporting it now, it's probably better dx to. Feel free to commit your suggestion ✅

docs/markdown/renderer.tsx
```diff
@@ -105,8 +105,7 @@ const Heading = (props: HeadingProps) => {
     store.add(nodeKey, { key: nodeKey, label, level, path: `#${slug}` });
   }
   const css = {
-    marginTop: 0,
-    '&:not(:first-of-type)': { marginTop: 30 },
+    '&:first-child': { marginTop: 0 },
```
lukebennett88, line 108: What is the reason behind changing this (other than simplifying things?)

nderkim: the previous style would result in:
``` tsx
// 0
<h1/>

// 0
<h2/>

// 0
<h3/>

// 30
<h2/>

// 30
<h3/>
```

but i think the intention was to just do:
``` jsx
// 0
<h1/>

<h2/>

<h3/>

<h2/>

<h3/>
```

packages/react-select/src/styles.ts
```diff
@@ -78,36 +78,12 @@ export interface StylesProps<
...
-  >;
-};
-
-export const defaultStyles: StylesFunctions<
-  unknown,
-  boolean,
-  GroupBase<unknown>
-> = {
+export const defaultStyles: {
+  [K in keyof StylesProps<any, any, any>]: (
```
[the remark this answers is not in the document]
nderkim: This was intentional to show that it doesn't matter what we pass in because we are only using the keys, which are unaffected by the generics.

JedWatson (approved):

Issue #4859, opened 2021-10-13
Not able to import /async, /creatable, .etc from an ES module


srmagura:
### [Minimal repro](https://github.com/srmagura/react-select-esm-repro)

If placed in an `.mjs` file, the following code results in an error.

```js
import AsyncSelect from "react-select/async";

console.log(AsyncSelect);
```

```text
internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^

Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import 'C:\Projects\OSS\Bugs\react-select-esm-repro\node_modules\react-select\async' is not supported resolving ES modules imported from C:\Projects\OSS\Bugs\react-select-esm-repro\index.mjs
Did you mean to import react-select-esm-repro/node_modules/react-select/async/dist/react-select.cjs.js?
    at finalizeResolution (internal/modules/esm/resolve.js:281:17)
    at moduleResolve (internal/modules/esm/resolve.js:708:10)
    at Loader.defaultResolve [as _resolve] (internal/modules/esm/resolve.js:819:11)
    at Loader.resolve (internal/modules/esm/loader.js:89:40)
    at Loader.getModuleJob (internal/modules/esm/loader.js:242:28)
    at ModuleWrap.<anonymous> (internal/modules/esm/module_job.js:73:40)
    at link (internal/modules/esm/module_job.js:72:36) {
  code: 'ERR_UNSUPPORTED_DIR_IMPORT',
  url: 'file:///C:/Projects/OSS/Bugs/react-select-esm-repro/node_modules/react-select/async'
}
```

While this reproduction uses Node, the issue also occurs when using Webpack 5 which follows Node's handling of ES module imports. Here's the Webpack error for Googleability:

```text
Module not found: Error: Can't resolve 'react-select/async' in 'C:\Projects\ITI\iti-react\iti-react\dist'
Did you mean 'react-select.esm.js'?
BREAKING CHANGE: The request 'react-select/async' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

## Suggested Solution

Utilize the `exports` field in `package.json` [as described in the Node documentation.](https://nodejs.org/api/packages.html#packages_subpath_exports)

[...]

PR #5559, opened 2023-02-14
Add `exports` field in package.json

gdiazdelaserna:
### TL;DR

Fixes the issue mentioned here: https://github.com/JedWatson/react-select/issues/4859

#### Issue
Consuming a React Select subpath from a JS module in a project using Webpack 5, throws an error saying the path in the  import statement is not fully specified.

#### Context
Webpack 5 introduced a default behavior that requires all import statements in JS modules to be [fully specified](https://webpack.js.org/configuration/module/#resolvefullyspecified).

Importing a React Select subpath (e.g. `import Creatable from 'react-select/creatable'`) is breaking that rule. Webpack 5 thinks that import statement is trying to import a file with a missing extension.

#### Solution
Inform Webpack that the subpaths are not file imports (and an extension shouldn't be required) by declaring an `exports` field in `package.json`.

The `exports` field is documented in these two places:
- [Node](https://nodejs.org/api/packages.html#exports)
- [Webpack](https://webpack.js.org/guides/package-exports/)

The new `exports` field includes all the subpaths mentioned in React Select's documentation. All other subpath imports will fail from now on.

Each subpath declared in the `exports` field includes 3 different file system paths to satisfy the different environments that might consume them:
- `require`: Will be used by environments using the `require` keyword to import files (e.g CommonJS)
- `import`: Will be used by environments using the `import` keyword to import files (e.g. ES Modules)
- `types`: Will be used by TypeScript to query type definitions

### DIFF
--- a/.changeset/early-pans-obey.md
+++ b/.changeset/early-pans-obey.md
@@ -1,5 +1,5 @@
 ---
-"react-select": patch
+'react-select': patch
 ---
 
 Add `exports` field in package.json
--- a/.changeset/early-pans-obey.md
+++ b/.changeset/early-pans-obey.md
@@ -1,0 +1,5 @@
+---
+"react-select": patch
+---
+
+Add `exports` field in package.json
--- a/packages/react-select/package.json
+++ b/packages/react-select/package.json
@@ -59,5 +59,37 @@
       "creatable/index.ts",
       "async-creatable/index.ts"
     ]
+  },
+  "exports": {
+    ".": {
+      "require": "./dist/react-select.cjs.js",
+      "import": "./dist/react-select.esm.js",
+      "types": "./dist/react-select.cjs.d.ts"
+    },
+    "./base": {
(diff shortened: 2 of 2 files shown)


Rall3n:
@gdiazdelaserna Has this been tested with webpack and vite?

gdiazdelaserna:
@Rall3n 

I have tested this change in multiple environments. I created this repo to show how this has been tested: 

https://github.com/gdiazdelaserna/react-select-exports-field-test

As you can see, I tested this in:
- Create React App
- Next.js
- Vite
- Webpack

This proves that:
- React Select continues to work correctly in all those environments (i.e. there are no regressions)
- The PR fixes the issue it intends to fix: React Select subpaths imports now work correctly when consumed from a JS module in a Webpack 5 app

Rall3n (approved):

davbrito:
As a workaround I put the path to the file on node_modules

image

PR #5582, opened 2023-03-15
Loop with arrow keys

Kageetai:
adds a new prop `loopWithArrowKeys` to default to `true`
Gives the option to disable looping with using arrow keys


kjetilkh-politiet (approved):
Looks good! Are there any reviewers left in this repo? Can anyone approve this?

kjetilkh-politiet:
Hello! We need this for a project as when used in combination with react-window it doesn't scroll when looping. This change would enable us to disable looping and prevent the issue. 

Are there any maintainers/verified reviewers who can look into this? @JedWatson @Methuselah96 @lukebennett88

PR #5625, opened 2023-04-30
Add appropriate roles to option and listbox elements

MssKnd:
This pull request aims to improve the accessibility of react-select by adding appropriate roles to the option and listbox elements.

Changes:

- Added the `option` role and `aria-selected` to the option-equivalent elements
- Added the `listbox` role to the listbox-equivalent elements

I noticed that while using react-select in my own project, I couldn't use getByRole with react-testing-library to target the option-equivalent elements. This led me to create this PR to address the issue.

By merging this PR, we can make react-select more accessible and provide a better experience for users with assistive technologies. Please let me know if any further changes or tests are needed. I look forward to contributing to this project and making it more inclusive for all users.

### DIFF
--- a/packages/react-select/src/components/Menu.tsx
+++ b/packages/react-select/src/components/Menu.tsx
@@ -448,6 +448,7 @@ export const MenuList = <
         'menu-list--is-multi': isMulti,
       })}
       ref={innerRef}
+      role="listbox"
       {...innerProps}
     >
       {children}
--- a/packages/react-select/src/components/Option.tsx
+++ b/packages/react-select/src/components/Option.tsx
@@ -98,6 +98,8 @@ const Option = <
         'option--is-selected': isSelected,
       })}
       ref={innerRef}
+      role="option"
+      aria-selected={isSelected}
       aria-disabled={isDisabled}
       {...innerProps}
     >


AriPerkkio (approved):
> I couldn't use getByRole with react-testing-library to target the option-equivalent elements.

I ran into this same exact problem.

As work-around, these roles can be added using `components` prop. It would be great to have these set by default as this PR does.

```tsx
import ReactSelect, { type OptionProps, type MenuListProps, components } from 'react-select';

export default function Select() {
  return (
    <ReactSelect components={{Option, MenuList}} />
  )
}

function Option({innerProps, isSelected, ...props}: OptionProps) {
    return (
        <components.Option
            {...props}
            isSelected={isSelected}
            innerProps={{...innerProps, role: 'option', 'aria-selected': isSelected}}
        />
    );
}

function MenuList({innerProps, ...props}: MenuListProps) {
    return (
        <components.MenuList {...props} innerProps={{...innerProps, role: 'listbox'}} />
    );
}
```

MssKnd:
@AriPerkkio 
Thank you for your nice information!
I am dealing with the same work-around right now.

csandman:
This should be fixed by #5758 by the way

MssKnd:
@csandman 
Thanks for letting me know!
I will close this PR.

PR #5626, opened 2023-05-02
Fix `exports` field

### LINKED ISSUE #4859: Not able to import /async, /creatable, .etc from an ES module

### LINKED ISSUE #5595: [ESM Support][5.7.2] Unable to use react-select in ESM project
There are a couple of issues we currently have with using `react-select` in our projects that use `"type": "module"` in `package.json` and have `"module"` and `"moduleResolution"` set to `"nodenext"` in `tsconfig.json`.

First of all, importing.

Rough example:
```
import ReactSelect, { Props } from 'react-select'

// ...

const SomeComponent = (props: Props) => {
  return <ReactSelect {...props} />
}
```
This now fails when building types with: `JSX element type 'ReactSelect' does not have any construct or call signatures.ts(2604)`

We can resolve this ourselves by changing the above example to:
```
import ReactSelect, { Props } from 'react-select'

const Select = ReactSelect.default || ReactSelect
// ...

const SomeComponent = (props: Props) => {
  return <Select {...props} />
}
```
but it is not ideal.

Doing the above change allows us to build types without issue.

There is another issue which happens when we start the app though, which fails with:
```
(node:15999) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
/Users/rafaldziegielewski/Documents/projekty/adminjs-design-system/node_modules/react-select/dist/react-select.esm.js:1
import { u as useStateManager } from './useStateManager-7e1e8489.esm.js';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at internalCompileFunction (node:internal/vm:74:18)
    at wrapSafe (node:internal/modules/cjs/loader:1141:20)
    at Module._compile (node:internal/modules/cjs/loader:1182:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1272:10)
    at Module.load (node:internal/modules/cjs/loader:1081:32)
    at Module._load (node:internal/modules/cjs/loader:922:12)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:169:29)
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)

Node.js v18.13.0
error Command failed with exit code 1.
```

The above error can be resolved by adding `"type": "module"` to every `package.json` of `react-select`, for example here:
https://github.com/JedWatson/react-select/blob/master/packages/react-select/base/package.json
and adding `.js` extensions to exports.

We managed to resolve this issue in our local environment by creating a patch:
https://gist.github.com/dziraf/160b203a8837b85a76a1f42e7b22d524
The issue still remains though as we're using it inside of one of our libraries and patches are unfortunately not installed recursively inside of `node_modules` when you install them.

emmatown:
closes #5595
closes #4859

The `exports` field added in #5559 was wrong in a few ways, [arethetypeswrong](https://arethetypeswrong.github.io/?p=react-select%405.7.2) explains this well:

> 🐛 Imports of all entrypoints under resolution modes that use the import condition in package.json "exports" resolved through a conditional package.json export, but only after failing to resolve through an earlier condition. This behavior is a [TypeScript bug](https://github.com/microsoft/TypeScript/issues/50762) and should not be relied upon.
> 
> 🚭 The implementation resolved at all entrypoints uses ESM syntax, but the detected module kind is CJS. This will be an error in Node (and potentially other runtimes and bundlers). The module kind was decided based on the nearest package.json’s lack of a "type": "module" field.

This PR makes react-select use Preconstruct's exports field feature which does things correctly, I've also checked the build output on this branch with [arethetypeswrong](https://arethetypeswrong.github.io) and it reports no errors. Of note is that there is still no bundle for Node ESM, the `.mjs` file used when the `import` condition matches just re-exports from the CJS bundle but fixes the default export (this is new to Preconstruct and is what `"importConditionDefaultExport": "default"` is doing), the lack of a Node ESM bundle is intentional so the [dual package hazard](https://nodejs.org/api/packages.html#packages_dual_package_hazard) can't happen. Bundlers will use the `module` condition to still have ESM there. There is also intentionally no `types` condition, that's unnecessary because Typescript will resolve to the `import` or `default` conditions and use the `.d.mts` and `.d.ts` files next to the implementations.

### DIFF
--- a/.changeset/nervous-islands-shake.md
+++ b/.changeset/nervous-islands-shake.md
@@ -1,0 +1,5 @@
+---
+'react-select': patch
+---
+
+Fix importing `react-select` in Node ESM
--- a/docs/generate-magical-types/generate/package.json
+++ b/docs/generate-magical-types/generate/package.json
@@ -1,3 +1,4 @@
 {
-  "main": "dist/react-select-generate-magical-types-generate.cjs.js"
+  "main": "dist/react-select-generate-magical-types-generate.cjs.js",
+  "module": "dist/react-select-generate-magical-types-generate.esm.js"
 }
--- a/docs/generate-magical-types/package.json
+++ b/docs/generate-magical-types/package.json
@@ -1,6 +1,19 @@
 {
   "name": "@react-select/generate-magical-types",
   "main": "dist/generate-magical-types.cjs.js",
+  "exports": {
+    "./generate": {
+      "module": "./generate/dist/react-select-generate-magical-types-generate.esm.js",
+      "import": "./generate/dist/react-select-generate-magical-types-generate.cjs.mjs",
+      "default": "./generate/dist/react-select-generate-magical-types-generate.cjs.js"
+    },
+    "./serialize": {
+      "module": "./serialize/dist/react-select-generate-magical-types-serialize.esm.js",
+      "import": "./serialize/dist/react-select-generate-magical-types-serialize.cjs.mjs",
+      "default": "./serialize/dist/react-select-generate-magical-types-serialize.cjs.js"
+    },
+    "./package.json": "./package.json"
+  },
   "//": "these deps aren't real, they're just to appease preconstruct",
   "dependencies": {
     "@babel/runtime": "*",
--- a/docs/generate-magical-types/serialize/package.json
+++ b/docs/generate-magical-types/serialize/package.json
@@ -1,3 +1,4 @@
 {
-  "main": "dist/react-select-generate-magical-types-serialize.cjs.js"
+  "main": "dist/react-select-generate-magical-types-serialize.cjs.js",
+  "module": "dist/react-select-generate-magical-types-serialize.esm.js"
 }
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "@emotion/babel-plugin": "^11.10.2",
     "@emotion/jest": "^11.1.0",
     "@manypkg/cli": "^0.19.2",
-    "@preconstruct/cli": "^2.6.0",
+    "@preconstruct/cli": "^2.6.2",
     "@testing-library/dom": "8.19.0",
     "@testing-library/jest-dom": "5.1.1",
     "@testing-library/react": "12.1.4",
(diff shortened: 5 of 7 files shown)


JedWatson (approved):
Thanks @emmatown

PR #5648, opened 2023-05-25
Add resultLimit to improve performance of react-select for many options

sorin-sabo:
Based on PR https://github.com/JedWatson/react-select/pull/4515
React-select with over 10.000 options has performance issues
Improve performance limiting displayed options based on user preference using resultLimit prop


utkarshsaraogi-airbase:
Hi! Any update on this issue? Or consumers are expected to handle this on their own?

PR #5657, opened 2023-06-05
support aria-placeholder

seanlaff:
The docs (https://react-select.com/props) define `aria-placeholder` as a supported prop, but as far as I could tell it wasn't actually supported. This changeset adds it.

This will enable `page.getPlaceholder("foo")` in playwright tests.


seanlaff:
Not sure why the e2e tests are stuck- I was going to retry the pipeline but I don't have permission

seanlaff:
@Methuselah96 Would you be able to re-run this pipeline for me to see if the e2e timeout resolves itself?

PR #5669, opened 2023-06-12
Add an option to introduce friction when using the delete key

murphpdx:
This feature requires that users hit the backspace key twice when deleting a value, unless the value is already selected. 

I created a feature request here: https://github.com/JedWatson/react-select/discussions/5670

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -92,6 +92,8 @@ export interface Props<
   autoFocus?: boolean;
   /** Remove the currently focused option when the user presses backspace when Select isClearable or isMulti */
   backspaceRemovesValue: boolean;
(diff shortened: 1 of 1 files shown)


packages/react-select/src/Select.tsx
```diff
@@ -1494,6 +1498,15 @@ export default class Select<
...
         if (inputValue) return;
+        if (backspaceFriction) {
+          if (focusedValue) {
+            this.removeValue(focusedValue);
+            return;
+          } else {
+            this.focusValue('previous');
+            return;
+          }
+        }
```
nick-keller, line 1509:
```suggestion
        if (backspaceFriction && !focusedValue) {
          this.focusValue('previous');
          return;
        }
```

nick-keller (changes requested):

websiddu:
Looking forward for this to getting merged :)

Issue #5549, opened 2023-01-30
On mobile device getting Ignored attempt to cancel a touchmove error


qaws5503:
# Describe
Hello everyone. When scrolling menu all works great with desktop, but when change to mobile it's getting error.

![截圖 2023-01-30 下午6 25 29](https://user-images.githubusercontent.com/57092368/215452324-75d1f025-100f-4ff5-a6ce-7c9a3823136c.png)

Does this error matters? or it's just some annoying complain.

# Reproduce
Just open the demo page [https://react-select.com/home](https://react-select.com/home), and change to mobile screen.
Then open any selector and start scrolling the popup menu. Then the console is running this error until the scroll event ends.

shaunaas:
I'm seeing this in my app error logs as far back as 10 months. Not just mobile - Safari, Edge, Chrome - Desktop.

PR #5672, opened 2023-06-13
Fix for calling non-cancellable scroll events

tu4mo:
Fixes https://github.com/JedWatson/react-select/issues/5549.


lukebennett88 (approved):
Thanks for the PR @tu4mo, looks good to me.

Issue #5743, opened 2023-09-01
Module augmentation not working with moduleResolution: bundler


FabianFrank:
There seems to be an issue with module augmentation and the exports field in package.json added in https://github.com/JedWatson/react-select/pull/5559 and improved in https://github.com/JedWatson/react-select/pull/5626. Module augmentation as described at https://react-select.com/typescript#custom-select-props appears to no longer work since `react-select/dist/declarations/src/Select` appears to no longer be the path of the `Select` module when the exports field in package.json is used during module resolution. I was unfortunately not able to figure out what the new path is. Deleting the exports field from package.json fixes the issue and module augmentation works as expected again.

csandman:
I am running into this issue as well on my wrapper for this package [`chakra-react-select`](https://github.com/csandman/chakra-react-select). I'm not sure this has anything to do with the change in #5626 though, as I'm having the same issue after reverting to `v5.7.2`. I'm really not too familiar with the inner workings of module augmentation, is there something that changed with it when using `moduleResolution: "bundler"`?

I'm also having the issue with `node16`/`nodenext`.

<img width="991" alt="image" src="https://github.com/JedWatson/react-select/assets/9214195/35ee3667-0b1a-4be6-a978-8a4ff922c05c">

PR #5744, opened 2023-09-01
Add type export for ./dist/declarations/src/Select

FabianFrank:
to fix module augmentation with moduleResolution node16. Fixes https://github.com/JedWatson/react-select/issues/5743.

### DIFF
--- a/packages/react-select/package.json
+++ b/packages/react-select/package.json
@@ -61,6 +61,9 @@
     ]
   },
   "exports": {
+    "./dist/declarations/src/Select": {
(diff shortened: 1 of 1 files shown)


FabianFrank:
I've only started yesterday to really dig into moduleResolution node16, I would appreciate it if @emmatown could weigh in since it seems she fixed up the export map in the past.

igl:
You can use a wildcard to export all the declarations

```
        "./dist/declarations/*": {
            "types": "./dist/declarations/*"
        },
```

Issue #5121, opened 2022-03-22
Accessibility - NVDA screenreader issue while reading the options on the Welcome Page.


SairamAlagapan:
This is a bug with regards to accessibility.

We are using NVDA Screenreader to read the options from Single listbox and it's being read out as Blank + Option + OptionName(Red) while it's being read out as Option + OptionName(Red) using VoiceOver. 

Requesting to look into this bug and provide a solution to prevent "Blank" being read.

Attaching a video for reference.

https://user-images.githubusercontent.com/59094456/159521879-dc1de7b9-1276-495c-b352-a210589cfb2a.mp4

Rall3n:
Hello @SairamAlagapan,

NVDA announcing "blank" is expected behavior when it reaches the end/start of a text field.

I think we can improve on this, but it would require some significant changes.

SairamAlagapan:
Hello @Rall3n,

Thank you for the response. If I may ask, when can I expect the changes to be completed? Will it be in the next release?

SairamAlagapan:
Hello @Rall3n,

Any updates on this?

SairamAlagapan:
Hello @Rall3n,

Any updates?

mellis481:
> NVDA announcing "blank" is expected behavior when it reaches the end/start of a text field.

To make sure it's clear, this behavior occurs if `isSearchable` is set to `true` (which is the default behavior).

And to add my opinion of this behavior... it creates an absolutely brutal experience for screen reader users.

SairamAlagapan:
Any update on this?

PR #5758, opened 2023-09-26
Accessibility - screenreader announces "blank" while reading the options

Ke1sy:
This PR resolves this issue: https://github.com/JedWatson/react-select/issues/5121

1. Added 'aria-activedescendant' for input and functionality to calculate it; 
2. Added role 'option' and  'aria-selected' for option;
3. Added role 'listbox' for menu;
4. Added tests for 'aria-activedescendant';
5. Changes in aria-live region:
- the instructions how to use select will be announced only one time when user focuses the input for the first time.
- instructions for menu or selected value will be announced only once after focusing them.
- removed aria-live for focused option because currently with correct aria-attributes it will be announced by screenreader natively as well as the status of this option (active or disabled).
- separated ariaContext into ariaFocused, ariaResults, ariaGuidance to avoid announcing redundant information and higlight only current change.

Video result using NVDA:
https://alina-andreeva.tinytake.com/msc/ODczMTIyNF8yMjEzMjU5MA

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -337,6 +337,7 @@ interface State<
   prevWasFocused: boolean;
   inputIsHiddenAfterUpdate: boolean | null | undefined;
   prevProps: Props<Option, IsMulti, Group> | void;
+  instancePrefix: string;
 }
 
 interface CategorizedOption<Option> {
@@ -448,10 +449,7 @@ function buildFocusableOptionsFromCategorizedOptions<
   );
 }
 
-function buildFocusableOptionsFromCategorizedOptionsWithIds<
-  Option,
-  Group extends GroupBase<Option>
->(
+function buildFocusableOptionsWithIds<Option, Group extends GroupBase<Option>>(
   categorizedOptions: readonly CategorizedGroupOrOption<Option, Group>[],
   optionId: string
 ) {
@@ -534,6 +532,17 @@ function getNextFocusedOption<
     ? lastFocusedOption
     : options[0];
 }
+
+const getFocusedOptionId = <Option,>(
+  focusableOptionsWithIds: FocusableOptionWithId<Option>[],
+  focusedOption: Option
+) => {
+  const focusedOptionId = focusableOptionsWithIds.find(
+    (option) => option.data === focusedOption
+  )?.id;
+  return focusedOptionId || null;
+};
+
 const getOptionLabel = <
   Option,
   IsMulti extends boolean,
@@ -632,6 +641,7 @@ export default class Select<
     prevWasFocused: false,
     inputIsHiddenAfterUpdate: undefined,
     prevProps: undefined,
+    instancePrefix: '',
   };
 
   // Misc. Instance Properties
@@ -642,7 +652,6 @@ export default class Select<
   commonProps: any; // TODO
   initialTouchX = 0;
   initialTouchY = 0;
-  instancePrefix = '';
   openAfterFocus = false;
   scrollToFocusedOptionOnUpdate = false;
   userIsDragging?: boolean;
@@ -672,16 +681,15 @@ export default class Select<
 
   constructor(props: Props<Option, IsMulti, Group>) {
     super(props);
-    this.instancePrefix =
+    this.state.instancePrefix =
       'react-select-' + (this.props.instanceId || ++instanceId);
     this.state.selectValue = cleanValue(props.value);
-    this.state.focusableOptionsWithIds = this.getFocusableOptionsWithIds();
     // Set focusedOption if menuIsOpen is set on init (e.g. defaultMenuIsOpen)
     if (props.menuIsOpen && this.state.selectValue.length) {
+      this.state.focusableOptionsWithIds = this.getFocusableOptionsWithIds();
+
       const focusableOptions = this.buildFocusableOptions();
-
       const optionIndex = focusableOptions.indexOf(this.state.selectValue[0]);
-
       this.state.focusedOption = focusableOptions[optionIndex];
       this.state.focusedOptionId = `${this.getElementId(
         'option'
@@ -700,6 +708,7 @@ export default class Select<
       ariaSelection,
       isFocused,
       prevWasFocused,
+      instancePrefix,
     } = state;
     const { options, value, menuIsOpen, inputValue, isMulti } = props;
     const selectValue = cleanValue(value);
@@ -714,13 +723,28 @@ export default class Select<
       const focusableOptions = menuIsOpen
         ? buildFocusableOptions(props, selectValue)
         : [];
+
+      const focusableOptionsWithIds = menuIsOpen
+        ? buildFocusableOptionsWithIds(
+            buildCategorizedOptions(props, selectValue),
+            `${instancePrefix}-option`
+          )
+        : [];
+
       const focusedValue = clearFocusValueOnUpdate
         ? getNextFocusedValue(state, selectValue)
         : null;
       const focusedOption = getNextFocusedOption(state, focusableOptions);
+      const focusedOptionId = getFocusedOptionId(
+        focusableOptionsWithIds,
+        focusedOption
+      );
+
       newMenuOptionsState = {
         selectValue,
         focusedOption,
+        focusedOptionId,
+        focusableOptionsWithIds,
         focusedValue,
         clearFocusValueOnUpdate: false,
       };
@@ -788,7 +812,7 @@ export default class Select<
     }
   }
   componentDidUpdate(prevProps: Props<Option, IsMulti, Group>) {
-    const { isDisabled, menuIsOpen, options } = this.props;
+    const { isDisabled, menuIsOpen } = this.props;
     const { isFocused } = this.state;
 
     if (
@@ -824,13 +848,6 @@ export default class Select<
       scrollIntoView(this.menuListRef, this.focusedOptionRef);
       this.scrollToFocusedOptionOnUpdate = false;
     }
-
-    if (options !== prevProps.options) {
-      // eslint-disable-next-line react/no-did-update-set-state
-      this.setState({
-        focusableOptionsWithIds: this.getFocusableOptionsWithIds(),
-      });
-    }
   }
   componentWillUnmount() {
     this.stopListeningComposition();
@@ -849,10 +866,6 @@ export default class Select<
     this.onInputChange('', {
       action: 'menu-close',
       prevInputValue: this.props.inputValue,
-    });
-
-    this.setState({
-      focusedOptionId: null,
     });
 
     this.props.onMenuClose();
@@ -1109,14 +1122,14 @@ export default class Select<
   }
 
   getFocusedOptionId = (focusedOption: Option) => {
-    const focusedOptionId = this.state.focusableOptionsWithIds.find(
-      (option) => option.data === focusedOption
-    )?.id;
-    return focusedOptionId || null;
+    return getFocusedOptionId(
+      this.state.focusableOptionsWithIds,
+      focusedOption
+    );
   };
 
   getFocusableOptionsWithIds = () => {
-    return buildFocusableOptionsFromCategorizedOptionsWithIds(
+    return buildFocusableOptionsWithIds(
       buildCategorizedOptions(this.props, this.state.selectValue),
       this.getElementId('option')
     );
@@ -1186,7 +1199,7 @@ export default class Select<
       | 'placeholder'
       | 'live-region'
   ) => {
-    return `${this.instancePrefix}-${element}`;
+    return `${this.state.instancePrefix}-${element}`;
   };
 
   getComponents = () => {
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -329,6 +329,8 @@ interface State<
   inputIsHidden: boolean;
   isFocused: boolean;
   focusedOption: Option | null;
+  focusedOptionId: string | null;
+  focusableOptionsWithIds: FocusableOptionWithId<Option>[];
   focusedValue: Option | null;
   selectValue: Options<Option>;
   clearFocusValueOnUpdate: boolean;
@@ -345,6 +347,11 @@ interface CategorizedOption<Option> {
   label: string;
   value: string;
   index: number;
+}
+
+interface FocusableOptionWithId<Option> {
+  data: Option;
+  id: string;
 }
 
 interface CategorizedGroup<Option, Group extends GroupBase<Option>> {
@@ -434,6 +441,34 @@ function buildFocusableOptionsFromCategorizedOptions<
         );
       } else {
         optionsAccumulator.push(categorizedOption.data);
+      }
+      return optionsAccumulator;
+    },
+    []
+  );
+}
+
+function buildFocusableOptionsFromCategorizedOptionsWithIds<
+  Option,
+  Group extends GroupBase<Option>
+>(
+  categorizedOptions: readonly CategorizedGroupOrOption<Option, Group>[],
+  optionId: string
+) {
+  return categorizedOptions.reduce<FocusableOptionWithId<Option>[]>(
+    (optionsAccumulator, categorizedOption) => {
+      if (categorizedOption.type === 'group') {
+        optionsAccumulator.push(
+          ...categorizedOption.options.map((option) => ({
+            data: option.data,
+            id: `${optionId}-${categorizedOption.index}-${option.index}`,
+          }))
+        );
+      } else {
+        optionsAccumulator.push({
+          data: categorizedOption.data,
+          id: `${optionId}-${categorizedOption.index}`,
+        });
       }
       return optionsAccumulator;
     },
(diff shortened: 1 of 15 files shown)


mellis481:
@lukebennett88 @JedWatson @dcousens @nderkim @Rall3n This PR contains the addition of vital ARIA attributes that significantly improve the UX for screen reader users. Please merge and release ASAP. Thanks! :slightly_smiling_face:

packages/react-select/src/accessibility/index.ts
```diff
@@ -168,11 +160,6 @@ export const defaultAriaLiveMessages = {
       return `value ${label} focused, ${getArrayIndex(selectValue, focused)}.`;
     }
 
-    if (context === 'menu') {
-      const disabled = isDisabled ? ' disabled' : '';
-      const status = `${isSelected ? 'selected' : 'focused'}${disabled}`;
-      return `option ${label} ${status}, ${getArrayIndex(options, focused)}.`;
-    }
```
lukebennett88, line 175: Just clarifying, is this being removed because we're now using `aria-activedescendant` to convey the same information?

lukebennett88:
Thanks so much @Ke1sy!
I've done an initial pass of the PR, and everything looks good to me so far. However, to make the review process a little smoother, would you mind adding a few comments on the PR to explain some of the changes and additions?

packages/react-select/src/accessibility/index.ts
Ke1sy: Yes, you're right, now the user will be informed about disabled options because of aria-activedescendant: 'Blue unavailable 2 of 10'

packages/react-select/src/accessibility/index.ts
```diff
@@ -98,25 +100,23 @@ export interface AriaLiveMessages<
...
             ? ', press Tab to select the option and exit the menu'
             : ''
         }.`;
       case 'input':
-        return `${props['aria-label'] || 'Select'} is focused ${
-          isSearchable ? ',type to refine list' : ''
-        }, press Down to open the menu, ${
-          isMulti ? ' press left to focus selected values' : ''
-        }`;
+        return isInitialFocus
```
Ke1sy, line 113: This change is added to avoid hearing the instructions each time user's focus is moved back to the input. The instructions will be announced to user only on initial focusing of input.

packages/react-select/src/components/LiveRegion.tsx
```diff
@@ -184,19 +189,18 @@ const LiveRegion = <
...
     tabSelectsValue,
+    isInitialFocus,
   ]);
 
-  const ariaContext = `${ariaFocused} ${ariaResults} ${ariaGuidance}`;
-
   const ScreenReaderText = (
     <Fragment>
       <span id="aria-selection">{ariaSelected}</span>
-      <span id="aria-context">{ariaContext}</span>
```
Ke1sy, line 194: I separeted ariaContext into 3 blocks to allow only sufficient information be announced depending on the change instead of announcing large block (ariaFocused, ariaResults, ariaGuidance) all the time which causes negative user experience

packages/react-select/src/Select.tsx
```diff
@@ -605,7 +652,6 @@ export default class Select<
   commonProps: any; // TODO
   initialTouchX = 0;
   initialTouchY = 0;
-  instancePrefix = '';
```
Ke1sy, line 608: I moved instancePrefix  to state to be able to use it inside of getDerivedStateFromProps

packages/react-select/src/Select.tsx
```diff
@@ -441,6 +449,31 @@ function buildFocusableOptionsFromCategorizedOptions<
   );
 }
 
+function buildFocusableOptionsWithIds<Option, Group extends GroupBase<Option>>(
```
Ke1sy, line 452: This function calculates ids for focusable options as id is needed in 'aria-activedescendant'.

packages/react-select/src/Select.tsx
```diff
@@ -635,15 +681,21 @@ export default class Select<
...
-
     // Set focusedOption if menuIsOpen is set on init (e.g. defaultMenuIsOpen)
     if (props.menuIsOpen && this.state.selectValue.length) {
+      const focusableOptionsWithIds: FocusableOptionWithId<Option>[] =
+        this.getFocusableOptionsWithIds();
       const focusableOptions = this.buildFocusableOptions();
       const optionIndex = focusableOptions.indexOf(this.state.selectValue[0]);
+      this.state.focusableOptionsWithIds = focusableOptionsWithIds;
       this.state.focusedOption = focusableOptions[optionIndex];
+      this.state.focusedOptionId = getFocusedOptionId(
```
Ke1sy, line 695: We need update `focusedOptionId` each time when `focusedOption `is updated

packages/react-select/src/Select.tsx
```diff
@@ -1437,7 +1524,13 @@ export default class Select<
     if (this.blockOptionHover || this.state.focusedOption === focusedOption) {
       return;
     }
-    this.setState({ focusedOption });
+    const options = this.getFocusableOptions();
+    const focusedOptionIndex = options.indexOf(focusedOption!);
+    this.setState({
+      focusedOption,
+      focusedOptionId:
+        focusedOptionIndex > -1 ? this.getFocusedOptionId(focusedOption) : null,
```
Ke1sy, line 1532: Updating of focusedOptionId (aria-activedescendant) during hover

packages/react-select/src/Select.tsx
```diff
@@ -672,13 +725,28 @@ export default class Select<
...
+      const focusedOptionId = getFocusedOptionId(
+        focusableOptionsWithIds,
+        focusedOption
+      );
+
       newMenuOptionsState = {
         selectValue,
         focusedOption,
+        focusedOptionId,
+        focusableOptionsWithIds,
```
Ke1sy, line 749: recalculation of `focusableOptionsWithIds `and `focusedOptionId` after some dependencies has changed (line 720)

packages/react-select/src/Select.tsx
```diff
@@ -921,6 +991,7 @@ export default class Select<
     this.setState({
       focusedOption: options[nextFocus],
       focusedValue: null,
+      focusedOptionId: this.getFocusedOptionId(options[nextFocus]),
```
Ke1sy, line 994: after keyboard navigation

packages/react-select/src/__tests__/constants.ts
```diff
@@ -87,3 +87,21 @@ export const OPTIONS_ACCENTED: readonly OptionAccented[] = [
...
 ];
+
+export interface GroupedOption {
+  readonly label: string;
+  readonly options:
+    | readonly OptionNumberValue[]
+    | readonly OptionBooleanValue[];
+}
+
+export const OPTIONS_GROUPED: readonly GroupedOption[] = [
```
Ke1sy, line 98: Added this to test the 'aria-activedescendant' for grouped options as ther hve different format of id that includes the group index

packages/react-select/src/Select.tsx
```diff
@@ -844,6 +913,7 @@ export default class Select<
         inputIsHiddenAfterUpdate: false,
         focusedValue: null,
         focusedOption: focusableOptions[openAtIndex],
+        focusedOptionId: this.getFocusedOptionId(focusableOptions[openAtIndex]),
```
Ke1sy, line 916: after search update

Ke1sy:
@lukebennett88 I added additional comments for changes, please feel free to ask if something else is not clear. Thanks!

csandman:
This PR is looking great! It would be awesome to see this package use more standard `aria-` attributes for it's accessibility implementation.

mellis481:
For posterity, the ideal SR experience our team found includes the following definition for the `ariaLiveMessages` prop (with the changes in this PR):
```
{
  guidance: ({ context, isMulti }) => {
    if (context === "input" && isMulti) {
      return " Press left or right to manage selections";
    }
    if (context === "value") {
      return " Press delete to remove the focused option";
    }
    return "";
  }
}
```
You can test it live in [this](https://codesandbox.io/s/react-codesandboxer-example-forked-xnwsfq?file=/index.js) example (again, based off the changes in this PR).

CC: @csandman

mellis481:
@lukebennett88 Can we get this bad boy merged???

lukebennett88:
I just tested in Safari with VoiceOver, and it appears that this update is causing a regression. Prior to making a selection for the first time, every item is announced in the same manner.
https://github.com/JedWatson/react-select/assets/3422401/dedf70b3-4f73-4db0-8caa-71adfc441d9e

mellis481:
@lukebennett88 Thanks for your response. Since you posted your message, my team (@Ke1sy is a member) have been investigating the issue you reported and I want to share an update.

To first summarize the issue... in Safari + VoiceOver only, when navigating through options in a listbox, the options _above_ the selected option (which uses `aria-selected`) are read out as expected, but options that are _below_ the selected option are not announced at all. Instead, the selected option value is constantly repeated on each arrow down. 

The changes @Ke1sy made involved removing much of the aria-live implementation and adding correct ARIA attributes/roles. The first rule of aria-live is to not use aria-live unless you have to, preferring semantic markup and ARIA attributes/roles so the direction we went with the PR is the right one. That said, after doing some investigation, we found a number of articles calling out poor support in Safari for certain ARIA. Here is an article that calls out an issue with how Safari handles `aria-selected`: https://a11ysupport.io/tech/aria/aria-selected_attribute.

We have reviewed our approach and, as far as we can tell, we've applied the correct ARIA. Because this issue is only reproduceable in Safari + VoiceOver, coupled with the fact that there are numerous issues that others have logged online regarding Safari and how it doesn't handle ARIA properly, the only conclusion my team has been able to make is that it's a issue with Safari + VO.

The current react-select avoids this issue by using an excessive amount of aria-live which creates a very bad UX for all screen reader users, not just those using Safari + VoiceOver. These changes are certainly a net positive. I understand it's hard to consider it acceptable that these changes result in a suboptimal (broken?) experience for Safari + VoiceOver users, but what do you do if/when that is because of an issue with Safari and/or VoiceOver?  What would you recommend for a path forward?

mellis481:
@lukebennett88 Thanks for eliciting some advice and sharing what react-aria is doing. My team will try to work on this soon. Stay tuned.

packages/react-select/src/Select.tsx
```diff
@@ -1624,6 +1719,8 @@ export default class Select<
       'aria-labelledby': this.props['aria-labelledby'],
       'aria-required': required,
       role: 'combobox',
+      'aria-activedescendant': this.state.focusedOptionId || '',
+
       ...(menuIsOpen && {
         'aria-controls': this.getElementId('listbox'),
         'aria-owns': this.getElementId('listbox'),
```
csandman, line 1726: This `aria-owns` isn't really necessary anymore and could be removed, but it doesn't really hurt.

> When a combobox's popup is displayed, ensure the [aria-controls](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls) attribute on the combobox element is set to the [`id`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#id) of the popup `listbox`, `tree`, `grid`, or `dialog` element. This is how the relationship between the element with the `combobox` role and the popup it controls is indicated. (Note: In older ARIA specs, this was `aria-owns` rather than `aria-controls`, so you may see `aria-owns` in older combobox implementations. The `aria-owns` in the code should be updated to `aria-controls`!)
> 
> — https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role

csandman:
Another thought I had about this (I know this PR is still a WIP while the Safari + VoiceOver issues are sorted out), there is one other aria pattern that might make sense for this package. For the `listbox` role, there is a pattern for using real grouped options. Specifically, if you add the `role="group"` attribute to the `Group` component, along with `aria-labelledby={headerProps.id}`, screen readers are capable of keeping track of the fact that items are intentionally grouped, and when a new group is navigated to, the group heading is read out. This pattern could be nice for accurately reflecting the grouped state in an accessibly way, instead of only having that grouping available for visually-able people.

There is a complete example of this pattern in the W3C guide for the [`listbox`](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/) pattern: https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/

I actually gave it a shot, specifically with VoiceOver on Chrome, and found the results to work pretty well. The only problem I ran into was that Chrome always thought whatever group I was in was 1 of 1. I believe this is due to the `role="listbox"` attribute being on the `Menu` component, with the only direct descendant being the `MenuList` component. This was easily fixed for me by moving that attribute to the `MenuList` component. I can't see any reason why this wouldn't work, but correct me if I'm wrong.

This change would be very small requirement wise overall:

- Move the `role="listbox"` attribute from the `Menu` component to the `MenuList` component (ensuring the `aria-controls` attribute is still pointing at the right element).
- Add the `role="group"` element to the outer div wrapper of the `Group` component.
- Add the `aria-labelledby` attribute on that same div using the `id` that's passed to the `GroupHeading` component.
- Add the `role="presentation"` attribute to the `GroupHeading` wrapping div. I'm not sure if this part is actually necessary, but they have it in the example implementation, and it definitely couldn't hurt either way.

---

The one other small thing I thought of, was adding the [`aria-multiselectable`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-multiselectable) attribute to whichever element ends up with the `listbox` role. I'm not actually sure if this would affect anything screen reader wise, but it does appear to be a recommended attribute for the listbox widget.

Ke1sy:
@csandman thank you for suggestions, I will take a look what I can do

mellis481:
@lukebennett88 @Ke1sy has updated this PR to fallback to an aria-live-based solution for Apple devices. Other devices will have the significantly improved screen reader UX that @Ke1sy implemented that uses correct ARIA/roles. So please review at your earliest convenience.

The group enhancements that @csandman proposed are not included in this PR. They can be done in a follow-up PR.

SairamAlagapan:
@mellis481 Can someone please merge this?

lukebennett88:
Just dropping in to say I've seen the PR has been updated to address the previous feedback and I will test and review when I get some free time.
Thanks again for all the hard work that went into this.

tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "strict": true,
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "types": ["./node_modules/user-agent-data-types"]
```
lukebennett88, line 11: Do we need this? I think the triple-slash directive is enough.
I just tried removing it and didn't get errors after running `yarn type-check`.

lukebennett88 (approved):
Amazing job @Ke1sy, I've tested this out in VoiceOver and NVDA and it all looks good to me.
I only have one minor nitpick question regarding the types.

Ke1sy:
@lukebennett88 th

> Amazing job @Ke1sy, I've tested this out in VoiceOver and NVDA and it all looks good to me. I only have one minor nitpick question regarding the types.

Thanks! 
@lukebennett88 you are right regarding the types, I rechecked and removed that line from tsconfig, hope we can merge this PR soon.

tsconfig.json
Ke1sy: removed it

packages/react-select/src/Select.tsx
```diff
@@ -1624,9 +1721,12 @@ export default class Select<
       'aria-labelledby': this.props['aria-labelledby'],
       'aria-required': required,
       role: 'combobox',
+      'aria-activedescendant': this.isAppleDevice
```
spathon, line 1724: When server-side rendering this hydration throws an error
`Warning: Extra attributes from the server: aria-activedescendant`
Setting either both to empty sting `''` or `undefined` solves
`'aria-activedescendant': this.isAppleDevice ? '' : this.state.focusedOptionId || ''`
or
`'aria-activedescendant': this.isAppleDevice ? undefined : this.state.focusedOptionId || undefined`

khanakia: any update on the issue ?

giuliocaccin:
I'm so happy this awesome merge request got merged, I wanted to say thanks, because it made it unnecessary to use `react-select-event` to us, and brought a lot of clarity in all our tests, thus greatly improving developer experience. Impressive. Thanks a lot

IlirEdis:
Hi there,
I'm getting `Warning: Extra attributes from the server: aria-activedescendant` in Nextjs v14.1.0 and seract-select v5.8.0

Screenshot 2024-02-05 at 13 46 29


Any help how i could fix this?

jacobsfletch:
> Hi there, I'm getting `Warning: Extra attributes from the server: aria-activedescendant` in Nextjs v14.1.0 and react-select v5.8.0

Same. Others too. Check out our conversation here: https://github.com/JedWatson/react-select/pull/5758#discussion_r1384745986

nynevi:
great work, but please mention changes like these in the release:
https://github.com/JedWatson/react-select/pull/5758/changes#diff-328abc07734b139d44b1f23562aa6721f763a74cb76a22139cd23b5c02ebaa7cL1973
https://github.com/JedWatson/react-select/pull/5758/changes#diff-328abc07734b139d44b1f23562aa6721f763a74cb76a22139cd23b5c02ebaa7cR2092-R2096

these are breaking changes for both the Menu and the MenuList. Menu had an id, no more, now that id is on MenuList. for a developer using a custom Menu and MenuList, these are breaking changes

PR #5782, opened 2023-10-19
Pass form prop to input so it works even if the select is ouside the form tag

zvin:
Hello!

When using `CreatableSelect` outside of a form passing it a `form` prop, this prop is not passed to the created `input` resulting in the field not being submitted as part of the form.
This PR fixes it.

My problem was with `CreatableSelect` but it might fix other use cases as well.

### DIFF
--- a/.changeset/serious-jokes-beam.md
+++ b/.changeset/serious-jokes-beam.md
@@ -1,0 +1,5 @@
+---
+'react-select': patch
+---
+
+Pass the form prop to the created input in renderFormField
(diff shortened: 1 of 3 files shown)


zvin:
Using a regular select will submit the `myField` value.
```jsx
<div>
  <form action="/someAction" id="myForm">
    <input type="submit" />
  </form>
  <select form="myForm" name="myField">
    <option value="x">X</option>
    <option value="y">Y</option>
  </select>
<div>
```

Using React-Select, it won't.
```jsx
<div>
  <form action="/someAction" id="myForm">
    <input type="submit" />
  </form>
  <Select form="myForm" name="myField" options={[{ value:"x", label: "X" }, { value:"y", label: "Y" }]} />
<div>
```

This PR fixes it.

zwhitchcox:
Please merge! @JedWatson

Issue #5775, opened 2023-10-13
react-select does not scroll into view selected option when `menuPortalTarget={document.body}`


mrudowski:
react-select does not scroll into view selected option when 
we use `menuPortalTarget={document.body}`

In docs example:
https://react-select.com/advanced#portaling

Or please check this codesandbox:
https://codesandbox.io/s/codesandboxer-example-forked-c5xp8p?file=/example.tsx

tdambra:
When will this change be live?

konradekk:
A workaround: find the focused option HTML element in the document tree and `scrollIntoView` on it after the render.

Little bit more precisely: use `setTimeout` to `getElementById` or first from `getElementsByClassName` (whichever suits you better) and apply `scrollIntoView` on it, with e.g. `{ behavior: "smooth", block: "center", inline: "start" }`.

Cosmitar:
I tried @konradekk's workaround and even when it works it also adds some unexpected behavior when the menu is hovered (the scroll jumps one option).
Since I was using custom options I decided to fix it inspired by [#2112](https://github.com/neetozone/neeto-ui/pull/2112/files)
Using the `innerRef` prop of the Option component and a `scrollIntoView` inside a `useEffect` hook when option is selected.

Brqqq:
I've tried scrolling into view the highlighted option on menu open. But it has weird consequences when the dropdown is inside a modal, which results in the whole page scrolling in unpredictable ways.

To solve it, I've used `npx patch-package` to patch the `Select-40119e12.esm.js` file

from
```
    _this.getMenuListRef = function (ref) {
      _this.menuListRef = ref;
    };
```

to 

```
    _this.getMenuListRef = function (ref) {
      _this.menuListRef = ref;
      if (_this.menuListRef && _this.focusedOptionRef && _this.scrollToFocusedOptionOnUpdate) {
        scrollIntoView(_this.menuListRef, _this.focusedOptionRef);
        _this.scrollToFocusedOptionOnUpdate = false;
      }
    };
```

If you override the `MenuPortal` component, be sure your component has a stable reference. So don't do:

```
<ReactSelect
    components={{ 
        MenuPortal: (props) => <components.MenuPortal {...props} />
    }}
/>
```

because this will cause unnecessary remounting of the `MenuPortal` on each render, which breaks scrolling. Instead don't override it at all or do it like this:

```
const CustomMenuPortal = props => {
  return <components.MenuPortal {...props}  />
};

[...]

Issue #5663, opened 2023-06-07
Scroll option into view while using portals


pmorim:
## Description
I've stumbled upon a bug where the "scroll option into view" feature does not work for portaled menus.
Although, it seems to work fine for non-portaled menus.

## Affected Versions
I've found this issue to have been introduced in `v4`, and it is still present in `v5`.

## Screen recording
![a gif of the bug](https://github.com/JedWatson/react-select/assets/77199990/d31bfc0e-c934-4664-9e5d-ac61704f6424)

## How to reproduce the bug
- Open the official docs
  - The same can be reproduced on the Storybook
- Navigate to `Advanced` > `Portaling`
- Open the Modal and its inner Menu
- Scroll down and pick an option that is outside of the initial scroll position
  - The last option for example - `Silver`
- Close the menu and open it again
- The menu will not scroll back to the selected option

## What is the expected behavior
The menu should scroll back to the selected option

## Notes
I have taken a look into the code and did not manage to find the issue. I wouldn't mind doing the PR myself if anyone can manage to point me in the right direction.

Issue #3648, opened 2019-06-24
React-select not scrolling to the selected value on menu re-open


ColbySong:
**react-select: 3.0.4**

After selecting a value and then re-opening the menu, the component does not scroll the selected value into view.

This behaviour occurs for both standard options and also the new options with group headers.

Options with group header issue:
https://codesandbox.io/s/p68rh?module=/example.js
- select an option far down the list
- re-open menu, selected element is highlighted but not scrolled into view when menu is opened

Standard options:
https://codesandbox.io/s/kwv9t?module=/example.js
- select option 'silver' which is far down the list
- re-open menu and you will see the view does not scroll to the selected value

similar issue reported by #3535 for custom options as well

ankush29:
solved https://github.com/JedWatson/react-select/pull/3652/commits/bb9c1d7bf53a68aa6bd4e6be5c40c3b0c14a7c01

csantos1113:
Same issue as https://github.com/JedWatson/react-select/issues/3656

csantos1113:
Thanks @ankush29 - will be fixed on PR https://github.com/JedWatson/react-select/pull/3652

imnikij:
Hey @csantos1113 I see this issue as still an open one, any idea by when this will be fixed ?

tonistenno:
Really waiting for this fix

alexmnv:
Had the same issue. The problem turned out to be that the option object passed to `value` prop was not  the same object present in `options` array, and `react-select` uses strict equality comparison (options.indexOf) to find selected option. E.g.:
```
<Select
  options={options}
  
  // wrong: new object is created. `options.indexOf(value)` fails to find the option index
  value={{
	id: selectedId,
	name: selectedName
  }}
/>

<Select
  options={options}
  
  // correct: `value` can be found in `options` array
  value={options[selectedIndex]}
/>
```

milosh86:
This issue is introduced in 3.0.4 release, probably this feature https://github.com/JedWatson/react-select/issues/3648 cause it.

Hodor9898:
I am having the same issue

TBear79:
Yes. I also have the same issue. I'm objects with same reference as value but it doesn't seem to be working. When is this going to be fixed?

AnthonyCrowcroft:
just rolled back to version 3.0.3 and it seems to still have the issue

AnthonyCrowcroft:
[...]

PR #5784, opened 2023-10-19
Scroll selected option into view when receiving refs

JefHellemansBlockbax:
Fixes:
- https://github.com/JedWatson/react-select/issues/5775
- https://github.com/JedWatson/react-select/issues/5663
- (probably) https://github.com/JedWatson/react-select/issues/3648

And is a less hacky solution than: https://github.com/JedWatson/react-select/pull/4477


JefHellemansBlockbax:
While looking at this again I noticed that another option is to make these refs part of the state. That way, when they get updated the `componentDidUpdate` function will trigger, causing the `scrollIntoView` call in there to be called. That is probably the best solution, and I'm willing to implement it in this PR if that helps. Feedback on both ideas (or either one really) would be much appreciated. We would love to get an updated version of `react-select` released in our product, as this functionality is currently broken for us. Based on the issues linked, it looks like we won't be the only ones benefitting from this change.

For now we'll implement a patch in our own code base, but that is something we would rather get rid of as soon as possible.

madelahsan (commented):
please merge it

Drustus (approved):

PR #5792, opened 2023-11-01
replace text 'npm' with 'yarn'

bunnyBites:
The build command is for 'yarn', but the label for that command has 'npm', which can be confusing for newcomers (me too).

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Features include:
 
 # Installation and usage
 
-The easiest way to use react-select is to install it from npm and build it into your app with Webpack.
+The easiest way to use react-select is to install it from yarn and build it into your app with Webpack.
 
 ```
 yarn add react-select


csandman:
That line isn't saying "install it using npm", it's saying "install it from the npm package registry": https://www.npmjs.com/package/react-select

So this change doesn't actually make sense. If anything, it would make more sense to just change the command to include both:

> The easiest way to use react-select is to install it from npm and build it into your app with Webpack.
> 
> ```
> npm install react-select
> # or
> yarn add react-select
> ```

htech9:
> That line isn't saying "install it using npm", it's saying "install it from the npm package registry": https://www.npmjs.com/package/react-select
> 
> So this change doesn't actually make sense. If anything, it would make more sense to just change the command to include both:
> 
> > The easiest way to use react-select is to install it from npm and build it into your app with Webpack.
> > ```
> > npm install react-select
> > # or
> > yarn add react-select
> > ```

I agree with this

htech9:
It seems that only yarn is referenced everywhere, so it looks like more that the initial PR makes sense.

 @JedWatson @gwyneplaine would you confirm this?

htech9 (approved):

Issue #5642, opened 2023-05-22
menuPlacement=auto is not recalculated when options change after menu is opened


joshhunt:
When using `AsyncSelect` with `menuPosition="auto"` towards the bottom of the page, once the options load the menu will be off screen.

![image](https://github.com/JedWatson/react-select/assets/46142/eb3363c9-4997-4bdd-bd89-e1bf39f137d8)

It appears the menu position is not recalculated if the options/size of the menu changes after its opened.

https://github.com/JedWatson/react-select/blob/3aee85cb8c5d045d47d45b0f9ebd0c58d4593c42/packages/react-select/src/components/Menu.tsx#L342-L363 

Reproduction sandbox https://codesandbox.io/s/react-select-menuplacement-repro-wqkuew?file=/src/App.js  https://wqkuew.csb.app/

ayuka-bg:
helllooo anybody alive here?

labkey-nicka:
This is addressed with #5736.

PR #5736, opened 2023-08-22
Issue #5733: MenuPlacer useLayoutEffect depend on isLoading

labkey-nicka:
Fixes #5733. This PR passes the `isLoading` prop down to the `MenuPlacer` component and then relies on `isLoading` as a dependency for running the `useLayoutEffect` to recompute the position of the menu. Subsequently, `getMenuPlacement` will recalculate the menu's position.

PR #5827, opened 2023-12-12
Fix: recalculate menu placement on options length

ayuka-bg:
https://github.com/JedWatson/react-select/issues/5642


labkey-nicka:
After discussing with @ayuka-bg I recommend #5736 in lieu of this change.

Issue #5329, opened 2022-09-02
Z-Index doesn't seem to be configurable


haveamission:
**Thanks for using react-select!**

If you are going to ask a question or want to propose a change or a new feature, then please don't file an issue for this.
Questions and feature requests have their own place in our discussions section.

## Are you reporting a bug or runtime error?

A bug. No matter what I try, looking back at the examples nor at the previous issue queue, I cannot affect z-index whatsoever.

<img width="793" alt="Screen Shot 2022-09-02 at 3 29 48 PM" src="https://user-images.githubusercontent.com/10475533/188224051-a292a21e-0a26-4a5e-86f4-fa4ebfefabf1.png">

Code is fairly standard:

```
const modelSelector = [
  { value: 'Ferrari 812 GTS', label: 'Ferrari 812 GTS' },
  { value: 'Ferrari 296 GTB', label: 'Ferrari 296 GTB' },
  { value: 'Ferrari 296 GTS', label: 'Ferrari 296 GTS' },
]

const ModelSelector = () => (
  <Select options={modelSelector} placeholder="Model" />
)
```

I've tried all sorts of style stuff, like this:

`const customStyles = {
  option: (provided: any, state: any) => ({
    ...provided,
    zIndex: 999,
  }),
  control: () => ({
    zIndex: 999,
  }),
  singleValue: (provided: any, state: any) => {
    const opacity = state.isDisabled ? 0.5 : 1
    const transition = 'opacity 300ms'

    return { ...provided, opacity, transition, zIndex: 99999 }
  },
}`

No luck.

This is in a React Native Web application, if that makes any difference.

How do I get z-index to work?

eckmLJE:
I believe you want to target `menuPortal` in the select's style object:

```tsx
<Select
  ...
  styles={{
    menuPortal: provided => ({
      zIndex: 100
    })
  }}
/>
```

The menu portal is the positioned container that holds the menu. Depending on your stacking context / z-index setup, you may need to use `menuPosition="fixed"` as well on the select.

GuySartorelli:
I opened a PR to address this late last year: https://github.com/JedWatson/react-select/pull/5830

eckmLJE:
[...]

PR #5830, opened 2023-12-19
Fix: Ensure menu is always in front of everything

GuySartorelli:
Fixes https://github.com/JedWatson/react-select/issues/5329

It's not uncommon for elements to have a z-index of >1, which would put them _in front_ of the dropdown menu.
There's no valid use case that I can think of where you'd actually want something to be in front of the dropdown menu, so the menu should have a really high z-index value to ensure it's on top of everything all the time.

In the unlikely event someone wants a specific element to be in front of the menu, they can either give that element a really high z-index, or change the z-index of menu for their specific use case - but we shouldn't have to customise the z-index to get the menu to display on top. The menu being on top should be the default behaviour.

### DIFF
--- a/packages/react-select/src/components/Menu.tsx
+++ b/packages/react-select/src/components/Menu.tsx
@@ -289,7 +289,7 @@ export const menuCSS = <
   [alignToControl(placement)]: '100%',
   position: 'absolute',
   width: '100%',
-  zIndex: 1,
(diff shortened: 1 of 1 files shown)


GuySartorelli:
Let me know if you want any tests, docs, etc - I didn't see a contribution guide so I just went with the least invasive change.

PR #5842, opened 2024-01-09
Fix event handling with react-beautiful-dnd

### LINKED ISSUE #5833: Select behaves inconsistently when receiving event with defaultPrevented
### Summary

In version 5.3 of react-select, a check for `defaultPrevented` was added to the onMouseDown event for the Select control, in order to prevent double events when clicking on the dropdown button or other interactive elements inside the Select: https://github.com/JedWatson/react-select/pull/5134.

This has prevented double events, but as a side-effect, it changed how Select responds to preventDefault. Now, if an `onMouseDown` event has default prevented before it reaches Select, the component behaves inconsistently. If the user presses on the dropdown, it still opens the select - but if the user presses the container, it doesn't.

This manifests most noticeably when a parent element calls preventDefault during the capture phase; an example of a library that does this is `react-beautiful-dnd`.

This may be in the realm of undefined behaviour, but an alternate approach to preventing duplicate events would provide more consistent behaviour for Select, and avoid edge cases like this where part of a select is interactive, but not others.

### Steps to reproduce

**Minimal reproduction:** https://codesandbox.io/p/sandbox/reproduction-select-with-a-preventdefault-parent-656c4l?file=%2Findex.tsx%3A57%2C28

Try pressing the selects, on the `input`, on the dropdown button, and on the main container. I've included a minimal example, as well as an example of the component inside react-beautiful-dnd

### Related issues
This issue triggered another consumer of react-select and RBD to refactor: https://github.com/grafana/grafana/pull/50025

itsdouges:
This pull request removes logic to bail out from an event if someone else listening to the event (usually in a capture phase) has called `preventDefault()`. This fixes interop between `react-select` and `react-beautiful-dnd`.

To ensure things don't regress, and to ensure to not contradict what was solved in https://github.com/JedWatson/react-select/pull/5134 I've added four e2e tests covering the various flows in both:

- an emulated react-beautiful-dnd environment, and
- direct react-calendar interop

All tests pass, the same behaviour remains, hopefully we're good to go.

**TODO**

- [ ] fix tests
- [ ] assess the solution and seek alternatives if needed

Closes https://github.com/JedWatson/react-select/issues/5833

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -1352,6 +1352,7 @@ export default class Select<
       return;
     }
     this.clearValue();
+    event.stopPropagation();
     event.preventDefault();
     this.openAfterFocus = false;
     if (event.type === 'touchend') {
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -1293,10 +1293,6 @@ export default class Select<
   onControlMouseDown = (
     event: React.MouseEvent<HTMLDivElement> | React.TouchEvent<HTMLDivElement>
   ) => {
-    // Event captured by dropdown indicator
-    if (event.defaultPrevented) {
-      return;
-    }
     const { openMenuOnClick } = this.props;
     if (!this.state.isFocused) {
       if (openMenuOnClick) {
--- a/.changeset/small-steaks-decide.md
+++ b/.changeset/small-steaks-decide.md
@@ -1,0 +1,5 @@
+---
+'react-select': patch
+---
+
+Select no longer stops the control mouse down event from firing if someone listening to the event calls `.preventDefault()`. This resolves issues seen with react-select interop-ing with other libraries, such as react-beautiful-dnd.
--- a/cypress/fixtures/selectors.json
+++ b/cypress/fixtures/selectors.json
@@ -24,5 +24,15 @@
   "multiSelectInput": "#react-select-multi-select-input",
   "placeHolderMulti": "#multi-select .react-select__placeholder",
   "toggleMenuMulti": "#multi-select .react-select__dropdown-indicator",
-  "focusedOption": ".react-select__option--is-focused"
+  "focusedOption": ".react-select__option--is-focused",
+  "preventDefaultTest": {
+    "valueContainer": ".prevent-default-test__value-container",
+    "indicator": ".prevent-default-test__indicator",
+    "menu": ".prevent-default-test__menu"
+  },
+  "bubblingTest": {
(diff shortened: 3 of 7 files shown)


itsdouges:
cc @rkulinski @flexdinesh

packages/react-select/src/Select.tsx
```diff
@@ -1293,10 +1293,6 @@ export default class Select<
   onControlMouseDown = (
     event: React.MouseEvent<HTMLDivElement> | React.TouchEvent<HTMLDivElement>
   ) => {
-    // Event captured by dropdown indicator
```
theKashey, line 1296: is there a test for this particular use case - dropdown indicator preventing default event (https://github.com/JedWatson/react-select/blob/06e34882638d1526b9f5a1238bb567a3e9460ce5/packages/react-select/src/Select.tsx#L1345)

Also, not quite clear why `stopPropagation` is not a case, `onMenuMouseDown` uses it (uses both) for example.

----

This code was added here for a reason, and I can see other pieces relying on this one. Basically - all `.preventDefault` in this file.

What if we try to change the game rules and just rely on a different information source?
```tsx
const createEventTracker = () => {
  const tracker = new WeakSet();
  return {
       mark: (event) => {
        tracker.add(event.target);
        Promise.resolve().then(() => tracker.delete(event.target);
      },
      isMarked: (event) => tracker.has(event.target);
 }
} 

/// ....
const mouseDownTracker = createEventTracker();

onDropdownIndicatorMouseDown = (e) => {
  // ...
  mouseDownTracker.mark(e);
}

onControlMouseDown = (event) => {
    // Event captured by dropdown indicator
    if (mouseDownTracker.isMarked(event)) {
      return;
    }
}
```

Might be a little too much code to support a single line, but this code will keep the old behavior and unlock our usecase.

itsdouges: This prevent default was added in https://github.com/JedWatson/react-select/pull/5134 — I can't imagine anything in Select is leaning on it?

We can definitely change the rules like you said if the behaviour is needed. I'm just not sure it is.

itsdouges: The failing tests might point to it being needed I suppose ;).

theKashey: Right, and you added tests for #5050 case, and they are passing. May be something else in between, including changes in DatePicker resolved the problem in the meanwhile

itsdouges: Yeah you're right, which is what the tests are showing now. I'll dig in soon.

Issue #4988, opened 2022-01-11
Accessibility: no way to clear the selection with a keyboard


dgreene1:
Caveat: I'm not an accessibility expert so perhaps there is an alternative mechanism available to clear selections. But the usual way (tabbing to the clear button) is not working.

**Reproduction Steps**

1. open NVDA
2. visit https://react-select.com/
3. tab to any select example that has a selected item and that is clearable (at the time of this writing the first example satisfies these conditions)
4. tab again

**Expected Behavior**

I would expect Step 4 to focus the "X" that clears the selected option.

**Actual Behavior**

You can't focus the clear button.

**Proposed Solution**

Make the clear `div` a `button` tag.

**Am I Willing To Submit A PR For This?**

Yes, absolutely. The company I work for has budget for fixing a11y issues in the open source libraries we use or that we are considering using.

dgreene1:
Yes, whatever keyboard key we choose (space works for me) we need to give it tabIndex.

would a PR be accepted for this?

Rall3n:
Hello @dgreene1 and @LauRoxx,

`react-select` tries to follow the [`combobox` W3C spec](https://www.w3.org/TR/wai-aria-practices/#combobox).
The spec dictates the following regarding keyboard interaction or more specific, tabbing:
> Tab: The combobox is in the page Tab sequence.
Note: The popup indicator icon or button (if present), the popup, and the popup descendants are excluded from the page Tab sequence.
> _[source](https://www.w3.org/TR/wai-aria-practices/#keyboard-interaction-6)_

In other words, sub-components should not be able to be focused by pressing the Tab key.

The intended way to clear the selection using a keyboard is by pressing the `Delete` key, with `Backspace` being an alternative ([source](https://www.w3.org/TR/wai-aria-practices/#listbox-popup-keyboard-interaction)).

> Would a PR be accepted for this?

You can create a PR which changes the `div` to a `button`, but that would only be acceptable if it behaves the same as without the PR in regards to default tabbing behavior. It should allow for developing users to change the behavior more easily by omitting relevant props (`aria-hidden`, `tabIndex`) instead of recreating the whole component.

dgreene1:
@Rall3n I agree to the spec but it’s just sad for the user that it’s a bit confusing. I concede though because following specs allows us to provide a common mental model for users so they can “port” their experiences from one application to another.

[...]

PR #5850, opened 2024-01-19
Fix: a11y, enabled clear indicator to be accessible via keyboard

yhy-1:
As discussed in https://github.com/JedWatson/react-select/issues/4988

note: Since focus needs to be set back to the select input, it announces that the "combo box has auto-complete select blank" instead of what is in the aria-live since focus takes precedence by the screen reader.

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -679,10 +679,7 @@ export default class Select<
   getInputRef: RefCallback<HTMLInputElement> = (ref) => {
     this.inputRef = ref;
   };
-  // clearIndicatorRef: HTMLButtonElement | null = null;
-  // getClearIndicatorRef: RefCallback<HTMLButtonElement> = (ref) => {
-  //   this.clearIndicatorRef = ref;
-  // };
+
   // Lifecycle
   // ------------------------------
 
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -176,6 +176,8 @@ export interface Props<
   instanceId?: number | string;
   /** Is the select value clearable */
   isClearable?: boolean;
+  /** enabled clear indicator to accessible via keyboard and screen-reader */
+  enableAccessibleClearIndicator?: boolean;
   /** Is the select disabled */
   isDisabled: boolean;
   /** Is the select in a state of loading (async) */
@@ -677,7 +679,10 @@ export default class Select<
   getInputRef: RefCallback<HTMLInputElement> = (ref) => {
     this.inputRef = ref;
   };
-
+  // clearIndicatorRef: HTMLButtonElement | null = null;
+  // getClearIndicatorRef: RefCallback<HTMLButtonElement> = (ref) => {
+  //   this.clearIndicatorRef = ref;
+  // };
   // Lifecycle
   // ------------------------------
 
@@ -734,7 +739,6 @@ export default class Select<
             `${instancePrefix}-option`
           )
         : [];
-
       const focusedValue = clearFocusValueOnUpdate
         ? getNextFocusedValue(state, selectValue)
         : null;
@@ -743,7 +747,6 @@ export default class Select<
         focusableOptionsWithIds,
         focusedOption
       );
-
       newMenuOptionsState = {
         selectValue,
         focusedOption,
@@ -764,8 +767,9 @@ export default class Select<
 
     let newAriaSelection = ariaSelection;
 
-    let hasKeptFocus = isFocused && prevWasFocused;
-
+    let hasKeptFocus =
+      isFocused && (prevWasFocused || ariaSelection?.action === 'clear');
+    //let isCleared = ariaSelection?.action === 'clear';
     if (isFocused && !hasKeptFocus) {
       // If `value` or `defaultValue` props are not empty then announce them
       // when the Select is initially focused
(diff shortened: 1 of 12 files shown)


packages/react-select/src/components/indicators.tsx
```diff
@@ -92,10 +92,18 @@ const baseCSS = <
         ':hover': {
           color: isFocused ? colors.neutral80 : colors.neutral40,
         },
+        border: 'none',
+        background: 'none',
+        ':focus': {
+          borderColor: !isFocused ? colors.primary : 'none',
+          outline: 'unset',
```
yhy-1, line 99: Added some focus css, this can be updated or removed

packages/react-select/src/Select.tsx
```diff
@@ -764,7 +764,8 @@ export default class Select<
 
     let newAriaSelection = ariaSelection;
 
-    let hasKeptFocus = isFocused && prevWasFocused;
+    let hasKeptFocus =
```
yhy-1, line 767: Ensure that since we are setting focus back to select input after clearing the value, initial-input-focus is not called again.

yhy-1:
@lukebennett88 @JedWatson @dcousens @nderkim @Rall3n @Methuselah96 , would you mind reviewing this PR.

yhy-1:
@lukebennett88 @JedWatson @dcousens @nderkim @Rall3n @Methuselah96 

Hi, would appreciate it if you have time to review this PR.

mellis481:
@lukebennett88 @JedWatson @dcousens @nderkim @Rall3n @Methuselah96 Can you please review/approve this PR? It's a critical accessibility feature that countless teams need.

jossmac:
I'm not an a11y expert, but I have some thoughts:
- accessible patterns should be the default, rather that making consumers opt-in e.g. `enableAccessibleClearIndicator`
- using hard-coded english for labels will become a problem e.g. `aria-label="clear"`
- my initial instinct is that `ESC` press, when the dialog is closed, should clear the value. i feel like the clear button would introduce an unexpected tab stop; like, a combobox doesn't have a focusable button for invoking the menu, you just press `DOWN|UP`

Interested to hear how others would approach this problem.

dcousens:
@yhy-1 as suggested by @jossmac, could you update your pull request to announce the `DEL` shortcut in the `aria-live`? (see https://github.com/JedWatson/react-select/blob/06e34882638d1526b9f5a1238bb567a3e9460ce5/packages/react-select/src/components/LiveRegion.tsx)

dcousens:
Sorry @yhy-1 that I missed your previous pings/notifications

yhy-1:
@jossmac @dcousens Hi, there isn't a consensus on the accessible patterns for this clear button, that is why it was made optional props in this PR. You can read in the discussion at https://github.com/JedWatson/react-select/issues/4988. Our company had flagged this as a usability issue. We want the ability to enable this to be keyboard accessible but have no means to do so. I can update the hard-code aria-label to add the ability for an aria-label. Would this PR be accepted? The aria-live already consists of so much stuff, that it would eventually get lost and won't meet the requirement for our company.

jossmac:
Hi @yhy-1, thanks for getting back to us.

I still feel that introducing an unexpected tab stop isn't the way to go. Consider an `<input type="search" />`, the clear behaviour is accessible via <kbd>ESC</kbd> press without the clear button being focusable.

> We want the ability to enable this to be keyboard accessible but have no means to do so.

Fair, this is partially true. Users may clear the value via <kbd>DEL</kbd> press, though it isn't currently discoverable.

jossmac:
Thanks for getting back to us. Though, I'm still not comfortable introducing an unexpected tab stop.

> there isn't a consensus on the accessible patterns for this clear button

After a little research, there's already a [keyboard interaction](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#keyboardinteraction) spec for this behaviour on the combobox pattern:

> <kbd>Escape</kbd>: Dismisses the popup if it is visible. Optionally, if the popup is hidden before <kbd>Escape</kbd> is pressed, clears the combobox.

@dcousens we should support the spec behaviour*. @yhy-1 This would meet the requirement for your company, right?

*Maybe in addition to <kbd>Delete</kbd> press? We'd need to prevent default on the <kbd>Escape</kbd> keyboard event, so consumers can respond accordingly when implementing modal dialog close behaviour etc.

mellis481:
@jossmac  It looks like currently:
- `DELETE` and `BACKSPACE` **will** clear input if the popup is **not** visible.
- `DELETE` and `BACKSPACE` will **not** clear input if the popup **is** visible.
- `ESCAPE` has no effect on clearing input.

I agree that react-select should support the clearing of input when `ESCAPE` is pressed when the popup is not visible.

zayprom:
@jossmac Understood your concerns. One thing is to look on WCAG recommendation about combobox, pattern, other thing is maybe also more general keyboard accessibility approaches. As a rule of thumb - element which is operable via mouse should be also operable via keyboard - meaning focus and potential actions. I understand concerns about clearable buttons being inner elements of react-select and potential "break" of tab sequence, but these elements are still operable via mouse. In many cases users are not always well educated about potential shortcuts as "Delete", "Backspace" or "Escape" and it can be frustrating not understanding why mouse can do it but keyboard not. Just my 2 cents. Thanks!

Issue #5176, opened 2022-05-02
defaultPrevented issue with react-beautiful-dnd


loun4:
Since the following change 150c02028aa8e47f4071357e44f4c0daeda44a18 
React Select can no longer be opened through `onControlMouseDown` and the `event.defaultPrevented` is always `true` when the component is rendered inside react-beautiful-dnd.

CSB: https://codesandbox.io/s/vertical-list-forked-mucu2g?file=/index.js

@rkulinski can you consider using internal flag instead of `defaultPrevented` (53cc8ce7f87ed660276361b6d89e8fb175c19ac6) to ensure that the event is captured by dropdown indicator ?

rkulinski:
Hey. That was my initial change - usage of internal flag. See this conversation
https://github.com/JedWatson/react-select/pull/5089

Possibly this would require some other kind of patch.

tlenclos:
Can reproduce also after upgrading to 5.3.

loun4:
@Rall3n @rkulinski Is there a plan to remove the `defaultPrevented` or provide a way to override the default behavior  ?

kamen-kirilov:
Is there any solution to that one? Still issue with **v5.6.0**

sreahard:
It would be great if there was a fix for this, I did work up a fiddle to get this working with react-beautiful-dnd, not ideal, but workable if any one else is running into this. https://codesandbox.io/s/vertical-list-forked-tqwc9v?file=/index.js

josephnle:
I found another workaround in react-beautiful-dnd. It's still ugly but I'm not yet sure what react-select can do to accommodate the case when a library like react-beautiful-dnd binds to global `mousedown`.

https://github.com/atlassian/react-beautiful-dnd/issues/1810#issuecomment-1077952496

ranafaraznaru:
Can someone tell me why we can't select text by double click if text is inside draggable div?

hoangmf:
Hi everyone,
It still seems like a problem. I'm faced with it.

htech9:
Since `react-beautiful-dnd` is now deprecated (see https://github.com/atlassian/react-beautiful-dnd/issues/2672)
would it be a good time to close this issue ?

PR #5851, opened 2024-01-22
Fix unspecific and unstable logic on control mouse down

### LINKED ISSUE #5833: Select behaves inconsistently when receiving event with defaultPrevented

mdo2:
Use internal member to check if control onMouseDown should act.

This is intended to be a temporal fix, as a better solution must be developed.

Closes #5833, #5176

Another PR about this: #5842

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -657,6 +657,7 @@ export default class Select<
   scrollToFocusedOptionOnUpdate = false;
   userIsDragging?: boolean;
   isAppleDevice = isAppleDevice();
+  mouseDownTriggeredBeyondControl = false;
 
   // Refs
   // ------------------------------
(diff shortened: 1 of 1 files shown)


mdo2:
@lukebennett88, @emmatown, sorry for bothering you, but can you review this small fix?

This issue is blocking my team and many others. Also, there exist two issues, at least complaining about this behavior.

I appreciate any help you can provide.

Issue #5562, opened 2023-02-16
aria-describedby not respected


aakashsigdel:
## Are you reporting a bug or runtime error? -> BUG

react-select doesn't respect `aria-describedby` as prop. This results in for errors not being pronounced by screen readers when focusing back on the field.

```jsx
    <AsyncSelect
      onChange={handleSelect}
      loadOptions={loadOptions}
      placeholder={placeholder}
      inputId={id}
      aria-describedby="id-of-another-element"
    />

```

Example:
https://codesandbox.io/s/react-select-v5-sandbox-forked-i1vu1l

cabillin:
+1 on this being an issue that I'd like to see fixed

Thanks for the workaround @wbolduc!

prathamesh-voltup:
Thank you! @wbolduc

robinmetral:
Duplicate of #1570. I agree it should be fixed, but it's probably best to keep discussions in the 8-year-old issue.

PR #5871, opened 2024-03-06
Add option for custom aria-describedby

Superman2971:
Provides support for issues: #5562 and #1570 

I've tested my fork on local repository to confirm this does indeed pass the provided `aria-describedby` to the react-select. Also tested and passed the `yarn lint`, `yarn type-check`, and `yarn test` per CONTRIBUTING.MD

Screenshot 2024-03-06 at 2 24 54 PM

Our team needs this implemented to meet a11y support for our use of react-select. We have some custom hint and error messaging that is related to the Select and needs to be accessibly connected through aria-describedby. I'm hoping we can contribute to the library and make this available in a minor update to the package. Please let me know what I can do to assist in getting this PR approved and ready to merge :) Cheers!

### DIFF
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -1738,7 +1738,9 @@ export default class Select<
             'aria-describedby': this.getElementId('live-region'),
           }
         : {
-            'aria-describedby': this.props['aria-describedby'] || this.getElementId('placeholder'),
+            'aria-describedby':
+              this.props['aria-describedby'] ||
+              this.getElementId('placeholder'),
           }),
     };
 
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -87,6 +87,8 @@ export interface Props<
   'aria-labelledby'?: AriaAttributes['aria-labelledby'];
   /** Used to set the priority with which screen reader should treat updates to live regions. The possible settings are: off, polite (default) or assertive */
   'aria-live'?: AriaAttributes['aria-live'];
+  /** HTML ID of an element that should be used as a description (for assistive tech) */
+  'aria-describedby'?: AriaAttributes['aria-describedby'];
   /** Customise the messages used by the aria-live component */
   ariaLiveMessages?: AriaLiveMessages<Option, IsMulti, Group>;
   /** Focus the control when it is mounted */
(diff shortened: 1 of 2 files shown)


Superman2971:
I'm afraid I can't re-run the unit tests and I'm unsure why the tests aren't running in CircleCI.

Screenshot 2024-03-06 at 3 35 45 PM
Screenshot 2024-03-06 at 3 33 25 PM

Locally all tests are running and pass (new test included)

Screenshot 2024-03-06 at 3 33 54 PM
Screenshot 2024-03-06 at 3 34 08 PM

Hopefully this won't block the PR from moving forward. Let me know any progress I can make for this on my end and I'll jump on it.

plahteenlahti:
Hey @Superman2971 I think your test is correct but you just need to run `prettier:check` and fix the error, which I think is that 

```
'aria-describedby': this.props['aria-describedby'] || this.getElementId('placeholder'),
```

should be on two lines to so that it goes under the max line length of prettiest default setting

Superman2971:
> you just need to run `prettier:check` and fix the error

Thank you @plahteenlahti! That was a good catch 👍 Now fixed 😄 ✅

alex84G:
We need this merge as well!

packages/react-select/src/Select.tsx
```diff
@@ -1738,7 +1740,9 @@ export default class Select<
             'aria-describedby': this.getElementId('live-region'),
           }
         : {
-            'aria-describedby': this.getElementId('placeholder'),
+            'aria-describedby':
+              this.props['aria-describedby'] ||
+              this.getElementId('placeholder'),
```
lukebennett88, line 1745: Perhaps this should be:

```suggestion
            'aria-describedby': [
              this.props['aria-describedby'],
              this.getElementId('placeholder'),
            ]
              .filter(Boolean)
              .join(' '),
```

oliverfarrell:
Is there anything in particular stopping this from being merged? Is it not something `react-select` is looking to support? There's definitely a use case for wanting a `<Select />` with `isSearchable={false}` to also be described by another element on the page and to not solely rely on `aria-errormessage`. As a custom `Input` is not rendered when `isSearchable` is false, it makes particular behaviour tricky to achieve.

Issue #5875, opened 2024-03-13
JAWS screen reader highlight/focus issue


juanthedeveloper:
Issue: Focus/highlight does not work properly for non searchable and searchable options when using JAWS screen reader

I think the issue lies with the dummy input used for non searchable options. The dummy input ends up outside the select component which leads JAWS to focus that. For searchable option, the highlighting isn't as bad since it does at least do it inside the component but ideally should do it for entire component. Tabbing to it is the easiest way to trigger the issue but also does it sometimes when clicking.

This is an accessibility issue.

Non searchable option video:
https://github.com/JedWatson/react-select/assets/86272014/ba360e6e-f5a0-4971-9c06-42af92c67e6b

PR #5876, opened 2024-03-14
not working

juanthedeveloper:
Fixes issue #5875


juanthedeveloper:
unit_tests fail due to lint, but I cant seem to fix it? Get the same warnings before and after my changes

PR #5960, opened 2024-09-18
Fix pop-value being sent for undefined values.

leonaves:
Because of this array access:

https://github.com/JedWatson/react-select/blob/884f1c42549faad7cb210169223b427ad6f0c9fd/packages/react-select/src/Select.tsx#L1091

`removedValue` could technically be undefined. This happens when the select box is empty and a user hits backspace. It can cause runtime errors in the consuming libraries if the options are objects rather than primitives. This stops those actions being sent.

(The other option would be to update the typing in the below part of the code to mark removedValue as optional, forcing consumers to protect against potential undefined values, but I believe this makes more sense, as we are not removing anything if the set of values is already empty, so sending a `pop-value` action doesn't really make sense):

https://github.com/JedWatson/react-select/blob/edf5265ee0158c026c9e8527a6d0490a5ac2ef23/packages/react-select/src/types.ts#L138-L142

### DIFF
--- a/.changeset/calm-snakes-collect.md
+++ b/.changeset/calm-snakes-collect.md
@@ -1,0 +1,5 @@
+---
+'react-select': patch
+---
+
+No longer send pop-value action when multi-select is empty. This correctly resolves typings with that event, where removedValue cannot be undefined.
--- a/packages/react-select/src/Select.tsx
+++ b/packages/react-select/src/Select.tsx
@@ -1096,10 +1096,12 @@ export default class Select<
       newValueArray[0] || null
     );
 
-    this.onChange(newValue, {
-      action: 'pop-value',
-      removedValue: lastSelectedValue,
-    });
+    if (lastSelectedValue) {
+      this.onChange(newValue, {
+        action: 'pop-value',
+        removedValue: lastSelectedValue,
+      });
+    }
   };
 
   // ==============================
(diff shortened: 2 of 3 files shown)


leonaves:
@lukebennett88 @emmatown @Methuselah96 I know this library isn't super actively maintained at the moment, but this is a tiny change with test coverage so I'd love a real quick look and a patch release! Would be really appreciated! 🙏

Methuselah96:
Hmm, this could technically be considered a breaking change, but this is probably the best answer. I'll approve and merge if no one objects in the next 24 hours.

leonaves:
@Methuselah96 Yeah I considered that, but as the exact shape of these events isn't actually documented anywhere (as far as I can see) other than the exported types, and the exported types actually declare that `removedValue` will never be undefined, I'd say it's fine to call it a patch.

Methuselah96:
Unfortunately react-select is popular enough that Hyrum's Law is in full effect. If it turns out people were depending on this, and there's on other sufficient solution, I'd be quick to revert it.

leonaves:
> I'll approve and merge if no one objects in the next 24 hours.

👀 😄

Methuselah96 (approved):

lights-a5:
Commenting here that this change broke my tests for some reason this morning. I haven't been able to delve too deeply into it or to find out what I'm doing that is breaking it. I can confirm that it was working on version 5.8.0 and not 5.8.1.

kibertoad:
@lights-a5 Can you check if your tests are correct in this case? maybe they were checking that the wrong behaviour is wrong and actually need to be adjusted

lights-a5:
The problem is me it turns out. The require for `Select` had a d3 in its name and my jest config has a moduleNameMapper mapped from d3 to a d3 library. The only change from you was the slug changed in the dist that included `d3` in it.

Just bad regex on my part. Sorry for the alarm!

PR #5980, opened 2024-12-04
Scrolling of the selected menu option into view for menuportals

m4g4:
A fix for: https://github.com/JedWatson/react-select/issues/5663


madelahsan (changes requested):
Please remove isFocused from line number 886 as it is not being used and the test is being failed because of it.
image

konradekk:
cross-ref: possibly fixes #5775, as well as #5663 and (probably) #3648 mentioned there.

Thris3n (approved):

PR #6004, opened 2025-02-08
Bugfix: menuShouldBlockScroll now applies the correct right-padding

### LINKED ISSUE #5342: Scrolllock shifts layout.
When i set my vertical scrollbars always on with `menuShouldBlockScroll`, there is a padding-right added to the body which shifts entire layout and i get white stripe on the right end of the page. 
Not sure how it should act, but if I set my OS settings to auto hide scrollbars it works okay.
Maybe there should be a prop to opt out from this padding adjustment?

robxbob:
# Bug Description
When `menuShouldBlockScroll=true`, the right padding for the `body` tag is computed. This is to account for the right scrollbar disappearing when this prop is set. However, it computes this new right padding by accounting for any right padding applied through the `styles` prop; this doesn't take into account for right padding added from classes.

# Bug Example
An example of the bug can be seen here where a class based right padding is applied using tailwind:
https://codesandbox.io/p/devbox/react-select-v5-sandbox-forked-fckm5s
Opening and closing the select will cause a shift in padding due to incomplete padding extraction.  

# Solution
https://codesandbox.io/p/devbox/react-select-v5-sandbox-forked-qnd3sz?workspaceId=ws_CRHnC6VmWcMy6krqXNjCPZ
- `window.getComputedStyle` is used to get the final computed right padding which accounts for padding from a class or style. 
- Removed the floating `paddingRight` property from `STYLE_KEYS` as we are no longer using this style anywhere in the code.

# Closes
#5342

### DIFF
--- a/packages/react-select/src/internal/useScrollLock.ts
+++ b/packages/react-select/src/internal/useScrollLock.ts
@@ -1,11 +1,6 @@
 import { useCallback, useEffect, useRef } from 'react';
 
-const STYLE_KEYS = [
-  'boxSizing',
-  'height',
-  'overflow',
-  'position',
-] as const;
+const STYLE_KEYS = ['boxSizing', 'height', 'overflow', 'position'] as const;
 
 const LOCK_STYLES = {
   boxSizing: 'border-box', // account for possible declaration `width: 100%;` on body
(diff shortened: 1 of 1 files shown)


man-trackunit (approved):
Yes please 🙏

PR #5860, opened 2024-02-06
fix: hydration problem caused by isAppleDevice

zhouhua:
try to fix #5859

PR #6051, opened 2025-07-11
fix: hydration problem caused by isAppleDevice

### LINKED ISSUE #5859: react-select v5.8.0: `Extra attributes from the server: aria-activedescendant`
When react-select is used in SSR, an error may occur during browser rendering: Extra attributes from the server: aria-activedescendant. This should be a hydration problem.

I think it's caused by this line of code: https://github.com/JedWatson/react-select/blob/06e34882638d1526b9f5a1238bb567a3e9460ce5/packages/react-select/src/Select.tsx#L1724

It may be written in a useEffect to avoid differences in browser and server environments.

jnyholm-sc:
New try based on https://github.com/JedWatson/react-select/pull/5860
To fix: https://github.com/JedWatson/react-select/issues/5859

### DIFF
--- a/.changeset/three-pants-attend.md
+++ b/.changeset/three-pants-attend.md
@@ -1,0 +1,5 @@
+---
+'react-select': patch
+---
+
+fix: hydration problem caused by isAppleDevice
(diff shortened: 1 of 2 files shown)


Methuselah96 (approved):
Thanks!

PR #3324, opened 2019-01-05
Auto focus disabled with a new prop

mnpreetkaur:
#2848 #3061 
Above issues could be resolved with this enhancement

@gwyneplaine I've added a new prop `disableAutoFocus`, which allows deactivation of auto focus of first or any option on opening of the option menu.

PR #6067, opened 2025-09-11
Add autoFocusOption prop to disable focusing an option until the user provides input

6uzm4n:
This PR builds on #4080 and #3705, updating it to support the most recent version of React Select.

The `autoFocusOption` prop has been tweaked to only disable the default selected option when no user input has been provided, which covers the main two cases where this prop is useful but retains the ability to fill the select using only the keyboard:
* When the menu is first open.
* When a value is selected in an `isMulti` select with `closeMenuOnSelect` set as false, as by default the inputValue is cleared.

The latter of the two has been the source of many long-standing issues (https://github.com/JedWatson/react-select/issues/2848, https://github.com/JedWatson/react-select/pull/3324, https://github.com/JedWatson/react-select/pull/3484, https://github.com/JedWatson/react-select/discussions/5447), mainly because it made two common patterns difficult at best and impossible at worst:
* Cleanly listening to the enter key to trigger a form submission.
* If used in conjunction with `tabSelectsValue` while having multiple selects with `tabIndex`.  it was impossible to tab to the next select without selecting an unprompted option.

Seeing the proposals of previous issues and PRs I've tried to keep the name of the prop as concise and clear as possible, but if anyone has any better name based on its behavior I'm open to suggestions.

This PR doesn't provide any new example in the docs, as the prop itself is quite simple to understand and I don't think there is any place in the docs where it makes sense to have it. However, if any maintainer thinks that an example would be nice to have, I will be happy to include it.


6uzm4n:
@Methuselah96 this is a highly requested feature that would be nice to have for a lot of people 🙏🏻

PR #6097, opened 2026-04-20
fix: stabilise MenuPortal updateComputedPosition to prevent infinite re-render loop

### LINKED ISSUE #6003: Maximum update depth exceeded
react-select version:
5.8.3

production sentry caught the error:
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

call stack:

1. /node_modules/react-select/dist/index-641ee5b8.esm.js in setMenuPortalElement at line 709:5

useLayoutEffect(function () {
    runAutoUpdate();
  }, [runAutoUpdate]);
  var setMenuPortalElement = useCallback(function (menuPortalElement) {
    menuPortalRef.current = menuPortalElement;
    runAutoUpdate();
  }, [runAutoUpdate]);
  // bail early if required elements aren't present
  if (!appendTo && menuPosition !== 'fixed' || !computedPosition) return null;

2. /node_modules/@floating-ui/dom/dist/floating-ui.dom.mjs in cleanupRef.current at line 627:1

resizeObserver = null;
    if (animationFrame) {
      cancelAnimationFrame(frameId);
    }
  };
}
/**
 * Resolves with an object of overflow side offsets that determine how much the
 * element is overflowing a given clipping boundary on each side.
 * - positive = overflowing the boundary by that number of pixels

3. /node_modules/@floating-ui/dom/dist/floating-ui.dom.mjs in cleanupRef.current at line 613:3

    update();
    }
    prevRefRect = nextRefRect;
    frameId = requestAnimationFrame(frameLoop);
  }
  update();
  return () => {
    var _resizeObserver2;
    ancestors.forEach(ancestor => {
      ancestorScroll && ancestor.removeEventListener('scroll', update);
      ancestorResize && ancestor.removeEventListener('resize', update);

4. /node_modules/react-select/dist/index-641ee5b8.esm.js in updateComputedPosition at line 684:7

if (!controlElement) return;
    var rect = getBoundingClientObj(controlElement);
    var scrollDistance = menuPosition === 'fixed' ? 0 : window.pageYOffset;
    var offset = rect[placement] + scrollDistance;
    if (offset !== (computedPosition === null || computedPosition === void 0 ? void 0 : computedPosition.offset) || rect.left !== (computedP {snip}
      setComputedPosition({
        offset: offset,
        rect: rect
      });
    }
  }, [controlElement, menuPosition, placement, computedPosition === null || computedPosition === void 0 ? void 0 : computedPosition.offset,  {snip}

5. then react-dom throws the update depth error

9KenM:
## Problem

`updateComputedPosition` closes over `computedPosition` and lists its
sub-properties (`offset`, `rect.left`, `rect.width`) as `useCallback`
dependencies. This means every call to `setComputedPosition` produces a
new callback identity, which invalidates the
`useLayoutEffect([updateComputedPosition])` and immediately re-fires the
callback.

Under subpixel-drifting conditions (zoomed-out viewport, scrollable
ancestor, or fractional DPI) the measured `offset` oscillates by <1px
between frames. Each oscillation triggers `setComputedPosition` →
identity change → effect re-run → repeat, until React throws:

> Maximum update depth exceeded.

Fixes #6003.

## Solution

Track `computedPosition` in a ref so the comparison inside
`updateComputedPosition` always reads the latest value without the
callback depending on it. This gives `updateComputedPosition` a stable
identity across renders and breaks the loop.

## Reproduction

1. Render a portaled `<Select menuPosition="fixed" />` inside a scrollable container
2. Set browser zoom to 90% or 110% (triggers subpixel rect drift)
3. Open the dropdown and scroll the container
4. Console shows: *Maximum update depth exceeded*


aminland:
Just to pile on, we are also seeing this issue and this fix seems to address the underlying issue in our use case.

PR #6102, opened 2026-07-16
Remove prop-types dependency

atl-mk:
This doesn't remove it entirely from the production package. It still comes through `react-transition-group`.

However, I don't see any need for a direct dependency on `prop-types` given the TS rewrite of v5?


atl-mk:
@Methuselah96 apologies, tagging you for visibility. You seem to be the most active maintainer

Methuselah96 (approved):

Methuselah96:
Thanks!

